PR #805, opened 2015-01-03
New package: blueman

DBError:


xtraeme:
I had an old package for a while... but didn't push it because it did not work correctly: https://gist.github.com/anonymous/540142a3a424ddca9bbc

Have you verified it works?

DBError:
It works now, possibly thanks to a python-gobject update. (or probably some other package)

xtraeme:
Please merge the changes on the template I pasted above; you are missing the required run-time dependencies (among others).

PR #806, opened 2015-01-03
tintin: updated to 2.01.1

hanspolo:


dominikh:
I'm getting a different checksum here: 34630a28422d4a58707607bfcbc6c132b5791fe722300ca394a0abd24750afd9

dominikh:
That'd be because xbps-src is downloading an HTML file from sourceforge, while your checksum is the correct one for the actual archive… I'll fix the template to use the correct distfile.

dominikh:
Merged as 06ae762.

PR #813, opened 2015-01-03
New Package: ssr-0.3.1

logenkain:
If you notice my commented configure_args  on line 8, is there a more elegant way to do this?  If someone doesn't want pulseaudio or jack installed I want it to be easy for them to recompile.  Preferably, the user wouldn't have to recompile at all.  Is there a  better way to do this?

EDIT:
Forgot to mention, the program doesn't seem to allow me to play back video on my R61 (screen goes black and wonky) on which I built it; however, everything works fine on my Desktop which is a rather fresh install.  I'm not sure if this is an R61 specific issue with SSR or if it's something I'm doing wrong in the build.


xtraeme:
The proper way is to enable pulseaudio and jack and then the binary will be linked to the library. Later on if the user installs `pulseaudio` or `jack` it will still work.

xtraeme:
You can enable both features safely and by default.

PR #827, opened 2015-01-07
profile-sync-daemon: version 5.61

graysky2:


Gottox:
Version is still 5.60

graysky2:
Sorry about that.  Fixed here: https://github.com/voidlinux/void-packages/pull/828

PR #828, opened 2015-01-07
profile-sync-daemon: update to v5.61

graysky2:


Gottox:
* [x] Please avoid opening multiple PRs for the same update.
* [ ] only install the systemd files if the systemd option is enabled. For example see ```grep -l build_option */template``` and [the corresponding manual section](https://github.com/voidlinux/void-packages/blob/master/Manual.md#build-options)
```
Installing systemd files...
install -p -d "/destdir//profile-sync-daemon-5.61/etc"
install -p -d "/destdir//profile-sync-daemon-5.61/usr/lib/systemd/system"
```

PR #834, opened 2015-01-09
Build gtk2 and gtk3 versions of webkitgtk

uggedal:
Builds locally under x86_64. i686 and armv7hf builds currently running locally.


xtraeme:
Looks good to me, thanks :+1:

uggedal:
Does not cross build :/ Will look into it in the morning.

uggedal:
I was mistaken, it does cross compile. All dependants which did cross build before does so now. Will merge this tonight.

xtraeme:
Great :+1:

PR #855, opened 2015-01-14
New package: lua-expat-1.3.0

raimundomartins:
Dependency of prosody, which I'll be uploading soon

### DIFF
--- a/srcpkgs/go/template
+++ b/srcpkgs/go/template
@@ -1,6 +1,6 @@
 # Template file for 'go'
 pkgname=go
-version=1.4
+version=1.4.1
 revision=1
 wrksrc=go
 hostmakedepends="git ca-certificates"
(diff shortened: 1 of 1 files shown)


srcpkgs/lua-expat/template
```diff
@@ -0,0 +1,14 @@
+# Template file for 'luaexpat'
+pkgname=luaexpat
+version=1.3.0
+revision=1
+build_style=gnu-makefile
+makedepends="lua51-devel expat-devel"
+depends="lua51 expat"
+short_desc="A SAX XML parser based on the Expat library for the Lua language"
+maintainer="Raimundo Martins <raimundoomartins@gmail.com>"
+license="GPL-compatible"
```
Gottox: Please either provide a license from /usr/share/licenses or install a license file using ```vlicense```

xtraeme:
The directory is named "lua-expat" but pkgname is named "luaexpat", please fix this.

Also the depends line is completely wrong, unless it's a pure lua module.

asergi:
Also, makedepends on 'lua51-devel'. If so, it shoud be named 'lua51-expat'.

xtraeme:
I agree with @asergi 

The lua packages should be renamed to match the python naming:

lua-foo -> (default lua version, 5.2)
lua5.1-foo -> for 5.1
lua5.3-foo -> for 5.3
...

raimundomartins:
How should I proceed to fix this? Just make another commit+push and pull request?
And what about the commit message, use the same?

xtraeme:
Reset your local repository and the force push the commit.

PR #859, opened 2015-01-15
avrdude

allan:


srcpkgs/avr-binutils/template
```diff
@@ -0,0 +1,42 @@
...
+maintainer="allan <mail@may.mooo.com>"
+homepage="http://www.gnu.org/software/binutils/"
+license="GPL-3"
+distfiles="$GNU_SITE/binutils/binutils-$version.tar.gz"
+checksum=cccf377168b41a52a76f46df18feb8f7285654b3c1bd69fc8265cb0fc6902f2d
+wrksrc="binutils-$version"
+only_for_archs="i686 x86_64"
+
+hostmakedepends="zlib-devel"
+depends="binutils"
```
xtraeme: Does this really depend on the native binutils?

allan: i deleted locales and info-pages because they are identical.  then depended on binutils to bring them back..

xtraeme: In that case, this dependency is not very useful. Please remove it.

srcpkgs/avr-binutils/template
```diff
@@ -0,0 +1,41 @@
...
+		--enable-threads \
+		--enable-install-libbfd \
+		--prefix=/usr \
+		--target=avr \
+		--with-pic
+
+	make configure-host
+}
+
+do_build() { make tooldir=/usr; }
```
xtraeme: Can you please follow the common style used in all templates?

```
func() {
   ....
}
```

Thanks.

srcpkgs/avr-gcc/template
```diff
@@ -0,0 +1,71 @@
...
+short_desc="The GNU C Compiler for AVR"
+maintainer="allan <mail@may.mooo.com>"
+homepage="http://gcc.gnu.org"
+license="GFDL-1.2, GPL-3, LGPL-2.1"
+distfiles="$GNU_SITE/gcc/gcc-$version/gcc-$version.tar.bz2"
+checksum=2020c98295856aa13fda0f2f3a4794490757fc24bcca918d52cc8b4917b972dd
+wrksrc="gcc-$version"
+only_for_archs="i686 x86_64"
+
+hostmakedepends="avr-binutils gcc-c++ zlib-devel libmpc-devel
```
xtraeme: `gcc-c++` is unnecessary, it's part of the bootstrap pkgs.

srcpkgs/avr-gcc/template
```diff
@@ -0,0 +1,71 @@
...
+		--with-plugin-ld=ld.gold \
+		--with-system-zlib \
+		--without-included-gettext
+}
+
+do_build() {
+	cd obj-avr
+
+	# well...
+	sed -i 's/-mtune=generic//g' Makefile
```
xtraeme: Rather than using `sed` I'd suggest you to use this before running `./configure`:

```
CFLAGS="${CFLAGS/-mtune=generic/}"
```

srcpkgs/avrdude/template
```diff
@@ -0,0 +1,15 @@
...
+revision=1
+build_style=gnu-configure
+short_desc="An utility to manipulate ROM and EEPROM of AVR microcontrollers"
+maintainer="allan <mail@may.mooo.com>"
+license="GPL-2"
+homepage="http://www.nongnu.org/avrdude/"
+distfiles="$NONGNU_SITE/$pkgname/${pkgname}-$version.tar.gz"
+checksum=9e98baca8e57cad402aaa1c7b61c8de750ed4f6fed577f7e4935db0430783d3b
+only_for_archs="i686 x86_64"
+hostmakedepends="bison flex elfutils-devel libusb-compat-devel"
```
xtraeme: `bison` is a dependency of `flex`, so that it's redundant.

xtraeme: Also probably, the other devel packages should be moved to `makedepends` instead.

srcpkgs/avr-libc/template
```diff
@@ -0,0 +1,24 @@
...
+hostmakedepends="avr-binutils avr-gcc"
+depends="avr-binutils avr-gcc"
+nostrip=yes
+
+do_configure() {
+	CC=avr-gcc CC_FOR_BUILD=avr-gcc AR=avr-ar AS=avr-as RANLIB=avr-ranlib \
+	./configure --host=avr --prefix=/usr
+}
+
+do_build() { make; }
```
xtraeme: In that package, you can use `build_style=gnu-configure` and remove `do_build()` and `do_install()`.

PR #873, opened 2015-01-16
LoginKit-git: rename LoginKit-git-devel to LoginKit-devel-git

Gottox:


xtraeme:
Looks good, thanks :+1:

PR #874, opened 2015-01-16
New package: xcape-1.1

oholiab:


xtraeme:
Please try to avoid git packages... there's a recent 1.1 version that can be packaged instead.

Thanks.

xtraeme:
Thanks for contributing your first source packages. You've created it correctly :+1:

PR #875, opened 2015-01-16
New packages: python-mpd2, sonata

v2px:


srcpkgs/python-mpd2/template
```diff
@@ -0,0 +1,34 @@
+# Template file for 'python-mpd2'
+pkgname=python-mpd2
+version=0.5.4
+revision=1
+noarch=yes
+wrksrc="${pkgname}-${version}"
```
xtraeme: This `wrksrc` assignment is redundant, by default wrksrc has this value.

srcpkgs/python-mpd2/template
```diff
@@ -0,0 +1,34 @@
...
+depends="python"
+short_desc="Provides a client interface for the Music Player Daemon"
+maintainer="Jens E. Becker <v2px@v2px.de>"
+homepage="https://github.com/Mic92/python-mpd2"
+license="GPL-3"
+distfiles="https://github.com/Mic92/python-mpd2/archive/v${version}.tar.gz"
+checksum=04dbc3b8d90746298e42852c5b7fb5be5e7af2ca03e812059639b6c060d98c53
+
+post_install() {
+	vlicense GPL.txt LICENSE
```
xtraeme: There's no need to install the license if it's GPL or LGPL licensed, remove `post_install`.

srcpkgs/sonata/template
```diff
@@ -0,0 +1,22 @@
+# Template file for 'sonata'
+pkgname=sonata
+version=1.7a2
+revision=1
+noarch=yes
+wrksrc="${pkgname}-${version}"
```
xtraeme: Ditto :-)

PR #877, opened 2015-01-16
tlp: add runit services

mid-kid:
Also removed 2 lines which changed the absolute path to /usr/bin in udev config and systemd services, while tlp is located in /usr/sbin.

### DIFF
--- a/srcpkgs/tlp/files/tlp/run
+++ b/srcpkgs/tlp/files/tlp/run
@@ -1,0 +1,3 @@
+#!/bin/sh
+tlp init start
+exec pause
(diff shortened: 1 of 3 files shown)


srcpkgs/tlp/files/tlp/run
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+tlp init start
```
xtraeme: This script is incomplete:

- Before executing pause(1), we need to make sure that "tlp init start" succeeded.
- It's useful to use `chpst -b tlp pause` to change `pause(1)` argv[0] to `tlp`.

Change it to:

```
#!/bin/sh
tlp init start || exit 1
exec chpst -b tlp pause
```

PR #879, opened 2015-01-17
New package: konversation-1.5.1_1 (A user friendly IRC client for KDE).

TheNumb:


srcpkgs/konversation/template
```diff
@@ -0,0 +1,13 @@
+# Template file for 'konversation'
+pkgname=konversation
+version=1.5.1
+revision=1
+build_style=cmake
+hostmakedepends="automoc4 cmake"
+makedepends="kdelibs-devel phonon-devel kdepimlibs-devel qca-devel docbook-xml"
```
xtraeme: `docbook-xml` should be in `hostmakedepends` instead, but not important because `qt` cannot be cross compiled right now.

Thanks :+1:

PR #888, opened 2015-01-18
 New package: xskat-4.0

pulux:
The card game Skat for Linux

xskat need gccmakedep to build


srcpkgs/xskat/template
```diff
@@ -0,0 +1,13 @@
+# Template file for 'xskat'
+pkgname=xskat
+version=4.0
+revision=1
+build_style=gnu-configure
+hostmakedepends="pkg-config xorg-util-macros"
+makedepends="libX11-devel imake xorg-cf-files gccmakedep"
```
xtraeme: libX11-devel should be the only pkg in `makedepends`. Move the other ones to `hostmakedepends`.

PR #889, opened 2015-01-18
New package: notion-3_2014052800

vtoc:


srcpkgs/notion/template
```diff
@@ -0,0 +1,27 @@
+# Template file for 'notion'
+
+pkgname="notion"
+_minversion="2014052800"
+version="3_${_minversion}"
+revision=1
+short_desc="Tabbed, tiling window manager"
+makedepends="make pkg-config groff lua-devel>=5.2 \
```
xtraeme: - You don't have to add `make` as build dependency, it's part of the bootstrap pkgs (base-chroot).
- `pkg-config` and `groff` must be moved to `hostmakedepends` instead.

srcpkgs/notion/template
```diff
@@ -0,0 +1,27 @@
...
+	 libX11-devel libXext-devel libSM-devel libXinerama-devel libXrandr-devel"
+depends="lua>=5.2 libX11 libXext libSM libXinerama libXrandr"
```
xtraeme: You don't need to specify the dependencies manually, unless xbps-src does not detect them (xbps-src scans all binaries and generates a dependency by using `objdump -p foo|grep NEEDED`).

Try:

```
$ ./xbps-src install notion
$ ./xbps-src show-deps notion
```
to see what dependencies are detected by xbps-src (after removing the depends line).

Gottox:
* [ ] add lua to hostmakedepends to make cross-compilation work.

PR #894, opened 2015-01-19
grub: fallback for failed grub-probe on zfs in mkconfig

vtoc:


xtraeme:
The package revision must be bumped, please do so.

PR #897, opened 2015-01-19
New package: spotify-0.9_1

teh-jazzman:
Version numbers vary significantly between 32 and 64 bit. 0.9 is the smallest common denominator.
xlint claims allow_unknown_shlibs is a custom variable...

### DIFF
--- a/srcpkgs/spotify/INSTALL
+++ b/srcpkgs/spotify/INSTALL
@@ -29,7 +29,7 @@ if test "$ACTION" = "post"; then
 	mv "${_BUILDDIR}/opt/spotify/spotify-client" /usr/share/spotify/spotify-client
 	for _s in 16 22 24 32 48 64 128 256 512; do
 		mkdir -p "/usr/share/icons/hicolor/${_s}x${_s}/apps"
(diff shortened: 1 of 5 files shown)


silvernode:
I've been waiting for this, thanks

xtraeme:
I'm not sure about legal terms with that one... are we allowed to pkg and distribute this, even in a non-free repository?

teh-jazzman:
Hmm, you're right. I thought it was at least in the official Debian/Ubuntu repos, but they are providing their own repo. How about I change the package to only contain the libgcrypt library and do the rest at installation-time via the INSTALL and REMOVE files?

Gottox:
@teh-jazzman that's a solution. We already do that with [chromium-pepper-flash](https://github.com/voidlinux/void-packages/tree/master/srcpkgs/chromium-pepper-flash) and [oracle-jdk](https://github.com/voidlinux/void-packages/tree/master/srcpkgs/oracle-jdk)

Gottox:
Great!

The only remaining issue is that the launcher icons aren't correctly installed. Will merge after the fix.

teh-jazzman:
I changed the icon names to fit the entry in the spotify.desktop file. I hope that fixes the problem. I currently don't run a DE, so I don't know if it is working or not. If not I'll build me a test environment later today.

Gottox:
Works for enlightenment :+1:

PR #898, opened 2015-01-19
New package: tcc-git (tcc is OLD, this fixes the issue).

necrophcodr:
So, tcc is old. Like getting REALLY old. It seems that i hadn't been maintained by the creator for a long while, but newer and more revised versions are available, at http://repo.or.cz/w/tinycc.git where they've added ARM support, amongst a ton of other things.
This package will want to replace tcc, because installing both is not exactly compatible.
However, is it a good idea to have this package? Or should it simply replace tcc entirely? Who knows.
Discuss.


Gottox:
We'll stick to the release. Feel free to add patches to tcc if something mandatory is missing

necrophcodr:
There's been tons of patches added already, like ARM support for one. Fully functional ARM support.

Why should this package stick to the "release", when so many other packages in void linux do not?

Gottox:
Most of the packages using git master don't have any releases, but are somehow relevant.

Can you give any examples for -git packages which do have older releases?

necrophcodr:
Yes. LoginKit, b2wm, and cwm.
Those releases are old though, and are missing essential functionality, as is the existing build of TinyCC in void linux, where the git version has:
Warning about conflicting compiler flags.  
ARM support.  
An actual stdbool.h that is standard conformant.  
Fixes to x86-64 ELF segment alignment (appearantly slightly broken in latest 'release'.  
Valgrind support.  
Working va_args implementation.  
Fix compilation on some of the *BSD systems, which were not supported before.  
Integer constant parsing is not C99 conformant.  

That, and a whole shitton more of fixes, optimizations, feature additions, and what have you. I'm not gonna go through the entire git log since 0.9.26, but there's happened a LOT.

I'm not sure why it makes sense to wait until the next release, considering how stable TCC really is, and with the additional benefit of supporting all platforms that Void Linux currently supports. But I digress.
I guess I'll talk to the developers of TCC about a new release then, and see if it can ever happen.

uggedal:
@necrophcodr: If it's so stable you should bug the upstream to cut a release.

Gottox:
I think it's a good idea to ask them for a new release.

There are a few reasons I'm against -git versions where there aren't needed:

* update-check will not work anymore

* switching between -git and release is a rather expensive task in contrast to just updating.

* Using a release means that the upstream maintainer thinks, that the application is ready to be packages. Using a -git package means that it will potentionally break when a package gets updated at the wrong time.

Just open a new PR once the new release ist out :)

PR #902, opened 2015-01-20
base-files: Prevent 'vkpurge list' from displaying errors at unknown kernel files

teh-jazzman:
I have some kernels from another distro on my boot partition and "vkpurge list" always throws "Invalid string..." errors when trying to list them. I re-used an apparently discarded local variable and dropped another in the process.


xtraeme:
Looks good, thanks :+1:

PR #903, opened 2015-01-21
New package: bochs-2.6.7

mid-kid:


mid-kid:
Travis CI derped, test it yourself instead.

srcpkgs/bochs/template
```diff
@@ -0,0 +1,18 @@
...
+homepage="http://bochs.sourceforge.net/"
+hostmakedepends="pkg-config"
+makedepends="libX11-devel libXrandr-devel gtk+-devel"
+configure_args="--enable-debugger --enable-disasm"
+distfiles="$SOURCEFORGE_SITE/bochs/bochs-${version}.tar.gz"
+checksum="a4354eeb699662da869b038e225160ec726b883f1b4f425a06cf56cbffbb8282"
+
+do_build() {
+	sed -i 's/^LIBS = /LIBS = -lpthread/g' Makefile
+	make -j 1
```
Gottox: Use `pre_build()` instead of `do_build()` and remove the make call. This way all build variables can be set for make.

To disable parallel build use the config varialbe` disable_parallel_build=yes`. See [the corresponding manual section](https://github.com/voidlinux/void-packages/blob/master/Manual.md#optional-variables) for further information on `disable_parallel_build`

mid-kid: I didn't even know that existed, thanks for informing me. Fixing it right away.

PR #905, opened 2015-01-21
gpgme: removed gnupg(version 1) dependency

vtoc:


xtraeme:
I'm not sure about this one... if gpgme needs gpg at run-time, and this does not depend on any of them it won't work.

vtoc:
it does need a gpg implementation at runtime. However, there is gpg1 and gpg2 and there is no reason why it should dpend on 1 instead of 2. Both implementation can not really co-exist as they share ~/.gnupg in an incompatible way as well as users usually expect that /usr/bin/gpg and /usr/bin/gpg2 point to the same implementation.

```
/usr/bin/gpgme-tool --gpg-binary=/usr/bin/gpg2
OK
LISTKEYS
...
```

xtraeme:
I wouldn't expect gpg and gpg2 to point to the same implementation.

Gottox:
Any process on this topic?

PR #911, opened 2015-01-23
zpaq: fix empty manual

necrophcodr:
Dropped pod2mdoc dependency, and use perl (=> /usr/bin/pod2man) instead, due to issues with pod2mdoc.
Appearantly it generated an empty file with only date and title, and this fixes that. The manual is now working (proper testing has been done on my end).


xtraeme:
Revision must be bumped.

necrophcodr:
bumped

PR #912, opened 2015-01-23
New Package:clementine-git-2015161

silvernode:


srcpkgs/clementine-git/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'clementine-git'
+pkgname=clementine-git
+version=2015161
```
Gottox: That's no valid date. use YYYYMMDD.

Gottox:
Personally I would prefere the release over the unstable master.

Are there any reasons why you choose to use the master?

uggedal:
+1 for stable release. We should only use -git packages for projects without releases.

silvernode:
April 23rd was the last stable release.  I tried to use it anyway, but there is an issue with Udisks, I noticed ArchLinux made a patch for it, but I'm not sure how to patch it myself. So I decided to go with unstable master which happened to compiled successfully.

xtraeme:
The latest release is recent enough to avoid a git package.

Please do so.

logenkain:
I believe he said he doesn't know how to get the release to build. Though I found the Arch patch he was refering to: https://projects.archlinux.org/svntogit/community.git/tree/trunk/udisks-namespace.patch?h=packages/clementine

Wish I could help but, I'm not sure how to deal with this either.  I took a look at the patch (I'm still rather new at reading patches) and I can't figure out what it's adding where.

I'm going to try to build the release to see what happens.  Will report back.

silvernode:
Try -Np0 -I in the patch_args variable. I noticed the Arch pkgbuild had that. I am having other issues with 1.2.3 other than udisks. Something else won't build now. Never had this issue before.

silvernode:
I'll close this for now and create a new request once a release of clementine is able to build.

PR #918, opened 2015-01-24
Tomahawk

Duncaen:


srcpkgs/jreen/template
```diff
@@ -0,0 +1,25 @@
+# Template file for 'jreen'
+pkgname=jreen
+version=1.2.0
+revision=1
+build_style=cmake
+configure_args="-DJREEN_FORCE_QT4=yes \
+ -DCMAKE_INSTALL_PREFIX=/usr \
```
xtraeme: - CMAKE_INSTALL_PREFIX and CMAKE_BUILD_TYPE are already defined in `build_style/cmake.sh`.
- pkg-config must be in `hostmakedepends` instead.

srcpkgs/websocketpp-devel/template
```diff
@@ -0,0 +1,13 @@
+# Template file for 'websocketpp'
+pkgname=websocketpp-devel
```
xtraeme: Any reason this pkg is not named `websocketpp` rather than `websocketpp-devel` ?

srcpkgs/jreen/template
Duncaen: ty removed it in multiple templates

srcpkgs/websocketpp-devel/template
Duncaen: It only contains headerfiles, i can rename it but i tought if its only for devel, name it devel.

xtraeme: In that case there's no need to add the `devel` suffix.

PR #924, opened 2015-01-26
New package: python-pyparsing 2.0.3

pulux:
New package: python-pyparsing 2.0.3


asergi:
`xbps-query -Rs python-parsing`

pulux:
Why do you use not the common name pyparsing?

PR #929, opened 2015-01-27
New package: root-devel-5.34.25

raimundomartins:
A C++ interpreter and framework from CERN.
It's root-devel because I don't know yet how to install without headers and *.so$


Gottox:
```
git rebase -i origin/master
# ... remove the commits you want to get rid of by deleting the lines (here 7b6a52e and d5eacfc)
git push -f
```

That way you don't have to close the PR but fix the commits instead :)

srcpkgs/root-devel/INSTALL
```diff
@@ -0,0 +1,2 @@
+# INSTALL
```
xtraeme: This is completely crap... I'd like to know the reason to run ldconfig(8).

srcpkgs/root-devel/template
```diff
@@ -0,0 +1,39 @@
+# Template file for 'root-devel'
+pkgname=root-devel
```
xtraeme: If the software is named *root* why did you name it *root-devel*?

srcpkgs/root-devel/template
```diff
@@ -0,0 +1,39 @@
...
+version=5.34.25
+revision=1
+build_style=configure
+wrksrc="root"
+makedepends="libGL glu-devel pcre-devel liblzma-devel gsl-devel fftw-devel libXpm-devel libXft-devel"
```
xtraeme: Adding `libGL` as a build dependency is completely wrong, it's unnecessary.

srcpkgs/root-devel/template
```diff
@@ -0,0 +1,39 @@
...
+		arm?l*) TARGET=linuxarm ;;
+	esac
+
+	./configure $TARGET --prefix=/usr ${features} ${sys_libs}
+}
+
+post_install() {
+	vmkdir etc/
+	vmkdir etc/ld.so.conf.d/
+	echo $(${DESTDIR}/usr/bin/root-config --libdir) > "${DESTDIR}"/etc/ld.so.conf.d/root.conf
```
xtraeme: Rather than using `ld.so.conf.d` cannot you set RPATH instead? this avoids the need to run ldconfig(8) via post-install.

srcpkgs/root-devel/template
raimundomartins: Why is it wrong? Without it the package configure doesn't detect opengl and builds without support for it.

srcpkgs/root-devel/template
raimundomartins: Like I said in the first comment:
It's root-devel because I don't know yet how to install without headers and *.so$
You mean to say that even so you prefer the package to be called root? Very well, I'll change it.

srcpkgs/root-devel/template
raimundomartins: I'm not familiar with RPATH, but from man ld:
 -rpath=dir
           Add a directory to the runtime library search path.  This is used
           when linking an ELF executable with shared objects.
It seems to me that this is useful only when compiling a library that links to an exotic path. That is not the case here. The idea of ldconfig and ld.so.conf.d is to make the framework discoverable at runtime by other applications.
This package's libs are in by default in ${LIBDIR}/root subdirectory, hence not automatically discoverable.

xtraeme:
No progress for 10 days... unless you fix the pkg I'll decline this PR.

raimundomartins:
Ah, I was waiting for your reply to my responses. Are you saying I can't have libGL in makedepends, and that I can't use ldconfig in install?
If that's the case, then you might as well close the PR, because without the former the package doesn't have opengl support, and without the latter the package can't be linked, so it's kind of pointless.

logenkain:
On Fri, Feb 06, 2015 at 04:11:29PM -0800, raimundomartins wrote:
> If that's the case, then you might as well close the PR, because without the former the package doesn't have opengl support

I don't think libGL should ever be a dependancy due to AMD/Nvidia
proprietary drivers conflicting with it.

Then again, I'm not sure what your package does.

xtraeme:
What I was saying is that you don't need to add libGL as build dependency, you should use *MesaLib-devel* instead, which depends on all mesa pkgs (including *libGL*).

About the ldconfig issue, that's why I suggested to set rpath in the binaries... to avoid the need of using ldconfig(8).

raimundomartins:
Ah! Ok, I don't know those things that well, I needed libGL so added that
one, no had mentioned mesalib :P ok, I'll try it today.
About rpath, I'll also try it, but I'm pretty sure it won't have the
intended effect. I've already explained why.

xtraeme:
Incomplete package... a -devel subpkg was necessary and wasn't provided. Feel free to resubmit it.

PR #931, opened 2015-01-27
runit-void: Install only active services by default

teh-jazzman:
When a user has deactivated a service with a `down` file, I think it is prudent to not install it by default.


xtraeme:
Good idea! :+1:

PR #938, opened 2015-01-28
New package: git-annex-5.20150113_1

oholiab:
This has got the tahoe-lafs special remote missing but it has pretty much all the other functionality and I've been using it for a week or so with no problems at all.

### DIFF
--- a/srcpkgs/git-annex/template
+++ b/srcpkgs/git-annex/template
@@ -1,0 +1,26 @@
+# Template file for 'git-annex'
+pkgname=git-annex
+version=5.20150113
+revision=1
+hostmakedepends="ghc cabal-install pkg-config"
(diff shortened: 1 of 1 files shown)


chneukirchen:
I think it would be better to use a sandbox to build, look into the pandoc template.

oholiab:
Okay, I'll give it a go :) I know next to nothing about the haskell ecosystem so would you mind if I hit you up for help if I get stuck?

oholiab:
Couple of things definitely need doing here then (after chats on IRC):
1. It doesn't need bind as a dep, just bind-utils. Misunderstanding on my part.
2. While I'm at it I'll talk to the author to try and get some more ideas of where I can trade out for more lightweight dependencies
3. Do the build in a cabal sandbox

xtraeme:
Stalled for a month... feel free to resubmit it, thanks.

PR #939, opened 2015-01-28
runit-void: Retain a rescue console

teh-jazzman:
This should provide a means to still access the system if the `runit-void` update goes awry.
Of course this would only be needed for a short while and so far noone else reported problems, so it might just as well be ignored :)


xtraeme:
I don't think we don't need such complication... the runit-void issue that happened some days ago is not expected to happen anymore.

teh-jazzman:
Okay :)

PR #945, opened 2015-02-02
New Package: pianobar-2014.09.28

silvernode:


srcpkgs/pianobar/template
```diff
@@ -0,0 +1,17 @@
...
+makedepends="faad2-devel libao-devel gnutls-devel libgcrypt-devel json-c-devel ffmpeg-devel"
+short_desc="A free/open-source, console-based client for Pandora radio"
+maintainer="Mollusk <mollusk@homebutter.com>"
+license="MIT"
+homepage="http://6xq.net/projects/pianobar/"
+distfiles="https://github.com/PromyLOPh/pianobar/archive/${version}.tar.gz"
+checksum="c4b9d84d1b0b80dc6ff2957f94b8fdf9219b063d154cf09337f1d15b137f3acb"
+pre_build(){
+	make
+	}
```
Gottox: * [ ] why is `pre_build()` needed?
* [ ] please indent correctly, `}` should not be indented.

silvernode:
Without pre_build() or some way to actually run make, pianobar would not build and I don't know why. The first thing I tried to do was build normally, however strange things happen while doing a build normally versus just typing make. Files that are able to be found while using plain make are not found while doing a normal build with xbps-src. Thus the build fails, so I decided to just go with what worked. I'll fix the indent, that was an obvious mistake that went overlooked.

Gottox:
by setting build_style=gnu-makefile, make is called through do_build anyway. see `common/build_style/gnu-makefile.sh`

`make` without any arguments will fail always fail except for native compile and it won't set the correct compilerflags.

silvernode:
Initially I tried with arguments specified by the pianobar developers but that didn't seem to matter. Build failed every time. It seems (and I could be wrong) that pianobar doesn't require special arguments. However, if I understand you correctly, regardless of the package in qyestion, there needs to be arguments otherwise the build will fail.

silvernode:
apparently the` Makefile` has an issue on line 127, which seems to conflict with line 8 of `gnu-makefile-sh`.  

line 127 of `Makefile` for pianobar: 
`@echo "    CC  $<"`
line 8 of `gun-makefile.sh`:
`CC="$CC" CXX="$CXX" LD="$LD" AR="$AR" RANLIB="$RANLIB" \`

xtraeme:
Either fix the software makefile to respect our vars or use `do_build()` and `do_install()`.

logenkain:
I was actually currious about this myself, is there no way to properly add compiler flags within a template?

xtraeme:
Conpiler/linker flags are set in the environment, if the makefile does not
respect them, you should change it to make so.

silvernode:
Should be good to go now. I chose the route of using do_build instead of changing the makefile, just to add the proper flags.

xtraeme:
This is still wrong :-)

xtraeme:
I pushed your package with the required tweaks, see my commit to see what you were missing.

Thanks!

PR #958, opened 2015-02-09
New package: moc

necrophcodr:
This adds the package "moc" - "music on console", which is a music player for the terminal. Somewhat like cmus, but not really at all.

Additional details are available on the website: http://moc.daper.net/

It has of course been tested to build, natively on all platforms. Cross building not supported due to no qemu-arm support in xbps-src yet.


xtraeme:
only_for_archs is wrong if you just want to disable cross compilation, use `nocross=yes` instead.

xtraeme:
Please remove the `_tarball` var, it's pointless.

PR #960, opened 2015-02-10
New package: libenet

necrophcodr:
Yadda yadda new package called enet that provides cross-platform reliable UDP, used in games and game engines like 0AD and Cube.

http://enet.bespin.org/


xtraeme:
Can you please stop adding nocross and those wrong configure_args? your previous pkg was cross compilable just by setting `build_style=gnu-configure`.

I would assume that those `configure_args` you've added are also wrong.

xtraeme:
Also the soname of the shared library must be added to `common/shlibs`, as well as creating a new `development` subpkg.

necrophcodr:
My apologies. It can be a bit difficult to read through all the possible ways of doing things right from time to time.

You are right, that a development subpkg should be added, and this is in no way the absolute correct way of doing things. I must've missed the common/shlib warning but It has been changed now.
I've been trying to get 0AD working, and though it required this package, I could not find it. So I attempted at adding the library, and got that part working, but I wanted to let it go upstream.

I hope that polishing it up works better for you.

necrophcodr:
Reopened from different branch. https://github.com/voidlinux/void-packages/pull/961

PR #961, opened 2015-02-10
New package: libenet

necrophcodr:
Yadda yadda new package called enet that provides cross-platform reliable UDP, used in games and game engines like 0AD and Cube.

http://enet.bespin.org/


xtraeme:
Much better now, thanks.

Please fix the indentation and add the .so symlink to devel pkg as final step.

PR #970, opened 2015-02-14
New package: lilo-24.1

nmeum:
Regarding this package, there are a few thinks that I would like to clarify. First of all LILO requires that `LC_ALL` is set to `C`, otherwise it doesn't build at all. Furthermore, LILO install itself to `/sbin` I moved it to `/usr/sbin` instead if I should move it to `/usr/bin` instead please let me know. Other utilities are also installed to `/usr/sbin`.  

One of these tools is `keytab-lilo` this tool is also installed by syslinux (even though syslinux installs it to `/usr/bin` and not to `/usr/sbin`). The version of `keytab-lilo` distributed by syslinux and the version distributed by LILO slightly differ, but I decided to remove it from syslinux instead of making these two packages conflict. So if you need `keytab-lilo` you have to install the LILO package.

In addition to that I decided not to install tux and debian images which are located in the `images/` subfolder in the LILO source tree, because a) I don't think that it is useful to have those b) The Makefile in `images/Makefile` does weird stuff and c) Gentoos ebuild doesn't install these either.

Last but not least I also decided to move and remove some files in `/etc`. For one thing, I removed scripts installed to `/etc/initramfs` and `/etc/kernel` because a) They assume that LILO is installed to `/sbin/lilo` and b) They seem to be debian specific or at least they don't seem useful on voidlinux. The last thing I did was move the example configuration file `/etc/lilo.conf_example` to `/etc/lilo.conf`.

If any of the things above are bullshit then let me know…


xtraeme:
Your package is trying to replace the `/usr/bin` directory with a file, resulting in that error.

nmeum:
Actually I fixed that a long time ago but didn't rebuild the package properly, however, after rebuilding from scratch I noticed that the URL was wrong. I fixed that and the package can be merged now, if you are ok with the other stuff…

xtraeme:
About `/etc/kernel/`, we do have `/etc/kernel.d/` which should be similar to what Debian uses. See grub or gummiboot as examples.

Thanks

nmeum:
cool, I guess I will write a hook for `/etc/kernel.d`.

PR #971, opened 2015-02-15
lilo: add kernel hooks

nmeum:
This adds a post-install and a post-remove kernel hooks for LILO. For each image that gets installed a file in in `/etc/lilo.d/` is created. Furthermore, a file called `/etc/lilo.d/default.conf` is created, this particular file just specifies the default image to boot. The files in `/etc/lilo.d` are not parsed by LILO by default, thus a temporary file is created to concat `/etc/lilo.conf` with `/etc/lilo.d/*.conf` this temporary file is passed to LILO using the `-C` flag. That's it.

### DIFF
--- a/srcpkgs/lilo/template
+++ b/srcpkgs/lilo/template
@@ -1,10 +1,11 @@
 # Template file for 'lilo'
 pkgname=lilo
 version=24.1
-revision=1
+revision=2
 only_for_archs="i686 x86_64"
 hostmakedepends="bin86"
 makedepends="device-mapper-devel"
+make_dirs="/etc/lilo.d 0755 root root"
 conf_files="/etc/lilo.conf"
 short_desc="Standard Linux boot loader"
 maintainer="Sören Tempel <soeren+void@soeren-tempel.net>"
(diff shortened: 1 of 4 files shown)


srcpkgs/lilo/template
```diff
@@ -21,12 +22,22 @@ pre_build() {
...
 
 do_install() {
 	make DESTDIR=${DESTDIR} install
-	rm -r ${DESTDIR}/etc/initramfs ${DESTDIR}/etc/kernel
 	mv ${DESTDIR}/sbin/* ${DESTDIR}/usr/sbin
-	mv ${DESTDIR}/etc/lilo.conf_example ${DESTDIR}/etc/lilo.conf
+
+	# Setup config files
+	rm ${DESTDIR}/etc/lilo.conf_example
+	vinstall ${FILESDIR}/lilo.conf 600 etc
```
xtraeme: `/etc/lilo.conf` must be added to `conf_files`, i.e:

`conf_files="/etc/lilo.conf"`

Other than that looks ok, thanks.

srcpkgs/lilo/template
```diff
@@ -1,10 +1,11 @@
 # Template file for 'lilo'
 pkgname=lilo
 version=24.1
-revision=1
+revision=2
 only_for_archs="i686 x86_64"
 hostmakedepends="bin86"
 makedepends="device-mapper-devel"
+make_dirs="/etc/lilo.d 0755 root root"
 conf_files="/etc/lilo.conf"
```
nmeum: `/etc/lilo.conf` is already a port of `conf_files`.

xtraeme: Oh true, missed it :+1:

nmeum:
The file path for the initramfs was wrong. I fixed that, now it works. You can merge it now if you want to.

nmeum:
Wait a second, I would just like to change a few things in the default config...

nmeum:
Ok, done. This can be merged for real now.

PR #978, opened 2015-02-17
New package: nim-0.10.2

allan:


srcpkgs/nim/template
```diff
@@ -0,0 +1,59 @@
+# Template file for 'nim'
+pkgname=nim
+version=0.10.2
+revision=1
+depends="libzip gcc gcc-c++"
```
xtraeme: `gcc` is a dependency of `gcc-c++`, you can remove it.

srcpkgs/nim/template
```diff
@@ -0,0 +1,59 @@
...
+makedepends="libzip"
```
xtraeme: Probably you do need to add `libzip-devel` to `makedepends` and remove `libzip` from `depends`.

srcpkgs/nim/template
```diff
@@ -0,0 +1,59 @@
...
+hostmakedepends="git ed"
+short_desc="Nim language"
```
xtraeme: Any chance to improve the short description somehow?

srcpkgs/nim/template
```diff
@@ -0,0 +1,59 @@
...
+
+post_extract() {
+	cd $wrksrc
+	test -d csources ||
+	git clone --depth=1 https://github.com/nim-lang/csources
+}
+
+do_build() {
+	if [ "$CROSS_BUILD" ]; then
+		ed >/dev/null 2>&1 config/nim.cfg <<-EDIT
```
xtraeme: Assuming cross compilation will always be for arm is completely wrong. Use something like this:

```
 arm*) ...
esac
```

srcpkgs/nim/template
allan: It doesn't need headers from libzip-devel.  Only used for the installer it compiles for itself. Maybe it should go in hostmakedepends then?

PR #991, opened 2015-02-20
Added armv5te profile

thypon:
This add armv5te profile needed by some cheap 10/15$ arm board.


xtraeme:
Do you know the build profile is only required to *compile* natively?

xtraeme:
This is unnecessary for now. I really doubt you would like to compile natively on armv5.

PR #1011, opened 2015-02-22
Mpv

Duncaen:


xtraeme:
Looks good... but you should check the pkg-config files and add the missing *-devel* pkgs if there's any that is needed.

PR #1016, opened 2015-02-24
New package: daemontools-0.76

bougyman:


bougyman:
For a clearer view of this whole thing: http://phabricator.rubyists.com/D1

srcpkgs/daemontools/template
```diff
@@ -0,0 +1,14 @@
+# Template file for 'daemontools'
+pkgname=daemontools
+version=0.76
+revision=2
+wrksrc=admin/${pkgname}-${version}
```
xtraeme: `wrksrc` this way is wrong and while running the `clean` target it won't be fully removed.

Use this instead:

```
wrksrc=admin
build_wrksrc=${pkgname}-${version}
```

bougyman: Will do.

bougyman: Done

Manual.md
```diff
@@ -544,6 +544,8 @@ be defined to set the allowed python versions to be built, i.e:
 
 - `waf` For packages that use the Python `waf` method with python2.
 
+- `slashpackage` For packages that use the /package heirarchy and package/compile to build.
```
xtraeme: `heirarchy` -> `hierarchy`

srcpkgs/daemontools/template
```diff
@@ -0,0 +1,13 @@
+# Template file for 'daemontools'
+pkgname=daemontools
+version=0.76
+revision=1
+wrksrc=admin
```
xtraeme: `wrksrc` and `build_wrksrc` could be set in the environment for this `build_style`, this way:

```
void-packages/common/environment/build-style/slashpackage.sh
wrksrc=admin
build_wrksrc=${pkgname}-${version}
```

And then you should be able to remove those two vars in templates that set `build_style=slashpackage`.

bougyman: Packages aren't always going to be in 'admin'. There 'prog', and a bunch of other top-level categories. Is it ok to set build_wrksrc in the build_style script and then all that has to be filled in is the category (wrksrc) ?

xtraeme:
Sure, I thought all them were using `admin`. In that case your change is ok.

xtraeme:
Are not there manpages for those bins in daemontools?

Manual.md
bougyman: Corrected

bougyman:
Nope, no manpages. It has a README that just points to the website I have as 'homepage', didn't see a reason to include it.

PR #1022, opened 2015-02-25
slashpackage.sh: Set compiler to $CC

bougyman:
If $CC is correct at this point this should fix cross compiling for anything slashpackage.


bougyman:
No good, doesn't work

PR #1024, opened 2015-02-25
daemontools: update to 0.76_2

bougyman:
* Added manpages (from http://smarden.org/djb)
* Added crosscompile patch (from openwrt)


xtraeme:
Patches are in p1 format and there's no patch_args set.

bougyman:
The patches are working as-is, from ${wrksrc} p0 seems to work fine.  Is there something I'm missing? They build fine here.

xtraeme:
Ok, sorry. WIll apply it in a some minutes.

PR #1030, opened 2015-02-26
New package: wmii-3.9.2

dookerdo:


srcpkgs/wmii/template
```diff
@@ -0,0 +1,38 @@
+# Template file for 'wmii'
+pkgname=wmii
+version=3.9.2
+revision=1
+wrksrc=${pkgname}+ixp-${version}
+#create_wrksrc=yes
```
uggedal: Remove unchanged comments from `xnew`.

srcpkgs/wmii/template
```diff
@@ -0,0 +1,38 @@
...
+#only_for_archs="i686 x86_64"
+#build_style=gnu-configure
+build_style=""
```
uggedal: You should use `build_style=gnu-makefile`.

srcpkgs/wmii/template
```diff
@@ -0,0 +1,38 @@
...
+#configure_args=""
+#make_build_args=""
+#make_install_args=""
+#conf_files=""
+#make_dirs="/var/log/dir 0755 root root"
+hostmakedepends=""
```
uggedal: Remove empty vars.

bougyman:
Please fix the xlint errors on this template.  see http://phabricator.rubyists.com/P4. Travis will fail with these as well.

srcpkgs/wmii/template
```diff
@@ -0,0 +1,38 @@
...
+makedepends="pkg-config libXext-devel libXrandr-devel libXft-devel libXrender-devel libX11-devel libXinerama-devel"
+depends="dmenu"
```
uggedal: I don't think dmenu is a hard dependency, it's bound to a key in the example/default scripts.

srcpkgs/wmii/template
```diff
@@ -0,0 +1,38 @@
...
+LDFLAGS="-lXext -lXft -lXrandr -lXrender -lX11 -lXinerama"
```
uggedal: I would rather sed/patch `LIBX11` etc in `config.mk` than override `LDFLAGS` directly.

srcpkgs/wmii/template
```diff
@@ -0,0 +1,38 @@
...
+short_desc="A lightweight, dynamic window manager for X11"
+maintainer="Zach L. <koni@archlinux.us>"
+license="GPL-3"
+homepage="https://code.google.com/p/wmii/"
+distfiles='https://wmii.googlecode.com/files/wmii+ixp-3.9.2.tbz'
+checksum=0466f1a5c534b0ee9fdace2c7a2d251845b7ec0ae80bdb8e68ae6584ebeebf2b
+
+do_build(){
```
uggedal: Use build style and remove this function. Move sed-ing to a `pre_build` function.

srcpkgs/wmii/template
```diff
@@ -0,0 +1,38 @@
...
+	cd ${wrksrc}
+	make clean
+	export CFLAGS="-I/usr/include/X11 -I/usr/include/X11/extensions -I/usr/include -I/usr/local/include -I/usr/include/freetype2"
+	export LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:/usr/lib/"
```
uggedal: Why?

srcpkgs/wmii/template
```diff
@@ -0,0 +1,38 @@
...
+	sed "4s#.*#PREFIX = /usr#" ${wrksrc}/config.mk
+	make PREFIX=/usr || exit
```
uggedal: No need for `|| exit`. Also, is explicit `PREFIX` needed after changing `config.mk`

srcpkgs/wmii/template
```diff
@@ -0,0 +1,38 @@
...
+homepage="https://code.google.com/p/wmii/"
+distfiles='https://wmii.googlecode.com/files/wmii+ixp-3.9.2.tbz'
+checksum=0466f1a5c534b0ee9fdace2c7a2d251845b7ec0ae80bdb8e68ae6584ebeebf2b
+
+do_build(){
+	cd ${wrksrc}
+	make clean
+	export CFLAGS="-I/usr/include/X11 -I/usr/include/X11/extensions -I/usr/include -I/usr/local/include -I/usr/include/freetype2"
+	export LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:/usr/lib/"
+	sed "4s#.*#PREFIX = /usr#" ${wrksrc}/config.mk
```
uggedal: `$PWD` is `${wrksrc}` so no need to prefix paths with it.

srcpkgs/wmii/template
dookerdo: it doesn't use a gnu style makefile

srcpkgs/wmii/template
```diff
@@ -0,0 +1,38 @@
...
+maintainer="Zach L. <koni@archlinux.us>"
+license="GPL-3"
+homepage="https://code.google.com/p/wmii/"
+distfiles='https://wmii.googlecode.com/files/wmii+ixp-3.9.2.tbz'
+checksum=0466f1a5c534b0ee9fdace2c7a2d251845b7ec0ae80bdb8e68ae6584ebeebf2b
+
+do_build(){
+	cd ${wrksrc}
+	make clean
+	export CFLAGS="-I/usr/include/X11 -I/usr/include/X11/extensions -I/usr/include -I/usr/local/include -I/usr/include/freetype2"
```
uggedal: See `LDFLAGS` note above, would be cleaner to change `config.mk` IMHO.

srcpkgs/wmii/template
uggedal: Does not matter what style it is. If it runs `make` for build and `make install` for install it will work (check `common/build-style/gnu-makefile.sh`).

dookerdo: alright, i'll change the template

srcpkgs/wmii/template
dookerdo: It's cleaner to sed into the config.mk instead of adding the LDFLAGS?

srcpkgs/wmii/template
dookerdo: It's not, you can run wmii without - it's just recommended by the project. Removed.

PR #1032, opened 2015-02-26
New package: wmii-3.9.2

dookerdo:


srcpkgs/wmii/template
```diff
@@ -0,0 +1,24 @@
+# Template file for 'wmii'
+pkgname=wmii
+version=3.9.2
+revision=1
+wrksrc=${pkgname}+ixp-${version}
+build_style=gnu-makefile
+makedepends="pkg-config libXrandr-devel libXft-devel libX11-devel libXinerama-devel"
```
xtraeme: - `pkg-config` must be in `hostmakedepends` instead.
- `libX11-devel` is redundant, it's a dependency of ther ones, remove it.

srcpkgs/wmii/template
```diff
@@ -0,0 +1,24 @@
...
+disable_parallel_build="yes"
```
xtraeme: You are using a custom `do_build()` that execs `make` without `$makejobs`... you can just remove `disable_parallel_build`.

PR #1040, opened 2015-03-01
go: files in /etc/profile.d don't need +x.

nmeum:
The `/etc/profile.d/go.sh` file had the executable bit `+x` set even though this isn't necessary for files in `/etc/profile.d`.


dominikh:
To be honest I'm not sure that file should exist at all. GOPATH is a very personal setting, and the value that file sets doesn't make much sense, anyway. All it does is make `go get`'s error less obvious to newbies, as now it'll be a permission error instead of the far more sensible "GOPATH is not set" one.

Unless there's good reason to keep that file, I'm voting to remove it.

bougyman:
The error without the path set is much more reasonable, I agree with
Dominik.

bougyman

nmeum:
alright, I will update this pull request in a second…

nmeum:
There you go.

dominikh:
Merged as 81e3ffa, thank you.

Issue #998, opened 2015-02-20
Package request: kvirc


thypon:
KVirc is an old stile (mirc like) irc tiling client written in Qt. Almost nearly KDE independ, only qt dependency.

PR #1041, opened 2015-03-02
New package: kvirc-4.2.0

thypon:
#998

### DIFF
--- a/srcpkgs/kvirc/template
+++ b/srcpkgs/kvirc/template
@@ -1,11 +1,10 @@
 # Template file for 'kvirc'
 pkgname=kvirc
 version=4.2.0
-revision=1
+revision=2
 build_style=cmake
-depends="python<3.0.0"
 hostmakedepends="cmake"
-makedepends="zlib-devel qt-devel"
+makedepends="zlib-devel qt-devel<5.0.0 libressl-devel python-devel<3.0.0 kdelibs-devel<5.0.0 perl"
 short_desc="Qt-based IRC client"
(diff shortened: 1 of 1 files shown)


thypon:
#998

srcpkgs/kvirc/template
```diff
@@ -0,0 +1,14 @@
+# Template file for 'kvirc'
+pkgname=kvirc
+version=4.2.0
+revision=1
+build_style=cmake
+depends="python<3.0.0"
```
Gottox: python points to python2.7 at the moment. Depend on the python3 package instead.

thypon: Python2 is an optdepends for plugins. 

There is a misunderstanding. KVirc depends on python2 (i put <3.0.0 meaning all that is less thant 3.0.0, then the opposite of python3).

http://www.kvirc.net/?id=install here the install guide. Maybe i could move python python,perl,ada,c,c++,fortran,java,objc,lto to optdepends.

Gottox: Ah, missread that. Keep in mind, that python is not installed at buildtime, therefore kvirc will be build without python support:
```
-- Cryptography:
--    Crypt Support And Engines   : Yes
--    OpenSSL Support             : Not available
-- Scripting:
--    Perl Support                : Yes, used for scripting+doc
--    Python Support              : No, not found.
```

xtraeme:
In that case you should also add `libressl-devel` to `makedepends`.

thypon:
Doesn't perl-devel exists in repositories?

PR #1056, opened 2015-03-04
bash: add -DSYS_BASHRC to CFLAGS

lemmi:
this enables bash to source a system-wide bashrc.
currently set to /etc/bash/bashrc.

### DIFF
--- a/srcpkgs/bash/template
+++ b/srcpkgs/bash/template
@@ -3,7 +3,7 @@ pkgname=bash
 _bash_distver=4.3
 _bash_patchlevel=033
 version=${_bash_distver}.${_bash_patchlevel}
-revision=2
+revision=3
 wrksrc=${pkgname}-${_bash_distver}
 build_pie=yes
 build_style=gnu-configure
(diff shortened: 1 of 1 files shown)


xtraeme:
Any reason to use `/etc/bash/bashrc` rather than just ` /etc/bashrc` ?

lemmi:
yes. i'd like to ship a default bashrc in that folder that also sources a bashrc.local at the end (or like with gentoo have another /etc/bash/bashrc.d with further scripts). the default should be as general as possible and updated by the distribution. users can have a their configuration outside the package management. keeping it in a dedicated folder keeps /etc cleaner and makes it more obvious how to keep it tidy.

srcpkgs/bash/template
```diff
@@ -18,6 +18,7 @@ homepage="http://www.gnu.org/software/bash/bash.html"
 license="GPL-3"
 distfiles="http://ftp.gnu.org/gnu/bash/bash-${_bash_distver}.tar.gz"
 checksum=afc687a28e0e24dc21b988fa159ff9dbcf6b7caa92ade8645cc6d5605cd024d4
+CFLAGS="${CFLAGS} -DSYS_BASHRC='\"/etc/bash/bashrc\"'"
```
xtraeme: There's no need to append CFLAGS to CFLAGS, it's enough to use:

`CFLAGS="-DSYS_BASHRC='\"/etc/bash/bashrc\"'"`

Please fix.

PR #1067, opened 2015-03-06
build-style: added go version in git

thypon:


dominikh:
When using the `go` build style, one should either be building from a tarball (and the `go get` step is only there to fetch dependencies), or build from master/tip/trunk. In both cases, the step you're adding is either useless (in the case of a tarball) or broken (in the case of a version such as "20150302", which is used for packages that are built straight from master.

thypon:
I missed something. Why for example hub is marked as non-git 2.2.0, while the build script downloads and build from master.

It's a common pattern in git binaries to use v${version} in tags to represent a version. I think putting "|| true" after "git checkout" could be a compromise.

dominikh:
`hub` isn't built from master, it's built from https://github.com/github/hub/archive/v${version}.tar.gz – `go get` is only used to download the dependencies of `hub`, which aren't versioned.

If a software has proper releases, it will be built from these releases, which usually means tarballs. `go get` isn't involved in that step.

If a software has no proper releases, packages are built from master in irregular intervals at the discretion of the package maintainer, in which case there's no `git checkout` needed. 

In neither case should the `go` build style need to check out any revision other than master (or the Go version tags, that `go get` itself takes care of.)

thypon:
Agree

PR #1068, opened 2015-03-06
New package: daemonize-1.7.5

thypon:
-rw-r--r-- root/root      1032 2015-03-06 18:54 ./props.plist
    -rw-r--r-- root/root      1107 2015-03-06 18:54 ./files.plist
    -rw-r--r-- 0/0            6095 2015-03-06 18:54 ./usr/share/man/man1/daemonize.1
    -rwxr-xr-x 0/0           13864 2015-03-06 18:54 ./usr/sbin/daemonize

### DIFF
--- a/srcpkgs/daemonize/template
+++ b/srcpkgs/daemonize/template
@@ -1,0 +1,17 @@
+# Template file for 'daemonize'
+pkgname=daemonize
+version=1.7.5
+revision=1
+build_style=gnu-configure
+hostmakedepends="git"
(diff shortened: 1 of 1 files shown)


xtraeme:
The website mentions that it's BSD licensed, is that correct?

thypon:
Yes and no.

It contains getopt.c that is in Public Domain and install.sh on MIT
https://github.com/bmc/daemonize/blob/master/LICENSE.md

PR #1078, opened 2015-03-07
New package: unison-2.48.3

allan:


xtraeme:
- Please use the common style for templates
- If it's a binary generated by ocaml it should not be stripped (nostrip=yes)

allan:
* The packaging guidelines of our favorite distribution `Fedora` say that OCaml binaries should be striped.
 * What is not common in this template, is it `CFLAGS= make...`?

xtraeme:
- I don't remember exactly but I think ocaml binaries did not work if they were stripped, that info could be outdated. If you've verified it behaves correctly I'll trust you.
- What I mean is `func() { ... }` rather than:

```
do_build() {
 ...
}
```

xtraeme:
no manual page?

allan:
here you go.  There has been some issues in the past where compiled with `ocaml -custom` which could not be stripped.
No man page.  The doc is built into the binary. Something like `unison -doc tutorial` and stuff.

xtraeme:
Perfect, thanks.

Issue #1035, opened 2015-02-27
Package Request: virt-manager


pulux:
The virt-manager application is a desktop user interface for managing virtual machines through libvirt. It primarily targets KVM VMs, but also manages Xen and LXC (linux containers).

Homepage: http://virt-manager.org/
Code: http://virt-manager.org/download/sources/virt-manager/virt-manager-1.1.0.tar.gz

PR #1079, opened 2015-03-08
Virt manager

thypon:
#1035


xtraeme:
I guess you are missing some `noarch=yes` on them.

xtraeme:
and `pycompile_module` as well.

xtraeme:
New package must set revision to 1.

Issue #1071, opened 2015-03-06
Request: lxc runit support


thypon:
Now only supports systemd service.

bougyman:
What do you propose? Lxc cannot be directly supervised/run in the foreground, it's not a fit for runit supervision. I'll take a look into the systemd service and see if I can come up with something.

thypon:
lxc supports an lxc-autostart directive supposed to work with the respective init daemon.

thypon:
the problem with lxc-autostart is that is a forking program. How to track and supervise it?

xtraeme:
Something like this should be really enough:

```
lxc-autostart/run:
#!/bin/sh
lxc-autostart || exit 1
exec chpst -b lxc-autostart pause
```

```
lxc-autostart/finish:
#!/bin/sh
lxc-autostart -s 15 || exit 1
```

PR #1080, opened 2015-03-08
lxc: added runit support

thypon:
#1071


chneukirchen:
Should be lxc-autostart -s -t 15?

PR #1083, opened 2015-03-08
New package: kdeutils-ark-14.12.3

thypon:


srcpkgs/kdeutils-ark/template
```diff
@@ -0,0 +1,13 @@
+# Template file for 'ark'
```
Gottox: s/ark/kdeutils-ark/

Gottox:
:+1:

PR #1087, opened 2015-03-09
mpv: revbump: rebuild against new ffmpeg

lemmi:
Warning: mpv was compiled against a different version of ffmpeg than the shared library it is linked against.


xtraeme:
mpv should work fine if the shared library is ABI compatible, this warning is simply noise.

lemmi:
yes it works, but the rebuild fixes the noise :)

PR #1092, opened 2015-03-10
New package: discount-2.1.8

necrophcodr:
This package adds the "much-needed" C implemented markdown parser for all your shell-based markdown parsing needs.

http://www.pell.portland.or.us/~orc/Code/discount/

### DIFF
--- a/srcpkgs/discount/template
+++ b/srcpkgs/discount/template
@@ -1,0 +1,21 @@
+# Template file for 'discount'
+pkgname=discount
+version=2.1.8
+revision=1
+build_style=configure
+configure_script="./configure.sh"
+configure_args="--prefix=/usr"
+short_desc="C implementation of John Gruber's Markdown"
+maintainer="necrophcodr <necrophcodr@necrophcodr.me>"
+license="3-clause-BSD"
+homepage="http://www.pell.portland.or.us/~orc/Code/discount"
+distfiles="${homepage}/${pkgname}-${version}.tar.gz"
+checksum=db9742460c4f33cf417ef794c8e41605f8fc498697c181aa2835f6944aba7137
+
+discount-devel_package() {
+	short_desc+=" - development files"
+	pkg_install() {
+		vmove usr/include
+		vmove usr/lib
+	}
+}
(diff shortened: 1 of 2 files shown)


srcpkgs/discount/template
```diff
@@ -0,0 +1,13 @@
+# Template file for 'discount'
+pkgname=discount
+version=2.1.8
+revision=1
+build_style=configure
+configure_script="./configure.sh"
+configure_args="--prefix=/usr"
+short_desc="C implementation o John Gruber's Markdown"
```
xtraeme: o -> of ?

Looks good, thanks.

necrophcodr:
Do not merge yet, installs development files too. Fuck. Fixing.

necrophcodr:
Alright, there we go. Revise, comment, so on so forth. Got a bit too fast on the previous commit, so it's a good thing I don't just write directly to the main master branch ever.

srcpkgs/discount/template
```diff
@@ -0,0 +1,22 @@
...
+homepage="http://www.pell.portland.or.us/~orc/Code/discount"
+distfiles="${homepage}/${pkgname}-${version}.tar.gz"
+checksum=db9742460c4f33cf417ef794c8e41605f8fc498697c181aa2835f6944aba7137
+
+discount-devel_package() {
+	depends="${pkgname}>=${version}_${revision}"
+	short_desc+=" - development files"
+	pkg_install() {
+		vmove usr/include
+		vmove usr/lib
```
uggedal: Does this package provide a shared library? If so it should be part of the main package. Should also be added to common/shlibs.

necrophcodr:
Only provides a static library as far as I can tell, so no shared libraries.
In reality, only 3 files are installed.

    /usr/include/mkdio.h
    /usr/lib/libmarkdown.a
    /usr/bin/markdown

Feel free to double check anyway.

xtraeme:
If this pkg does not provide a shared library there's no reason to make the `-devel` pkg depend on it, i.e remove the `depends` line from `discount-devel_package()`.

PR #1098, opened 2015-03-11
ncurses: Generate pkg-config files for devel package

teh-jazzman:


xtraeme:
What's the reason to close this? seemed ok.

teh-jazzman:
I noticed that the .pc files are only created if pkg-config is installed and the directory /usr/lib/pkgconfig exists. Am I correct in assuming the first is somehow problematic since ncurses is a bootstrap package?

teh-jazzman:
It is building correctly now, but I am not sure if putting pkg-config as a dependency is alright

teh-jazzman:
Dammit, sorry. Apparently I somehow managed to clutter the pull request with old stuff again. I will make a clean one when I know if this is viable.

xtraeme:
Indeed, adding a pkg-config dependency is not a good idea due to base-chroot.

teh-jazzman:
Turns out there is a simple solution. I'll run some tests and then create a new (clean) pull request. Thanks.

PR #1099, opened 2015-03-11
updating poppler to 0.32

Gottox:


xtraeme:
You did not update the checksum in -qt[45].

Other than that looks ok :+1:

PR #1101, opened 2015-03-11
New package: filteraudio-git-20150310_1

teh-jazzman:
Audio filter library for use with several TOX clients.

### DIFF
--- a/srcpkgs/filteraudio-git/template
+++ b/srcpkgs/filteraudio-git/template
@@ -1,0 +1,26 @@
+# Template file for 'filteraudio-git'
+pkgname="filteraudio-git"
+version="20150313"
+revision=1
+short_desc="Lightweight audio filtering library for TOX"
+maintainer="Stefan Mühlinghaus <jazzman@alphabreed.com>"
+license="3-clause-BSD"
+homepage="https://tox.im"
+hostmakedepends="git"
+build_style="gnu-makefile"
+
+do_fetch() {
+	git clone https://github.com/irungentoo/filter_audio.git ${wrksrc}
+}
+
+# Development package
+filteraudio-devel-git_package() {
+	short_desc+=" - development files"
+	depends="${sourcepkg}>=${version}_${revision}"
+	pkg_install() {
+		vmove usr/include
+		vmove usr/lib/*.a
+		vmove usr/lib/pkgconfig
+	}
+}
+
--- a/common/shlibs
+++ b/common/shlibs
@@ -1960,3 +1960,4 @@ libusbredirhost.so.1 usbredir-0.7_1
 libprotobuf-c.so.1 protobuf-c-1.1.0_1
 libutf8proc.so.1 libutf8proc-1.1.6_1
 libnsutils.so.0 libnsutils-0.0.1_1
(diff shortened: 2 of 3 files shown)


srcpkgs/filteraudio-git/template
```diff
@@ -0,0 +1,27 @@
...
+hostmakedepends="git"
+build_style="gnu-makefile"
+wrksrc="filter_audio"
+
+do_fetch() {
+	git clone https://github.com/irungentoo/filter_audio.git
+}
+
+# Development package
+filteraudio-git-devel_package() {
```
Gottox: please rename the subpackage to filteraudio-devel-git.

srcpkgs/filteraudio-git/template
```diff
@@ -0,0 +1,27 @@
...
+short_desc="Lightweight audio filtering library for TOX"
+maintainer="Stefan Mühlinghaus <jazzman@alphabreed.com>"
+license="BSD 3-clause"
+homepage="https://tox.im"
+hostmakedepends="git"
+build_style="gnu-makefile"
+wrksrc="filter_audio"
+
+do_fetch() {
+	git clone https://github.com/irungentoo/filter_audio.git
```
xtraeme: Use this:

```
git clone https://github.com/irungentoo/filter_audio.git ${wrksrc}
```

and then you can remove `wrksrc`.

srcpkgs/filteraudio-git/template
```diff
@@ -0,0 +1,27 @@
+# Template file for 'filteraudio-git'
+pkgname="filteraudio-git"
+version="20150313"
+revision=1
+short_desc="Lightweight audio filtering library for TOX"
+maintainer="Stefan Mühlinghaus <jazzman@alphabreed.com>"
+license="BSD 3-clause"
```
xtraeme: 3-clause-BSD instead.

xtraeme:
Any reason to pkg this even if it's not required by tox? I'd like to avoid adding more -git pkgs if possible.

teh-jazzman:
It is required by two TOX clients. I'm not too comfortable with the git-packages myself, but they are not doing proper releases or even proper version numbers. The three TOX clients have the same problem by the way.

Alternatively I could take the precompiled binaries for the clients from https://wiki.tox.im/Binaries and package those. They have toxcore and filteraudio already statically linked in and we would not need the library packages. They only provide two of the three TOX clients available for Linux though and do not say on the website what the current versions are.

teh-jazzman:
Another option would be to put the filteraudio library in the toxcore-git package. That way there would only be one package for the TOX-specific dependencies. As far as I know it is not used for any other software anyway, but of course I cannot be sure about that.

Gottox:
please solve the merge conflicts, will merge afterwards

teh-jazzman:
The conflicts are in common/shlibs. To properly resolve them I would need to merge with the master branch, which by now is quite a bit ahead of this branch. Any tipps on how to do that without spamming hundrets of commits into this pull request?

teh-jazzman:
Nevermind, I got it to work. But git and I will never be friends...

Gottox:
:smile: I learned to love it

PR #1105, opened 2015-03-12
base-system-busybox: enable ps, pgrep and pkill

nmeum:
This enables the `ps`, `pgrep` and `pkill` busybox applets, therefore busybox now conflicts with `procps-ng`. This wasn't possible before, because `runit-void` depended on `procps-ng`, however, when voidlinux/void-runit#19 gets merged then `runit-void` no longer depends on `procps-ng`.

**Warning:** Before merging this the `runit-void` package needs to be updated.


srcpkgs/base-system-busybox/files/dotconfig
```diff
@@ -50,7 +50,7 @@ CONFIG_FEATURE_SUID_CONFIG_QUIET=y
 # CONFIG_FEATURE_PREFER_APPLETS is not set
 CONFIG_BUSYBOX_EXEC_PATH="/proc/self/exe"
 CONFIG_FEATURE_SYSLOG=y
-CONFIG_FEATURE_HAVE_RPC=y
```
nmeum: No idea why this was modified by `make menuconfig` as well…

xtraeme:
**runit-void** depends on **procps-ng** exactly for **pkill** and **sysctl**.

If you read the comments in the *base-system-busybox* pkg you'll find that I explicitely decided to depend on *procps-ng* exactly to avoid adding such conditionals.

nmeum:
I interpreted this as a TODO but well…fair enough…

PR #1114, opened 2015-03-13
ncurses: generate .pc files for devel package without pkg-config dependency

teh-jazzman:


srcpkgs/ncurses/template
```diff
@@ -24,16 +24,20 @@ do_configure() {
 	# Override config.sub with ours.
 	cp -f ${XBPS_CROSSPFDIR}/config.sub ${wrksrc}
 
+	# Make sure .pc files are generated
+	export PKG_CONFIG=/usr/bin/pkg-config
```
xtraeme: Rather than exporting the var, pass it to the configure script instead.

teh-jazzman:
I wonder if I will ever learn to properly use git/github *sigh*
I will make a clean pull request if this is OK in principle.

xtraeme:
This pull request is broken, re-submit a proper one, thanks.

PR #1116, opened 2015-03-13
msp430

allan:


xtraeme:
I'm not sure if `$XBPS_BUILDDIR` will be cleaned up after building this package... use `create_wrksrc=yes` instead.

srcpkgs/msp430-toolchain/template
```diff
@@ -0,0 +1,71 @@
...
+homepage="http://www.ti.com/tool/msp430-gcc-opensource"
+license="GPL-2"
+distfiles="http://software-dl.ti.com/msp430/msp430_public_sw/mcu/msp430/MSPGCC/latest/exports/msp430-gcc-source.tar.bz2
+	   http://software-dl.ti.com/msp430/msp430_public_sw/mcu/msp430/MSPGCC/latest/exports/msp430-gcc-support-files.zip"
+checksum="67179535e68951bfb82f8bb1bf099a716e921f11b1ad773c87a08b6054ed31b0
+	  303b743fae597c8861c6bd85bb52d9b4ad51d6537bbc3f010b2e09c6a624651e"
+nocross=yes
+nostrip=yes
+create_wrksrc=yes
+hostmakedepends="unzip expect libgcc-32bit flex"
```
xtraeme: Just curiosity why libgcc-32bit is required?

allan: after trial-and-erroring for hours this helped at some point.  but fortunately it is not needed anymore. let me push..

srcpkgs/msp430-toolchain/template
```diff
@@ -0,0 +1,70 @@
+# Template file for 'msp430-toolchain'
+pkgname=msp430-toolchain
+version=3_02_03_00
```
xtraeme: Please replace underscores with dots.

We've recently changed xbps-src to forbid underscores and dashes in version.

PR #1117, opened 2015-03-13
runit-void: Explicitly depend on procps-ng and awk

nmeum:
See: https://github.com/voidlinux/void-runit#dependencies


xtraeme:
Holding this up until I fix virtual packages as dependencies in xbps-src.

PR #1118, opened 2015-03-14
New package: couchdb-1.6.1

arete:


srcpkgs/couchdb/template
```diff
@@ -0,0 +1,30 @@
+# Template file for 'couchdb'
+pkgname=couchdb
+version=1.6.1
+revision=1
+wrksrc="apache-couchdb-$version"
+build_style=gnu-configure
+configure_args="--prefix=/usr --sysconfdir=/etc --localstatedir=/var "
```
xtraeme: This line is redundant, those setting are set by default.

srcpkgs/couchdb/template
```diff
@@ -0,0 +1,30 @@
...
+conf_files="/etc/couchdb/default.ini /etc/couchdb/local.ini"
+couchdb_homedir="/var/lib/couchdb"
+hostmakedepends="erlang"
+makedepends="libressl-devel>=2.1.4  icu-devel js-devel"
+depends="erlang icu libressl js"
+short_desc="A document-oriented database"
+maintainer="Gerardo Di Iorio <arete74@gmail.com>"
+license="APACHE"
```
xtraeme: Apache-2.0

srcpkgs/couchdb/template
```diff
@@ -0,0 +1,30 @@
...
+version=1.6.1
+revision=1
+wrksrc="apache-couchdb-$version"
+build_style=gnu-configure
+configure_args="--prefix=/usr --sysconfdir=/etc --localstatedir=/var "
+conf_files="/etc/couchdb/default.ini /etc/couchdb/local.ini"
+couchdb_homedir="/var/lib/couchdb"
+hostmakedepends="erlang"
+makedepends="libressl-devel>=2.1.4  icu-devel js-devel"
+depends="erlang icu libressl js"
```
xtraeme: This depends line I guess is also wrong, dependencies should be guessed automagically if the binaries are dynamically linked.

srcpkgs/couchdb/INSTALL
```diff
@@ -0,0 +1,8 @@
+case "$ACTION" in
+        post)
+                if [ "$UPDATE" = "yes" ]; then
```
xtraeme: You should remove this if conditional, it makes no sense to run it only when the pkg is being updated.

arete:
ok i have update the files

Issue #1089, opened 2015-03-09
Package request: schroot


thypon:

bougyman:
?

thypon:
https://wiki.debian.org/Schroot "Schroot allows users to execute commands or interactive shells in different chroots"

PR #1135, opened 2015-03-14
New package: schroot-1.6.10

thypon:
#1089


srcpkgs/schroot/template
```diff
@@ -0,0 +1,72 @@
+# Template file for 'schroot'
+pkgname=schroot
+version=1.6.10
+revision=1
+build_style=gnu-configure
+configure_args="
+	--prefix=/usr
```
xtraeme: `--prefix --libexecdir --sysconfdir and --localstatedir` are already defined by the build style `gnu-configure`, remove them.

PR #1151, opened 2015-03-17
eclipse-ecj: depend on java-runtime

nmeum:


srcpkgs/eclipse-ecj/template
```diff
@@ -1,9 +1,10 @@
 # Template build file for 'eclipse-ecj'.
 pkgname=eclipse-ecj
 version=4.9
-revision=1
+revision=2
 noarch=yes
 build_style=fetch
+depend="virtual?java-runtime"
```
xtraeme: depend -> depends

nmeum:
Fixed that typo.

PR #1153, opened 2015-03-17
Fix two java-runtime dependcies

nmeum:


nmeum:
The fact that the lightzone-bin test fails is not really my fault as you might see…should I fix this nonetheless?

PR #1155, opened 2015-03-17
base-system-busybox: update to 1.23.1

nmeum:


nmeum:
Don't blame me, blame `make oldconfig`.

bougyman:
I thought it might be something like that. I like the added vi undo queue but i'm biased.

PR #1156, opened 2015-03-17
New package: fping-3.10

nmeum:
See: http://fping.org/.


nmeum:
Turns out: This is not under BSD. It looked BSDish, but it is licensed under a custom license. Since that means that it isn't free as in freedom because that license isn't approved by the OSI or by the FSF. I don't know how to handle this,, any hints?

bougyman:
non-free seems good.

nmeum:
Done. I wasn't sure what to call the license so I just decided to call it 'fping'. Should I install that license using `vlicense`?

xtraeme:
The license used in fping is a MIT-alike one, so that it's completely free.

PR #1161, opened 2015-03-18
xinit: log to /dev/stdout

nmeum:


nmeum:
Hm, actually this doesn't seem to work as expected…

PR #1166, opened 2015-03-19
x264: update to 20141218.2245.

lemmi:
disable asm only for armv6hr target, enable opencl (needs perl), update to latest stable.


Gottox:
build fails for both armv6 and armv7 for me.

lemmi:
ok, i assumed crosscompilation was dead for me. maybe this works only on x86 then. i'll give another shot on the weekend.

xtraeme:
Looks ok at first glance :+1:

PR #1168, opened 2015-03-19
New package: monkey-1.5.6

necrophcodr:
This adds a new HTTP embeddable web server, with support for all Void Linux architectures out of the box.
Additionally it supports IPv6, SSL, CGI and FastCGI, and more.

http://monkey-project.com/


srcpkgs/monkey/template
```diff
@@ -0,0 +1,23 @@
+# Template file for 'monkey'
+pkgname=monkey
+version=1.5.6
+revision=1
+build_style=configure
+configure_args+="--prefix=/usr --plugdir=/usr/share/monkey/plugins "
+configure_args+="--sysconfdir=/usr/share/monkey/conf "
```
xtraeme: `--sysconfdir` should be set to `/etc`, this is where configuration files are expected to be stored.

necrophcodr:
@xtraeme 
Then it should be /etc/monkey probably. The project installs configuration files, just like apache, in the sysconfdir, so yeah. I was a bit split on the file paths though, but that does seem like a better choice yeah. Feel free to comment further, so I can make any required changes.

xtraeme:
- mandir is wrong, --mandir=/usr/share/man
- ssl does not seem to be enabled anywhere
- there's no runit service
- some dirs should be created via `make_dirs`

and perhaps more :-)

necrophcodr:
It is true that there's no runit service. I'll look into that. As for the SSL, I might've overlooked something there as it may actually be a plugin. I'll look into all those, and address them as I can.

Thanks!

necrophcodr:
I was also missing the pid file. It seems that it was configured to be placed in /builddir/monkey-1.5.6/logs/monkey.pid.

Not very useful. That is also fixed now. I'll test the packaging of it and so on, and see if it works, then re-push the commit.

xtraeme:
pidfile is irrelevant with runit because it's expected that the application is running in non forking/daemon mode.

necrophcodr:
It is, but monkey may do some shit with it anyway. I can't disable it easily, so I fixed it instead. Runit can do what it does.

Monkey exports some pidfile that it makes use of for.. Something. I'm not sure what exactly. But it's fixed now. And it's pushed. Comment and critique, as I may have done terrible shit, but we'll see. I decided to look at some stuff Apache did, so this is kind of based on some of that stuff for now.

srcpkgs/monkey/files/monkey/run
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+exec chpst -uhttp /usr/bin/monkey
```
xtraeme: - no need to hardcode the path, use simply `exec chpst -u httpd monkey`

srcpkgs/monkey/template
```diff
@@ -0,0 +1,38 @@
+# Template file for 'monkey'
+pkgname=monkey
+version=1.5.6
+revision=1
+build_style=configure
+makedepends="mbedtls-devel"
+depends="mbedtls"
```
xtraeme: This depends line is wrong... if monkey links to mbedtls the dependency will added automagically by xbps-src.

srcpkgs/monkey/template
```diff
@@ -0,0 +1,38 @@
...
+configure_args+="--sysconfdir=/etc/monkey/ --enable-plugins=mbedtls "
+configure_args+="--datadir=/srv/httpd --mandir=/usr/share/man "
+configure_args+="--pidfile=/var/run/monkey.pid"
+short_desc="Cross-arch embeddable lightweight HTTP server"
+maintainer="necrophcodr <necrophcodr@necrophcodr.me>"
+license="APL-2.0"
+homepage="http://monkey-project.com"
+distfiles="${homepage}/releases/1.5/${pkgname}-${version}.tar.gz"
+checksum=7c3d845306aa74ee6effd7ab6169d16ac4e6450e564954d0d0baa2d1e9be1a22
+conflicts="apache>=0"
```
xtraeme: Seriously this conflict does not make any sense :-)

srcpkgs/monkey/files/monkey/run
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+exec chpst -uhttpd /usr/bin/monkey
```
xtraeme: s,/usr/bin,,g

srcpkgs/monkey/template
```diff
@@ -0,0 +1,38 @@
+# Template file for 'monkey'
+pkgname=monkey
+version=1.5.6
+revision=1
+build_style=configure
+makedepends="mbedtls-devel"
+configure_args+="--prefix=/usr --plugdir=/usr/share/monkey/plugins "
+configure_args+="--sysconfdir=/etc/monkey/ --enable-plugins=mbedtls "
+configure_args+="--datadir=/srv/httpd --mandir=/usr/share/man "
```
xtraeme: Are you sure that datadir is set as www rootdir?

necrophcodr:
@xtraeme According to the configuration lines printed when it executes, yes, datadir is originally path/to/code/htdocs, so it should probably be /srv/httpd when changed. Datadir is www rootdir.

I have a user local installation of monkey, and it reads the correct directories as well when the switch is used. If it is ALSO used for something else, I do not know.

xtraeme:
ok, will take a look at this soon.

Thanks :+1:

PR #1183, opened 2015-03-22
plan9port: move to nonfree repository

nmeum:
Since it is licensed under a custom license it is not free.


chneukirchen:
The code is freely licensed (OSI/DSFG approved), and the restriction on the bitmap fonts is questionable at best. I see no issue here.

Issue #1188, opened 2015-03-22
redshift: default deps


pulux:
redshift has in template huge deps:

`"GConf-devel libXxf86vm-devel pygtk-devel python-xdg geoclue-devel>=0.12.0_7"`

Who need GeoClue and GConf in Default, redshift needs only `libXxf86vm-devel` to build others are optional.

What is the best way to provide the Minimal-Version?

xtraeme:
Split the UI into a -gtk subpackage.

pulux:
see #1195

PR #1192, opened 2015-03-23
redshift: update to 1.9.1 and split GUI

pulux:


Gottox:
Can you explain why splitting those packages makes sense?

pulux:
so we have a small working app without much deps, and a full featured version with GConf, geoclue, ...
se #1188

Gottox:
I'm not sure about that. @xtraeme mentioned, that it should be split into a -gtk subpackage, this *was* already the case.

Different sets of dependencies are rather done with build_options than through splitting the package.

I don't think this should be merged.

xtraeme:
I don't see any benefit neither :-)

I thought the pkg did not have a -gtk subpackage, but this is not the case.

pulux:
ok

PR #1195, opened 2015-03-24
redshift: update to 1.10 and add build_options=geoclue

pulux:


srcpkgs/redshift/template
```diff
@@ -1,12 +1,12 @@
...
-revision=2
+version=1.10
+revision=1
 build_style=gnu-configure
-configure_args="--enable-geoclue"
-hostmakedepends="pkg-config automake libtool gettext-devel git"
-makedepends="GConf-devel libXxf86vm-devel pygtk-devel python-xdg geoclue-devel>=0.12.0_7"
-short_desc="Adjusts the color temperature of your screen according to your surroundings"
+hostmakedepends="pkg-config automake libtool gettext-devel git intltool"
+configure_args="--enable-gui $(vopt_if geoclue --enable-geoclue)"
```
Gottox: You can use `vopt_enable geoclue` here

srcpkgs/redshift/template
```diff
@@ -1,12 +1,12 @@
...
+makedepends="libXxf86vm-devel $(vopt_if geoclue geoclue-devel) $(vopt_if geoclue GConf-devel) python3.4 pygtk-devel python-xdg"
```
Gottox: Can be simplyfied to `vopt_if geoclue "geoclue-devel GConf-devel"`

PR #1196, opened 2015-03-25
New package: toxic-0.5.2_1

teh-jazzman:
Changed toxcore-git as well to pick a reliable commit hash which can be linked against toxic.

Am I correct in assumuming that when I have a (more or less) fixed version number and I pick a corresponding commit hash I can refrain from adding the "-git" suffix to the package name even though the sources are pulled from git?


Gottox:
You can refrain the -git suffix if it is a release.

Personally, I'm not a friend of accessing the sources via git at all as it breaks parts of the security model (at least if you rely on tags or branches) and adds an extra dependency to the build.

For the -git packages it's fine, but releases should be downloaded via http. For the release use the links provided on the release page: https://github.com/Tox/toxic/releases

teh-jazzman:
Nope, I tried those. They are broken :)
Same goes for the tags. I needed to cherry-pick a working commit hash. I hope that will change in future releases. When it does I'll gladly change the code source to use the releases.

Gottox:
That sounds very odd...

teh-jazzman:
As far as I can tell the developement process at the moment ist quite chaotic and very much in flux. It boils down to this:

* Not every toxcore (the library) commit can be compiled
* Few current toxic (the client) commits work with current toxcore commits
* The toxic releases and tags are (at least) too old and hence incompatible
* Old toxcore commits may be compatible with the toxic releases (haven't tested) but then would probably not be compatibe with other clients
* The commits I picked for toxcore and toxic work together and with at least one other client as well

teh-jazzman:
Changed the way github sources are downloaded as proposed in #1198, removing the git dependency and allowing to check the downloaded files via sha256 checksum.
Also updated the common/shlibs relation for the toxcore libraries.

srcpkgs/toxic/template
```diff
@@ -0,0 +1,19 @@
...
+pkgname="toxic"
+version="0.5.2"
+revision=1
+_commithash="0c39e7b158ff74a2ef4ed62a72b5b7a3c878ac7f"
+short_desc="NCurses-based TOX instant messanger client"
+maintainer="Stefan Mühlinghaus <jazzman@alphabreed.com>"
+license="GPL-3"
+homepage="https://wiki.tox.im/Toxic"
+depends="toxcore-git>=20150313_2"
+makedepends="toxcore-devel-git>=20150313_2 libX11-devel freealut-devel libconfig-devel ncurses-libs ncurses-devel>=5.9_10 libopenal-devel libnotify-devel libvpx-devel libsodium-devel opus-devel"
```
xtraeme: ncurses-libs is redundant

srcpkgs/toxic/template
```diff
@@ -0,0 +1,19 @@
...
+pkgname="toxic"
+version="0.5.2"
+revision=1
+_commithash="0c39e7b158ff74a2ef4ed62a72b5b7a3c878ac7f"
+short_desc="NCurses-based TOX instant messanger client"
+maintainer="Stefan Mühlinghaus <jazzman@alphabreed.com>"
+license="GPL-3"
+homepage="https://wiki.tox.im/Toxic"
+depends="toxcore-git>=20150313_2"
+makedepends="toxcore-devel-git>=20150313_2 libX11-devel freealut-devel libconfig-devel ncurses-devel>=5.9_10 libopenal-devel libnotify-devel libvpx-devel libsodium-devel opus-devel"
```
xtraeme: version comparators in `hostmakedepends` and `makedepends` are not supported anymore.

See https://github.com/voidlinux/void-packages/commit/1c5f4690e682941ec4d2e247a831ba881adb2993

PR #1197, opened 2015-03-25
base-system-busybox update and misc

nmeum:
Update busybox and try to fix a few 'compatibility issues' with base-system.


xtraeme:
If you submit an update to a new package revision must be set to 1, not 5 nor 22.

I'll accept it this time but please take this into account for the next PRs.

Thanks.

nmeum:
[Originally](https://github.com/nmeum/void-packages/commit/9bb6f198e2091b5527e130230264f965abc2191f) I set the revision to 1. I just bumped it for every other commit I made. Isn't it more correct to do it this way?

Issue #1198, opened 2015-03-25
Verify files checked out from git


nmeum:
Some packages fetch their source code from git. For example: atom, runit-void, socklog-void, 2bwm-git, etc. The problem with that is that the source code isn't verified at all. This makes it possible for an attacker to easily compromise the source code used to built the package and thus compromising the resulting binaries. If voidlinux starts getting more users the built server might become a juicy target and nobody would probably notice if some packages where built from compromised git sources.

Furthermore. the version numbers for git packages don't indicate which commits are included in the package since they are mostly just date timestamps (e.g void-runit) therefore package builds are not really reproducible either. 

That's why I think that it is a very bad idea to ship packages built directly from git. They should be removed and built from tarballs instead. If a tarball isn't available because upstream hasn't released one yet than the package shouldn't be included until a release has been made.

Of cause this isn't a major security issue but always checking out a specific commit might be a good idea.

thypon:
Could you explain how the attack could work? The git history is signed and is simple to say if their integrity is preserved or not and the repository that contains each package is supposed to be trusted.

nmeum:
For example: If somebody does a man in the middle attack on the buildserver and hands over a modified git repo to the client than that wouldn't be noticed in any way, right? Because after cloning the repository no additional checks regarding the integrity of the checkout are performed. At least this is my current viewpoint. Of cause my suggestion to remove all packages built from git is kind of provocative but I simply wanted to raise awareness for these two issues.

> The git history is signed

To my knowledge the git history isn't cryptographically signed. It is possible to sign commits with GNUPG but that's completely optional and not done by most repositories.

thypon:
When i say the git history is signed i don't think about cryptographic signature but hash signature.
Downloading a binary tar.gz + sha256 =~ download git repo + commit hash. Right now i have personally used tags that are commit hash aliases but it's also quite simple to use the commit hash to verify that the tag alias point to the same thing.

nmeum:
The templates don't even look at the commit hashes currently.

logenkain:
> The templates don't even look at the commit hashes currently.

I suppose, if we have a place to store it, we could take the source ball
of a snapshot and save it on Void servers somewhere.  Use the sha256
from that and be done with it.  I believe I have seen some arch packages
done this way.

Though I agree, even if only for cosmetic reasons, we should avoid having git
packages in the main repos.  Would be better if we made our own "releases" and
keep them on void servers until we need to update them or the dev
[...]

PR #1210, opened 2015-03-27
redshift: Fetch tarballs.

nmeum:
Build redshift from tarballs.

See also: #1198


srcpkgs/redshift/template
```diff
@@ -1,19 +1,17 @@
 # Template file for 'redshift'
 pkgname=redshift
 version=1.10
-revision=1
+revision=2
```
Gottox: no need to revbump here too

PR #1213, opened 2015-03-27
libstrophe: remove git make dependency

nmeum:
The initial libstrophe package used git to fetch the sources. The current version doesn't so git isn't required as a hostmakedependency anymore.


srcpkgs/libstrophe/template
```diff
@@ -1,9 +1,9 @@
 # Template file for 'libstrophe'
 pkgname=libstrophe
 version=0.8.7
-revision=4
+revision=5
```
Gottox: No need to revbump, the binary won't change.

PR #1214, opened 2015-03-27
xbps-src: Set the MAKEFLAGS environment variable

nmeum:
This ensures that packages with custom build functions also use our provided MAKEFLAGS. Without this commit it was necessary to always append the "${makejobs}" variable manually to make. For compatibility reasons that is still possible but should probably be removed in the future by making "makejobs" a local variable.

See also:
https://projects.archlinux.org/pacman.git/tree/scripts/makepkg.sh.in#n1538

---

This also means that the `${makejobs}` variable can be removed from build-styles. I noticed that some build-styles also use a `${make_build_args}` variable. I am not 100% sure what this variable does but maybe it should be append to `MAKEFLAGS` as well?

### DIFF
--- a/common/xbps-src/shutils/common.sh
+++ b/common/xbps-src/shutils/common.sh
@@ -217,7 +217,7 @@ setup_pkg() {
     [ -z "$pkg" ] && return 1
 
     # Start with a sane environment
-    unset -v PKG_BUILD_OPTIONS XBPS_CROSS_CFLAGS XBPS_CROSS_CXXFLAGS XBPS_CROSS_CPPFLAGS XBPS_CROSS_LDFLAGS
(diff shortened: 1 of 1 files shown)


xtraeme:
Looks good, but incomplete. Just setting the var won't magically do anything... submit the changes where MAKEFLAGS could be used.

chneukirchen:
I think make picks up MAKEFLAGS automatically. But this requires some testing, I guess.

```
If the environment variable 'MAKEFILES' is defined, 'make' considers its
value as a list of names (separated by whitespace) of additional
makefiles to be read before the others.
```

nmeum:
Yeah, that's what I thought as well `make(1)` should pickup `MAKEFLAGS` automatically if the environment variable exists, thus exporting it should do the job.

xtraeme:
Exporting the var and not unsetting it is still not enough.

At the beginning of setup_pkg(), MAKEFLAGS should be unset too.

PR #1216, opened 2015-03-28
New package: i3-gaps-git

skrzyp:
This is my second package template for Void Linux.

It contains i3-gaps-git, which is an fork of i3 window manager with additional features: gaps, smart borders, i3bar height, etc.

I'm still unsure how can I mark -git packages versions in `version=""`


xtraeme:
I won't accept more -git packages unless it's absolutely necessary.

If this pkg does not have any tag, tell the developers to create them.

PR #1221, opened 2015-03-28
New package: libetpan 1.6

skrzyp:
This is my third Void Linux template and first library template.

It contains libetpan, a library to access NNTP and IMAP4 servers, used by Claws Mail and other e-mail clients.

This is also a preparation for Claws Mail template.


srcpkgs/libetpan/template
```diff
@@ -0,0 +1,29 @@
...
+pre_configure() {
+	./autogen.sh
+}
+
+libetpan-devel_package() {
+	depends="${pkgname}>=${version}_${revision}"
+		short_desc+=" - development files"
+		pkg_install() {
+			vmove usr/include
+				vmove "usr/lib/*.so"
```
dominikh: What's with the indentation here?

srcpkgs/libetpan/template
```diff
@@ -0,0 +1,29 @@
...
+pkgname="libetpan"
+version="1.6"
+revision=1
+build_style=gnu-configure
+hostmakedepends="libtool automake autoconf"
+short_desc="Mail Framework library for C language"
+maintainer="Jakub Skrzypnik <jot.skrzyp@gmail.com>"
+license="BSD"
+homepage="http://www.etpan.org/libetpan.html"
+distfiles="https://github.com/dinhviethoa/libetpan/archive/1.6.tar.gz"
```
dominikh: use ${version} instead of 1.6 in the URL

dominikh:
Squash all those commits

srcpkgs/libetpan/template
```diff
@@ -0,0 +1,29 @@
...
+}
+
+libetpan-devel_package() {
+	depends="${pkgname}>=${version}_${revision}"
+	short_desc+=" - development files"
+	pkg_install() {
+		vmove usr/include
+		vmove "usr/lib/*.so"
+		vmove "usr/lib/*.a"
+		vmove "usr/lib/*.la"
```
xtraeme: This line is wrong... libtool archives (*.la) are removed by default. I really doubt you have tested this.

You are also missing the soname change in common/shlibs.

xtraeme:
If this pkg installs pkg-config files those should be part of the -devel pkg.

PR #1224, opened 2015-03-28
New package: octave-3.8.2

crocusino:


xtraeme:
This is completely wrong... packages should add `development` build dependencies instead, i.e `pcre-devel`.  This development pkg provides the headers and depends on the package providing the shared library to link against.

srcpkgs/octave/template
```diff
@@ -0,0 +1,13 @@
+# Template file for 'octave'
+pkgname=octave
+version=3.8.2
+revision=1
+build_style=gnu-configure
+hostmakedepends="gcc-c++ gcc-fortran perl"
```
xtraeme: You don't need to add `gcc-c++` to hostmakedepends, it's part of base-chroot.

srcpkgs/octave/template
```diff
@@ -0,0 +1,13 @@
...
+makedepends="ghostscript less gnuplot-wx libgraphicsmagick-devel pcre-devel readline-devel fftw-devel blas-devel lapack-devel"
```
xtraeme: - I really doubt you need `less` in makedepends.
- You should add `ghostscript-devel` instead.

crocusino: - ghostscript (seems just 'gs', not a library) is necessary as a dep, otherwise some graphics may fail
- either 'gnuplot' binary is used or a native GUI is built (which is not desired)
I know that neither of the two is linked to octave, but both are necessary for running

xtraeme:
If ghostscript or gnuplot-wx are required at runtime, those should be added to `depends` instead.

Gottox:
Please rebase.

crocusino:
Strange, it builds properly for me, I just checked it.

Gottox:
The rebase has gone wrong.

Gottox:
```
git pull --all
git rebase -i origin/master    <- delete all lines which shouldn't be in the pullreq
# Do NOT merge
git push -f
```

Gottox:
forgot what I said. It seems that github itself has changed something :)

Gottox:
any progress on this issue?

xtraeme:
- 32 commits with unrelated changes and merges. Please rebase and submit a proper PR.
- void uses development packages, so that headers, static libraries and other files only required for development must be moved to a -devel subpkg.

This pkg is very incomplete, but can be used as starting point.

xtraeme:
No progress for nearly 1 month... closing.

Thanks for your effort.

PR #1227, opened 2015-03-29
Random updates

Gottox:
I'm too tired to push this without review. :)

### DIFF
--- a/srcpkgs/afl/template
+++ b/srcpkgs/afl/template
@@ -1,6 +1,6 @@
 # Template file for 'afl'
 pkgname=afl
-version=1.56b
+version=1.58b
 revision=1
 build_style=gnu-makefile
 # x86 only currently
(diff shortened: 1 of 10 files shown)


srcpkgs/afl/template
```diff
@@ -1,6 +1,6 @@
 # Template file for 'afl'
 pkgname=afl
-version=1.56b
+version=1.58b
```
xtraeme: Checksum not updated.

srcpkgs/appdata-tools/template
```diff
@@ -1,6 +1,6 @@
 # Template build file for 'appdata-tools'.
 pkgname=appdata-tools
-version=0.1.7
+version=0.1.8
```
xtraeme: Checksum not updated.

srcpkgs/exiftool/template
```diff
@@ -1,6 +1,6 @@
 # Template file for 'exiftool'
 pkgname=exiftool
-version=9.89
+version=9.90
```
xtraeme: Checksum not updated.

PR #1231, opened 2015-03-29
Replace several -git packages with proper releases

dominikh:
Build cwm, apvlv, lumina and libvdpau-sunxi from tagged releases.

Cc'ing the respective package maintainers for review: @chneukirchen @xtraeme @Ypnose @logenkain

@xtraeme, is there any way with xbps to properly rename a package? That is, when we rename foo-git to foo, `xbps-install -Su` should replace an installed foo-git with foo, even if the new version is lower.


xtraeme:
@dominikh there's no support for this on xbps (currently).

Ypnose:
About cwm, I would prefer a version like <code>version=$(date +%Y%m%d)</code> rather than OpenBSD release. With <code>5.6</code>, every changes would increment the <code>revision</code> and it's less readable, compared to the build date. And reaching something like <code>revision=10</code> with <code>version=5.6</code> is not very "clean" to me.
Anyway, I'm not the man in charge here, that's a suggestion.

xtraeme:
I'm the man in charge, and as such, I prefer to use a proper release in all packages on this PR.

KTHXBYE

Ypnose:
Ok then. Forgot my comment.

chneukirchen:
OK for lumina.

dominikh:
Rebased and merged.

PR #1232, opened 2015-03-29
New package: utox-0.2.o_1

teh-jazzman:
Second TOX client


srcpkgs/utox/template
```diff
@@ -0,0 +1,17 @@
+# Template file for 'utox'
+pkgname="utox"
+version="0.2.o"
+revision=1
+_commithash="bdd1a009518e325a73a527da152793102b765ef6"
```
Gottox: Why exactly this commit? - Why not stick to the 0.2.o Tag?

teh-jazzman: The contents of the tag can change, at which point the .tar.gz file would no longer validate against the checksum. The commit is more reliable in that regard.

dominikh:
> The contents of the tag can change, at which point the .tar.gz file would no longer validate against the checksum

I consider that a feature. If the tag changes, something's up, and the maintainer should inquire what happened, instead of continuing to build a commit that upstream no longer considers version X.

srcpkgs/utox/template
dominikh: > The contents of the tag can change, at which point the .tar.gz file would no longer validate against the checksum

I consider that a feature. If the tag changes, something's up, and the maintainer should inquire what happened, instead of continuing to build a commit that upstream no longer considers version X.

srcpkgs/utox/template
```diff
@@ -0,0 +1,17 @@
...
+short_desc="Lightweight TOX instant messenger client written in C"
+maintainer="Stefan Mühlinghaus <jazzman@alphabreed.com>"
+license="GPL-3"
+homepage="https://wiki.tox.im/UTox"
+depends="toxcore-git>=20150313_2"
```
dominikh: Is this `depends` line actually necessary? Does xbps-src not detect the .so dependency correctly?

srcpkgs/utox/template
Gottox: 1. If the content of a tag changes, the maintainer should be slapped in the face. It's the as you would update a versioned .tar.gz without changing the version. Tags are meant to be immutable, that's why you need to add a -f flag to overwrite them.
2. The commit hash doesn't even point to the release tag.

teh-jazzman: Point taken. I changed the template to fetch the tag instead of the commit.

srcpkgs/utox/template
teh-jazzman: You are right, it probably was not neccessary. That was just me trying to make extra sure the current version is used. I removed the `depends` line.

PR #1235, opened 2015-03-29
New package: qtox-20150325_1

teh-jazzman:
Third (and last) TOX client


dominikh:
Essentially the same comments and resolutions as in #1232

teh-jazzman:
I removed the `depends` line but still have to fetch the commit since qtox has no releases/tags and `HEAD` is too volatile (and would not work with this version of the toxcore library anyway)

PR #1246, opened 2015-03-31
newsbeuter: added podbeuter patch

### LINKED ISSUE #1244: Newsbeuter: They broke podbeuter
The newsbeuter team doesn't make releases very often.  The current version has a segfault with podbeuter, which is part of newsbeuter.

I suggest we downgrade the package to the previous version if/until they make another proper release.

The current version is "r2.9", the last fully working version was "r2.8"
So if I want to downgrade, should I just add reverts above version as shown below?

reverts="r2.9"
version=r2.9

And if I understand this right, xbps will tell people who have this installed that it is downgrading, and it will also automagicaly update the template once a new revision is released? (As far as I'm aware, the r2.9 bump was done automatically.)

Thanks,

Logen Kain

logenkain:
Added patch to resolve #1244


xtraeme:
The pkg revision must be bumped.

PR #1253, opened 2015-04-01
New package: rxvt-unicode-terminfo

skrzyp:
It contains only terminfo data `/usr/share/terminfo` for headless systems (e.g. servers) which does not require X, but used to accept connections from other hosts trough urxvt.

As side note, it's a bit urgent to me, while I have configured a simple server with Void right now and I need that package to connect with this server comfortably, using urxvt :smile:

### DIFF
--- a/srcpkgs/rxvt-unicode/template
+++ b/srcpkgs/rxvt-unicode/template
@@ -1,7 +1,7 @@
 # Template build file for 'rxvt-unicode'.
 pkgname=rxvt-unicode
 version=9.21
-revision=2
+revision=3
 build_style=gnu-configure
 configure_args="
  --with-terminfo=/usr/share/terminfo --enable-256-color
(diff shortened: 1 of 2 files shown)


xtraeme:
Package must be revbumped and a symlink to the subpkg must be added.

dominikh:
Should rxvt-unicode then depend on rxvt-unicode-terminfo?

xtraeme:
@dominikh agreed

srcpkgs/rxvt-unicode/template
```diff
@@ -92,3 +92,10 @@ do_configure() {
 		./configure $configure_args
 	fi
 }
+
+rxvt-unicode-terminfo_package() {
+	short_desc+=" - terminfo data for headless systems"
```
dominikh: Personally I'd just say "terminfo data", dropping the "for headless systems" part. On headed systems, you still need it.

xtraeme: I agree, and this is missing `noarch=yes` as well.

dominikh:
6b3c149

PR #1258, opened 2015-04-02
mopidy: update to 1.0.0

allan:


srcpkgs/mopidy/template
```diff
@@ -1,6 +1,6 @@
 # Template file for 'mopidy'
 pkgname=mopidy
-version=0.19.5
+version=1.0.0
 revision=1
 build_style=python-module
```
asergi: Revision must be set to 1.

srcpkgs/mopidy/template
```diff
@@ -8,8 +8,7 @@ maintainer="allan <mail@may.mooo.com>"
 license="Apache-2.0"
 homepage="http://www.mopidy.com"
 distfiles="${PYPI_SITE}/M/Mopidy/Mopidy-${version}.tar.gz"
-checksum=940604eadd6435ebf7a85bbc0f15ea57a7fdf5e7e276da17732955ab10050627
-
+checksum=509ec28f176ceea5dc213c16069763ba5496744bf96e117c292fee38f5007fee
 noarch=yes
 python_versions=2.7
```
asergi: By default, python_versions is set to 2.7, so you can remove it.

asergi: You can remove python-devel from both makedepends and hostmakedepends.
Also, python-pykka from hostmakedepends, too.

srcpkgs/mopidy/template
dominikh: Revision *is* set to 1?

PR #1260, opened 2015-04-02
New package: openimageio-1.5.13

lemmi:
my first PR for a shared lib, please pay closer attention for mistakes with the subpkg, not sure if this is all correct. can't cross compile because a dependency is not available (ilmbase).
this is a dependency missing to compile blender with the cycles-renderer.


xtraeme:
This is good enough, thanks :+1:

PR #1266, opened 2015-04-03
New package: vorbisgain-0.37

Grauwolf:


srcpkgs/vorbisgain/template
```diff
@@ -0,0 +1,14 @@
+# Template file for 'vorbisgain'
+pkgname="vorbisgain"
+version="0.37"
+revision=1
+build_style=gnu-configure
+hostmakedepends="pkg-config"
+makedepends="libvorbis-devel"
+depends="libvorbis"
```
xtraeme: This depends line is unnecessary.

srcpkgs/vorbisgain/template
```diff
@@ -0,0 +1,14 @@
...
+short_desc="A utility that computes the ReplayGain values for Ogg Vorbis files"
```
xtraeme: A -> An ?

Grauwolf: the first syllable of the word is pronounced 'you' so 'A utility' should be right

srcpkgs/vorbisgain/template
Grauwolf: Sorry. My bad.

PR #1267, opened 2015-04-03
NOT TO BE MERGED YET: Gnome 3.16

Gottox:
This updates Gnome to 3.16.

Compiling gnome works fine but both gnome-session and gdm fail to start.


Gottox:
No progress for no. Reopening once we've got Gnome3.16 running.

PR #1269, opened 2015-04-03
mosquitto: update to 1.4.1

k00mi:


dominikh:
Merged as b2d0b01, thank you.

Issue #1230, opened 2015-03-29
Pulseaudio: add system wide service


thypon:
Headless server @ https://wiki.gentoo.org/wiki/PulseAudio

nmeum:
Running Pulseaudio as a system wide service is actually not recommend. Doing so should only be considered in certain edge cases. Furthermore, additional users and groups are required to do so.

See also: The more or less official documentation http://www.freedesktop.org/wiki/Software/PulseAudio/Documentation/User/SystemWide/

thypon:
In my configuration i need to use the audio in two different users.

dominikh:
Feel free to write your own service ;) Void shouldn't provide for all the edge cases out there.

PR #1271, opened 2015-04-04
pulseaudio: add systemwide service.

thypon:
#1230

### DIFF
--- a/srcpkgs/pulseaudio/files/pulseaudio/run
+++ b/srcpkgs/pulseaudio/files/pulseaudio/run
@@ -1,0 +1,2 @@
+#!/bin/sh
+exec pulseaudio --system
--- a/srcpkgs/pulseaudio/template
+++ b/srcpkgs/pulseaudio/template
@@ -1,7 +1,7 @@
 # Template file for 'pulseaudio'
 pkgname=pulseaudio
 version=6.0
-revision=3
+revision=4
 # Package build options
 build_options="systemd"
 build_style=gnu-configure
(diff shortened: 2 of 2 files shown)


Gottox:
Pulseaudio deprecate the usage as a systemwird service. Anyway, there are usecases which can be covered by Void. So I give :+1: here.

srcpkgs/pulseaudio/files/pulseaudio/run
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+exec /usr/bin/pulseaudio --system
```
xtraeme: Do not hardcode the patch, use `exec pulseaudio ...`

thypon: It's a rule or a preference? I think that eliminating explicit executable path creates security risk in case of PATH modification.

xtraeme:
It's mandatory because the bin should be available in PATH, which is set in /etc/runit/[12].

thypon:
Rebase?

xtraeme:
As you prefer.

thypon:
Rebased.

xtraeme:
Please rebase it again to test it with your new travis-ci infrastructure, thanks.

thypon:
It fails with i686.

xtraeme:
Not your fault, the cross-i686-pc-linux-gnu cross compiler pkg seems to be fucked up.

Thanks.

PR #1272, opened 2015-04-05
New package: remind-03.01.13

Grauwolf:


bougyman:
tested with armv7hf and armv6hf as well as x86s

PR #1275, opened 2015-04-07
New package: gradle-2.3

asiekierka:


srcpkgs/gradle/template
```diff
@@ -0,0 +1,22 @@
...
+license="Apache-2.0"
+homepage="http://gradle.org/"
+distfiles="https://services.gradle.org/distributions/gradle-${version}-bin.zip"
+checksum=010dd9f31849abc3d5644e282943b1c1c355f8e2635c5789833979ce590a3774
+
+do_install() {
+	vmkdir "usr/bin"
+	vmkdir "usr/lib/gradle"
+	vmkdir "usr/share/doc/gradle"
+	vinstall ${FILESDIR}/gradle 755 "usr/bin"
```
xtraeme: use vbin instead.

PR #1278, opened 2015-04-07
New package: arandr-0.1.7.1

mid-kid:
(Forgot the version number in the commit, sorry)


Duncaen:
git commit --amend

mid-kid:
Fixed.

asergi:
Please apply the following changes:
- add noarch=yes
- remove 'python' from both {make,}depends
- add pycompile_module="screenlayout"

mid-kid:
Changed as per your request.
I'm new to packaging in void, so I don't really know what the standards are, and I'm too lazy to read through the whole document.

srcpkgs/arandr/template
```diff
@@ -0,0 +1,14 @@
...
+version="0.1.7.1"
+revision=1
+build_style=python-module
+short_desc="Graphical frontend for XRandR"
+maintainer="mid-kid <esteve.varela@gmail.com>"
+license="GPL-3"
+homepage="http://christian.amsuess.com/tools/arandr/"
+distfiles="http://christian.amsuess.com/tools/arandr/files/arandr-${version}.tar.gz"
+checksum="ab73aa90d05d48b95a01afdad37b7c21a2cdf4af98f7d78b2502b819382748da"
+makedepends="python-docutils"
```
xtraeme: makedepends -> hostmakedepends

asergi: Missed that :+1:

mid-kid: I don't get the difference... It's not really detailed in the manual aside from the fact that it's the dependencies for the "host" and "target", or I missed the part where that is explained.

Duncaen:
Its for crosscompiling, makedepenencies are normally -devel packages that you need in the chroot, hostmakedependencies are things the host has to run like, python-docutils, unzip or git for example.
Your welcome to come to our IRC channel #xbps on freenode ;)

mid-kid:
Ah, now I understand. Thanks.

xtraeme:
Applied, thanks.

PR #1276, opened 2015-04-07
New package: intellij-idea-community-edition-14.1.1

asiekierka:

PR #1282, opened 2015-04-08
Revert "New package: intellij-idea-community-edition-14.1.1"

bougyman:
Reverts voidlinux/void-packages#1276


bougyman:
Need to take into account the arch-specific binaries because stripping fails: http://build.voidlinux.eu/builders/i686_builder/builds/11101/steps/shell_3/logs/stdio

PR #1292, opened 2015-04-09
use proot instead of xbps-u{chroot,unshare}

Gottox:
This is just a proof of concept of doing the isolation in userspace. What do you think about this?


Gottox:
If we want to rely on kernelspace isolation, the developers announced, that they will write a drop-in replacement for proot using namespaces.

PR #1295, opened 2015-04-10
New package: cvm-0.97

bougyman:


srcpkgs/cvm/template
```diff
@@ -0,0 +1,37 @@
+# Template file for 'cvm'
+pkgname=cvm
+version=0.97
+revision=1
+build_style=gnu-makefile
+makedepends="bglibs-devel"
+nocross=yes
+hostmakedepends="libtool"
+depends="bglibs"
```
xtraeme: This depends line is unnecessary, if cvm is linked to any shared library provided by bglibs, the dependency will be added automagically thanks to common/shlibs.

srcpkgs/cvm/template
```diff
@@ -0,0 +1,37 @@
...
+short_desc="Credential Validation Modules"
+maintainer="bougyman <bougyman@voidlinux.eu>"
+license="GPL-3"
+homepage="http://untroubled.org/cvm/"
+distfiles="http://untroubled.org/cvm/cvm-0.97.tar.gz"
```
xtraeme: 0.97 -> ${version}

PR #1311, opened 2015-04-11
New package: : SoulseekQt-2014-11-30

jarbowski:
Hi,
I think Void needs a soulseek client.
I would have prefered to make a nicotine-plus package, but it's too complex for me.
It's my first attempt to make a Void package, so be indulgent, please.
I don't know how to upload the Files directory with the png and the desktop files, I'll try later.

### DIFF
--- a/srcpkgs/soulseekqt
+++ b/srcpkgs/soulseekqt
@@ -1,0 +1,32 @@
+# Template file for 'soulseekqt'
+pkgname=soulseekqt
+version=20141130
+revision=1
+_pkgver=2014-11-30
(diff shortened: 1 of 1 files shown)


srcpkgs/soulseekqt
```diff
@@ -0,0 +1,32 @@
+# Template file for 'soulseekqt'
```
Gottox: srcpkgs must contain directories named after the packages not files.

jarbowski:
I'm sorry, could you explicitly point the issue, I can't understand.
Thank you.

PR #1312, opened 2015-04-11
New package: preload-0.6.4

### LINKED ISSUE #1291: Package request: preload
http://en.wikipedia.org/wiki/Preload_%28software%29

bougyman:
Closes #1291


thypon:
Build on my machine without errors.

srcpkgs/preload/files/preload/run
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+exec 2>&1
+[ -r conf ] && . conf
```
xtraeme: This must be `". ./conf"`

srcpkgs/preload/template
```diff
@@ -0,0 +1,22 @@
+# Template file for 'preload'
+pkgname=preload
+version=0.6.4
+revision=1
+build_style=gnu-configure
+hostmakedepends="help2man pkg-config"
+makedepends="libglib-devel"
+depends="sv-helper"
```
xtraeme: Can you please stop adding sv-helper as dependency? this should not be the default, it's an optional dependency and I don't need it.

xtraeme:
What I mean is that if you need to use rsvlogd in the log service, you can create it manually. There's no need to force it by default.

bougyman:
Closing this, accidentally pushed this upstream. It has been fixed, though, per all your comments plus cross-build for i686 works, now. Thanks!

PR #1314, opened 2015-04-12
New package: zeal-20141123

Grauwolf:


srcpkgs/zeal/template
```diff
@@ -0,0 +1,26 @@
+# Template file for 'zeal'
+pkgname=zeal
+version=20141123
+revision=1
+wrksrc="${pkgname}-${version}/zeal"
+hostmakedepends="qt5-qmake make"
+makedepends="qt5-devel qt5-webkit-devel xcb-util-keysyms-devel libarchive-devel"
+depends="bsdtar qt5-plugin-sqlite"
```
xtraeme: Why does it depend on bsdtar?

srcpkgs/zeal/template
```diff
@@ -0,0 +1,26 @@
+# Template file for 'zeal'
+pkgname=zeal
+version=20141123
+revision=1
+wrksrc="${pkgname}-${version}/zeal"
```
xtraeme: This is not the proper way to set wrksrc. wrksrc must be set to a top-level directory in $XBPS_BUILDDIR (/builddir), otherwise the clean target won't clean this up correctly.

Use this:

```
build_wrksrc=zeal
```

and remove `wrksrc`.

srcpkgs/zeal/template
```diff
@@ -0,0 +1,26 @@
...
+hostmakedepends="qt5-qmake make"
```
xtraeme: You don't really need to add `make` to `hostmakedepends`. GNU make is already a dependency of `base-chroot`, and is part of the bootstrap packages.

Remove it.

srcpkgs/zeal/template
Grauwolf: While zeal will start without having bsdtar installed you won't be able to download and extract any docset. I could also create a INSTALL.msg which says that but i thought adding bsdtar to the depends makes much more sense in this case.

Grauwolf:
bsdtar is still in there. the other two things are fixed

PR #1323, opened 2015-04-15
Add update file for wesnoth

hanspolo:


srcpkgs/wesnoth/update
```diff
@@ -0,0 +1,3 @@
+pkgname=wesnoth
+version=1.12.2
```
Gottox: above lines can be dropped

srcpkgs/wesnoth/update
```diff
@@ -0,0 +1,3 @@
...
+ignore=1.13
```
Gottox: missing \n at the end.

xtraeme:
What will happen when 1.15 is available? ignoring just 1.13 is wrong.

Gottox:
ignore="1.1[13679]"

PR #1334, opened 2015-04-17
Evolution depends on gnome-keyring

pullmoll:
Or to be more specific it depends on /usr/share/dbus-1/services/org.freedesktop.secrets.service, i.e. the gnome-keyring-daemon run with --components=secrets
Without this dependency installed you will see the missing service org.freedesktop.secrets reported in Evolution when trying to connect to a server that requires authentication.


xtraeme:
The changes to `etc/defaults.conf` shouldn't be there.

pullmoll:
This was of course unintended. I reverted the changes now.

pullmoll:
I don't know why my qt-creator template try appears here. It is unrelated. Would I have to create another clone of my fork first?

xtraeme:
The qtcreator changes are irrelevant for this PR, please submit another one for this pkg.

PR #1338, opened 2015-04-18
New Package: dtrx-7.1

### LINKED ISSUE #1335: Package Request: dtrx
"Do The Right Extraction."

http://brettcsmith.org/2007/dtrx/

This is a tiny Python script used to automate the extraction of various archive and compression formats. It's a minor convenience package.

logenkain:
This closes #1335


srcpkgs/dtrx/template
```diff
@@ -0,0 +1,15 @@
+# Template file for 'dtrx'
+pkgname=dtrx
+version=7.1
+revision=1
+noarch=yes
+build_style=python-module
+pycompile_module="dtrx"
+makedepends="python"
```
Gottox: python is a `hostmakedepends`.

xtraeme:
python -> python-setuptools
makedepends -> hostmakedepends

logenkain:
I was looking at the Deluge Template and noticed that python-setuptools is in all three depends lines.  Is that how I should have this template?

xtraeme:
Commonly `python-setuptools` in `hostmakedepends` is enough for pure python modules (no C extensions).

xtraeme:
I noticed some things from the output of travis builds:

- `pycompile_module` is unnecessary, remove it.
- the ARM packages are empty, i.e no files in DESTDIR.

PR #1339, opened 2015-04-18
Fix evolution dependencies

pullmoll:
Evolution depends on gnome-keyring for connecting to servers which require authentication.
Also an icon theme seems to be required in the makedepends. I used adwaita-icon-theme.


xtraeme:
This pull request is a mess, please a submit a proper pull request with only a relevant commit for evolution-data-server.

Try to create a branch in your repository fork and then force push from it to that PR.

PR #1340, opened 2015-04-19
New package: ddclient-3.8.2

ayghor:


ayghor:
With this we can get rid of noip2 and my ugly patch. It supports a number of dynamic dns services!

srcpkgs/ddclient/template
```diff
@@ -0,0 +1,39 @@
...
+short_desc="Perl client used to update dynamic DNS"
+license="GPL-2"
+maintainer="rogi <rogi@skylittlesystem.org>"
+homepage="${SOURCEFORGE_SITE}/p/ddclient/wiki/Home/"
+distfiles="${SOURCEFORGE_SITE}/$pkgname/$pkgname-$version.tar.bz2"
+checksum=f343d2297b97b769949d4d6c3f603a8c52433acf2104245538808a2ea003ed5e
+
+do_install() {
+	vlicense COPYING
+	vlicense COPYRIGHT
```
xtraeme: If this package is GPL-*, there's no need to install the license files.

srcpkgs/ddclient/template
```diff
@@ -0,0 +1,39 @@
...
+
+	vdoc Changelog
```
xtraeme: Please do not install Changelog, it's rather useless.

xtraeme:
Please fix the error reported by travis.

ayghor:
It was hard to find teh error on travis log! Teh message was white and was about trailing whitespaces. I ran xlint here, it should pass now.

PR #1343, opened 2015-04-19
gnumeric: updated to 1.12.22

hanspolo:


xtraeme:
Please add `gdk-pixbuf-devel` to `hostmakedepends` to fix cross compilation.

hanspolo:
I updated it.

xtraeme:
Thanks :+1:

xtraeme:
Looks like gdx-pixbuf-devel wasn't enough, but it's ok for now.

PR #1345, opened 2015-04-19
New package: dispcalGUI-2.6.0.0

lemmi:


srcpkgs/dispcalGUI/template
```diff
@@ -0,0 +1,16 @@
+# Template file for 'dispcalGUI'
+pkgname=dispcalGUI
+version=2.6.0.0
+revision=1
+build_style=python-module
+python_versions="2.7"
+nocross=yes
+hostmakedepends="python"
+makedepends="python-devel xorg-server-devel libXinerama-devel libXrandr-devel"
```
xtraeme: - python_versions is 2.7 by default, remove it.
- xorg-server-devel shouldn't be used as build dependency.
- as Ypnose said, replace `dispcalGUI` with `${pkgname}`.

PR #1346, opened 2015-04-19
Fix missing dependency for evolution

pullmoll:
Evolution depends on gnome-keyring for connecting to servers which require authentication.
To be more specific it depends on /usr/share/dbus-1/services/org.freedesktop.secrets.service, i.e. the gnome-keyring-daemon run with --components=secrets
Also an icon theme is required as a make dependency. I chose the adwaita-icon-theme.


pullmoll:
The build process needs to include ./libjbig in the include path for the headers, i.e.:
do_build() {
    CFLAGS+=-I${wrksrc}/libjbig make -e
}
But this is not sufficient, because there is no install target in the jbigkit Makefile(s) at all.

The latter requires a hand-made do_install() script, which would install the resulting binaries, libraries etc. to the ${PKGDESTDIR}/usr/wherever directories. I'll leave these remaining steps to someone else because I'm not sober, though :)

pullmoll:
This is ridiculous. Why des github associate a push on a separate branch with this pull request? It is on branch testing and not related. *grr*

PR #1356, opened 2015-04-20
New package: axel-2.4

pullmoll:


pullmoll:
Doesn't work for some unknown reason: can't connect to http sites. I'll close it until I can solve the issue.

PR #1357, opened 2015-04-20
DO NOT MERGE - cupsd: don't run as root

### LINKED ISSUE #1347: cups: runs as root
Right now cups runs as root. 

this is the `run` file:
```
#!/bin/sh
exec cupsd -f
```

The user and the group used shoudl be `cups:lp`

bougyman:
Fixes #1347


bougyman:
Need some testing on this. Was there a reason to run as root and not cups?

xtraeme:
This is pointless, cups uses a non priviledged port so that it cannot be started as !root.

Issue #1273, opened 2015-04-05
Enable Yama LSM


thypon:
Yama LSM could be a nice security addition.

bougyman:
+1

xtraeme:
If you want to have this by default please submit a proper pull request with all required changes.

thypon:
```
diff --git a/.config b/.config
index d9df953..0a39df8 100644
--- a/.config
+++ b/.config
@@ -873,6 +873,7 @@ CONFIG_IPV6_GRE=m
 CONFIG_IPV6_MULTIPLE_TABLES=y
 CONFIG_IPV6_SUBTREES=y
 # CONFIG_IPV6_MROUTE is not set
+# CONFIG_NETLABEL is not set
 # CONFIG_NETWORK_SECMARK is not set
 CONFIG_NET_PTP_CLASSIFY=y
 # CONFIG_NETWORK_PHY_TIMESTAMPING is not set
@@ -1128,6 +1129,7 @@ CONFIG_IP_NF_TARGET_CLUSTERIP=m
 CONFIG_IP_NF_TARGET_ECN=m
 CONFIG_IP_NF_TARGET_TTL=m
 CONFIG_IP_NF_RAW=m
+# CONFIG_IP_NF_SECURITY is not set
 CONFIG_IP_NF_ARPTABLES=m
 CONFIG_IP_NF_ARPFILTER=m
 CONFIG_IP_NF_ARP_MANGLE=m
@@ -1163,6 +1165,7 @@ CONFIG_IP6_NF_TARGET_REJECT=m
 CONFIG_IP6_NF_TARGET_SYNPROXY=m
 CONFIG_IP6_NF_MANGLE=m
 CONFIG_IP6_NF_RAW=m
+# CONFIG_IP6_NF_SECURITY is not set
 CONFIG_IP6_NF_NAT=m
 CONFIG_IP6_NF_TARGET_MASQUERADE=m
 CONFIG_IP6_NF_TARGET_NPT=m
@@ -6490,6 +6493,7 @@ CONFIG_FSNOTIFY=y
 CONFIG_DNOTIFY=y
 CONFIG_INOTIFY_USER=y
 CONFIG_FANOTIFY=y
+# CONFIG_FANOTIFY_ACCESS_PERMISSIONS is not set
 CONFIG_QUOTA=y
 CONFIG_QUOTA_NETLINK_INTERFACE=y
 # CONFIG_PRINT_QUOTA_WARNING is not set
[...]

PR #1368, opened 2015-04-22
Yama LSM

thypon:
#1273


srcpkgs/linux3.14/files/i386-dotconfig
```diff
@@ -2019,7 +2022,7 @@ CONFIG_I2O_PROC=m
 CONFIG_MACINTOSH_DRIVERS=y
 CONFIG_MAC_EMUMOUSEBTN=m
 CONFIG_NETDEVICES=y
-CONFIG_MII=m
+CONFIG_MII=y
```
xtraeme: any reason for this?

srcpkgs/linux3.14/files/i386-dotconfig
```diff
@@ -3085,7 +3088,7 @@ CONFIG_HSI_CHAR=m
 #
 # PPS support
 #
-CONFIG_PPS=m
+CONFIG_PPS=y
```
xtraeme: and this one?

thypon: It was enabled when i put yama by default. Some dependency i think.

srcpkgs/linux3.14/files/i386-dotconfig
thypon: Autoadded by menuconfig when i enabled Yama LSM, not intended, but a side effect. It happens only on 3.14, we could exclude it.

xtraeme:
Ok, but this PR cannot be applied. Try to rebase it :-)

xtraeme:
And while you are at it, please update linux3.18 as well.

thypon:
timeout

PR #1379, opened 2015-04-23
New package: axel-2.4

pullmoll:
Doesn't compile on mips and mipsel.

### DIFF
--- a/srcpkgs/axel/template
+++ b/srcpkgs/axel/template
@@ -4,6 +4,7 @@ version=2.4
 revision=1
 wrksrc=axel-2.4
 build_style=gnu-configure
+conf_files=/etc/axelrc
 short_desc="Download accelerator for FTP and HTTP"
 maintainer="Jürgen Buchmüller <pullmoll@t-online.de>"
 license="GPL-2"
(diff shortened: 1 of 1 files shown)


srcpkgs/axel/template
```diff
@@ -0,0 +1,21 @@
...
+revision=1
+wrksrc=axel-2.4
+build_style=gnu-configure
+short_desc="Download accelerator for FTP and HTTP"
+maintainer="Jürgen Buchmüller <pullmoll@t-online.de>"
+license="GPL-2"
+homepage="https://axel.alioth.debian.org/"
+distfiles="${DEBIAN_SITE}/main/a/axel/axel_${version}.orig.tar.gz"
+checksum="359a57ab4e354bcb6075430d977c59d33eb3e2f1415a811948fa8ae657ca8036"
+only_for_archs="x86_64 x86_64-musl i686 i686-musl armv6hf armv6hf-musl armv7hf armv7hf-musl"
```
xtraeme: Please remove this line, it's pointless.

srcpkgs/axel/template
```diff
@@ -0,0 +1,21 @@
...
+
+do_configure() {
+	./configure --prefix=/usr
```
xtraeme: If you don't pass `${configure_args}` the cross compilation will fail, I'd suggest you to remove `./configure` and rename `do_configure()` to `post_configure()`.

xtraeme:
`/etc/axelrc` must be added to `conf_files` as well, and please disable stripping via configure.

pullmoll:
Hmmm.. removing strip requires a dummy STRIP= in Makefile.settings, perhaps echo. I'll try.

srcpkgs/axel/template
```diff
@@ -0,0 +1,20 @@
+# Template file for 'axel'
+pkgname=axel
+version=2.4
+revision=1
+wrksrc=axel-2.4
```
xtraeme: This line is redundant, `wrksrc` by default is set to `${pkgname}-${version}`.

xtraeme:
You should follow the standard commit log style:

```
<pkgname>: ...
```

pullmoll:
Perhaps xlint can be improved to detect redundant wrksrc?

xtraeme:
I agree // cc @chneukirchen

PR #1382, opened 2015-04-23
CopyAgent: update SHA256

### LINKED ISSUE #1366: CopyAgent sha256sum mismatch
The file Copy.tgz downloaded from https://copy.com/install/linux/Copy.tgz hash is f33c0297c2d8f7e90ec481beed88a918446a87507c9ea29f8feae21a4e85f72e
The template says
checksum=f474099d86baadd05758fa33164dae44b0127933f73c6a6a6e2f243bbf62bc42
Since there is no version number, I guess this will happen every now and then.

bougyman:
Fixes #1366 

This really sucks that they don't expose a version number.


xtraeme:
The tarball should be renamed with a proper version to make this work with future updates:

```
distfiles="url>${pkgname}-${version}.tar.gz"
```

xtraeme:
No need to use an additional variable, but it's ok too :-)

Issue #1369, opened 2015-04-22
Package request: nfc-tools


thypon:
A lot of usefull tools/libraries to hack with nfc

https://github.com/nfc-tools

bougyman:
This is a lot of individual packages, not just one.

thypon:
Yep.

PR #1385, opened 2015-04-23
Nfc tools

thypon:
#1369 here the two most used nfc-tools lib/apps


srcpkgs/libnfc/template
```diff
@@ -0,0 +1,31 @@
...
+hostmakedepends="cmake pkg-config"
+makedepends="libusb-devel libusb-compat-devel"
+maintainer="Andrea Brancaleoni <miwaxe@gmail.com>"
+license="GPL-3"
+homepage="http://nfc-tools.org/"
+distfiles="https://github.com/nfc-tools/libnfc/archive/libnfc-${version}.tar.gz"
+checksum=30de35b4f1af3f57dab40d91ffb2275664a35859ff2b014ba7b226aa3f5465f5
+
+pre_configure() {
+	sed -i "s|MODE=\"0664\", GROUP=\"plugdev\"|TAG+=\"uaccess\"|g" contrib/udev/42-pn53x.rules
```
xtraeme: We don't have a `plugdev` group.

srcpkgs/mfoc/template
```diff
@@ -0,0 +1,21 @@
...
+makedepends="libnfc-devel"
+short_desc="Mifare Classic Offline Cracker"
+maintainer="Andrea Brancaleoni <miwaxe@gmail.com>"
+license="GPL-3"
+homepage="http://nfc-tools.org"
+distfiles="https://github.com/nfc-tools/mfoc/archive/${pkgname}-${version}.tar.gz"
+checksum=2dfd8ffa4a8b357807680d190a91c8cf3db54b4211a781edc1108af401dbaad7
+
+pre_configure() {
+	aclocal
```
xtraeme: You can replace those commands with `autoreconf -fi`

srcpkgs/mfoc/template
```diff
@@ -0,0 +1,21 @@
+# Template file for 'mfoc'
+pkgname=mfoc
+version=0.10.7
+revision=1
+wrksrc="${pkgname}-${pkgname}-${version}"
+build_style=gnu-configure
+hostmakedepends="autoconf automake pkg-config"
```
xtraeme: `autoconf` is a dependency of `automake`, so that it's redundant.

srcpkgs/libnfc/template
thypon: In this sed i removed plugdev and add tag uaccess

srcpkgs/mfoc/template
thypon: check

srcpkgs/mfoc/template
thypon: check

srcpkgs/libnfc/template
xtraeme: You are right, I saw this backwards :-)

thypon:
Now it's okay

Issue #1252, opened 2015-04-01
Package request: pdftk


thypon:
https://www.pdflabs.com/tools/pdftk-server/

xtraeme:
Pushed in #1386

PR #1386, opened 2015-04-24
New package: pdftk-2.02

thypon:
#1252

### DIFF
--- a/srcpkgs/pdftk/template
+++ b/srcpkgs/pdftk/template
@@ -1,0 +1,24 @@
+# Template file for 'pdftk'
+pkgname=pdftk
+version=2.02
+revision=1
+wrksrc="${pkgname}-${version}-dist"
+build_wrksrc=${pkgname}
+build_style=gnu-makefile
+hostmakedepends="unzip gcc-gcj"
+makedepends="libgcj-devel"
+nocross=yes
+short_desc="Command-line tool for working with PDFs"
+maintainer="Andrea Brancaleoni <miwaxe@gmail.com>"
+license="GPL-2"
+homepage="https://www.pdflabs.com/tools/pdftk-server/"
+distfiles="http://www.pdflabs.com/tools/pdftk-the-pdf-toolkit/${pkgname}-${version}-src.zip"
+checksum=118f6a25fd3acaafb58824dce6f97cdc07e56050e666b90e4c4ef426ea37b8c1
+
+pre_configure() {
+	cp ${FILESDIR}/Makefile Makefile
+}
+do_install() {
+	vbin pdftk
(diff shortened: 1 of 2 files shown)


srcpkgs/pdftk/template
```diff
@@ -0,0 +1,23 @@
+# Template file for 'pdftk'
+pkgname=pdftk
+version=2.02
+revision=1
+wrksrc="${pkgname}-${version}-dist/pdftk"
+build_style=gnu-makefile
+hostmakedepends="unzip"
+makedepends="gcc-gcj"
```
xtraeme: - `gcc-gcj` should be in `hostmakedepends` instead.
- I really doubt this pkg needs `libgcj-devel` as runtime dependency.

srcpkgs/pdftk/template
```diff
@@ -0,0 +1,23 @@
+# Template file for 'pdftk'
+pkgname=pdftk
+version=2.02
+revision=1
+wrksrc="${pkgname}-${version}-dist/pdftk"
```
xtraeme: This is wrong, `wrksrc` should always be set to a top level directory in ${XBPS_BUILDDIR} (/builddir). Use this:

```
wrksrc=${pkgname}-${version}-dist
build_wrksrc=${pkgname}
```

srcpkgs/pdftk/template
thypon: ops, my fault, i thought it was needed by makedepends.

thypon:
Why the hell you rebase master....

thypon:
Doing the same thing on my Worskstation with "./xbps-src pkg pdftk" doesn't fail.

thypon:
the funny thing is that this ends the build on my pc:
```
[gattuso@razzi packages]$ tar tvf hostdir/binpkgs/pdftk/pdftk-2.02_1.x86_64.xbps 
-rw-r--r-- root/root      1252 2015-04-24 22:48 ./props.plist
-rw-r--r-- root/root      1097 2015-04-24 22:48 ./files.plist
-rw-r--r-- 0/0           23432 2015-04-24 22:48 ./usr/share/man/man1/pdftk.1
-rwxr-xr-x 0/0         2998976 2015-04-24 22:48 ./usr/bin/pdftk
```

thypon:
I think that it fails on
```
gcj -Wno-all -fsource=1.3 -O2 --encoding=UTF-8 --classpath="/usr/share/java/libgcj.jar:/builddir/pdftk-2.02-dist/java:." -c java_lib.jar
```
because there is a per command timeout. This parti is very heavyweight and takes like 5/10 seconds on my 4770k.

PR #1393, opened 2015-04-24
atom: fix .desktop file

thypon:


xtraeme:
Please add `nocross=yes`.

thypon:
Fail with i686; have i to disable it or there is some bug again?

Issue #1186, opened 2015-03-22
Package request: rng-tools


thypon:
http://sourceforge.net/projects/gkernel/

PR #1396, opened 2015-04-24
New package: rng-tools-5

thypon:
#1186

### DIFF
--- a/srcpkgs/rng-tools/files/rngd/run
+++ b/srcpkgs/rng-tools/files/rngd/run
@@ -1,0 +1,14 @@
+#!/bin/sh
+RNGD_OPTS=""
+[ -r conf ] && . ./conf
+
+a=0
+n=$(nproc)
+
+while [ $a -lt $n ]
+do
+	rngd $RNGD_OPTS
+	a=`expr $a + 1`
+done
+
+exec chpst -b rng-tools pause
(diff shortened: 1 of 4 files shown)


srcpkgs/rng-tools/files/rngd/run
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+. /etc/conf.d/rngd
+
+a=0
+n=$(nproc)
+
+while [ $a -lt $n ]
+do
+	rngd $RNGD_OPTS
+	a=`expr $a + 1`
```
xtraeme: `a=$((a+1))`

srcpkgs/rng-tools/template
```diff
@@ -0,0 +1,17 @@
...
+version=5
+revision=1
+build_style=gnu-configure
+makedepends="libgcrypt-devel"
+short_desc="Random number generator related utilities"
+maintainer="Andrea Brancaleoni <miwaxe@gmail.com>"
+license="GPL-3"
+homepage="http://sourceforge.net/projects/gkernel"
+distfiles="${SOURCEFORGE_SITE}/gkernel/${pkgname}-${version}.tar.gz"
+checksum=60a102b6603bbcce2da341470cad42eeaa9564a16b4490e7867026ca11a3078e
```
xtraeme: `/etc/conf.d/rng` is useless, it's empty... I would remove it completely and use `<service>/conf` instead.

thypon: Could you show an example?

xtraeme:
Sure, in `service/run`:

```
[ -r conf ] && . ./conf
... ${OPTS:=defopts} ...
```

PR #1400, opened 2015-04-24
New package: SDL_gfx-2.0.25

pullmoll:
Again, hopefully proper now.


pullmoll:
It looks like the comparison really has to be ${sourcepkg}>=${version}_${revision} to actually work. I stumbled upon this by trying to actually use SDL_gfx-devel in the coming srcpkgs/widelands/template.

xtraeme:
The comparision will work with a version comparator or a fixed version.

srcpkgs/SDL_gfx/template
```diff
@@ -0,0 +1,38 @@
+# Template file for 'SDL_gfx'
+pkgname=SDL_gfx
+version=2.0.25
+revision=1
+build_style=gnu-configure
+hostmakedepends="pkg-config"
+makedepends="SDL-devel"
+short_desc="Graphics drawing primitves for SDL"
```
xtraeme: typo "primitives"

pullmoll:
@xtraeme who or what adds the line `libSDL_gfx.so.2.0.25 SDL_gfx-2.0.25_2` to common/shlibs?
Should I commit it with the package that needs SDL_gfx?

dominikh:
The person contributing the package containing that file adds the line to the file. It's done in the commit that adds the package. I.e. you and unfortunately the commit that already got merged :) So yeah, fix that in a follow-up commit.

PR #1415, opened 2015-04-26
New package: widelands-18

pullmoll:
Forgot this addition to common/shlibs in the original pull request.


pullmoll:
Argh.. As expected now travis-ci timeouts because it hasn't seen output for more than 10 minutes.
Next try will be without redirecting stderr so that some warnings are printed :)

pullmoll:
Hmm.. it looks like now `common/travis/show_files.sh $BOOTSTRAP $ARCH` makes the output exceed the 4 megabytes limit and I can't do anything about it.

srcpkgs/widelands/files/widelands
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+# Wrapper script for widelands
+WL_DATA=/usr/share/games/widelans
```
xtraeme: widelans -> widelands

srcpkgs/widelands/template
```diff
@@ -0,0 +1,44 @@
...
+	( ${make_cmd} ${makejobs} ${make_build_args} ${make_build_target} ) > $_logfile
+}
+do_install() {
+	local _logfile
+	: ${make_cmd:=make}
+	: ${make_install_target:=install}
+	_logfile=/tmp/${pkgname}-${version}_${revision}-install.log
+	cd build
+	echo "- install output redirected to $_logfile"
+	( ${make_cmd} DESTDIR=${DESTDIR} ${make_install_args} ${make_install_target} ) > $_logfile
```
xtraeme: This `do_build` and `do_install` stuff is totally crap, remove this completely.

srcpkgs/widelands/template
```diff
@@ -0,0 +1,44 @@
...
+}
+post_install() {
+	vlicense COPYING
```
xtraeme: Again, if license is GPL there's no need to install license. The GPL licenses are already part of `base-files`.

srcpkgs/widelands/template
```diff
@@ -0,0 +1,44 @@
...
+	vinstall ${FILESDIR}/${pkgname}.desktop 644 usr/share/applications
+	vbin ${FILESDIR}/${pkgname}
+}
+do_clean() {
+	# FIXME: archive these logs somewhere?
+	rm -f -v /tmp/${pkgname}-${version}_${revision}-{build,install}.log
```
xtraeme: Ditto for `do_clean`.

xbps-src already creates logs for any phase in ${XBPS_STATEDIR}, aka `${XBPS_BUILDDIR}/.xbps-<pkgname>`.

A package should not do this manually.

srcpkgs/widelands/template
```diff
@@ -0,0 +1,44 @@
+# Template file for 'widelands'
+pkgname=widelands
+version=18
+revision=1
+wrksrc=${pkgname}-build${version}-src
+build_style=cmake
+configure_args="-DCMAKE_BUILD_TYPE=Release -DLUA_LIBRARY=/usr/lib/liblua5.1.a -DLUA_INCLUDE_DIR=/usr/include/lua5.1"
```
xtraeme: - CMAKE_BUILD_TYPE=Release is already set in `common/build-style/cmake.sh`.

srcpkgs/widelands/template
```diff
@@ -0,0 +1,44 @@
...
+hostmakedepends="cmake pkg-config"
+makedepends="boost-devel doxygen python minizip-devel libGL glew-devel ncurses-devel readline-devel SDL-devel SDL_gfx-devel SDL_image-devel SDL_net-devel SDL_mixer-devel SDL_ttf-devel lua51-devel libpng-devel gettext zlib-devel"
```
xtraeme: - I really doubt doxygen is required to build.
- python must be in hostmakedepends instead.
- remove libGL and gettext

srcpkgs/widelands/template
pullmoll: I did this just to make travis-ci not terminate the build due to verbosity. This build always hits the 4 megabytes limit.

xtraeme: travis-ci is not ready to handle such big packages. The source packages must not contain workarounds for travis.

common/hooks/post-install/03-remove-empty-dirs.sh
```diff
@@ -1,12 +1,14 @@
 # This hooks removes empty dirs and warns about them.
 
 hook() {
-	for f in $(find ${PKGDESTDIR} -type d -empty|sort -r); do
+	for f in $(find ${PKGDESTDIR} -type d -empty|sort -r|tr " " "|"); do
+		f=${f//|/ }
```
xtraeme: I'd want this var quoted, just in case.

srcpkgs/widelands/template
xtraeme: There's a lot of redundant dependencies in this line:

- readline-devel depends on ncurses-devel
- SDL_xxx-devel depends on SDL-devel

srcpkgs/widelands/template
pullmoll: Is there a way to tell travis-ci to not even try to build a package?

xtraeme: There's no way AFAIK.

srcpkgs/widelands/template
pullmoll: True. And I think png-devel depends on zlib-devel. I'll remove these and build locally before pushing the changes.

common/hooks/post-install/03-remove-empty-dirs.sh
xtraeme: Please make this one a separate pull request, I'd like to apply this independently.

srcpkgs/widelands/template
pullmoll: Ditto for SDL_mixer-devel. I guess I need to keep the SDL_xxx-devel in makedepends.

pullmoll: DUH! *facepalms* Of course the other way round: remove SDL-devel. I should get a coffee first :)

common/hooks/post-install/03-remove-empty-dirs.sh
pullmoll: Done. I'll also create a clean new pull request for widelands-18 after this.

chneukirchen:
FTR, you can use "xmindep pkg1 pkg2 pkg3" from xtools to get the minimal set of dependencies.

pullmoll:
@chneukirchen Great stuff. So using `xmindep $(grep makedepends= srcpkgs/mypkg/template|cut -d'"' -f2)` besides xlint is on my record.

PR #1418, opened 2015-04-26
libcgroup: add cgred service and group

bougyman:
* added system group cgred
* added service /etc/sv/cgred
* Two conf files, /etc/cgrules.conf and /etc/cgconfig.conf

### DIFF
--- a/srcpkgs/libcgroup/files/cgred/run
+++ b/srcpkgs/libcgroup/files/cgred/run
@@ -1,0 +1,5 @@
+#!/bin/sh
+# Default logs to syslog with facility DAEMON
+# man cgrulesengd for options list and descriptions.
+: ${OPTS:=-s}
+exec cgrulesengd $OPTS -g cgred --nodaemon 2>&1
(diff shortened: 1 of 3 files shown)


srcpkgs/libcgroup/files/cgred/run
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+[ -r ./conf ] && . ./conf
+exec cgrulesengd $OPTS -g cgred --nodaemon 2>&1
```
xtraeme: `${OPTS}` -> `${OPTS:=-s}` and remove `conf`. Those `conf` files should not be provided by pkgs unless it's really necessary.

srcpkgs/libcgroup/INSTALL
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+case ${ACTION} in
+post)
+	chown root:cgred /usr/bin/cgexec
```
xtraeme: Use relative paths please, otherwise it won't work with `xbps-install -r`.

xtraeme:
There's no libtool wrapper, what you are seeing is that libtool is trying to link against host libraries, rather than target libraries.

Try adding this:

```
hostmakedepends="flex libtool"
...
pre_configure() {
    libtoolize -f
}
```

xtraeme:
This is ok now :+1:

PR #1420, opened 2015-04-26
New Package: CuteMarkEd-0.11.0

### LINKED ISSUE #1401: Package request: cutemarked

logenkain:
closes #1401 

I want to point out line 5.

I did a "#?" to get rid of the "v" in the version string.  Is this fine?  Or is another way prefered to drop the first letter from the version string?

### DIFF
--- a/srcpkgs/CuteMarkEd/template
+++ b/srcpkgs/CuteMarkEd/template
@@ -1,0 +1,27 @@
+# Template file for 'CuteMarkEd'
+pkgname=CuteMarkEd
+version=v0.11.0
+revision=1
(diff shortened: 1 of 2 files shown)


logenkain:
Looks like arm can't cross-compile qt5 dev files.  Moving the dev files to hostmakedepends seems to almost work, but arm doesn't seem to have discount-devel either.

logenkain:
Unless someone tells me how to make it work, does this mean I should add:
only_for_archs="i686 x86_64"

xtraeme:
- `${var#?}` is ok
- Use `nocross=yes` rather than `only_for_archs`

PR #1421, opened 2015-04-26
common/environment: set FORCE_UNSAFE_CONFIGURE=1 to make some builds work with unshare.

### LINKED ISSUE #1403: tar: build fails with unshare
It needs `configure_args="... FORCE_UNSAFE_CONFIGURE=1"` or configure will fail; at least with `XBPS_CHROOT=proot`.

Gottox:
fixes #1403


xtraeme:
I would prefer if the `unshare` chroot-style would change uid/gid to the current user instead.

Gottox:
True. But for that, we need su/sudo in masterdir. What do you think
about dropping support for unshare? Its use cases are covered by proot too.

xtraeme:
:+1:

PR #1425, opened 2015-04-27
xbps-src: fix handling of path names containing spaces

pullmoll:


common/hooks/post-install/03-remove-empty-dirs.sh
```diff
@@ -1,7 +1,8 @@
 # This hooks removes empty dirs and warns about them.
 
 hook() {
-	for f in $(find ${PKGDESTDIR} -type d -empty|sort -r); do
+	for f in $(find ${PKGDESTDIR} -type d -empty|sort -r|tr " " "|"); do
+		f="${f//|/ }"
```
Gottox: How about this instead:
```bash
find "${PKGDESTDIR}" -type d -empty|sort -r | while read f; do
    ...
done
```

pullmoll: Not sure if this is sufficient. I tried similar, but with -print0 and sort -z, and that failed...

PR #1426, opened 2015-04-27
New package: gsimplecal-2.1

jarbowski:


dominikh:
**Open** or **closed**, not both, thanks. This is just pointless spam.

jarbowski:
@dominikh 
Hi, I'm sorry but I don't know what to do: must I create two different branches for the gtk2 and gtk3 versions of gsimplecal-2.1 or one pull request for both is good?

dominikh:
Do we need both the gtk2 and gtk3 versions? @xtraeme, what's our position regarding GTK2 vs GTK3?

If we do want both, they go in the same branch/PR.

Neither of that requires constantly closing and opening the PR, however. Just force push your changes to your fork and that's it. Also rebase and squash your commits, there's no reason to have all 10 attempts in the history.

xtraeme:
@dominikh in an ideal world would be nice to have non-conflicting gtk2 and gtk3 packages... but I don't have a preference. I let the package submitter decide what's best.

jarbowski:
I did this because the option exits. It's always good to have choice.

srcpkgs/gsimplecal-gtk2/template
```diff
@@ -0,0 +1,20 @@
...
+wrksrc="${pkgname%-*}-${version}"
+build_style=gnu-configure
+configure_args="--enable-gtk2"
+makedepends="gtk+-devel"
+hostmakedepends="automake pkg-config"
+conflicts="gsimplecal-gtk3>=0"
+short_desc="Simple and lightweight GTK calendar"
+maintainer="Jarbowski <jarbowski@gmail.com>"
+license="BSD"
+homepage="http://dmedvinsky.github.com/$pkgname"
```
Duncaen: homepage is wrong

xtraeme:
I won't apply this PR with 20 commits not even following the commit rules.

Please rebase and submit a proper commit.

PR #1427, opened 2015-04-27
New package: cuda-7.0.28

tonkazoid:
sorry for the fuss

### DIFF
--- a/srcpkgs/miruo/template
+++ b/srcpkgs/miruo/template
@@ -1,0 +1,17 @@
+# Template file for 'miruo'
+pkgname=miruo
+version=0.9.6
+revision=1
+build_style=gnu-configure
+makedepends="libpcap-devel"
+short_desc="Pretty-print TCP session monitor/analyzer"
+maintainer="Christian Neukirchen <chneukirchen@gmail.com>"
+license="GPL-3"
+homepage="https://github.com/KLab/miruo"
+distfiles="https://github.com/KLab/${pkgname}/archive/${version}.tar.gz"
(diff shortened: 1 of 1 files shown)


srcpkgs/cuda/template
```diff
@@ -0,0 +1,42 @@
+# Template file for 'cuda'
+pkgname=cuda
+version=7.0.28
+revision=1
+maintainer="Georg Schabel <gescha@posteo.de>"
+homepage="http://www.nvidia.com/object/cuda_home.html"
+short_desc="NVIDIA's GPU programming toolkit"
+license="Propietary NVIDIA license"
```
Gottox: typo: "proprietary"

Gottox:
Does this licence allow redistribution?

tonkazoid:
http://docs.nvidia.com/cuda/eula/index.html#redistribution-rights
http://docs.nvidia.com/cuda/eula/index.html#linux-freebsd-exception

Oh and the "propietary". Blindly Copied from here https://github.com/voidlinux/void-packages/blob/ff31d5a1a200f252a6a14db42eef2269265199fb/srcpkgs/nvidia/template#L9

xtraeme:
This can't be applied again, no idea why.

tonkazoid:
Strange.. 
Another thing i've noticed: I'm currently using it without a problem in blender. But after the first installation i need to start blender as root once for the GPU/CPU selection to show up for all users. I don't know if i can fix this with the template.

tonkazoid:
most people fix the "run as root once"-problem with nvidia-modprobe. added it to the nvidia package and removed the libavcodec53 from shlibs, this package doesn't exist anymore.

PR #1440, opened 2015-04-28
rofi: updated to 0.15.4

hanspolo:


srcpkgs/rofi/template
```diff
@@ -1,16 +1,16 @@
 # Template file for 'rofi'
 pkgname=rofi
-version=0.15.2
+version=0.15.4
 revision=1
 build_style=gnu-configure
 hostmakedepends="pkg-config"
-makedepends="pango-devel libXinerama-devel libXft-devel"
+makedepends="pango-devel libXinerama-devel libXft-devel autoconf"
```
xtraeme: autoconf must be in hostmakedepends instead.

Issue #1355, opened 2015-04-20
Package Request: ttf-inconsolata


Wqer555:
A fantastic programming font. 

http://levien.com/type/myfonts/inconsolata.html

logenkain:
On Mon, Apr 20, 2015 at 03:05:02PM -0700, Wqer555 wrote:
> http://levien.com/type/myfonts/inconsolata.html

It is a nice looking font, but here is a quote from that page:

"It will be released under the new Open Font License of SIL."

This tells me that it technically doesn't have a license right now, I
would feel uncomfortable trying to put it into any distribution without
contacting the author first.

Not saying we can't do it, but I want to bring it to the attention of
anyone who decides to package it.

chneukirchen:
logenkain <notifications@github.com> writes:

> On Mon, Apr 20, 2015 at 03:05:02PM -0700, Wqer555 wrote:
>> http://levien.com/type/myfonts/inconsolata.html
>
> It is a nice looking font, but here is a quote from that page:
>
> "It will be released under the new Open Font License of SIL."
>
> This tells me that it technically doesn't have a license right now, I
> would feel uncomfortable trying to put it into any distribution without
> contacting the author first.
>
> Not saying we can't do it, but I want to bring it to the attention of
> anyone who decides to package it.

Debian packages this as Open Font License SIL.  Should be fine.

-- 
Christian Neukirchen  <chneukirchen@gmail.com>  http://chneukirchen.org

PR #1445, opened 2015-04-28
New Package: inconsolata-fonts-otf-001.010

tonkazoid:
for #1355


chneukirchen:
Please rename this font-inconsolata-otf.

srcpkgs/font-inconsolata-otf/template
```diff
@@ -0,0 +1,28 @@
+# Template build file for 'font-inconsolata-otf'.
+pkgname=font-inconsolata-otf
+version=001.010
+revision=1
+maintainer="Georg Schabel <gescha@posteo.de>"
+short_desc="Sans-serif monotype font designed for code listings"
+homepage="http://www.levien.com/type/myfonts/inconsolata.html"
+license="OFL-1.1"
```
xtraeme: I'd prefer to move the license to base-files. This license seems generic enough to not duplicate it in multiple packages.

srcpkgs/font-inconsolata-otf/template
```diff
@@ -0,0 +1,26 @@
...
+
+distfiles="http://levien.com/type/myfonts/Inconsolata.otf"
+checksum=1561e616c414a1b82d6e6dfbd18e5726fd65028913ade191e5fa38b6ec375a1a
+
+makedepends="font-util xbps-triggers"
```
xtraeme: This pull request is ok, but there's only a minor nit: `makedepends` is unnecessary because `font-util` is only necessary at run-time.

srcpkgs/font-inconsolata-otf/template
```diff
@@ -0,0 +1,26 @@
...
+depends="${makedepends}"
+noarch="yes"
+build_style=fetch
+font_dirs="/usr/share/fonts/OTF"
+
+do_install() {
+	vmkdir usr/share/fonts/OTF
+	install -m644 ${XBPS_SRCDISTDIR}/${pkgname}-${version}/Inconsolata.otf \
```
xtraeme: Also if you use `build_style=fetch` the `distfiles` are copied to `$wrksrc`.

PR #1452, opened 2015-04-28
steam: Moved 64 bit package to multilib/nonfree

teh-jazzman:
I hope moving the package from one repo to another does not screw anything up.

Can we be reasonably sure that users of `multilib/nonfree` also use `multilib`? If not, would a package dependency from `multilib/nonfree` on `multilib` be feasable?


Gottox:
Can you explain the reason for moving the package into another repo?

teh-jazzman:
The 64 bit package depends on 32 bit packages from the `multilib` repository. Those would need to be installed manually since the `multilib` is not neccessarily activated and thus the dependencies could not automatically be installed. Moving the package itself to `multilib` would resolve that problem. Archlinux is doing the same thing as can be seen [here](https://www.archlinux.org/packages/?q=steam) (i686 in Community, x86_64 in Multilib).

Unfortunately, since `multilib/nonfree` is the "right" place for it, it would be good if there was a way to make sure that `multilib` is active as well.

PR #1454, opened 2015-04-29
New package: beep-1.3

thypon:


srcpkgs/beep/INSTALL
```diff
@@ -0,0 +1,6 @@
+# INSTALL
+case "$ACTION" in
+post)
+	/usr/sbin/setcap cap_dac_override,cap_sys_tty_config+ep /usr/bin/beep
```
xtraeme: Do not hardcode path to setcap please.

srcpkgs/beep/template
```diff
@@ -0,0 +1,16 @@
...
+short_desc="Beep music composer"
+maintainer="Andrea Brancaleoni <miwaxe@gmail.com>"
+license="GPL-3"
+depends="libcap-progs"
+homepage="http://www.johnath.com/beep/"
+distfiles="http://www.johnath.com/${pkgname}/${pkgname}-${version}.tar.gz"
+checksum=59acef7a987de5557cefd1a904666cc2691f132929af39e65450b182a581ec2d
+
+do_install() {
+	vbin beep
```
xtraeme: No manpage?

thypon: With vman yield a strange error saying it can't parse the man. I can install with vinstall but it's more a workaround that a real solution.

srcpkgs/beep/INSTALL
thypon: It's not the first time you say this thing about not hardcoding path. Maybe we could write something about into Manual.md. I feel more secure to call /usr/sbin/setcap instead of setcap generally.

xtraeme: Hardcoding the path might not use the cmd from rootdir.

srcpkgs/beep/template
xtraeme: I hate pkgs that don't have manual pages. If there's none in the official distribution, fine, otherwise I'd like to have a manpage :-)

thypon: There is, but i think there is a problem with `vman`. If you don't bother with `vinstall` i'll go with it.

chneukirchen: Just gunzip the manpage first.

thypon: In what phase?

xtraeme: do_install

PR #1462, opened 2015-04-30
New package: scantailor-0.9.11.1

pullmoll:


srcpkgs/scantailor/template
```diff
@@ -0,0 +1,26 @@
+# Template file for 'scantailor'
+pkgname=scantailor
+version=0.9.11.1
+revision=1
+wrksrc=${pkgname}-RELEASE_${version//./_}
+build_style=cmake
+hostmakedepends="cmake pkg-config"
+makedepends="libpng-devel tiff-devel boost-devel qt-devel libXrender-devel"
+depends="libGL"
```
xtraeme: This dependency is wrong.

srcpkgs/scantailor/template
```diff
@@ -0,0 +1,26 @@
+# Template file for 'scantailor'
+pkgname=scantailor
+version=0.9.11.1
+revision=1
+wrksrc=${pkgname}-RELEASE_${version//./_}
+build_style=cmake
+hostmakedepends="cmake pkg-config"
```
xtraeme: Add qt-qmake and you can remove probably `pre_configure`.

pullmoll: Here on `x86_64` building the package with `-a i686` fails because qmake isn't found. On travis-ci it builds for i686.

`xmindep` here tells me that `qt-qmake` is redundant. So perhaps in those cases, qmake dependent builds, I just always have to use `-m masterdir-x86` and set `nocross=y`?

pullmoll:
This seems to do the trick: `hostdepends+="qt-devel qt-qmake"`. I can now build it with `-a i686`. The arm* family still fails because of `=> ERROR: qt-devel-4.8.6_12: cannot be cross compiled, exiting...`

PR #1466, opened 2015-04-30
talloc: fix compilation on musl

jprjr:


xtraeme:
The patch should be added to /patches instead, this way you don't need to apply it manually (in -p0 format).

jprjr:
Done and done, fixed+rebased

PR #1467, opened 2015-04-30
libXvMC: unbreak musl cross compilation

jprjr:


xtraeme:
Please bump revision.

jprjr:
Can do! Just pushed a rebased commit with the revision bumped.

PR #1469, opened 2015-04-30
armv5te: fix build problems

thypon:
I'm trying to build `base-system`. Right now i build 115 packages without problems: i have problem with libressl.

Right now i have this issue that seems to be fixed in upstream: maybe i could apply the diff related to that commit:
https://github.com/libressl-portable/portable/issues/41


xtraeme:
The cross-vpkg-dummy is ok, the coreutils one is completely wrong.

thypon:
Without reverting that commit in da2d731 ./xbps-src fails to find `coreutils-doc_package` function and fails itself. Reverting it doesn't create coreutils-doc package and doesn't fail to build it.

thypon:
```
index: added `coreutils-8.23_6' (armv5tel).
index: 25 packages registered.
=> coreutils-8.23_6: removing autodeps, please wait...
=> coreutils-8.23_6: removing autocrossdeps, please wait...
=> coreutils-8.23_6: cleaning build directory...
=> coreutils: removing files from destdir...
=> ERROR: coreutils-doc: missing coreutils-doc_package() function!
```

This is the error right now.

srcpkgs/cross-vpkg-dummy/template
```diff
@@ -1,7 +1,7 @@
 # Template file for 'cross-vpkg-dummy'
 pkgname=cross-vpkg-dummy
 version=0.17
-revision=1
+revision=2
```
xtraeme: Please bump version instead, rather than revision.

xtraeme:
This commit https://github.com/voidlinux/void-packages/commit/fdc356cb607a68c3be1cb7f6c3bcec373dcc0920 makes your glib changes unnecessary, remove them.

srcpkgs/glibc/template
```diff
@@ -58,7 +58,7 @@ do_configure() {
 		i686) export CFLAGS+=" -mno-tls-direct-seg-refs";;
 		# Force hard float ABI.
 		# To build for soft float: --with-float=soft --without-fp.
-		armv5*) configure_args+=" --with-march=armv5te --without-fp --with-float=soft";;
+		armv5*) CFLAGS='-O2 -pipe';;
```
xtraeme: what's the reason for this? this makes no sense at all.

xtraeme:
The cross compiler, cross profile and relevant pkgs now build correctly for armv5tel.

PR #1470, opened 2015-05-01
qt5: Create a qt5-doc package to not throw away /usr/share/doc/qt5

pullmoll:
Cosmetics: use a sub packages description style where the first letter is always uppercase.

There was a mix of lowercase and uppercase first letters for the sub packages, which I thought is a bit inadequate for a matured distribution like void linux.

The reason to keep /usr/share/doc/qt5 is to try to modify qtcreator to also install docs with `make docs` and later `make install_docs` targets. The latter is not working right yet, because some `qthelp` or `qhelpgenerator` binary seems to be missing. I'm on the hunt for this in the depths of qt5, though.

In any case qtcreator without the *.qch help files is not really complete. You can't use F1 or context sensitive help without those files. I'm not entirely sure aboute the doxygen hostmakedepends. Have to double-check the logs to see if it is used at all... sorry to be a PITA :)

Intention is to make qtcreator depend on qt5-doc and add `make_build_args="all doc"` and in do_install() `make INSTALL_ROT=$(DESTDIR)/usr install install_docs` to get everything in place.


xtraeme:
Looks ok, but you did not revbump it.

pullmoll:
Hold on. Docs are not yet built, just the configuration files. I need to get qdoc and qhelgenerator installed in masterdir/usr/bin somehow.

pullmoll:
I tink this should work now, but please take a look at qt5-doc_package() if I did it right.

srcpkgs/qt5/template
```diff
@@ -116,8 +115,28 @@ qt5-declarative_package() {
...
+	short_desc+=" - Documentation"
+	# This requires qdoc from qtbase and qhelpgenerator from qttools
+	depends="qt5-${version}_${revision} qt5-tools-${version}_${revision}"
+	pkg_install() {
+		cd ${wrksrc}
+		vmove usr/share/doc
+		# Temporarily install binaries qdoc and qhelpgenerator
+		mkdir -p /usr/lib/qt5/bin
+		install -m755 qtbase/bin/qdoc /usr/lib/qt5/bin
+		install -m755 qttools/bin/qhelpgenerator /usr/lib/qt5/bin
```
xtraeme: I'd prefer you use a temporary directory in `${wrksrc}` instead, rather than masterdir, otherwise it could be polluted.

pullmoll: In that case I'll have to hack around the Makefiles, because `qdoc` is run as explicitly `/usr/lib/qt5/bin/qdoc`. Otherwise I would just have changed PATH. I'll see what I can come up with.

pullmoll: DUH! It comes to me I could try and symlink both files in a single directory below ${wrksrc}, then overriding QT_INSTALL_BIN could work.

xtraeme:
This is even worse than it was before :-)

What you have to do is simply create a tempdir in ${wrksrc}, setup the symlinks on it, set LD_LIBRARY_PATH to that tempdir and run `make installdocs` or whatever.

I don't get why you need to do this complicated dance.

pullmoll:
xtraeme said:
> I  don't get why you need to do this complicated dance.

Proably because I'm a noob :) I still lack full overview regarding xbps-src and what part of the file hierarchy is accesible when and where. Retrying following your hints now.

pullmoll:
Ok, I stick up on this problem because I can't get it working. It's beyond my skills.
Here is the latest try which still fails when qhelpgenerator doesn't find the libqminimal.so plugin at some stage. Going into the chroot and trying to run qhelpgenerator with the same LD_LIBRARY_PATH prepended does work, i.e. it finds the plugin and the required libraries... 
```shell
qt5-doc_package() {
	short_desc+=" - Documentation"
	# This requires qdoc from qtbase and qhelpgenerator from qttools
	depends="qt5-tools-${version}_${revision} qt5-plugin-sqlite-${version}_${revision}"
	pkg_install() {
		cd $wrksrc
		vmove usr/share/doc
		# Original (installed) binaries directory
		local qtbin=/usr/lib/qt5/bin
		# Temporary binaries directory
		local tmpbin=$wrksrc/docbuild/bin
		local ld_extra="$wrksrc/qtbase/lib:$wrksrc/qttools/lib"
		local plugins="$wrksrc/qtbase/plugins"
		mkdir -p $tmpbin/platforms
		mkdir -p $tmpbin/sqldrivers
		ln -sfv $wrksrc/qtbase/bin/qdoc $tmpbin/qdoc
		ln -sfv $wrksrc/qttools/bin/qhelpgenerator $tmpbin/qhelpgenerator
		ln -sfv $plugins/platforms/libqminimal.so $tmpbin/platforms/libqminimal.so
		ln -sfv $plugins/sqldrivers/libsqlite.so $tmpbin/sqldrivers/libsqlite.so
		# Patch Makefiles to refer to $tmpbin/{qdoc,qhelpgenerator}
		for m in $(find . -name "Makefile*"); do
			mv $m $m.orig
			sed $m.orig \
				-e "s;$qtbin/qdoc;$tmpbin/qdoc;g" \
				-e "s;$qtbin/qhelpgenerator;$tmpbin/qhelpgenerator;g" \
				> $m
		done
		export LD_LIBRARY_PATH="$ld_extra:$LD_LIBRARY_PATH"
		make ${makejobs} docs
		make INSTALL_ROOT=$PKGDESTDIR install_docs
		# Restore original Makefiles just in case...
		for m in $(find . -name "Makefile*.orig"); do
			mv $m ${m%.orig}
		done
	}
}
```
The *right* solution would be to install qt5-tools and qt5-plugin-sqlite, untar the sources again and only then `make docs`. According to the [Qt Wiki](http://wiki.qt.io/Building_Qt_Documentation) the docs can be built *after installation* and I can't manage to fake a (partial) installation.

I don't know if it is feasible to to write a qt-5.INSTALL which downloads the source tarball, extracts it and runs `make docs` inside. That should work without problems.

pullmoll:
Finally got it working. This was truly an odyssey ;)

PR #1475, opened 2015-05-01
new package: argp-standalone

jprjr:
This is an implementation of argp, for libc implementations that lack it (musl)

### DIFF
--- a/srcpkgs/argp-standalone/template
+++ b/srcpkgs/argp-standalone/template
@@ -1,0 +1,19 @@
+# Template file for 'argp-standalone'
+pkgname=argp-standalone
+version=1.3
+revision=1
+build_style=gnu-configure
+short_desc="Implementation of ARGP"
+maintainer="John Regan <john@jrjrtech.com>"
+homepage="https://www.freshports.org/devel/argp-standalone/"
+license="Public Domain"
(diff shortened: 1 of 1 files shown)


xtraeme:
There's no reason to name it `foo-devel`, just name it `argp-standalone`.

jprjr:
No problem, just pushed a new, rebased commit with the correct name

xtraeme:
- build_pie is unnecessary.
- wrksrc is set by default to `${pkgname}-${version}`, redundant.
- nostrip is unnecessary.

jprjr:
Okey-doke - fix+rebased

srcpkgs/argp-standalone/template
```diff
@@ -0,0 +1,19 @@
...
+revision=1
+build_style=gnu-configure
+short_desc="Implementation of ARGP"
+maintainer="John Regan <john@jrjrtech.com>"
+homepage="https://www.freshports.org/devel/argp-standalone/"
+license="Public Domain"
+distfiles="http://www.lysator.liu.se/~nisse/misc/argp-standalone-${version}.tar.gz"
+checksum=dec79694da1319acd2238ce95df57f3680fea2482096e483323fddf3d818d8be
+
+export CFLAGS="-fPIC"
```
xtraeme: There's no need to use `export`.

jprjr:
Removed `export`

Issue #720, opened 2014-12-12
Enable Vala support for CTAGS package


radare:
See https://aur.archlinux.org/packages/ct/ctags-vala/PKGBUILD

chneukirchen:
Upsteam source for this patch is dead.

Upstream bug: https://github.com/universal-ctags/ctags/issues/621

We cannot fix this in Void currently.

PR #1477, opened 2015-05-01
ctags: enable vala support

thypon:
#720


chneukirchen:
I can't see any vala support in ctags.

thypon:
My fault. A patch is needed. Close for now.

PR #1482, opened 2015-05-02
pulseaudio: armv5te/aarch64 fix

thypon:


Gottox:
Looks okay to me. I'm not sure if a revbump is needed here.

PR #1488, opened 2015-05-02
add pool import service

pulux:
This is necessary  for LUKS encrypted pools.

### DIFF
--- a/srcpkgs/zfs/files/zfs-import/finish
+++ b/srcpkgs/zfs/files/zfs-import/finish
@@ -1,0 +1,5 @@
+#!/bin/sh
+set -e
+exec 2>&1
+zfs unshare -a
+zfs unmount -a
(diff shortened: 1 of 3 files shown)


srcpkgs/zfs/files/zfs-import/run
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+exec 2>&1
+zpool import -a
+exec /sbin/pause
```
xtraeme:
```
#!/bin/sh
exec 2>&1
zpool import -a || exit 1
exec chpst -b zpool-import pause
```

xtraeme:
For your next commits please follow the commit message rules: `"<pkgname>: ..."`.

Thanks.

srcpkgs/zfs/files/zfs-import/finish
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+exec 2>&1
+zfs unshare -a
```
xtraeme: I would rework this one as:
```
#!/bin/sh
set -e
...
```

This way if `zfs unshare -a` works won't go through the next command without returning error.

PR #1489, opened 2015-05-02
sdlmame: update to 0.161

pullmoll:
+ Depend on lua-devel for new build system (genie)
+ Fix xlint warnings
+ Add sub package sdlmess for Multi Emulator Super System


xtraeme:
sdlmame now owns `/usr/share/man/man6/mess.6` which should be owned by `sdlmess`. Please fix.

pullmoll:
Ok. I saw I forgot tools `unidasm` (mame) and `imgtool` (mess). Will move the mess tools man pages as well. It'll take some time to test build here :)

PR #1490, opened 2015-05-02
New package: jpeginfo-1.6.1

bougyman:


bougyman:
Merged without the double commit from CLI.

PR #1492, opened 2015-05-03
tar: added xattrs/acl support

thypon:


chneukirchen:
+1 if this works.

chneukirchen:
(Still, xbps uses libarchive to extract tar.)

thypon:
:-S, libarchive support xattrs like bsdtar, but i think it's not saved in a compat way to how GNU tar saves them. I have to check.

srcpkgs/tar/template
```diff
@@ -1,15 +1,16 @@
...
 pkgname=tar
 version=1.28
-revision=5
+revision=6
 bootstrap=yes
 build_pie=yes
 build_style=gnu-configure
 configure_args="gl_cv_struct_dirent_d_ino=yes"
 conflicts="chroot-tar>=0"
+makedepends="attr-devel acl-devel"
```
xtraeme: acl-devel depends on attr-devel, so that attr-devel is redundant.

PR #1493, opened 2015-05-03
thinkfan: add service script

pulux:
add the service script for thinkfan, configuration must be create by user

### DIFF
--- a/srcpkgs/thinkfan/template
+++ b/srcpkgs/thinkfan/template
@@ -1,7 +1,7 @@
 # Template file for 'thinkfan'
 pkgname=thinkfan
 version=0.9.2
-revision=1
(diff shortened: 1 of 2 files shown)


srcpkgs/thinkfan/template
```diff
@@ -26,3 +26,7 @@ do_install() {
 	vmkdir usr/lib/modprobe.d
 	vinstall ${FILESDIR}/thinkpad_acpi.conf 644 usr/lib/modprobe.d
 }
+
+post_install() {
+	vsv thinkfan
```
xtraeme: There's no need to use `post_install`, move `vsv` at the end of `do_install()`.

PR #1494, opened 2015-05-03
Tomahawk qt5

Duncaen:
quazip and qca and their qt5 version have conflicting so names, is there a better way to avoid conflicts?


xtraeme:
There's no other way to avoid conflicting sonames. Multiple sonames can be declared in `common/shlibs` and will work if any of them matches the existing pkgname.

xtraeme:
Looks good :+1:

PR #1495, opened 2015-05-03
disable OTG on Linux

Gottox:
This solves issues with the USB-Armory, see https://gist.github.com/Gottox/b7e7b39541f8dad81cc5

USB OTG support only useful on non-desktop devices. At least Debian and Arch have them disabled.


thypon:
Is there something to disable from the linux command line?

xtraeme:
Seems reasonable :+1:

PR #1501, opened 2015-05-04
elfutils: update to latest, fix compilation for musl libc

jprjr:
This depends on #1475 

Most of the patches are from OpenWRT's patchset here: https://github.com/openwrt/packages/tree/master/libs/elfutils/patches along with the Musl-Cross patches from here: https://github.com/GregorR/musl-cross/blob/master/patches/patch-elfutils.sh

This is my first time doing a big patch like this, apologies in advance if I missed anything or did anything incorrectly. I'm using the `$XBPS_TARGET_MACHINE` variable to check if the system is musl-based or not, no idea if that's the right way to check. I expect I'll need to make some changes before this gets merged.

### DIFF
--- a/srcpkgs/elfutils/template
+++ b/srcpkgs/elfutils/template
@@ -1,6 +1,6 @@
 # Template file for 'elfutils'
 pkgname=elfutils
-version=0.160
+version=0.161
 revision=1
 build_style=gnu-configure
 configure_args="--program-prefix=eu-"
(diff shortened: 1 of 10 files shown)


srcpkgs/elfutils/template
```diff
@@ -11,7 +11,20 @@ maintainer="Juan RP <xtraeme@voidlinux.eu>"
 license="GPL-3"
 homepage="https://fedorahosted.org/elfutils/"
 distfiles="https://fedorahosted.org/releases/e/l/${pkgname}/${version}/${pkgname}-${version}.tar.bz2"
-checksum=741b556863c069ceab2d81eb54aeda8c34f46728859704eaf9baef8503e9a9d1
+checksum=570c91a1783fa5386aaa2dfdd08dda1de777c2b63bf3b9c1437d635ffdd7a070
+
+if [ -n "$XBPS_TARGET_MACHINE" ]; then
+    case "$XBPS_TARGET_MACHINE" in
```
xtraeme: You can simplify this part as:

```
*-musl) ...
esac
```

jprjr:
Simplified + pushed

xtraeme:
As you can see, the build failed on travis. This could be fixed just adding:

```
CFLAGS="-Wno-error"
```
to the template.

xtraeme:
Or specifically for this `strerror_r` warning:

```
CFLAGS="-Wno-unused-result"
```

jprjr:
Added+rebased

PR #1506, opened 2015-05-05
New package: freefall-4.d74aae4e

GiedriusS:


srcpkgs/freefall/template
```diff
@@ -0,0 +1,24 @@
+# Template file for 'freefall'
+pkgname="freefall"
+version="4.d74aae4e"
```
Gottox: where does this version come from?

srcpkgs/freefall/template
```diff
@@ -0,0 +1,24 @@
...
+revision=1
+create_wrksrc=yes
+short_desc="Disk protection for HP/DELL machines"
+maintainer="Giedrius Statkevičius <giedriuswork@gmail.com>"
+license="GPL-2"
+homepage="http://www.kernel.org"
+checksum="6fad3e93be9fdb5fbf87ced466961d0c3190614bf5e8efdb0fa20071f4dc06ba"
+distfiles="https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/plain/Documentation/laptops/freefall.c"
```
Gottox: The package will break as soon as this file is updated. Use a specific revision of this file.

srcpkgs/freefall/INSTALL.msg
```diff
@@ -0,0 +1,2 @@
+By default freefall service uses /dev/sda so edit /etc/sv/freefall/run and add
+argument to freefall if you want to use a different device
```
Gottox: Not a good idea, xbps will overwrite the file on the next update.

srcpkgs/freefall/files/freefall/run
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+exec freefall
```
Gottox: add ${OPTS} to allow custom arguments. See: https://github.com/GiedriusS/void-packages/blob/master/srcpkgs/mongodb/files/mongodb/run for an example

xtraeme:
The run script should be something like this:

```
#!/bin/sh
[ -r conf ] && . ./conf
exec freefall ${OPTS:=/dev/sda}
```

PR #1507, opened 2015-05-05
llvm: fix cross-compilation under musl

jprjr:
This uses patches from Alpine Linux's LLVM package, with the Alpine-specific stuff removed: http://git.alpinelinux.org/cgit/aports/tree/testing/llvm

### DIFF
--- a/srcpkgs/llvm/files/patches/cfe/cfe-004-add-musl-triples.patch
+++ b/srcpkgs/llvm/files/patches/cfe/cfe-004-add-musl-triples.patch
@@ -1,0 +1,37 @@
+diff --git lib/Driver/ToolChains.cpp lib/Driver/ToolChains.cpp
+index 07fc182..232f99e 100644
+--- lib/Driver/ToolChains.cpp
++++ lib/Driver/ToolChains.cpp
+@@ -1296,7 +1296,10 @@ bool Generic_GCC::GCCInstallationDetector::getBiarchSibling(Multilib &M) const {
+   static const char *const ARMTriples[] = { "arm-linux-gnueabi",
+                                             "arm-linux-androideabi" };
+   static const char *const ARMHFTriples[] = { "arm-linux-gnueabihf",
+-                                              "armv7hl-redhat-linux-gnueabi" };
++                                              "armv7hl-redhat-linux-gnueabi",
++                                              "armv6-linux-musleabihf",
++                                              "armv7-linux-musleabihf",
++                                              "arm-linux-musleabihf" };
+   static const char *const ARMebLibDirs[] = { "/lib" };
+   static const char *const ARMebTriples[] = { "armeb-linux-gnueabi",
+                                               "armeb-linux-androideabi" };
+@@ -1308,7 +1311,8 @@ bool Generic_GCC::GCCInstallationDetector::getBiarchSibling(Multilib &M) const {
+     "x86_64-linux-gnu", "x86_64-unknown-linux-gnu", "x86_64-pc-linux-gnu",
+     "x86_64-redhat-linux6E", "x86_64-redhat-linux", "x86_64-suse-linux",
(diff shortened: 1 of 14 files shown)


srcpkgs/llvm/files/patches/cfe/cfe-005-fix-dynamic-linker-paths.patch
```diff
@@ -0,0 +1,58 @@
...
+     return "/lib/ld-linux-aarch64_be.so.1";
+   else if (ToolChain.getArch() == llvm::Triple::arm ||
+            ToolChain.getArch() == llvm::Triple::thumb) {
+-    if (ToolChain.getTriple().getEnvironment() == llvm::Triple::GNUEABIHF)
++    switch (ToolChain.getTriple().getEnvironment()) {
++    case llvm::Triple::Musl:
++    case llvm::Triple::MuslEABI:
++      return "/lib/ld-musl-arm.so.3";
++    case llvm::Triple::MuslEABIHF:
++      return "/lib/ld-musl-armhf.so.3";
```
xtraeme: These are wrong, should be `ld-musl-armhf.so.1` instead. We don't support ARM softfloat with musl currently, so that I'm not sure about the softfloat one, I would expect it's wrong too.

srcpkgs/llvm/files/patches/cfe/cfe-004-add-musl-triples.patch
```diff
@@ -0,0 +1,36 @@
...
+--- lib/Driver/ToolChains.cpp
++++ lib/Driver/ToolChains.cpp
+@@ -1296,7 +1296,9 @@ bool Generic_GCC::GCCInstallationDetector::getBiarchSibling(Multilib &M) const {
+   static const char *const ARMTriples[] = { "arm-linux-gnueabi",
+                                             "arm-linux-androideabi" };
+   static const char *const ARMHFTriples[] = { "arm-linux-gnueabihf",
+-                                              "armv7hl-redhat-linux-gnueabi" };
++                                              "armv7hl-redhat-linux-gnueabi",
++                                              "armv6-linux-musleabihf",
++                                              "armv7-linux-musleabihf"};
```
xtraeme: Please add `arm-linux-musleabihf` to match our future musl cross compiler target.

Right now it's `arm-linux-musleabi` but I'll change this soon to `arm-linux-musleabihf`.

srcpkgs/llvm/files/patches/compiler-rt/compiler-rt-001-musl-disable-sanitizers.patch
```diff
@@ -0,0 +1,25 @@
+diff --git lib/Makefile.mk lib/Makefile.mk
+index ed9690d..aacb2b1 100644
+--- lib/Makefile.mk
++++ lib/Makefile.mk
+@@ -10,13 +10,13 @@
+ SubDirs :=
+ 
+ # Add submodules.
+-SubDirs += asan
++# SubDirs += asan
```
xtraeme: Does that mean now that those submodules are also disabled for glibc?

xtraeme:
I could only find the issues mentioned above, well done.

Thanks

srcpkgs/llvm/files/patches/compiler-rt/compiler-rt-001-musl-disable-sanitizers.patch
jprjr: I think so. Maybe only apply that patch with the target is musl?

xtraeme: Yes, please.

xtraeme:
@jprjr are you on IRC? please join irc.freenode.net #xbps

jprjr:
Fixed + rebased

PR #1509, opened 2015-05-05
xbps-src: Implement XBPS_DISTFILES_MIRROR

### LINKED ISSUE #1372: Idea: Implement XBPS_DISTFILES_MIRROR
To reduce download times and impact on many source servers, it would help if a (list of) mirror site(s) for source packages could be provided through etc/conf.

For example for me `http://mirror.netcologne.de/gentoo/distfiles` carries many of the source files required to build packages. Another example may be some source file repository in a local `file:///some/where/distfiles` path or some local network http or ftp server.

The location(s) specified in XBPS_DISTFILES_MIRROR should be queried for a distfile first and the download location specified in the template should be used only as a fallback, if the download itself or the checksum of the downloaded distfile fails.

pullmoll:
Closes #1372
- You may specify one or more alternate download locations for distfiles
- For the file:// scheme to work, manual mount below masterdir/host is required
- Works best with a local network HTTP or FTP server


xtraeme:
My concern about this patch is this needs manual intervention to make it work in the chroot.

xtraeme:
I could add an option to xbps-uunshare/uchroot to bind mound an additional dir, but first we'd need to have a fixed path in /host.

This could be added for xbps-0.45.

pullmoll:
For the file:// scheme, yes. HTTP and FTP mirrors work fine. Unfortunately I can't bind mount a specified path automagically in xbps-src, because that would need su(do).

xtraeme:
No, you don't need to be root with xbps-uunshare/uchroot.

pullmoll:
Okay, you know better :) An alternate way could be to define XBPS_HOSTDIR to some other local path and under that path mount your distfiles path below e.g. source/distfiles...

xtraeme:
This is a good idea.

pullmoll:
Since you know better, perhaps modify the expert description along those lines?

xtraeme:
I would suggest `/host/sources/mirror` or `/host/sources/distfiles`.

xtraeme:
Sure, will do once the implementation is finished.

pullmoll:
Ah, ok, so you would expect a fixed mountpoint below `/host/sources/somepath` and mount the first (and only) file:// URL's path there?

xtraeme:
@pullmoll are you on IRC? I'd like to talk with you privately.

pullmoll:
Nope, I don't have/use IRC :/

xtraeme:
Please join irc.freenode.net #xbps

pullmoll:
Sorry, I can't now, because I'm away soon. We could discuss by email, if you prefer, or I could try to set up some IRC client next time.

xtraeme:
Ok, no problem. I was going to offer you commit rights to voidlinux.

pullmoll:
No, wait. I'm aboard for just a week or so and wouldn't trust myself enought to commit to void-packages :) I'd rather give me some time until you or others don't have to fix my pull requests anymore.

xtraeme:
Sure, that's ok.

xtraeme:
I've implemented `-b src:dest` in `xbps-uunshare(8)` and `xbps-uchroot(8)` for 0.45.

That means that this issue can be implemented properly once 0.45 is released.

pullmoll:
Great! So one way to go now could be to parse XBPS_DISTFILES_MIRROR for file scheme URLs and

+ create a corresponding path below `masterdir/hostdir/sources/distfiles`
+ bind mount host `/some/path` on `masterdir/hostdir/sources/distfiles/some/path`, probably read only
+ modify `00-distfiles.sh` to prepend not just `/host`, but `/host/sources/distfiles` to the file URLs path.

And when exiting xbps-src, automatically unmount - perhaps also remove mountpoints.

pullmoll:
@xtraeme In 79a361a I tried to add a `setup_distfiles_mirror()` function to create mount points and add corresponding options to `XBPS_CHROOT_CMD_ARGS`. I hope this is (close to) how you meant it should work.

I'm not sure if calling it only when `-z "$IN_CHROOT"` is right, though.

xtraeme:
Now that `xbps-0.45` is out this could be merged, but I'd like to review it a bit more :-)

pullmoll:
It doesn't work as is. I tried again now with xbps-0.45. The problem is in the timing when to bind mount the file:// URLs and exactly where this should happen. I still lack a bit of oversight in the xbps-src and common scripts...

PR #1514, opened 2015-05-06
New package: dhex-0.68

localhost:


srcpkgs/dhex/template
```diff
@@ -0,0 +1,23 @@
+# Template file for 'dhex'
+pkgname=dhex
+version=0.68
+revision=1
+build_style=gnu-makefile
+make_install_args="DESTDIR=/usr"
```
xtraeme: This line is redundant because you are using a custom `do_install()`.

xtraeme:
Also please indent with tabs as suggested by xlint on travis.

Thanks :-)

localhost:
Thanks for your feedback. I added a commit with your suggestions. :)

xtraeme:
Please remove `build_style=gnu-makefile` and use this:

```
do_build() {
  make CC=$CC CFLAGS="$CFLAGS" LDFLAGS="$LDFLAGS" ${makejobs}
}
```

This should fix cross compilation and use our {C,LD}FLAGS for building.

localhost:
Hmm, didn't work - same problem with `gnu/stubs-32.h` not being found. Did I make an error integrating your suggestion?

Sorry for the trouble regarding this pull request...

pullmoll:
Hi ::1 - I tried disabling the Makefile's fixed `CFLAGS=`, `LDFLAGS=` and `CPPFLAGS=` lines. This seems to be sufficient to get the thing compile:
```shell
do_build() {
        sed -i Makefile \
                -e "s;^CFLAGS=;#CFLAGS=;" \
                -e "s;^LDFLASG=;#LDFLASG=;" \
                -e "s;^CPPFLAGS=;#CPPFLAGS=;"
        make CC=$CC CFLAGS="$CFLAGS" LDFLAGS="$LDFLAGS" ${makejobs}
}
```
HTH

PR #1517, opened 2015-05-06
inetutils: split into packages

pullmoll:


pullmoll:
Is it possible to define conflicts based on the build options, i.e. as late as in pre_configure()?

xtraeme:
No, it's not possible.

xtraeme:
I would prefer splitting the utilities into subpkgs instead, rather than build options.

xtraeme:
So that all utils are built by default but the pkg creates multiple subpkgs with a specific utility, this way you can also set conflicts.

pullmoll:
I'm not sure about `--without-wrap`. If it is removed, I think we'd get the `tcpd` binary which could then be used to wrap `/etc/inetd.conf` services like finger etc.

I'm not acquainted with `inetd` and don't know how to write an all new `/etc/inetd.conf` or what should be in there. I have to go do some RTFM or googling, unless we could just take an example from another Linux or *BSD distribution and adapt it for Void. Same goes for `hosts.{allow,deny}`.

Gottox:
I prefer to put at least server and client apps in the same packages.

pullmoll:
How about adding pseudo packages `inetutils-xxx-all` for client + server for protocol `xxx`?
I think that many times users just want to have a certain client and don't need the corresponding daemon, like in the ftp or telnet for example.

Gottox:
```
~ ls -lh `xbps-query -f inetutils | grep '^/usr/bin\|libexec'`
-rwxr-xr-x 1 root root   52 Sep  7  2014 /usr/bin/dnsdomainname
-rwxr-xr-x 1 root root 121K Sep  7  2014 /usr/bin/ftp
-rwxr-xr-x 1 root root  48K Sep  7  2014 /usr/bin/inetutils-hostname
-rwsr-xr-x 1 root root  58K Sep  7  2014 /usr/bin/rcp
-rwsr-xr-x 1 root root  52K Sep  7  2014 /usr/bin/rlogin
-rwsr-xr-x 1 root root  48K Sep  7  2014 /usr/bin/rsh
-rwxr-xr-x 1 root root  60K Sep  7  2014 /usr/bin/talk
-rwxr-xr-x 1 root root 120K Sep  7  2014 /usr/bin/telnet
-rwxr-xr-x 1 root root 135K Sep  7  2014 /usr/libexec/ftpd
-rwxr-xr-x 1 root root  65K Sep  7  2014 /usr/libexec/rlogind
-rwxr-xr-x 1 root root  61K Sep  7  2014 /usr/libexec/rshd
-rwxr-xr-x 1 root root  65K Sep  7  2014 /usr/libexec/talkd
-rwxr-xr-x 1 root root  90K Sep  7  2014 /usr/libexec/telnetd
```
I think it's okay to split the different protocols, but I don't see why saving a few dozen kb is worth it.

pullmoll:
The idea was to be able to install e.g. inetutils-ifconfig or inetutils-tftp(d), but with concise conflict reports. This isn't possible with build_options, thus the splitting.

xtraeme:
I like the current approach.

srcpkgs/inetutils/template
```diff
@@ -1,35 +1,190 @@
...
+}
+inetutils-rcp_package() {
+	pkg_descr+=" - rcp client (remote copy)"
+	pkg_install() {
+		vbin src/rcp
+		vman man/rcp.1
+	}
+}
+inetutils-rexec_package() {
+	pkg_descr+=" - rexec client and server (remote exec)"
```
xtraeme: pkg_descr -> short_desc

srcpkgs/inetutils/template
```diff
@@ -1,35 +1,190 @@
...
+	pkg_install() {
+		vbin src/inetd
+		vman man/inetd.8
+		# TODO: Install default config files with all services disabled?
+		# Add config_files="/etc/inetd.conf /etc/hosts.deny /etc/hosts.allow"
+	}
+}
+inetutils-syslog_package() {
+	pkg_descr+=" - logger tool and syslog daemon"
+	conflicts="util-linux>=0 rsyslog>=0"
```
xtraeme: There's no need to conflict against rsyslog.

srcpkgs/inetutils/template
```diff
@@ -1,35 +1,190 @@
...
+	pkg_descr+=" - whois tool"
+	conflicts="whois>=0"
+	pkg_install() {
+		vbin whois/whois
+		vman man/whois.1
+	}
+}
+inetutils-inetd_package() {
+	pkg_descr+=" - inet daemon"
+	conflicts="xinetd>=0"
```
xtraeme: There's no need to conflict against xinetd, both pkgs can coexist.

srcpkgs/inetutils/template
```diff
@@ -1,35 +1,190 @@
...
+inetutils-dnsdomainname_package() {
+	pkg_descr+=" - dnsdomainname tool"
+	pkg_install() {
+		vbin src/dnsdomainname
+		vman man/dnsdomainname.1
+	}
+}
+inetutils-hostname_package() {
+	pkg_descr+=" - hostname tool"
+	pkg_install() {
```
xtraeme: `conflicts="coreutils>=0"`

chneukirchen:
Imo coreutils hostname should be dropped and inetutils-hostname be used for base-system instead.
coreutils hostname does not implement `hostname -f`, which quite some tools expect.

Gottox:
+1

srcpkgs/inetutils/template
```diff
@@ -1,35 +1,208 @@
...
+	short_descr+=" - inet daemon"
+	pkg_install() {
+		vbin src/inetd
+		vman man/inetd.8
+		# TODO: Install default config files with all services disabled?
+		# Add config_files="/etc/inetd.conf /etc/hosts.deny /etc/hosts.allow"
+	}
+}
+inetutils-syslog_package() {
+	short_descr+=" - logger tool and syslog daemon"
```
xtraeme: short_descr -> short_desc

srcpkgs/inetutils/template
```diff
@@ -1,35 +1,208 @@
...
+			vman man/rlogind.8
+			vinstall ${FILESDIR}/rlogin.xinetd 644 etc/xninet.d
+			# TODO: write etc/inet.d/rlogin.conf or add an etc/inetd.conf line
+		fi
+	}
+}
+inetutils-rsh_package() {
+	short_descr+=" - rsh client and server (remote shell)"
+	pkg_install() {
+		if [ -e src/rsh -a -e src/rshd ]; then
```
xtraeme: Why is it wrong? because in the false branch the pkg will be completely empty.

srcpkgs/inetutils/template
```diff
@@ -1,35 +1,208 @@
...
+		vman man/ftpd.8
+		if [ "$build_option_systemd" ]; then
+			vinstall ${FILESDIR}/ftpd.service 644 usr/lib/systemd/system
+		fi
+		# TODO: write etc/inet.d/ftp.conf or add an etc/inetd.conf line
+	}
+}
+inetutils-ping_package() {
+	short_descr+=" - ping tool (IPv4 and IPv6)"
+	conflicts="iputils>=0 busybox>=0"
```
Gottox: Does not conflict with busybox but with base-system-busybox.

srcpkgs/inetutils/template
```diff
@@ -1,35 +1,208 @@
...
 homepage="http://www.gnu.org/software/inetutils"
-distfiles="http://ftp.gnu.org/gnu/${pkgname}/${pkgname}-${version}.tar.gz"
+distfiles="${GNU_SITE}/${pkgname}/${pkgname}-${version}.tar.gz"
 checksum=0423ea40a160cd4630f817e247df5c08ecc625378d2bb56a0509117a36a89913
 
 pre_configure() {
 	sed -i 's|_GL_WARN_ON_USE (gets|//_GL_WARN_ON_USE (gets|' lib/stdio.in.h
+	# Create a definition to allow the ifconfig program to build properly
+	echo '#define PATH_PROCNET_DEV "/proc/net/dev"' >> ifconfig/system/linux.h
+	case "$XBPS_TARGET_ARCH" in
```
xtraeme: XBPS_TARGET_ARCH -> XBPS_TARGET_MACHINE

srcpkgs/inetutils/template
pullmoll: Hmmm...
```shell
./xbps-src -a x86_64-musl pkg inetutils
...
subpackages: inetutils-dnsdomainname
inetutils-ftp
inetutils-hostname
inetutils-ifconfig
inetutils-inetd
inetutils-ping

inetutils-rexec


inetutils-syslog

inetutils-telnet
inetutils-tftp
inetutils-traceroute
inetutils-uucpd
inetutils-whois
…
```
So removing from subpackages worked right. And also configure tells:
```shell
Summary of build decisions:

  Clients:

    dnsdomainname  yes
    ftp            yes  -lreadline
    hostname       yes
    ifconfig       yes
    logger         yes
    ping           yes  
    ping6          yes  
    rcp            no
    rexec          yes
    rlogin         no
    rsh            no
    talk           no
    telnet         yes  -lcurses
    tftp           yes
    traceroute     yes  
    whois          yes

  Servers:

    ftpd           yes  PAM 
    inetd          yes
    rexecd         yes  PAM
    rlogind        no
    rshd           no
    syslogd        yes
    talkd          no
    telnetd        yes  -lcurses
    tftpd          yes
    uucpd          yes
```

Yet when it comes to packaging this happens:
```
=> inetutils-ping-1.9.2_2: running post-install hook: 06-strip-and-debug-pkgs ...
   Stripped executable: /usr/bin/ping6
   Stripped executable: /usr/bin/ping
=> inetutils-ping-1.9.2_2: running post-install hook: 98-lib32 ...
=> inetutils-rcp-1.9.2_2: running pre-install hook: 00-lib32 ...
=> inetutils-rcp-1.9.2_2: running pre-install hook: 02-script-wrapper ...
=> inetutils-rcp-1.9.2_2: running pkg_install ...
=> ERROR: inetutils-rcp-1.9.2_2: vinstall: cannot find 'src/rcp'...
=> ERROR: inetutils-rcp-1.9.2_2: failed to run pkg_install() at line 140.
```

Huh?

xtraeme:
The main pkg (inetutils) only contains the info documentation... this is useless.

I would suggest you to make inetutils depend on all subpkgs, this way it's a meta-pkg.

srcpkgs/inetutils/template
```diff
@@ -1,35 +1,208 @@
...
+	case "$XBPS_TARGET_MACHINE" in
+	*-musl)
+		# Disable subpackages which won't build
+		for subpkg in rcp rlogin rsh rshd talk talkd; do
+			subpackages=${subpackages/$pkgname-$subpkg/}
+			depends=${depends/$pkgname-$subpkg/}
+			configure_args+=" --disable-$subpkg"
+		done
+		export subpackages
+		echo "subpackages: $(echo $subpackages|fmt)"
```
xtraeme: This won't work if you run `xbps-src` progressively by running a phase, i.e:

    $ ./xbps-src build ...
    $ ./xbps-src pkg ...

Please take a lookg at `gcc` to set subpackages properly.

xtraeme:
Any reason to close it?

pullmoll:
I thought you had merged the latest already.

PR #1522, opened 2015-05-07
mediainfo: combine -gui package

tonkazoid:
I tried a workaround to keep the -gui package with setting a variable in -gui_package and then check for it in do_{build,install}. Didn't work, thus combined them.

### DIFF
--- a/srcpkgs/mediainfo/template
+++ b/srcpkgs/mediainfo/template
@@ -2,7 +2,7 @@
 
 pkgname=mediainfo
 version=0.7.73
-revision=1
+revision=2
 short_desc="Display technical and tag data for video and audio files"
 homepage="http://mediaarea.net/MediaInfo"
 maintainer="Georg Schabel <gescha@posteo.de>"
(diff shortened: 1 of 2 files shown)


srcpkgs/mediainfo/template
```diff
@@ -11,35 +11,42 @@ license="BSD"
...
 checksum=1fde5b2972d923f7f87c2b2e738e8a086603243770961b077cf558f9836e612e
 
-build_style=gnu-configure
 hostmakedepends="automake libtool pkg-config"
 makedepends="libmediainfo-devel zlib-devel
  wxGTK-devel wxWidgets-devel"
+depends="desktop-file-utils hicolor-icon-theme"
 create_wrksrc=yes
-build_wrksrc="MediaInfo/Project/GNU/CLI"
+_gnudir="MediaInfo/Project/GNU"
```
xtraeme: You are missing `replaces="mediainfo-gui>=0"`

srcpkgs/mediainfo/template
```diff
@@ -11,35 +11,42 @@ license="BSD"
...
 
-pre_configure() {
-	autoreconf -fi
+do_configure() {
+	for d in CLI GUI; do
+		cd $wrksrc/$_gnudir/$d
+		autoreconf -fi
+		./configure --prefix=/usr
```
xtraeme: `./configure ${configure_args}` instead

PR #1523, opened 2015-05-07
New package: btsync-2.0.105

JuhaniImberg:


chneukirchen:
I think we are not allowed to redistribute this.

tonkazoid:
alternative?
https://github.com/voidlinux/void-packages/tree/master/srcpkgs/syncthing

"Syncthing replaces proprietary sync and cloud services with something open, trustworthy and decentralized."

JuhaniImberg:
So I changed the package to work like `chromium-pepper-flash`, it downloads and installs the program only as an post-install hook so you would be redistributing just an installation script.

@tonkazoid I would use that but I need support for Windows Phone, otherwise it looks way better.

Gottox:
INSTALL and REMOVE are missing newline at the end, otherwise looks reasonable too me.

PR #1524, opened 2015-05-07
libclc-git: make noarch

jprjr:


Gottox:
According to http://llvm.org/releases/1.3/docs/BytecodeFormat.html, the format should be portable. So this should be okay :+1:

PR #1525, opened 2015-05-07
New package: TwitchNotifier-0.3.1

GiedriusS:


srcpkgs/TwitchNotifier/template
```diff
@@ -0,0 +1,15 @@
+# Template file for 'TwitchNotifier'
+pkgname="TwitchNotifier"
+version="0.3.1"
+revision=1
+build_style=python-module
+hostmakedepends="python3.4-devel"
+makedepends="python3.4-devel"
+depends="virtual?python3.4-gobject>=3.14.0_2 virtual?python3.4-requests>=2.7.0_1 virtual?libnotify>=0.7.6_4"
```
xtraeme: virtual is wrong here.

GiedriusS: Why? Those packages aren't needed for packaging up TwitchNotifier.

xtraeme: Also there's no reason to force a greater than or equal version.

xtraeme:
`depends` only check if those pkgs are available as binary packages, so that they aren't installed to the masterdir, just verified and if are not available they are built.

`virtual?dep` is not meant to be used to skip this step.

GiedriusS:
Should I remake the whole commit or could I just add a new commit on top of it with the changes that remove virtual and >=?

xtraeme:
Up to you.

xtraeme:
You are missing `noarch=yes`

GiedriusS:
Sorry for causing so much issues.

PR #1530, opened 2015-05-08
llvm: make with RTTI

bougyman:


xtraeme:
Already applied by @bougyman

PR #1544, opened 2015-05-11
New package: drbd-utils-8.9.2

Gottox:
drbd is messy. So I did my best to clean up. :)


xtraeme:
This is ok, thanks :+1:

PR #1545, opened 2015-05-11
New package: ponyc-0.1.3

bougyman:


bougyman:
Wanted review on the show-files for this one: http://sprunge.us/PGLi

bougyman:
merged to master.

srcpkgs/ponyc/template
```diff
@@ -0,0 +1,25 @@
+# Template file for 'ponyc'
+pkgname=ponyc
+version=0.1.3
+revision=1
+only_for_archs="x86_64"
+build_style=gnu-makefile
```
Gottox: Either use gnu-makefile or use do_{build,install}(), not both. using gnu-makefile is prefered though.

PR #1546, opened 2015-05-11
nim: update to 0.11.2

allan:


allan:
the libzip soname bump from last week broke it

PR #1550, opened 2015-05-12
darktable: enable on musl

lemmi:
the patch removes unused stderr from the openmp pragma. 
musl defines stderr as (stderr). somehow this is not a valid identifier anymore.

### DIFF
--- a/srcpkgs/darktable/template
+++ b/srcpkgs/darktable/template
@@ -1,7 +1,7 @@
 # Template file for 'darktable'
 pkgname=darktable
 version=1.6.6
-revision=1
+revision=2
 build_style=cmake
 #this makes sure to use -march=generic and -msse3
 configure_args="-DBINARY_PACKAGE_BUILD=ON"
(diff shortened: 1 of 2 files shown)


srcpkgs/darktable/template
```diff
@@ -15,8 +15,8 @@ makedepends="gtk+-devel glib-devel exiv2-devel lcms2-devel
 	 libwebp-devel libsoup-devel lensfun-devel sqlite-devel librsvg-devel
 	 lua-devel json-glib-devel libgomp-devel colord-devel
 	 libopenjpeg-devel libopenexr-devel libgraphicsmagick-devel
-	 SDL-devel libsecret-devel"
-only_for_archs="i686 x86_64"  # too much SSE2 stuff all over
+	 SDL-devel libsecret-devel glu glu-devel"
```
xtraeme: There's no need to add `glu`, `glu-devel` depends on it already.

lemmi:
then something else is wrong. without that it won't find GL/glu.h and has undefined references to gluOrtho...something. this only happens on musl

lemmi:
after making sure my hostdir was empty it finally worked with just glu-devel :)

PR #1552, opened 2015-05-12
Node add build options

Gottox:
Node.js and io.js now support build options to enable/disable shared dependencies. As the node/iojs dependencies happily break APIs every now and then this makes maintainance easier.

### DIFF
--- a/srcpkgs/nodejs/template
+++ b/srcpkgs/nodejs/template
@@ -4,7 +4,9 @@ version=0.12.3
 revision=1
 wrksrc=node-v${version}
 hostmakedepends="pkg-config python"
-makedepends="libressl-devel zlib-devel python-devel libuv-devel"
+makedepends="zlib-devel python-devel
+ $(vopt_if ssl libressl-devel) $(vopt_if libuv libuv-devel)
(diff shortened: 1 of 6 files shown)


srcpkgs/nodejs/template
```diff
@@ -12,6 +14,9 @@ license="MIT"
 homepage="http://nodejs.org/"
 distfiles="${homepage}/dist/v${version}/node-v${version}.tar.gz"
 checksum=ac7e78ade93e633e7ed628532bb8e650caba0c9c33af33581957f3382e2a772d
+build_options="libressl libuv http_parser"
```
xtraeme: Why not enabling them by default? I don't see the point for these options.

Gottox: shared http_parser is broken on node.js and libressl is broken with io.js. I'd like to build with shared v8 in future too.

Those dependencies will break from time to time. buildoptions allow to quickly switch from broken shared builds to the bundled packages.

xtraeme: Makes sense, thanks.

Please rename libressl opt to just ssl, to match common/options.description, and maybe add the description for those opts to the template.

PR #1553, opened 2015-05-12
qtcreator: depend on qt5-*-devel and qt5-doc

pullmoll:
A developer using qtcreator will most probably want the majority of available qt-*-devel packages.
qt5-doc contains the *.qch help files and qt5-plugin-sqlite is required to access them.


xtraeme:
Makes sense :-)

pullmoll:
I'm not sure about qt4. Personally I use the qt-doc and all qt-*-devel as well, but I read somewhere qt4 is projected EOL end of this year.

PR #1555, opened 2015-05-12
New package: calligra-2.9.2

### LINKED ISSUE #1324: Package Request: krita
https://krita.org/download/krita-desktop/

Gottox:
Takes hours to build. closes #1324.


silvernode:
is it possible to just build krita without the whole suite?

Gottox:
Maybe. But why?

silvernode:
Well, if nothing else I for one was looking forward to Krita. For awhile I wasn't aware it was even part of a suite. Also having calligra split up while additionally offering a suite meta package could be beneficial. However I supposed that would be like splitting up libreoffice. I'm not sure if it makes sense to do this or not.

Gottox:
Splitting those packages makes sense. Step by step, it has to compile first :)

PR #1566, opened 2015-05-13
mpfr: Import upstream fixes for 3.1.2 (p11)

pullmoll:
Imported from LFS (http://www.linuxfromscratch.org/lfs/view/development/chapter06/mpfr.html)
For the source see: http://www.mpfr.org/mpfr-current/allpatches


uggedal:
This is missing [p11](http://www.mpfr.org/mpfr-current/patch11).

pullmoll:
Thank you! I'll add that to the patch file and rebase.

PR #1567, opened 2015-05-13
binutils: switch to tar.bz2 to reduce bandwidth

pullmoll:
Just to save some bandwidth / download time: tar.bz2 is 23485kB while the tar.gz is 30088kB.


pullmoll:
If this is `A Good Idea ™`, I could do the same for a number packages with distfiles from `${GNU_SITE}`...

xtraeme:
Seems a good idea, thanks :-)

pullmoll:
I'll create a pull request with a number of $GNU_SITE packages then.

PR #1574, opened 2015-05-14
New package: mmv-1.01b

pullmoll:


chneukirchen:
Looks good, do you want to add the symlinks for mcp, mad, mln too?

pullmoll:
Okay, will add them.

pullmoll:
Aw, its usr/share/man/man1 ... wait

pullmoll:
IMHO mmv should be even in base-system :)

chneukirchen:
Not before ed is. ;)

PR #1575, opened 2015-05-14
New package: google-talkplugin-5.41.0.0

norvegh:


srcpkgs/google-talkplugin/template
```diff
@@ -0,0 +1,46 @@
...
+do_fetch()
+{
+	mkdir -p $wrksrc
+	cd $wrksrc
+	wget $_file
+	if ! sha256sum -c ${FILESDIR}/${XBPS_TARGET_MACHINE}.sum; then
+		echo "INVALID CHECKSUM"
+		return -1
+	fi
+}
```
xtraeme: This is again unnecessary, please check the manual and see other packages to change this apropiately.

xtraeme:
I'm really unsure about this license and redistribution and packaging.

srcpkgs/google-talkplugin/template
norvegh: OK, it works with .rpm distfiles, but it will extract the files directly below builddir (not in google-talkplugin-$version), and those will not be cleaned up when the build finishes. So shall I just delete everything manually in the builddir in post_install()?

xtraeme: This is what `create_wrksrc` is for.

silvernode:
Maybe a download script would be more appropriate in this scenario?

norvegh:
Yes, you are right, this is not redistributable. I will check whether I can fix a download script.

norvegh:
First attempt with the install (download) script. Please comment how to make it better. When the package is uninstalled there will be empty directories left (like /usr/lib/iceweasel/plugins/). Shall the REMOVE script check for that and remove all those?

bougyman:
Yes, the REMOVE script should clean up anything necessary. I'm happy to see this package come to void.

xtraeme:
This pkg does not allow redistribution, you should set `restricted=yes` in the template and remove the INSTALL/REMOVE stuff.

xtraeme:
Stalled for 3 months, feel free to resubmit it with my suggested changes.

PR #1576, opened 2015-05-14
New package: bftpd-4.4

norvegh:


srcpkgs/bftpd/template
```diff
@@ -0,0 +1,33 @@
...
+	wget http://sourceforge.net/projects/bftpd/files/bftpd/bftpd-${version}/bftpd-${version}.tar.gz/download
+	if ! sha256sum -c $FILESDIR/sum; then
+		echo "INVALID CHECKSUM"
+		rm -f download
+		return -1
+	fi
+	tar xvfz download
+	rm download
+	mv bftpd bftpd-${version}
+}
```
xtraeme: Check the `Manual` and specifically for `distfiles` and `checksum`.

You are using wget and do_fetch that is only required in 0.01% real cases.

norvegh:
The purpose is that it is a simple and secure FTP server that actually works out of the box.

srcpkgs/bftpd/template
```diff
@@ -0,0 +1,25 @@
...
+maintainer="Norbert Vegh <vegh@norvegh.com>"
+license="GPL-2"
+homepage="http://bftpd.sourceforge.net/"
+checksum=b805ebbdd3de993ca14d733ad9ad007ac342a1646b257d91a53ec87a36c7f741
+distfiles=http://${SOURCEFORGE_SITE}/bftpd/bftpd-${version}.tar.gz
+
+post_extract()
+{
+	cd ${XBPS_BUILDDIR}
+	mv bftpd bftpd-${version}
```
xtraeme: This `post_extract` is still unnnecessary.

srcpkgs/bftpd/files/sum
```diff
@@ -0,0 +1 @@
+b805ebbdd3de993ca14d733ad9ad007ac342a1646b257d91a53ec87a36c7f741  download
```
xtraeme: This file is unnecessary...

srcpkgs/bftpd/template
```diff
@@ -0,0 +1,25 @@
...
+}
+
+build_style=configure
```
xtraeme: And probably you should set `build_style=gnu-configure` and remove `configure_args`.

srcpkgs/bftpd/template
norvegh: The tar.gz extracts into bftpd, without version numbering, so the build process will not find it. How else can I solve this?

xtraeme: `wrksrc=bftpd`

Issue #1257, opened 2015-04-02
Package request: rxvt-unicode-terminfo (again!), aspell-pl, {fontconfig,freetype,cairo}-ultimate


skrzyp:
While I'm leaving you at this weekend, I can't post any new templates until next Tuesday, so I'll leave some requests which may be good not only for me:

* [x] **rxvt-unicode-terminfo**: add `rxvt-unicode-256color` definition (it exists in ArchLinux's package)
* [ ] **aspell-pl**: dictionary of my language. But a little note: use [this dict files](http://sjp.pl/slownik/ort/) instead of GNU dictionaries, they're more recent and frequently updated.
* [ ] **fontconfig-ultimate**: bumping the #1053, but using [bohoomil/fontconfig-ultimate](https://github.com/bohoomil/fontconfig-ultimate). It really makes a very big (positive, ofc) difference in font rendering, mostly with [kori's infinality config](https://github.com/kori/shell.d/blob/master/infinality). Just try it yourself and decide to package it or not

At the end - Merry Easter :smiley:

Gottox:
xbps-query -Rs rxvt-unicode-terminfo
[-] rxvt-unicode-terminfo-9.21_4 rxvt clone supporting Xft fonts and Unicode - terminfo data

thypon:
#1053 <- duplicate.

thypon:
#1554

xtraeme:
All of them are now available.

Issue #1053, opened 2015-03-04
Patch request: infinality


thypon:
Could we include infinality patchset or the ubuntu patchset to improve font rendering on lcd displays?

http://www.infinality.net/blog/
https://wiki.archlinux.org/index.php/Infinality

bougyman:
can you link to the patchset you're recommending?

thypon:
https://github.com/bohoomil/fontconfig-ultimate This is the patchset updated very ofter from bohoomil.

This instead is the patches from ubuntu would be much more tricky.

nmeum:
:+1:

skrzyp:
:+1: 
It's quite easy, just add patches for freetype and cairo, but fontconfig part might be more tricky.

thypon:
Right now the patch is alredy inside the void-packages repository. To apply simply download it and:

```
xbulk -oultimate cairo freetype fontconfig
xbps-install -R hostdir/binpkgs/ cairo freetype fontconfig
```

thypon:
I'm considering to create a separate repository containing these updated packages.

nkoehring:
@thypon I unfortunately didn't find out, what this `xbulk` command is. Is it `./common/xbps-src/shutils/bulk.sh` ?

Duncaen:
@nkoehring the patches are obsolete and already removed again from the templates. (https://www.freetype.org/index.html#news)

`xbulk` is part of `xtools`.

PR #1582, opened 2015-05-15
Infinality

thypon:
#1053 #1257


srcpkgs/freetype/template
```diff
@@ -12,6 +12,30 @@ license="GPL-2"
 distfiles="${SOURCEFORGE_SITE}/$pkgname/$pkgname-$version.tar.bz2"
 checksum=387bcc4b780b12484aa2ec9f7db1a55d8286eb5639f45fbc0fbba7a4e5a1afb9
 
+build_options="ultimate"
+desc_option_ultimate="Enable ultimate patchset"
+build_options_default="ultimate"
```
xtraeme: I will refuse to merge this if this is enabled by default.

Gottox: What do you think patched packages {cairo,freetype,fontconfig}-ultimate which provide {cairo,freetype,fontconfig}?

xtraeme:
I won't maintain anything related to those unofficial patches, just FYI.

srcpkgs/freetype/template
xtraeme: This will cause more harm than good.

thypon:
I could help mantaining them; i <3 perfect font rendering. An alternative could be using the ubuntu patchset. @xtraeme do you prefer them enabled by default?

thypon:
Added steam-fonts to fix a missrendering with ultimate patchset.

skrzyp:
:+1:  Respect for your effort - that weird way of compiling those patches always took off me from making their templates.
FYI, not only you love those perfect font rendering, especially with [kori's config](https://github.com/kori/shell.d/blob/master/infinality)

thypon:
@xtraeme Given is disableable i'd prefer to include these option by default. All the major desktop distribution has his own patchset to render correctly fonts, but i think that infinality is the more cross distribution patchset / configurable approach.

skrzyp:
@thypon : What's going on with this port? Maybe we should provide is at alternative package which can replace fontconfig, instead of making fontconfig flavor?

Sorry for asking too much, but I just can't wait for it ^^

thypon:
You can clone and use it right now like i'm doing since a week waiting that @xtreame decide something.

Right now is disabled by default then you need to enable with the "-oultimate" flag.

```
git clone https://github.com/thypon/void-packages
cd void-packages/
git checkout infinality
./xbps-src binary-bootstrap
xbulk -oultimate cairo fontconfig freetype steam-fonts
xbps-install --repository=hostdir/binpkgs/infinality/ cairo fontconfig freetype steam-fonts
```

I really don't want to mantain a not so used/out of sync alternative package like in archlinux. I really hope we could merge this one as default or use the patchset from fedora/ubuntu to obtain a good font rendering per default out of the box.

skrzyp:
Thank you so much, @thypon!
Maybe you should convince @xtraeme by providing additional infos about it and screenshots?
Also, you could hang out with us on IRC :>

thypon:
Merged manually

Duncaen:
Infinality is unmaintained and was dropped from void-packages, fontconfig improved and merged parts of it.

PR #1586, opened 2015-05-16
New package: fortune-mod-void-001

tonkazoid:


chneukirchen:
Looks ok, but better use YYYYMMDD as version number.

chneukirchen:
Applied.

tonkazoid:
i dont' know where it got wrong, but `hostmakedepends` should be `makedepends`, no?

https://github.com/voidlinux/void-packages/blob/d9cc272634fcb9c3dcca5f9357c1e9a0b82a5dbe/srcpkgs/fortune-mod-void/template
https://github.com/voidlinux/void-packages/blob/master/srcpkgs/fortune-mod-void/template

chneukirchen:
No. I fixed that when applying.

tonkazoid:
And xtraeme fixed 'depends' :P. I still mix up makedepends with
hostmakedepends sometimes.

PR #1588, opened 2015-05-16
New package: quodlibet-3.4.0

jarbowski:


xtraeme:
- This package is missing `noarch=yes`
- This package is missing runtime dependencies

jarbowski:
It isn't missing noarch=yes
And according to Arch pkgbuild, only some optional dependencies are missing: gst-libav, zeitgeist (only x86 in Void), libgpod, cddb-py, python2-musicbrainz2, kakasi.
I've added python-inotify and libkeybinder2.

xtraeme:
Last time I merged one of your pkgs had to loose almost 1h fixing the pkg.

jarbowski:
Sorry for that, it won't happen again. I think everything is okay now, it works well on my box.

PR #1590, opened 2015-05-17
New packages: cubietruck-kernel

pullmoll:
This kernel is based on linux-sunxi, heavily modified by dan-and and others.
The latest git revision is 3.4.107 (pre-release).


xtraeme:
IMHO we'd need to change cubieboard2-kernel and cubietruck-kernel to use the upstream linux kernel.

Even if there's no 3d hardware acceleration will be useful to have an unified kernel for common armv7 platforms.

pullmoll:
Okay, then we should perhaps have both ways for the user to choose from:
+ sunxi-kernel (main line kernel)
+ sunxi-kernel-desktop (linux-sunxi based with patches, mali400 support etc. - i.e. this one)

and corresponding to that then two sets of cubie{board2,truck}-base and cubie{board2,truck}-base-desktop

For me a headless cubietruck is not what I want, because I intend(ed) to use it as a media center...

What do you think?

xtraeme:
What is not clear to me is if this `dan-and` linux kernel is official at all or seems to be maintained by some independent devs.

I'd think that probably using this kernel would be ok for now.

My long term goal is creating an unified armv7 linux kernel pkg using the official/upstream linux kernel.

pullmoll:
The kernel is not official any more or less than the linux-sunxi fork, it's just the latest effort used by much of the cubie* community. Before I knew about void, I tried to create a cubietruck desktop setup based on Igor Pecovnik's work, which is a debootstrap with either main line or dan-and's kernel.

If you agree, I would rename the packages like this then
- `cubietruck-kernel -> sun7i-kernel-desktop`
- `cubietruck-base -> cubietruck-desktop`

And perhaps also create new `cubietruck-base` and `cubieboard2-desktop` which depend on the corresponding kernel.

I can also try to change sun7i-kernel to use main line after I got this one done and working.

xtraeme:
I would skip the `-desktop` kernels and use this one as it seems to be ok for cubietruck.

If you want you can invest your time in making `linux4.0` support all possible ARMv7 platforms.

pullmoll:
xtraeme: If you want you can invest your time in making linux4.0 support all possible ARMv7 platforms. 

I have only 2 cubietrucks and a beagle bone black board here, so I can't test any other platforms. But then it's probably just choosing the right DTB (device tree blob) for the architecture. There were dozens when I had last looked at it...

xtraeme:
I've got a cubie2 and bbb.

We could support at least those 3 platforms by default in the unified ARMv7 kernel.

The unified kernel pkg should provide all .dtb files for multiple platforms (device tree blob files).

pullmoll:
BTW I can't tell yet if this kernel is ok for the cubietruck, because I didn't manage to mklive using my home built packages. The kernel used to work with Debian wheezy and jessie as well as Ubuntu.

xtraeme:
You shouldn't use `mklive`, but rather `mkrootfs` and optionally `mkimage`.

See the scripts, probably you'll have to add some minor changes.

pullmoll:
I meant to say "I used void-mklive scripts". For me mkrootfs.sh fails when I tried to supply my own repo like `XBPS_REPOSITORY="--repository=/mnt/data/void/binpkgs" sh build-rootfs.sh`
```
Running xbps-install -S  --cachedir=/var/cache/xbps --repository=/mnt/data/void/binpkgs -r /tmp/tmp.hSINkykhuc for target armv7l ...
Running xbps-query -R -r /tmp/tmp.hSINkykhuc  --cachedir=/var/cache/xbps --repository=/mnt/data/void/binpkgs -ppkgver base-system for target armv7l ...
FATAL: Failed to run xbps-query -R -r /tmp/tmp.hSINkykhuc  --cachedir=/var/cache/xbps --repository=/mnt/data/void/binpkgs -ppkgver base-system
```

xtraeme:
You should use it as:

    # ./mkrootfs -r /mnt/data/void/binpkgs [opts] <platform>
    # ./mkimage <platform>

pullmoll:
Okay, will try that

xtraeme:
So that just how I thought, our sun7i-kernel pkg should work just fine for cubietruck.

Please take a look at the `sun7i-kernel` in archlinuxarm.org (github).

xtraeme:
This is unnecessary, the sun7i-kernel should work just fine.

PR #1591, opened 2015-05-17
cubietruck-base: depend on cubietruck-kernel

pullmoll:


xtraeme:
This should work with our sun7i-kernel right now.

PR #1592, opened 2015-05-17
atom: update to 0.199.0.

thypon:


thypon:
Error with travis, should build in the buildbots.

PR #1599, opened 2015-05-18
New package: pidgin-libnotify-0.14

derlaft:
Patches taken from archlinux


xtraeme:
Thanks for your first package, it's ok :-)

Please set `--disable-static` in `configure_args` to avoid installing the static library, it's pointless.

derlaft:
Added.

PR #1600, opened 2015-05-18
xbps-src: unified ccache directory

pullmoll:
- I turns out that ccache is very well capable of recognizing different
  compilers and using only files stored to the cache by each compiler.
- To avoid large parts of a cache becoming useless just because a compiler
  was rebuilt during some (bootstrap) stage, rely on the contents of
  the compiler binary for identification instead of mtime + size.


xtraeme:
Looks good, please fix the indentation :-)

PR #1601, opened 2015-05-18
New packages: nomacs, nomacs-qt5

derlaft:
Is ${_pkg} a good way to do it?

//Thanks devs, it's probably the distro I was wishing for a long time

### DIFF
--- a/srcpkgs/nomacs-qt5/template
+++ b/srcpkgs/nomacs-qt5/template
@@ -1,0 +1,19 @@
+# Template file for 'nomacs-qt5'
+pkgname=nomacs-qt5
+revision=1
+version=2.4.4
+_pkg=${pkgname//-qt5/}
+wrksrc=${_pkg}-${version}
+build_style=cmake
+configure_args="
+ -DENABLE_QT5=1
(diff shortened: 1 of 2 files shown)


srcpkgs/nomacs-qt5/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'nomacs-qt5'
+pkgname=nomacs-qt5
+version=2.4.4
+_pkg=${pkgname//-qt5/}
+wrksrc=${_pkg}-${version}
+revision=1
```
xtraeme: Please move `revision` just below `version`.

`xbps-checkvers(8)` needs those vars in consecutive lines to work properly.

xtraeme: Also both pkgs install the same files, so that they have conflicting files.

Either add `conflicts="pkg>=0"` or use `provides/replaces`.

derlaft:
fixed

xtraeme:
You can remove ´lib32disabled`, it's unnecessary here.

Thanks.

PR #1607, opened 2015-05-19
New package: kompare and libkomparediff2

norvegh:


srcpkgs/libkomparediff2/template
```diff
@@ -0,0 +1,24 @@
...
+makedepends="kdelibs-devel qt-devel phonon-devel"
+
+libkomparediff2-devel_package()
+{
+		depends="libkomparediff2>=${version}_${revision}"
+		short_desc+=" - development files"
+		pkg_install()
+		{
+				vmove usr/include/libkomparediff2
+				vmove usr/lib/cmake/libkomparediff2
```
xtraeme: A development pkg must contain *all* files required for development.

xtraeme: And please use the common style used in all other source packages.

norvegh: OK, what else is missing? The libkomparediff2 package has only the .so files. If I add those also to the devel package wouldn't it cause a conflict?

norvegh: I think I figured it out.

xtraeme:
This PR cannot be applied, please rebase.

srcpkgs/libkomparediff2/template
```diff
@@ -0,0 +1,23 @@
...
+license="GPL-2+"
+homepage="http://www.kde.org/"
+distfiles="http://download.kde.org/stable/${version}/src/${pkgname}-${version}.tar.xz"
+checksum=70c670a69a0b31ab8fb12ccda54662e0f435fed9d9a2b390de68150a1c1e0e6a
+build_style=cmake
+hostmakedepends="cmake automoc4 pkg-config"
+makedepends="kdelibs-devel qt-devel phonon-devel"
+
+libkomparediff2-devel_package() {
+		depends="libkomparediff2>=${version}_${revision}"
```
xtraeme: - `depends="..."` -> `depends="${sourcepkg}>=${version}_${revision}"`
- Please fix the indentation.
- `usr/include/...` -> `usr/include`
- `usr/lib/cmake/...` -> `usr/lib/cmake`

xtraeme:
I believe there's a newer version, 4.14.3 or the 15.04 ones.

PR #1608, opened 2015-05-19
micropython: update to 1.4.3.

necrophcodr:
Title says nearly everything.

I'd like to point out though that I was unable to test building with simple tar files, which is why I resolved to do_fetch with git instead. It seems to work better, and the resulting binary contains proper version number.


beefcurtains:
Why use git instead of the download? The url is https://github.com/micropython/micropython/archive/v1.4.3.tar.gz
Also, you should probably rebase your branch on the latest upstream master.

necrophcodr:
I would love to, but it results in problematic errors. It appears that either way, errors are a result. Strangely, these errors are not present when compiling outside the chroot.

```
CPP ../py/qstrdefs.h

../py/qstrdefs.h:27:25: fatal error: py/mpconfig.h: No such file or directory

 #include "py/mpconfig.h"
```

Pretty strange considering the file clearly is right there, so I'm not sure what is going on there.

xtraeme:
Add this:

```
do_build() {
   make CC=$CC
}
```

necrophcodr:
Thanks, that does appear to have fixed the issue. @beefcurtains the resulting binary will not have any version number as it should have, giving no indication what version of micropython the user is actually running.  
Using the git clone resolves this entirely.

beefcurtains:
Why do you want to change the name of the binary in an update? If you want to know the version, you can always ``xbps-query -s micropython``.

necrophcodr:
Not the name of the binary, but the version imprinted when running it. It displays a version when in REPL, but without the git version, there's no version tag.

This won't affect the sys.version part, but what are the cons of using a git clone as opposed to just fetching it with tar, considering the simple gains?

PR #1609, opened 2015-05-19
New package: kate-4.14.2

norvegh:


common/shlibs
```diff
@@ -2031,3 +2031,7 @@ libiec61883.so.0 libiec61883-1.2.0_1
 libffado.so.2 libffado-2.2.1_1
 libossp-uuid.so.16 libossp-1.6.2_1
 libplayerctl-1.0.so.0 playerctl-0.4.2_1
+libkateinterfaces.so.4 kate-4.14.2_1
```
xtraeme: You don't really need to add a shlib here unless the shlib is expected to be linkable by other applications and/or libraries. 

Resuming, what you've added in `common/shlibs` is unnecessary.

norvegh: Actually now I see that the package also contains include files. Does it mean that I need to make a devel package as well? I checked the kate debian packages, and they have separate packages for for libkateinterfaces4, libkatepartinterfaces4 etc. But I did not find a dev package that contains those header files. This is all very confusing to me.

norvegh:
Added a devel package.

PR #1613, opened 2015-05-19
New package: fail2ban-0.8.14.

necrophcodr:
This provides the current stable version of fail2ban, with no batteries included.

Critiques and comments, lets have 'em!

### DIFF
--- a/srcpkgs/fail2ban/template
+++ b/srcpkgs/fail2ban/template
@@ -1,0 +1,84 @@
+# Template file for 'fail2ban'
+pkgname=fail2ban
+version=0.8.14
+revision=1
+noarch=yes
+build_style=python-module
+hostmakedepends="pkg-config python"
+depends="python"
+short_desc="Authentication failure monitor system"
+pycompile_dirs="/usr/share/fail2ban/"
+conf_files="/etc/fail2ban/action.d/apf.conf /etc/fail2ban/action.d/badips.conf \
+/etc/fail2ban/action.d/blocklist_de.conf /etc/fail2ban/action.d/bsd-ipfw.conf \
+/etc/fail2ban/action.d/complain.conf /etc/fail2ban/action.d/dshield.conf \
+/etc/fail2ban/action.d/dummy.conf /etc/fail2ban/action.d/firewallcmd-ipset.conf \
+/etc/fail2ban/action.d/firewallcmd-new.conf \
+/etc/fail2ban/action.d/hostsdeny.conf  /etc/fail2ban/action.d/ipfilter.conf \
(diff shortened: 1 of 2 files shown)


chneukirchen:
Which ssh logfile will this use?

necrophcodr:
None. That's up to the individual to configure. No jails are included in this release, as is the case with the default package, and the Ubuntu and probably Debian packages.

All that is included are actions and filters, the rest is something one can configure.

xtraeme:
- This package is `noarch`
- This package is missing `conf_files`
- This package is missing `pycompile_dirs`

necrophcodr:
I know about the `conf_files` part. I did not know about the rest. Thanks!

I'm just trying to get the `conf_files` part done because that is one of those that sucks, with like 50 configuration files to enter. I'm guessing directories aren't a go in that area.

Anyway, is `noarch` and `pycompile_dirs` documented someplace? I'll read up on them then.

necrophcodr:
How about that? Please let me know if there's a better way to do conf_dirs because oh my god

xtraeme:
There's no way to allow wildcards for `conf_files` right now.

Would be nice to support that in xbps-src tho.

srcpkgs/fail2ban/template
```diff
@@ -0,0 +1,84 @@
+# Template file for 'fail2ban'
+pkgname=fail2ban
+version=0.8.14
+revision=1
+noarch=yes
+build_style=python-module
+hostmakedepends="pkg-config python"
+depends="python"
+short_desc="Authentication failure monitor system"
+pycompile_dirs="/usr/share/fail2ban/client /usr/share/fail2ban/server /usr/share/fail2ban/common /usr/share/fail2ban/testcases"
```
xtraeme: You can just use `pycompile_dirs=/usr/share/fail2ban"`

srcpkgs/fail2ban/template
```diff
@@ -0,0 +1,84 @@
...
+conf_files="etc/fail2ban/action.d/apf.conf etc/fail2ban/action.d/badips.conf \
```
xtraeme: `conf_files` expect an absolute path, i.e `/etc/foo`.

necrophcodr: It did the thing wrong. Alright, I'll have it fixed in a bit.

Gottox:
please rebase.

necrophcodr:
Rebase done.

necrophcodr:
Rebase completed. Waiting for travis to light up green.

Gottox:
Here we go!

PR #1614, opened 2015-05-19
New package: dtc-1.4.1

thypon:


common/shlibs
```diff
@@ -2031,3 +2031,4 @@ libiec61883.so.0 libiec61883-1.2.0_1
 libffado.so.2 libffado-2.2.1_1
 libossp-uuid.so.16 libossp-1.6.2_1
 libplayerctl-1.0.so.0 playerctl-0.4.2_1
+libfdt-1.4.1.so dtc-1.4.1_1
```
xtraeme: This soname is wrong:

```
   SONAME libfdt.so.1 from /usr/lib/libfdt-1.4.1.so
```

That means that you'll have to move the .so files/symlinks to the main pkg.

thypon: fixed

PR #1615, opened 2015-05-19
edbrowse: update to 3.5.4.

CMB:
This also contains a patch from upstream, so that edbrowse will
build under musl.  This will go away with the next release.


chneukirchen:
Very nice! Are you using Void Linux too?

CMB:
Christian Neukirchen <notifications@github.com> writes:

> Very nice! Are you using Void Linux too?

Yes, as of a couple weeks ago, and I'm really enjoying it too.
I lurk in #xbps (and elsewhere) as teiresias.

-- Chris

PR #1626, opened 2015-05-21
New package: sopwith-1.8.4

beefcurtains:
Upstream package can't be cross compiled, so I made a patch that fixes it.


srcpkgs/sopwith/template
```diff
@@ -0,0 +1,24 @@
+# Template file for 'sopwith'
+pkgname=sopwith
+version=1.8.4
+revision=1
+build_style=gnu-configure
+depends="SDL"
+makedepends="${depends}-devel"
+hostmakedepends="autoconf automake ${makedepends}"
```
xtraeme: This is completely wrong... you only really need `SDL-devel` in `makedepends`.

beefcurtains: That'll fail when autoreconf is run.

xtraeme: In this case just drop `depends`. It's unnecessary with `xbps-src`.

beefcurtains: Thank you :heart: and sorry, my bad!

srcpkgs/sopwith/template
```diff
@@ -0,0 +1,23 @@
+# Template file for 'sopwith'
+pkgname=sopwith
+version=1.8.4
+revision=1
+build_style=gnu-configure
+makedepends="${depends}-devel"
```
xtraeme: This is wrong...

beefcurtains: I know, already fixed.

srcpkgs/sopwith/template
```diff
@@ -0,0 +1,23 @@
+# Template file for 'sopwith'
+pkgname=sopwith
+version=1.8.4
+revision=1
+build_style=gnu-configure
+makedepends="SDL-devel"
+hostmakedepends="autoconf automake ${makedepends}"
```
xtraeme: `autoconf` is redundant because it's a dependency of `automake`

beefcurtains: Sorry about that, should I also rebase the whole thing?

xtraeme: it is ok

PR #1627, opened 2015-05-21
New package: cdm-git-20140705

thypon:


xtraeme:
No git packages please.

thypon:
Unfortunatly the mantainer never released a "version" even if it's in mantainance mode. If it's neded i'll fork and release a version but i don't think it's needed.

thypon:
Closed for now. I'll try with another DM.

PR #1631, opened 2015-05-22
New package: ipset-6.24

beefcurtains:
Administration tool for IP sets in the Linux kernel


srcpkgs/ipset/template
```diff
@@ -0,0 +1,32 @@
+# Template file for 'ipset'
+pkgname=ipset
+version=6.24
+revision=1
+build_style=gnu-configure
+configure_args="--with-kmod=no"
```
xtraeme: What's the reason to disable kmod?

srcpkgs/ipset/template
```diff
@@ -0,0 +1,32 @@
...
+hostmakedepends="pkg-config"
+makedepends="libmnl-devel"
+short_desc="Administration tool for IP sets"
+maintainer="beefcurtains <beefcurtains@users.noreply.github.com>"
+license="GPL-2"
+homepage="http://ipset.netfilter.org/"
+distfiles="${homepage}ipset-6.24.tar.bz2"
```
xtraeme: `6.24` -> `${version}`

srcpkgs/ipset/template
beefcurtains: The kernel module should already be present, or built into the kernel when this tool is run.

xtraeme: You cannot always assume that:

```
[juan@leysa void-packages]$ grep -i ip_set /boot/config-4.0.4_1 
CONFIG_IP_SET=m
...
```

So that if you disable kmod the kernel module perhaps won't be loaded automagically.

beefcurtains: Dynamic loading seems to rely on the presence of dl.h (regardless of the kmod switch), but I can't find that anywhere.

xtraeme: Dynamic loading via libltdl (which you are missing in `makedepends` ) has nothing to do with kernel modules.

Please remove `--with-kmod=no` and add `libltdl-devel libkmod-devel` to `makedepends`.

xtraeme: And also please add `configure_args="--sbindir=/usr/bin"`.

beefcurtains: Ah, yes. It has plugins which it can optionally link dynamically (and that facility should indeed be provided by libltdl). However, building the kernel module requires access to the kernel sources during the build (not ``libkmod-devel``). I don't really see the point as more often than not, the kernel module is already built and installed.

``libltdl-devel`` provides ``ltdl.h``, but it's looking specifically for ``dl.h`` and fails. It also looks for the ancient GNU dld library, which no longer even exists. The same thing happens on Arch and Debian (who also specify --with-kmod=no, btw), because the dynamic loader support is so old it doesn't build right on a modern system. The resulting binary is 468KB on my system and the library is 506KB, so I can see why no one has bothered to fix that yet.

No part of the code deals with loading a kernel module, I think Linux invokes insmod by itself whenever it has to.

xtraeme: Ok, the info you gave makes sense.

xtraeme: This is good then, thanks :+1:

beefcurtains: Using /usr/bin doesn't make a whole lot sense. ``ebtables``, ``iptables`` and ``ip6tables`` are also in /usr/sbin. Changing the path may break existing firewall scripts. Using the tool requires root access, anyway.

xtraeme: We are migrating `/sbin` to `/bin` progressively, for new pkgs I'd like to see `/bin` by default.

There's no point in using `/sbin` anymore...

beefcurtains: Alright, thank you very much :heart: :heart:

PR #1633, opened 2015-05-22
New package: re2c-0.14.2

beefcurtains:
This program partially builds itself and isn't meant to be cross compiled, but if the host compiler is the same as the target compiler it should be fine.


xtraeme:
Looks good too :+1:

PR #1634, opened 2015-05-22
bandwith / download times + $GNU_SITE cleanup

pullmoll:
+ To reduce bandwidth and download times, packages use bz2/lzma/xz where possible.
+ Also use ${GNU_SITE} and ${KERNEL_SITE} where appropriate.
+ Clean up xlint warnings.
+ inetutils: update to 1.9.3 and split into subpkgs


beefcurtains:
Why increment the revision in cases where only the url has changed, not anything in the resulting binary package?

pullmoll:
Because otherwise the pkg isn't rebuilt and the download thus not verified.

PR #1639, opened 2015-05-23
New package: peerflix-0.30.0

thypon:


beefcurtains:
Peerflix has [quite a few dependencies](https://github.com/mafintosh/peerflix/blob/master/package.json) and npm has its own unique way of dealing with them. What I don't understand is why you need an xbps package to run npm for you, there's no reason you can't let several different package managers install stuff simultaneously. Now if any of the dependencies have an update (which is at least once or twice a week), you'll need to update this as well. If not automated, this creates extra busywork.

Also if npm packages are copied verbatim (without a prefix), there will be conflicts down the road. The following are also npm packages: bash, bison, dash, firefox, gcc, i3, java, etc.

thypon:
It installs all of them in a peerflix prefix. If you tried it you could see.

beefcurtains:
I meant package names, not binary names. However, having a /usr/$pkgname seems backwards to me. Might as well install GoboLinux if that's what you prefer.

thypon:
I don't get what are you saying; i don't even know gobolinux. Atom is packaged exactly in this way in voidlinux; maybe you could switch debian to have atom and other nodejs amenities in 2030.

thypon:
NPM, pip, rubygems are a way to have libraries in development phase; then a proper distribution needs a proper package to avoid installing build tools (gcc, git) to have atom (for example).

xtraeme:
Afaik, this package can be directly installed via npm even for a specific user via `npm`. It's pointless to package this for xbps.

I'd like to avoid packaging redundant copies that can be installed by its package manager, in fact, I'm starting to think that refusing such packages is a good idea, that means:

- for python, refuse pkgs that can be installed easily via `pip`
- for perl, refuse pkgs that can be installed easily via `cpan`
- for nodejs, refuse pkgs that can be installed easily via `npm`
- ...

So that I don't see the point for this package.

beefcurtains:
I can see the need for such packages if they're a build dependency, though.

thypon:
@xtraeme it's easy also to install a lot of package for go with. Why we don't remove also all the go applications?

It's not about how easy is to install things but how easy it's to manage and upgrade packages. Having a unique and centralized package manager it's a value; at least i can manage/upgrade all the things i need w/o changing my workflow for every single development package manager X. Could be good if xbps could, at least, trace all the diff done during installation with (EG) ptrace by npm and at least track the files installed in the filesystem.

dominikh:
> Why we don't remove also all the go applications?
You're comparing apples and oranges here. For one, Go doesn't have a package manager, it just runs `$DVCS clone` and compiles code. Go programs also don't have any Go-specific runtime dependencies, especially not a Go interpreter. Node, Python and so on do.

Nevertheless, I agree that we should be packaging **applications**, even if they can be installed via npm, pip etc. Using multiple package managers to install applications has several drawbacks, most significantly the risk of file conflicts. If I pip install a program that installs to /usr/bin/foo and gem install another program that also installs to /usr/bin/foo, I just corrupted my installation. Using xbps would avoid this (implicitly). Not to mention that, in my opinion, it's an insane idea to use language-specific package managers to install programs system-wide (such a package might just as well overwrite /usr/bin/cp).

I also don't think that we should discriminate applications based on what language they were written in. The enduser doesn't care about it, nor should he have to deal with that language's package manager. peerflix and rtorrent, the latter of which we package, should be treated equally.

We might as well kick out most of the programs written in C, all it takes to install them is curl, ./configure and a compiler.

beefcurtains:
It doesn't matter what language an application is written in, what matters is how it deals with dependencies and if it's already available to you via another package manager. Npm drags in the entire dependency tree for each package you install, which may even include multiple different versions of any given package. You could still create a repository of such packages and have the applications installed into /opt/${pkgname} where such packages would normally reside.
And if your package manager thinks it's allowed to install with prefix=/usr instead of prefix=/usr/local it's broken and should be fixed. Npm already uses /usr/local for global installs. In any case, using xbps to install these applications does not solve the conflict situation. You're merely shuffling chairs and the result is more issues here.  
rtorrent (written in C++) uses gnu autotools, so it has none of the problems that npm packages have.

thypon:
@beefcurtains all languages deal with dependencies in a different. Go package/binaries are compiled _statically_ on _behalf a different version of libraries_ (for each binary). Until we start to compile also Go binaries non-static your complains does not apply.

@dominikh comparing nodejs vs go was obviously provocative but something is true:
1. The dependencies in both of them are (with _NPM_ in node and _go get_ in go) are at a fixed version when "compiled" or "installed"
2. It's false that go binaries doesn't need a runtime like nodejs. It's easy to see what's the golang runtime:
```
ldd /usr/bin/hub 
        linux-vdso.so.1 (0x00007ffd36fbd000)
        libpthread.so.0 => /usr/lib/libpthread.so.0 (0x00007fbce1d21000)
        libc.so.6 => /usr/lib/libc.so.6 (0x00007fbce197e000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fbce1f3f000)
```

dominikh:
I'd really rather not derail this conversation, but: I said "Go-specific" runtime. Requiring the libc is not Go-specific. Also, it's quite easy to build Go programs that do not link to the libc either and create completely static binaries. Provocative arguments are fine, as long as they're not invalid :-) It's not an apt comparison, because it suffers none of the issues that provoked this discussion in the first place.

thypon:
Closed for now until someone @xtraeme change his mind on these packages or we choose democratically how to handle these packages. I'm currently using it in a archlinux schroot.

PR #1646, opened 2015-05-25
New Package:  jwm-s1103

silvernode:


srcpkgs/jwm/template
```diff
@@ -0,0 +1,21 @@
+# Template file for 'jwm'
+
+pkgname="jwm"
+version="s1103"
+revision=1
+build_style=gnu-configure
+hostmakedepends="pkg-config"
+makedepends="automake cairo-devel libjpeg-turbo-devel
```
xtraeme: automake must be in hostmakedepends instead.

srcpkgs/jwm/template
```diff
@@ -0,0 +1,21 @@
...
+pkgname="jwm"
+version="s1103"
+revision=1
+build_style=gnu-configure
+hostmakedepends="pkg-config"
+makedepends="automake cairo-devel libjpeg-turbo-devel 
+libpng-devel libXext-devel libXrender-devel 
+libXmu-devel libXft-devel libXinerama-devel
+libXpm-devel fribidi-devel"
+short_desc="JWM is a light-weight window manager for the X11 Window System"
```
xtraeme: please remove "JWM is"... it's redundant.

silvernode:
it's been awhile, I pretty sure I just forgot 32 bit deps

xtraeme:
You are also missing `conf_files="/etc/system.jwmrc"`

xtraeme:
Also, you should replace `cairo-devel` with `librsvg-devel`... otherwise `cairo` support is disabled.

PR #1654, opened 2015-05-26
New package: detox-1.2.0

ViktorF:


beefcurtains:
[Fix pending](https://github.com/ViktorF/void-packages/pull/1)

srcpkgs/detox/template
```diff
@@ -0,0 +1,15 @@
...
+version="1.2.0"
+revision=1
+build_style=gnu-configure
+configure_args="--prefix=/usr --sysconfdir=/etc/detox"
+patch_args="-Np1"
+short_desc="Utility designed to clean up filenames"
+maintainer="Martin Födinger <foedinger.martin@autistici.org>"
+license="BSD"
+homepage="http://detox.sourceforge.net/"
+makedepends="flex libfl-devel"
```
xtraeme: - move `flex` to `hostmakedepends`
- remove `libfll-devel` (unnecessary)

srcpkgs/detox/template
```diff
@@ -0,0 +1,15 @@
+# Template file for 'detox'
+
+pkgname="detox"
+version="1.2.0"
+revision=1
+build_style=gnu-configure
+configure_args="--prefix=/usr --sysconfdir=/etc/detox"
```
xtraeme: - You can remove `--prefix=/usr`, it's set by default for `build_style=gnu-configure`

srcpkgs/detox/template
beefcurtains: Removing ``libfl-devel`` breaks cross compilation, it's a dependency of flex which should be installed on the host *and* the target. The latter doesn't happen in that case and the [build fails](https://travis-ci.org/beefcurtains/void-packages/builds/64162431).

PR #1656, opened 2015-05-26
New package: gdb-python

pullmoll:
QtCreator does not work with the non-python enabled gdb and depending on an option is not possible.
+ Remove option python from gdb and instead create a distinct package
+ Later make qtcreator depend on the new gdb-python package


xtraeme:
This is a mess... I'll enable the python option by default then.

pullmoll:
Sorry to be a PITA :(

xtraeme:
No problem :-)

pullmoll:
You said you didn't want what you just did, thus I tried the approach on a different path. Glad you changed your mind ;-)

PR #1660, opened 2015-05-26
perl: force rebuild, fix xlint

beefcurtains:
I wonder why it says in #1659 perl is missing on armv?hf-musl, maybe this'll fix it.


xtraeme:
perl has just been rebuilt right now, thanks.

PR #1662, opened 2015-05-27
New package: dumb-1.0

beefcurtains:
I've successfully built and ran a frontend for this, it produces quality vibrations.


srcpkgs/libdumb/template
```diff
@@ -0,0 +1,38 @@
+# Template file for 'libdumb'
+pkgname=libdumb
+version=1.0
+revision=1
+
+# dumb breaks xlint with its directory structure
+wrksrc=dumb-${version}
```
xtraeme: This is wrong... please use:

```
wrksrc=dump-${version}
build_wrksrc=cmake
```

xtraeme:
The package is ok, but would be nice if the generated shared library is `versioned`, i.e `libfoo.so.major.minor` rather than `libfoo.so`.

Otherwise xbps won't be able to detect soname bumps.

beefcurtains:
Agreed. Dumb doesn't do that, what do you suggest? ``libdumb.so.1`` and in ``-devel`` a symlink ``libdumb.so -> libdumb.so.1``?

xtraeme:
The build system must be changed to use `-Wl,-soname libdumb.so.1` and then you can make a `libdumb.so` symlink and move it to `-devel`.

beefcurtains:
:heart: :heart: Fixed, tested, works perfectly! Thank you *SO MUCH*! :heart: :heart:

beefcurtains:
Please don't merge this yet, I should change the name back to ``dumb`` first (renamed it before filing this PR because I thought that'd be better but it violates the naming policy). I'd also like to take a stab at including an example frontend (e.g. dumb2wav).

srcpkgs/dumb/template
```diff
@@ -0,0 +1,35 @@
...
+esac
+
+pre_install() {
+	vdoc ../readme.txt
+	vdoc ../release.txt
+	vlicense ../licence.txt
+}
+
+dumb-devel_package() {
+	depends="${pkgname}>=${version}_${revision}"
```
xtraeme: rather than `pkgname` you can use `sourcepkg` that expands to the main pkg name.

pullmoll: I also made this type of mistake (`pkgname` instead of `sourcpkg`) in several packages and will identify and update them...

beefcurtains: [A change](https://github.com/chneukirchen/xtools/commit/0f2ee339ae133e7de9045d3c7d480335a8144cd7) has already landed in xtools that should prevent this in the future.

beefcurtains:
There's stuff in there that shouldn't be in there, not sure how that happened.

beefcurtains:
It's not pushing the commits:
```
$ git log|head -n 1
commit e477feeff8b7edd773f93df8e794a4d90d6d1994
$ git push --force origin dumb
Everything up-to-date
```

beefcurtains:
Please re-review :space_invader:

PR #1669, opened 2015-05-28
New package: hostaliases-2

thypon:


srcpkgs/hostaliases/template
```diff
@@ -0,0 +1,13 @@
...
+version=1
+revision=1
+build_style=gnu-makefile
+short_desc="HOSTALIASES libc independant support"
+maintainer="Andrea Brancaleoni <miwaxe@gmail.com>"
+license="MIT"
+homepage="https://github.com/thypon/hostaliases"
+distfiles="https://github.com/thypon/hostaliases/archive/v${version}.tar.gz"
+checksum=019afa5850a8a0699d5ff955f66e97b208af72f3b87370a7e81dc17701367b12
+LDFLAGS+=" -shared"
```
xtraeme: Can you please fix this in `hostaliases` itself? if you're building a shared library most likely you need `-shared` and `-fPIC`.

thypon: It's actually like that in hostaliases Makefile but `xbps-src` overwrites them.

xtraeme: You should change your Makefile to this instead:

```
CFLAGS += ...
LDFLAGS += ...
```

PR #1670, opened 2015-05-28
New package: tesseract-ocr-3.02.02

pullmoll:


xtraeme:
The musl build failed due to this:

```
svutil.cpp:234:21: error: ‘FD_ZERO’ was not declared in this scope
```

That file needs to include `sys/select.h`.

PR #1671, opened 2015-05-28
leptonica: fix depends for subpackages

pullmoll:


beefcurtains:
What's wrong with ``${pkgname}>=${version}_${revision}``? Asking because I do the same basic thing in ``ipset`` and the forthcoming ``dumb``.

PR #1675, opened 2015-05-28
leptonica: use $sourcepkg in subpackages' depends

pullmoll:


xtraeme:
There's no need to revbump it because `$sourcepkg == leptonica`.

Please change this next time you update it.

pullmoll:
Ok, closing this then.

PR #1678, opened 2015-05-29
DO NOT MERGE guile: try to hack for building on musl

pullmoll:
The binary is built with the strtol_l wrapper patched in,
but when generating the texi files guild segfaults here.

If anyone finds out why this is, guile should otherwise build with musl.


pullmoll:
FWIW the same segfault happens if I patch `# define USE_GNU_LOCALE_API` and disable it by replacing with `# undef USE_GNU_LOCALE_API` in libguile/i18n.c.
In other words: it's not above patch which is buggy, but something else.

pullmoll:
Further tests and `gdb`'ing the `guile` executable in the chroot reveals that the Boehm garbage colloector `gc` is where it crashes.
To be more specific it crashes right when calling `GC_init()` through the `GC_INIT()` macro.
The `GC_init()` function calls `GC_init_linux_data_start()` which at its end calls `GC_find_limit()`. There a temporary signal handler for catching `SIGSEGV` and `SIGBUS` is installed and this doesn't seem to work, because the (expected?) segment violation is reported.

xtraeme:
Take a look at Alpine: http://git.alpinelinux.org/cgit/aports/tree/main/gc/

pullmoll:
I think this patch is already in `srcpkgs/gc/patches/fix-boehm-gc.patch`!?

pullmoll:
I gave up on trying to get this working.

beefcurtains:
I'm still doing this, it just hasn't had top priority. Some observations:
* Guile tries to allocate a string for the command "vm-run", but when the relevant functions are finished it has become a null string
* Guile then throws an exception but a handler has not yet been initialized (no try/catch block)
* Because of this a segmentation fault occurs (it jumps to an uninitialized pointer iirc).
* The functions that create string buffers implicitly invoke a gc sweep, way too early

The good news is that it's fully reproducible with gc and guile compiled with ``-g -O0``, without using xbps-src. The bad news is, using current git versions for either or both results in the same crash. I may be able to create a workaround (create the translated string directly), but that'll likely result in another crash when the next command is read. Something is going wrong inside gc, but it's a lot of code to step through and examine. Comparing against a glibc build might help.

chneukirchen:
It gets further with GC_DONT_GC=yes, so the GC seems indeed to be the issue.

chneukirchen:
guile --without-threads segfaults as well.

chneukirchen:
fixed via 3df71f2

PR #1679, opened 2015-05-29
New package: geda-1.8.2

pullmoll:
I chose the all lowercase name, while the project sometimes uses gEDA.


srcpkgs/geda/template
```diff
@@ -0,0 +1,46 @@
+# Template file for 'geda'
+pkgname=geda
+version=1.8.2
+revision=1
+wrksrc=${pkgname}-gaf-${version}
+build_style=gnu-configure
+hostmakedepends="pkg-config guile desktop-file-utils"
+makedepends="gc-devel gtk+-devel glib-devel guile-devel"
```
xtraeme: `gtk+-devel` depends on `libglib-devel` already, remove `glib-devel`.

srcpkgs/geda/template
```diff
@@ -0,0 +1,46 @@
...
+depends="guile geda-data>=0"
```
xtraeme: `>=0` is redundant here, it's added automagically.

srcpkgs/geda/template
pullmoll: `xmindep gtk+-devel glib-devel` returns no reduction, but I'll change it and see.

xtraeme: - libglib-devel is the pkg that provides the headers and static libs, so symlinks.
- glib-devel contains some utilities to generate glib stuff and depends on libglib-devel.

pullmoll:
For now *-musl remains broken, because I have no fix for guile.

PR #1680, opened 2015-05-30
New package: ffmpegthumbnailer-2.0.10

pullmoll:


xtraeme:
This file is in a wrong path: `
/home/travis/build/voidlinux/void-packages/masterdir/destdir//ffmpegthumbnailer-2.0.10/usr/share/man1/ffmpegthumbnailer.1`

xtraeme:
Please run:

```
$ cat usr/lib/pkgconfig/libffmpegthumbnailer.pc
```

pullmoll:
Hm.. moving usr/man to usr/share/man is not sufficient then. I'll have to patch something...

xtraeme:
I'd like to see the contents of `libffmpegthumbnailer.pc` yet.

pullmoll:
I got you wrong and thought _I_ should have a look:
```
prefix=/usr
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include

Name: libffmpegthumbnailer
Description: Generates thumbnails for movie files
Version: 2.0.10
Libs: -L${libdir} -lffmpegthumbnailer
Libs.private: -L -lavcodec -lavutil -lavformat -lswscale /usr/lib64/libjpeg.so /usr/lib64/libpng.so /usr/lib64/libz.so
Cflags: -I${includedir}
```

xtraeme:
As you can see, to link against `libffmpegthumbnailer` some additional libs are required `libavcodec, libjpeg, etc`.... that means that the `-devel` pkg is missing those deps.

This is a general problem for all new packagers that don't pay attention to the required dependencies in development packages.

xtraeme:
Solution to what I said above:

```
foo-devel_package() {
   depends="${makedepends} ${sourcepkg}>=${version}_${revision}"
   ...
}
```

pullmoll:
Ok, I see the problem.. without that you would always have to manually add the (unrecorded) dependencies. I'll add that and rebase

xtraeme:
Exactly.

pullmoll:
Could the build scripts detect this situation and warn, just like in the case of missing common/shlibs entries?

pullmoll:
Hmm.. the libffmpegthumbnailer.pc file doesn't seem to have changed here.

xtraeme:
Dependency names in pkg-config files wouldn't match package names, so that a mapping would still be necessary.

pullmoll:
like here with ffmpeg-devel giving libavcodec, libavutil and libavformat.. I see.

PR #1681, opened 2015-05-30
New package: libopenraw-0.0.9

pullmoll:
With this and ffmpegthumbnailer the tumbler package can be completed.
Why the cross builds don't find boost escapes me. Perhaps because the package doesn't recognize `--with-libtool-sysroot`?

### DIFF
--- a/srcpkgs/libopenraw/template
+++ b/srcpkgs/libopenraw/template
@@ -1,0 +1,37 @@
+# Template file for 'libopenraw'
+pkgname=libopenraw
+version=0.0.9
+revision=1
+build_style=gnu-configure
+configure_args="--with-boost=${XBPS_CROSS_BASE}/usr"
+hostmakedepends="pkg-config curl"
+makedepends="glib-devel gdk-pixbuf-devel boost-devel libxml2-devel libjpeg-turbo-devel"
+triggers="gtk-pixbuf-loaders"
+short_desc="Library for camera RAW files decoding"
+maintainer="Jürgen Buchmüller <pullmoll@t-online.de>"
+license="GPL-3"
+homepage="http://${pkgname}.freedesktop.org/wiki/"
+distfiles="http://${pkgname}.freedesktop.org/download/${pkgname}-${version}.tar.bz2"
+checksum=49fd1adf0a0228c7a17a79bf98d8d03664195feae1e50f4ddd1b20162626e18f
+
+LDFLAGS="-L${XBPS_CROSS_BASE}/usr/lib"
+post_configure() {
+	# Fix local posix_{open,close} name clash with unistd.h
+	sed -e "s;posix_open;s_posix_open;g" \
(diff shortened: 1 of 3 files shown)


xtraeme:
Probably because the `configure` script is hardcoding `-I/usr/include`, try to see if there's an option to specify the boost prefix, i.e `--with-boost=${XBPS_CROSS_BASE}/usr` or similar.

pullmoll:
Ok, this required two fixes:
+ explicitly set `--with-boost=${XBPS_CROSS_BASE}/usr` as you suggested
+ fix a name clash for musl unistd.h names `posix_{open,close}` used for local functions

I could perhaps report the latter upstream and ask them to rename the functions.

srcpkgs/libopenraw/template
```diff
@@ -0,0 +1,38 @@
...
+makedepends="glib-devel gdk-pixbuf-devel boost-devel libxml2-devel libjpeg-turbo-devel"
+short_desc="Library for camera RAW files decoding"
+maintainer="Jürgen Buchmüller <pullmoll@t-online.de>"
+license="GPL-3"
+homepage="http://${pkgname}.freedesktop.org/wiki/"
+distfiles="http://${pkgname}.freedesktop.org/download/${pkgname}-${version}.tar.bz2"
+checksum=49fd1adf0a0228c7a17a79bf98d8d03664195feae1e50f4ddd1b20162626e18f
+
+if [ "${CROSS_BUILD}" ]; then
+	configure_args+=" --with-sysroot=${XBPS_CROSS_BASE}"
```
xtraeme: - `--with-sysroot` is always sets with `build_style=gnu-configure`, please remove it.
- This conditional is also unnecessary, just keep `LDFLAGS`.

srcpkgs/libopenraw/template
```diff
@@ -0,0 +1,38 @@
...
+	LDFLAGS="-L${XBPS_CROSS_BASE}/usr/lib"
+fi
+post_configure() {
+	case "${XBPS_TARGET_MACHINE}" in
+	*-musl)	# Fix local posix_{open,close} name clash with unistd.h
+		sed -e "s;posix_open;s_posix_open;g" \
+			-e "s;posix_close;s_posix_close;g" \
+			-i ${wrksrc}/lib/io/posix_io.c
+		;;
+	esac
```
xtraeme: That fix shouldn't be applied only for musl, also for glibc.

xtraeme:
One more thing, this package needs:

```
triggers="gtk-pixbuf-loaders"
```

Because this installs a new image format for gdk-pixbuf and its cache must be updated.

pullmoll:
Is this `gdk-pixbuf-loaders` or really `gtk-...`?

xtraeme:
`triggers="gtk-pixbuf-loaders"` is not a typo

PR #1688, opened 2015-05-30
New package: vim-huge-python3

christer:
vim-huge but using python3 instead of 2.


srcpkgs/vim/template
```diff
@@ -6,7 +6,7 @@ version=${_distver}.${_patchver}
 revision=2
 hostmakedepends="mercurial pkg-config"
 makedepends="ncurses-devel acl-devel libXt-devel gtk+-devel perl
- ruby-devel python-devel lua-devel"
+ ruby-devel python-devel python3.4-devel lua-devel"
```
beefcurtains: Why both?

xtraeme:
Looks good, but please bump the revision.

Thanks :+1:

PR #1691, opened 2015-05-31
linux4.0: make sure $hdrdest dirs exist

pullmoll:
Build failed in x86_64-musl masterdir when running `./xbps-src -m /tmp/musl -N pkg base-system`

### DIFF
--- a/srcpkgs/linux4.0/template
+++ b/srcpkgs/linux4.0/template
@@ -50,9 +50,9 @@ do_configure() {
 			;;
 	esac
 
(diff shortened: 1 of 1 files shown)


xtraeme:
I'd like to see a build log... because I never could reproduce it :-)

pullmoll:
There you go: http://sprunge.us/AEQi

xtraeme:
Ok, thanks.

Please don't revbump it, the pkg won't change.

PR #1694, opened 2015-05-31
New package: arptables-0.0.4

beefcurtains:
Runit service shamelessly lifted from ``ebtables``.

### DIFF
--- a/srcpkgs/arptables/files/arptables/run
+++ b/srcpkgs/arptables/files/arptables/run
@@ -1,0 +1,6 @@
+#!/bin/sh
+RULES=/etc/iptables/arptables.rules
+[ -r conf ] && . ./conf
+[ -r $RULES ] || exit 0
+arptables-restore < $RULES
+exec chpst -b arptables pause
--- a/srcpkgs/arptables/template
+++ b/srcpkgs/arptables/template
@@ -1,0 +1,20 @@
+# Template file for 'arptables'
+pkgname=arptables
+version=0.0.4
+revision=1
+wrksrc="${pkgname}-v${version}"
+build_style=gnu-makefile
+depends="perl"
+short_desc="ARP table administration"
+maintainer="beefcurtains <beefcurtains@users.noreply.github.com>"
+license="GPL-2"
+homepage="http://ebtables.sourceforge.net/"
+distfiles="ftp://ftp.netfilter.org/pub/arptables/arptables-v0.0.4.tar.gz"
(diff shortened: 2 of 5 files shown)


srcpkgs/arptables/template
```diff
@@ -0,0 +1,22 @@
...
+homepage="http://ebtables.sourceforge.net/"
+distfiles="ftp://ftp.netfilter.org/pub/arptables/arptables-v0.0.4.tar.gz"
+checksum=277985e29ecd93bd759a58242cad0e02ba9d4a6e1b7795235e3b507661bc0049
+
+do_install() {
+	vbin arptables
+	vbin arptables-restore
+	vbin arptables-save
+	vman arptables.8
+	touch /etc/arptables.conf
```
xtraeme: Either you have two options here:

- remove this line and update the runit scripts to check for this file
- add this file to `conf_files`

thypon:
:+1:

srcpkgs/arptables/template
beefcurtains: Fixed. It was copy/pasted from ebtables, which may not have the most exemplary template.

beefcurtains:
Travis shows commit id ``a93e0eb`` instead of ``67faa1d``, not sure how to fix this.

srcpkgs/arptables/files/arptables/run
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+arptables-restore < "/etc/arptables.conf"
```
xtraeme: So what happens if that files doesn't exist?

beefcurtains: It'll fail, of course. That's why it's created in do_install().

beefcurtains:
Perhaps it'd be better if a configuration file could specify the location of the rules and whether saving should be skipped or not.
Then again, I could also just leave the whole runit part out. It's really just a convenience thing and I've merely included it for completeness.

xtraeme:
I'd suggest you the following:

- Don't ship the empty `/etc/arptables.conf` conf file.
- Update the runit scripts to exit 1 if the file does not exist.

beefcurtains:
Thanks! That makes a lot of sense. I also took the liberty of changing the default config path to ``/etc/iptables/arptables.rules`` so it's more consistent with other forks. ``ebtables`` should be updated accordingly. Eventually ``nftables`` will replace them all :smile_cat: 
For those who don't want to save the current tables on shutdown, I've added an option SKIP_SAVE that'll do just that.

Thoughts?

PR #1696, opened 2015-05-31
new package: qutebrowser-git

mattiasgiese:


pullmoll:
Install xtools and run `xlint srcpkgs/qutebrowser-git/template` to see a list of things to fix

mattiasgiese:
Thanks for the hint :-)

GiedriusS:
I'm pretty sure that -git packages are not allowed.

Gottox:
What's the reason to use -git over a release?

mattiasgiese:
I have seen some -git packages in the repository so i thought it might be okay.

mattiasgiese:
Well, The-Compiler does not release as often and the current git master has some interesting features like speed enhancements and history lookup.

Gottox:
It's a small world. I talked to @the-compiler about Void on 31c3. I think he'll releases a new version if you ask him kindly.

We deprecated the usage of -git packages some months ago. Therefore if it's not definitely needed we won't accept any non-release packages.

The-Compiler:
But then I'll have to adjust all my issue milestones! :wink:

No, you guys are right, it's probably time for a v0.3 earlier than I wanted.

xtraeme:
`-git` pkgs won't be accepted unless there's a good reason, use the released version.

The-Compiler:
I just released qutebrowser v0.3.0 - so probably you'll get it very soon when @uggedal or someone else had time to update the package :smile:

Sorry it took me another four weeks!

uggedal:
0.3.0 pushed.

The-Compiler:
That was quick - thank you! :+1:

PR #1697, opened 2015-05-31
Manual.md: Fix build_style desc for gnu-makefile

pullmoll:


dominikh:
Merged as 6bf4be3, thanks.

PR #1698, opened 2015-05-31
new package: ttf and otf versions of source-code-pro

mattiasgiese:


srcpkgs/font-sourcecodepro/template
```diff
@@ -0,0 +1,22 @@
...
+revision=1
+noarch=yes
+wrksrc=source-code-pro-${version}
+hostmakedepends="unzip"
+depends="font-util xbps-triggers>=0.58"
+font_dirs="/usr/share/fonts/OTF /usr/share/fonts/TTF"
+short_desc="Monospaced font family for user interface and coding environments"
+maintainer="Mattias Giese <mattias@s99.be>"
+homepage="http://adobe-fonts.github.io/source-code-pro/"
+license="SIL Open Font License 1.1"
```
beefcurtains: The license doesn't exist in ``/usr/share/licenses``, so you should add a call to ``vlicense``.

srcpkgs/font-sourcecodepro/template
```diff
@@ -0,0 +1,22 @@
...
+distfiles="https://github.com/adobe-fonts/source-code-pro/archive/1.017R.zip"
```
beefcurtains: Please use ${version}, thanks.

srcpkgs/font-sourcecodepro/template
xtraeme: `base-files` contains already `/usr/share/licenses/OFL-1.1`:

SIL OPEN FONT LICENSE Version 1.1 - 26 February 2007

So that `vlicense` is unnecessary, and license should be set to `OFL-1.1`

PR #1699, opened 2015-05-31
ebtables: flush rules on exit

thypon:
#1694 
This is the common behavior (in my experience) for all the *tables services in the distributions i tested/i use: fedora/archlinux/debian/ubuntu.


beefcurtains:
This is retarded. Nothing needs to be flushed on shutdown, if you just want to purge the rules you can just flush it by invoking ebtables -F yourself. No need to have runit do that for you.

thypon:
It's simple the default behaviour on all the other distributions i use and i ever used. I don't see problem in complying.

Gottox:
It sounds reasonable to me to disable ebtables when the service is stopped (through sv stop ... for example)

beefcurtains:
But it's not, it shouldn't even be considered a "service". The init system is just used to invoke save and restore as needed. You cannot disable ebtables when it's built into the kernel, and it should just remain functional until the system halts (or at least until all network connectivity is ceased).

xtraeme:
I agree with @beefcurtains

Gottox:
Shorewall for example works like this. But I understand, that both methods are valid.

thypon:
I think should work like the other distribution to not introduce confusion. Plus there is an edge case where saving rules on exit doesn't work: if i'm experimenting with new rules in case i mess up when i reboot/restart the service i have all the rules messed up by my latest rule.

I consider saving more or less like a COMMIT for my *tables rules.

thypon:
Let's say i enable the libvirtd service, then i enable the ebtables service. Libvirtd add rules to the end of the rules list, ebtables service save rules. What happens after 10/20 resets?

beefcurtains:
Enabling the service means you'd like the rules to be saved across reboots. If you don't want that, don't enable the service. It's nice to still have an easy way to load the rules automatically, though. That's why I added the SKIP_SAVE option in my arptables pr.

thypon:
Enabling the service (w/o) knowing what the service exactly does means flushing the rules. That's for me and all the other sysadmins coming from other distribution that has developed a common sense for these things.

Then you could say it's wrong and all the other things. I think it's not wrong enough to change this common thing and the solution you propose has drawbacks. If your solutions would be no save on exit and no flush i'd be ok with that but saving (auto committing) on exit it's actually only wrong in the cases i explained above.

beefcurtains:
You have a couple of options here:
* Not enabling both libvirtd and iptables/ebtables at the same time (have yet to reproduce libvirt mucking with ebtables but I'll take your word for it)
* Fixing libvirt so it's not too retarded to recognize its own rules (at least in iptables there's -m comment)
* Using the "append-only" or "immutable" extended attribute flag on the rules file
* Use etckeeper to always boot into a known good state (e.g. git checkout -f HEAD before anything else)

I don't buy into the bandwagon of doing what other distributions are doing, implying that if others are doing a thing it must be the right thing.
Given that these services ({ip,ip6,eb,arp}tables) are disabled by default (right?), what would be an acceptable meet-in-the-middle solution for you?

thypon:
Neither i do, but *pointer_to_what_i_said_upper.

No save during exit is acceptable, neither flush. I can flush myself. Saving on exit (and not flushing/disabling the service) not only it's the countrary of the common sense but imho it is also dangerous. But hey, they are my opinions: iptables right know does exacltly what imho should do ebtables and arptables. At least we have to choose one thing and stick with that. Having services the behaves quite differently with similar utilities (arptables, iptables, ebtables) isn't good.

In this case i don't see a real advantage to save on exit besides creating confusion to newcomers.

beefcurtains:
If your use case is such that your own rules are loaded on boot and changes are forgotten, the service of ``arptables`` should work for you after doing a ``echo SKIP_SAVE=y >/etc/sv/arptables/conf``. I'd like to harmonize the services and make them work for the most common use cases.

PR #1702, opened 2015-06-01
ebtables: include patch from Alpine to fix *-musl

beefcurtains:
[Original source](http://git.alpinelinux.org/cgit/aports/commit/main/ebtables/musl-fixes.patch?id=0f5076024a9700644ac9d542d2ca643fa38e77da). Thanks to Natanael Copa.

### DIFF
--- a/srcpkgs/ebtables/template
+++ b/srcpkgs/ebtables/template
@@ -2,28 +2,31 @@
 pkgname=ebtables
 version=2.0.10.4
 _version=${version%.*}-${version##*.}
-revision=2
+revision=3
 build_style=gnu-makefile
+conf_files="/etc/ebtables.conf"
 short_desc="Filtering tool for a Linux-based bridging firewall"
-maintainer="Enno Boland <eb@s01.de>"
+maintainer="beefcurtains <beefcurtains@users.noreply.github.com>"
 license="GPL-2"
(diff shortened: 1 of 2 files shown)


srcpkgs/ebtables/template
```diff
@@ -2,28 +2,31 @@
 pkgname=ebtables
 version=2.0.10.4
 _version=${version%.*}-${version##*.}
-revision=2
+revision=3
 build_style=gnu-makefile
+conf_files="/etc/ebtables.conf"
 short_desc="Filtering tool for a Linux-based bridging firewall"
-maintainer="Enno Boland <eb@s01.de>"
+maintainer="beefcurtains <beefcurtains@users.noreply.github.com>"
```
Gottox: noreply addresses make no sense.
Do really you want to take over the maintainership?

Gottox:
Thanks for the fixes! Must have been drunk when I made this template. :D

srcpkgs/ebtables/template
beefcurtains: You actually *can* send me email on that address. However, I'd be happy to revert this change. No hard feelings, I hope?

Gottox: Didn't know that.

No, it's fine. I changed it back myself. :)

PR #1704, opened 2015-06-01
evolution: revbump + hook to build with musl

pullmoll:
Similar to evolution-data-server also evolution is not happy with musl iconv()
Convince configure to go ahead and write a default iconv-detect.h

Also add gnome-icon-theme to hostdepends and depends, as configure wants either gnome-icon-theme or adwaita-icon-theme installed.

The dependency on gnome-keyring is required for authentication with servers.

### DIFF
--- a/srcpkgs/evolution/template
+++ b/srcpkgs/evolution/template
@@ -1,19 +1,19 @@
 # Template file for 'evolution'
 pkgname=evolution
 version=3.12.11
-revision=1
+revision=2
 build_style=gnu-configure
 configure_args="--disable-pst-import --with-openldap --disable-bogofilter
  --with-krb5=/usr --disable-schemas-compile --disable-static --disable-spamassassin
  --disable-text-highlight --disable-gtkspell"
-hostmakedepends="pkg-config intltool itstool gnome-doc-utils gobject-introspection"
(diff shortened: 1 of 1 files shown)


srcpkgs/evolution/template
```diff
@@ -21,6 +21,23 @@ homepage="http://www.gnome.org"
...
+	case "$XBPS_TARGET_MACHINE" in
+	*-musl)
+		# Create a iconv format detection header
+		local hdr=${wrksrc}/iconv-detect.h
+		echo '/* This is an auto-generated header, DO NOT EDIT! */' > $hdr
+		echo  >>$hdr
+		echo '#define ICONV_ISO_D_FORMAT "iso-%d-%d"' >>$hdr
+		echo '#define ICONV_ISO_S_FORMAT "iso-%d-%s"' >>$hdr
+		echo '#define ICONV_10646 "iso-10646"' >>$hdr
+	esac
```
xtraeme: Duplicating this stuff is not what I'd like to use...

`common/environment` and `common/hooks` could be used to generalize this for musl.

pullmoll:
You think of adding `ac_cv_libiconv_utf8=yes` to configure_args for every musl build?

xtraeme:
To set custom autoconf overrides tests see `common/environment/configure/autoconf-cache/musl-linux`.

pullmoll:
Done in `evolution-data-server` pr

PR #1703, opened 2015-06-01
evolution-data-server: revbump, patch + hook to build with musl

pullmoll:
The evolution-data-server configure script doesn't trust musl iconv() implementation.
Tell it to assume it's good by setting `ac_cv_libiconv_utf8=hopefully`
A missing macro `ICONV_10646` is conditionally defined in camel/camel-inconv.h to `iso-10646`, which I hope is right for musl's iconv as well.

PR #1705, opened 2015-06-01
gnome-keyring: fix build on musl

pullmoll:
To get evolution compiled with musl libc, this patch is required as well, because evolution depends on gnome-keyring for authenticating with servers.
Then the evolution-data-server pr #1703 is required and finally evolution #1704 should build.

The result can be previewed when installing `evolution` from `--repository=http://muslrepo.voidlinux.de/current` which has all required packages.

### DIFF
--- a/srcpkgs/gnome-keyring/template
+++ b/srcpkgs/gnome-keyring/template
@@ -1,7 +1,7 @@
 # Template file for 'gnome-keyring'
 pkgname=gnome-keyring
 version=3.14.0
-revision=1
+revision=2
 lib32disabled=yes
 build_style=gnu-configure
 configure_args="--with-pam-dir=/usr/lib/security --disable-schemas-compile"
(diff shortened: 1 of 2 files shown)


srcpkgs/gnome-keyring/template
```diff
@@ -13,6 +13,15 @@ depends="libcap-progs dconf>=0.18 gcr>=3.12"
...
 maintainer="Juan RP <xtraeme@voidlinux.eu>"
 homepage="http://www.gnome.org"
-license="GPL, LGPL"
+license="GPL-2, LGPL-2.1"
 distfiles="${GNOME_SITE}/$pkgname/${version%.*}/$pkgname-$version.tar.xz"
 checksum=c7059516cc47192e8bc5f1095e8e57cd3388825a4584ea8ad67a97763b7c9040
+
+post_extract() {
+	case "$XBPS_TARGET_MACHINE" in
+	*-musl)	# Add sys/select.h for FD_ISSET
```
xtraeme: I'd like to see a proper patch rather than conditional + manual patching.

PR #1713, opened 2015-06-02
chroot-util-linux: disable programs that use pam

beefcurtains:
If the host has ``pam-devel`` installed, bootstrapping fails because chroot-util-linux builds programs that rely on pam. This PR disables those programs explicitly.


xtraeme:
Revbump please

PR #1719, opened 2015-06-02
runit-swap: update to 1.2.0.

thypon:


srcpkgs/runit-swap/template
```diff
@@ -1,6 +1,6 @@
 # Template file for 'runit-swap'
 pkgname=runit-swap
-version=1.1.0
+version=1.2.0
 revision=3
```
xtraeme: Please reset revision.

PR #1726, opened 2015-06-02
moved to usr/bin + removed systemd support

thypon:


srcpkgs/irqbalance/template
```diff
@@ -1,8 +1,7 @@
 # Template file for 'irqbalance'
 pkgname=irqbalance
 version=1.0.9
-revision=1
-build_options="systemd"
+revision=3
```
beefcurtains: Why revbump twice?

thypon: because the first revbump is for move to usr/bin the second is for removed systemd support. So you could rebase on it cherry-picking only the first commit.

beefcurtains: Okay, never mind then :space_invader:

beefcurtains:
Looks good! :+1:

thypon:
I'm not sure for systemd support. If it's ok also for @xtreame i merge this one.

PR #1728, opened 2015-06-02
rfkill: move to usr/bin.

thypon:


beefcurtains:
Looks good, :+1: 

(Should check wicd for breakage.)

PR #1729, opened 2015-06-02
autofs: move to usr/bin.

thypon:


srcpkgs/autofs/template
```diff
@@ -17,6 +17,7 @@ license="GPL-2"
 homepage="${KERNEL_SITE}/daemons/autofs"
 distfiles="${KERNEL_SITE}/daemons/autofs/v5/autofs-${version}.tar.xz"
 checksum=e08ff0ef9ff365d87b6c33dca136d20e4e07675197a324f2816894cb437021c3
+only_for_archs="i686 x86_64 armv5tel armv6l armv7l aarch64"
```
beefcurtains: If it doesn't work with musl, we'll figure out why. Disabling a package for musl is only a last resort. If you're adding this just to pass Travis, please don't. If there's a legitimate reason why this package will never work without glibc, I'd be keen to hear why.

thypon: It's like that because:

* we have travis
* there is no autofs in alpine. I'm not a musl expert and i can't solve the problems easily, when someone succed he could remove this line.

beefcurtains: * A PR can be merged without the OK from Travis
* The first thing to do in order to fix the build for musl, is to remove that line from the template. It serves no purpose other than to make the "merge" button look prettier.

thypon:
#1733 let's see what's the package merged first.

PR #1731, opened 2015-06-02
iptables: move to usr/bin.

thypon:


srcpkgs/iptables/template
```diff
@@ -51,6 +52,5 @@ iptables-devel_package() {
 		vmove "usr/lib/*.so"
 		vmove usr/include
 		vmove usr/lib/pkgconfig
-		vmove usr/share/man/man3
```
beefcurtains: These manpages should remain in iptables-devel.

thypon: If only it'd inside. The build only fails if i leave that line. Try it yourself.

srcpkgs/iptables/template
```diff
@@ -13,6 +12,7 @@ homepage="http://www.netfilter.org/"
 license="GPL-2"
 distfiles="$homepage/projects/iptables/files/$pkgname-$version.tar.bz2"
 checksum=52004c68021da9a599feed27f65defcfb22128f7da2c0531c0f75de0f479d3e0
+configure_args="--sbindir=/usr/bin"
```
beefcurtains: This overrides the configure_args above, that's why the package broke.

thypon: ops my fault i think

PR #1733, opened 2015-06-02
fsarchiver: move to usr/bin.

thypon:


srcpkgs/fsarchiver/template
```diff
@@ -10,5 +10,6 @@ short_desc="Safe and flexible file-system backup and deployment tool"
 maintainer="Juan RP <xtraeme@voidlinux.eu>"
 license="GPL-2"
 homepage="http://www.fsarchiver.org/"
-distfiles="${SOURCEFORGE_SITE}/$pkgname/$pkgname-$version.tar.gz"
+distfiles="${DEBIAN_SITE}/main/f/${pkgname}/${pkgname}_${version}.orig.tar.gz"
```
beefcurtains: I'm aware of the [problems](https://plus.google.com/+gimp/posts/cxhB1PScFpe) [with](http://danluu.com/slashdot-sourceforge/) [sourceforge](https://sourceforge.net/blog/gimp-win-project-wasnt-hijacked-just-abandoned/) lately, but we shouldn't just start piggybacking on Debian, either.
The URL is ``${SOURCEFORGE_SITE}/${pkgname}/${pkgname}-src/${version}/${pkgname}-${version}.tar.gz``.

PR #1738, opened 2015-06-02
radvd: move to usr/bin.

thypon:


beefcurtains:
xlint is telling you to indent with tabs...

PR #1739, opened 2015-06-02
usbmuxd: move to usr/bin.

thypon:


kylesusername:
why exactly are you moving all of these to /usr/bin?

thypon:
because /usr/sbin would became a symlink to /usr/bin shortly.

kylesusername:
But if these commands are meant to be run as root, then /usr/sbin is their standardized home.

beefcurtains:
@kylesusername sbin has been deprecated in void for some time, the templates just hadn't been updated to reflect that until recently. xtraeme is through 75% of his list now.

thypon:
Nowadays this division is more or less ridiculous. The capability system is in place synce years and unless you want a single prefixed directory for every single capability (like rbin for NET_RAW capability for example) it's here only to complicate things.

kylesusername:
@beefcurtains Ah, okay. Good to know. 
@thypon Pardon me for not knowing about this deprecation.

dominikh:
Can't we/you update all these packages in a single PR (one commit per changed package)? That'd avoid a lot of work, and merge commits.

thypon:
Yep, we could. But i like to do this thing gradually and with some peer review. On archlinux this change broke the distro badly for some day (imho because was too repentine).

PR #1740, opened 2015-06-02
dhcpcd: removed systemd service

thypon:


beefcurtains:
Change the URL to ``http://ftp.osuosl.org/pub/blfs/conglomeration/${pkgname}/${pkgname}-${version}.tar.bz2``, the original site appears to be down.

PR #1741, opened 2015-06-02
New Package: font-hanazono-20141012

diogoleal:


beefcurtains:
Commit messages should start with the name of the package that's been modified.
Also, could you split this PR in half? These two don't appear to be related.

diogoleal:
The problem was that I created a template for font-hanazono and then created one for processing. Then github linked the two commits in one.

I will take a git reset previous to that sent the processing.

beefcurtains:
This is 8 commits, could you please rebase?

diogoleal:
This is already solved. 
The problem was that I sent two packages in the same branch and then I made a correction in the template to use vinstall() function instead of install command

xtraeme:
This PR contains 8 commits, please rebase and then force push the relevant commit... not 20K.

PR #1749, opened 2015-06-03
New Package: lmms-1.1.3

### LINKED ISSUE #1718: Package Request: lmms
https://lmms.io/

logenkain:
closes #1718 

I'm not sure if comments are allowed in templates, but I put a couple in there in case someone package libstk. Or if wine gets a 64 bit version.

If there is a way to deal with wine let me know, as far as I know lmms requires wine-devel during build in order to support VST plugins via wine.  I have no idea if there is a way to do this without making the whole program 32bit only.
I suspect there is a way I can add an extra depend for the 32-bit version only so people can use VST's with at least the 32bit version, but now I'm thinking I should have glossed over the manual before commiting.  Let me know if there are any ideas to deal with wine wine.


logenkain:
I just found this
"lib32depends"
I'm guessing I should add "lib32depends='wine-devel'", but I'm going to wait for reviews before changing the template.  Or if there is a better way, please let me know.

srcpkgs/lmms/template
```diff
@@ -0,0 +1,18 @@
+# Template file for 'lmms'
+pkgname=lmms
+version=1.1.3
+revision=1
+build_style=cmake
+nocross=yes
```
beefcurtains: If this is here to stop it from compiling on musl, this is not how you stop it from compiling on musl.

logenkain: On Wed, Jun 03, 2015 at 12:05:16AM -0700, beefcurtains wrote:
> If this is here to stop it from compiling on musl, this is not how you stop it from compiling on musl.

I have the nocross for arm.  The musl is new to me.

xtraeme:
Take a look at how `wine` is packaged.

logenkain:
On Wed, Jun 03, 2015 at 09:07:00AM -0700, Juan RP wrote:
> Take a look at how `wine` is packaged.

I'm not sure if I understand what you're going for.  Are you suggesting
I make it 32-bit only?

I love this style of learning, but I'm afraid I'm still not sure what
I'm supposed to be referencing.

logenkain:
Two things:
One, I'm not sure what to do about wine short of making lmms 32-bit only, so let's just ignore it.  It's an expirmental feature anyway.
Two, how do I bypass musl?

logenkain:
I noticed xlint triggers on only_for_arches=i686, not sure if that's supposed to be.

logenkain:
I think I'll close this so I can get my branch up to date.  xbps-src doesn't seem to work anymore, I'm guessing all those updates changed something.

beefcurtains:
If it's indeed okay to build without wine support, it certainly can be packaged.
Reopen this issue if you're still interested.

PR #1763, opened 2015-06-04
New package: subtle-contrib-125

thypon:


thypon:
Broken :-(

pullmoll:
You need to either disable -Werror or change the `#include <sys/poll.h>` to `#include <poll.h>`. The latter can be unconditionally, i.e. works with glibc as well.

thypon:
It's subtle and not subtle-contrib that fails. Testing subtle-contrib i found some bug in the scripts from `subtle-contrib`.

pullmoll:
Ah, ok, got you wrong then.. sorry :)

Issue #1760, opened 2015-06-04
build hostmakedepends are not always installed while building


Gottox:
See this error log: https://gist.github.com/Gottox/9a9cf5433086f8dc5d16
The template contains intltool as hostmakedepend, but it isn't installed

```
[tox@buildbox void-packages]$ xs chroot
=> Using `/home/tox/void-packages/hostdir/binpkgs/gnome-3.16' as local repository.
[*] Updating `http://repo.voidlinux.eu/current/x86_64-repodata' ...
[*] Updating `http://repo.voidlinux.eu/current/nonfree/x86_64-repodata' ...
[*] Updating `http://muslrepo.voidlinux.eu/current/x86_64-repodata' ...
ERROR: [reposync] failed to fetch file `http://muslrepo.voidlinux.eu/current/x86_64-repodata': Not Found
[*] Updating `http://muslrepo.voidlinux.eu/current/nonfree/x86_64-repodata' ...
ERROR: [reposync] failed to fetch file `http://muslrepo.voidlinux.eu/current/nonfree/x86_64-repodata': Not Found
[*] Updating `http://repo.voidlinux.eu/current/multilib/x86_64-repodata' ...
[*] Updating `http://repo.voidlinux.eu/current/multilib/nonfree/x86_64-repodata' ...
[tox@/home/tox/void-packages/masterdir /]$ xbps-query -Rs intltool
[-] intltool-0.51.0_1 Toolbox for internationalization
[tox@/home/tox/void-packages/masterdir /]$ 
```

xtraeme:
I wasn't able to reproduce it since this issue was submitted.

Gottox:
Hadn't had the issue since then. I'm opening a new issue once I'm able to reproduce it.

PR #1765, opened 2015-06-04
Gnome 3.16

Gottox:
Tested under VirtualBox and Native - No showstoppers so far.


thypon:
<3 it :-)

Gottox:
Seems that travis suffers from #1760

Issue #1635, opened 2015-05-22
Package request: polygen


thypon:
https://packages.debian.org/unstable/games/polygen

Generative Grammars Sentence Generator

beefcurtains:
The downloads on the homepage are gone, but there's a github mirror with releases here: https://github.com/piger/polygen

thypon:
We should use the version on debian site.

PR #1767, opened 2015-06-04
New package: polygen-1.0.6.ds2

thypon:


thypon:
#1635

beefcurtains:
Why nocross?
Why grab from debian?
Why install another copy of GPL2?

thypon:
1. Doesn't crossbuild because of ocaml. All the other packages inside voidlinux ocaml based doesn't crosscompile IMK.
2. The debian version contains also some grammar to use/test it out of the box. The vanilla version isn't updated since 11 years wouldn't be a problem using the debian version.
3. It has a license, it's good to include it. It's not the full GPL2.

PR #1769, opened 2015-06-05
qt5: unbreak musl

pullmoll:
It built locally on musl, but untested yet. Specificially the `resolv_compat.h` code may be doing something wrong. You get the idea, anyway.

### DIFF
--- a/srcpkgs/qt5/patches/qt-musl-pthread_getattr_np.patch
+++ b/srcpkgs/qt5/patches/qt-musl-pthread_getattr_np.patch
@@ -1,0 +1,13 @@
+The musl libc also has pthread_getattr_np()
+
+--- qtwebengine/src/3rdparty/chromium/third_party/WebKit/Source/platform/heap/ThreadState.cpp	2015-02-17 05:58:05.000000000 +0100
++++ qtwebengine/src/3rdparty/chromium/third_party/WebKit/Source/platform/heap/ThreadState.cpp	2015-06-05 02:29:15.253256411 +0200
+@@ -54,7 +54,7 @@
+
+ static void* getStackStart()
+ {
+-#if defined(__GLIBC__) || OS(ANDROID)
++#if defined(__GLIBC__) || OS(ANDROID) || OS(LINUX_MUSL)
+     pthread_attr_t attr;
+     if (!pthread_getattr_np(pthread_self(), &attr)) {
+         void* base;
--- a/srcpkgs/qt5/patches/qt-musl-resolve.patch
+++ b/srcpkgs/qt5/patches/qt-musl-resolve.patch
@@ -1,0 +1,29 @@
+The musl resolver does not define res_ninit() and res_nclose() functions
+like glibc does. A wrapper for musl to mimick GLIBC's function should do
+the trick.
+
+--- qtwebengine/src/3rdparty/chromium/net/base/dns_reloader.cc	2015-06-03 12:34:26.979892244 +0200
++++ qtwebengine/src/3rdparty/chromium/net/base/dns_reloader.cc	2015-06-03 12:32:07.091899808 +0200
+@@ -8,6 +8,9 @@
+     !defined(OS_ANDROID)
+
+ #include <resolv.h>
++#if defined(OS_LINUX) && !defined(__GLIBC__)
++#include "net/dns/resolv_compat.h"
++#endif
+
+ #include "base/basictypes.h"
+ #include "base/lazy_instance.h"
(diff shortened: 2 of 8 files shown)


srcpkgs/qt5/patches/qt-musl-pthread_getattr_np.patch
```diff
@@ -0,0 +1,13 @@
+The musl libc also has pthread_getattr_np()
+
+--- qtwebengine/src/3rdparty/chromium/third_party/WebKit/Source/platform/heap/ThreadState.cpp	2015-02-17 05:58:05.000000000 +0100
++++ qtwebengine/src/3rdparty/chromium/third_party/WebKit/Source/platform/heap/ThreadState.cpp	2015-06-05 02:29:15.253256411 +0200
+@@ -54,7 +54,7 @@
+ 
+ static void* getStackStart()
+ {
+-#if defined(__GLIBC__) || OS(ANDROID)
++#if defined(__GLIBC__) || OS(ANDROID) || defined(__MUSL__)
```
xtraeme: This __MUSL__ is completely wrong and will never be accepted upstream...

srcpkgs/qt5/patches/qt-musl-resolve.patch
```diff
@@ -0,0 +1,29 @@
...
+like glibc does. A wrapper for musl to mimick GLIBC's function should do
+the trick.
+
+--- qtwebengine/src/3rdparty/chromium/net/base/dns_reloader.cc	2015-06-03 12:34:26.979892244 +0200
++++ qtwebengine/src/3rdparty/chromium/net/base/dns_reloader.cc	2015-06-03 12:32:07.091899808 +0200
+@@ -8,6 +8,9 @@
+     !defined(OS_ANDROID)
+ 
+ #include <resolv.h>
++#if !defined(__GLIBC__)
```
xtraeme: I would change those to `#if defined(OS_LINUX) && !defined(__GLIBC__)`

srcpkgs/qt5/patches/qt-musl-pthread_getattr_np.patch
pullmoll: I just needed some NAME and locally insert a #define NAME in case of *-musl.
I'll choose something less offending then.

srcpkgs/qt5/patches/qt-musl-resolve.patch
pullmoll: ok

beefcurtains:
Excellent work! One question, though: why inline those functions? Does that work around a compile time error?

pullmoll:
`static inline` is a work around 1)  the need to separate function declaration and implementation 2) the avoidal of multiple definitions in case the header is included, but the functions are not used in a source file. We used this a lot in the MAME project ;)

I also made it this way because res_ninit() and res_nclose() are very short and can very well be inlined.

pullmoll:
FWIW the local build seems to hang printing
```=> WARNING: qt5-doc-5.4.1_5: binpkg is being created, waiting for 1s...```
I stopped it and will try again, but I think this can be merged.

**Comment by pullmoll** [comment]
It worked the second time round now and can be merged IMO.

File: srcpkgs/qt5/template [hunk 3]
```diff
@@ -24,13 +24,23 @@ makedepends="
...
-case "$XBPS_TARGET_MACHINE" in
-	*-musl) broken="http://build.voidlinux.eu/builders/x86_64-musl_builder/builds/723/steps/shell_3/logs/stdio";;
-esac
-
 pre_configure() {
 	sed -i "s|-O2|${CXXFLAGS}|" qtbase/mkspecs/common/{g++,gcc}-base.conf
 	sed -i "/^QMAKE_LFLAGS\s/s|+=|+= ${LDFLAGS}|g" qtbase/mkspecs/common/gcc-base.conf
+
+	# Compatibility functions res_ninit() and res_nclose() for musl libc
+	cp ${FILESDIR}/resolv_compat.h ${wrksrc}/qtwebengine/src/3rdparty/chromium/net/dns
```
beefcurtains: Stupid question perhaps, but shouldn't this only be done for musl?

pullmoll: Actually in recent adittions xtraeme seemed to tend to quite the opposite, i.e. create patches that are glibc compatible and only take effect when building against musl libc.
In this case the additional file doesn't hurt glibc builds, because it isn't included at all.
There is no special handling except detecting the LINUX && != GLIBC case.

beefcurtains: Oh, I see. My bad, thanks for explaining!

Issue #1686, opened 2015-05-30
Package Request: Lutris - Open Gaming Platform


JayVii:
https://lutris.net/
"Lutris is an open gaming platform for Linux. It helps you install and manage your games in a unified interface.
Our goal is to support every game which runs on Linux,
from native to Windows games (via Wine) to emulators and
browser games. The desktop application and the website
are libre software, your contributions are welcome!"
Currently, there are about 428 games in the database: https://lutris.net/games/

The Client is available for Ubuntu/Debian, Fedora/OpenSUSE, Arch, Gentoo, Slackware, as Python-installer, or as Source from Github:
https://lutris.net/downloads/

JayVii:
Lutris saved me a lot of headaches on my Slackware-Machine, as I'd usually have to compile the games for it. Lutris however downloads the binaries as "fat-packages" (including most libraries and packages you need), so you have minimal effort. Also, they usually provide the newest versions available, which is another plus.

PR #1772, opened 2015-06-05
New Package: lutris-0.3.6.3

JayVii:
#1686
cleaner pull-request :P

### DIFF
--- a/srcpkgs/lutris/template
+++ b/srcpkgs/lutris/template
@@ -1,0 +1,16 @@
+# Template file for 'Lutris'
+pkgname=lutris
+version=0.3.6.3
+revision=1
+build_style=python-module
+pycompile_module="lutris"
+short_desc="Open gaming platform for managing games in a unified way"
+maintainer="Jan Wey. <janwey.git@gmail.com>"
+homepage="https://lutris.net/"
+license="GPL-3"
+distfiles="http://lutris.net/releases/${pkgname}_${version}.tar.gz"
+checksum=a9061403ef8844b2c1443b1c5f15fe98af224d652d66268b47ae8e215e289058
+depends="python python-gobject python-yaml python-xdg gvfs"
+makedepends="python python-devel python-xdg"
+hostmakedepends="python-setuptools"
+wrksrc=lutris


srcpkgs/lutris/template
```diff
@@ -0,0 +1,44 @@
...
+
+do_install() {
+	# Installing /usr/bin and /usr/share
+		vbin lutris/bin/lutris
+		vinstall lutris/data/apps.lutris.gschema.xml 755 /usr/share/lutris/
+		vinstall lutris/data/net.lutris.xboxdrv.policy 755 /usr/share/lutris/
+		vinstall lutris/data/net.lutris.xboxdrvctl.policy 755 /usr/share/lutris/
+		vcopy lutris/data/media /usr/share/lutris/
+		vcopy lutris/data/ui /usr/share/lutris/
+		vcopy lutris/debian /usr/share/lutris/
```
beefcurtains: You sure? This is usually the stuff to create a Debian package.

JayVii: That's the .tar.gz package from https://lutris.net/downloads/
it comes with a python-script that can be used as installer and my do_install() does similar things.
The "debian" folder contains changelogs, license-files, dependency lists.

I have lutris running on my device right now, using a package made from this template

beefcurtains: I think you can safely ignore whatever sits in debian/

JayVii: surely, i'll remove it. I'll hold back the commit for now, in case somebody finds something else that should be addressed. :)

beefcurtains: Commits are free, and you can always rebase. Or just do it the easy way and do ``git commit --amend`` (you'll need to ``git push --force origin lutris``)

JayVii: alright, commit is up :)

beefcurtains: You can pretty much omit the whole ``doc`` section.

JayVii: SO, we really should get rid of all thatis not mandatory for actually running the application? (docs, manuals, etc)

beefcurtains: Not everything, just the useless baggage.

beefcurtains:
The elephant in the room here, is that this doesn't compile a GPLv3 program. Please use the appropriate build style for this project.

JayVii:
I am aware of this, yet I cannot  quite see how. I'm not experienced in these things, could easily be total BS:

 but as far as i can see turns their "Makefile" just the .tar.gz into a .deb https://github.com/lutris/lutris/blob/master/Makefile

A similar thing is done with .rpm 
https://github.com/lutris/lutris/blob/master/INSTALL.rst
https://github.com/lutris/lutris/blob/master/lutris.spec

Other than that, it tells you to "run the script" 
https://github.com/lutris/lutris/blob/master/INSTALL

Their "/bin/lutris" isn't actually a compiled binary, but a python-script
https://github.com/lutris/lutris/blob/master/bin/lutris

beefcurtains:
When you compare the rpmspec and the python-module build-style, they have something in common. In both cases it boils down to this:
```
python setup.py build
```
So supposedly if the host has [setuptools](https://pythonhosted.org/an_example_pypi_project/setuptools.html) installed, it could start a build. So ``python-setuptools`` goes in ``hostmakedepends`` and the other build depends (``python-devel``, ``python-xdg``) go in ``makedepends``. Setup.py handles installation too, so you can remove ``do_install`` and do whatever else needs to be done before/after install in the relevant pre/post hooks.

Try this, see if you can capture what the rpmspec does and translate that into a template.

JayVii:
I just ran said script on my main rig (Debian-X86_64), "python-setuptools" is installed.
output: http://pastebin.com/raw.php?i=1PDTvhFe
As you can see, nothing didn't really change, except for a folder "build" that now is filled with some essentials (actually everything important but the .desktop and the image-files).
however, this is nothing different from my template, it doesn't even rename anything :S

beefcurtains:
I cannot derive your conclusion from the exhibit you presented. But if it's pure python, you should drop the architecture restriction. And if everything but the ``.desktop`` file and icon is included, your ``post_install`` is going to be much smaller than the current ``do_install``.

srcpkgs/lutris/INSTALL
```diff
@@ -0,0 +1,5 @@
+case "${ACTION}" in
+post)
+	chmod -R 755 /usr/share/glib-2.0/schemas/
```
xtraeme: This is really wrong.

JayVii: @xtraeme I did this because of permission-errors leading lutris to crash upon downloading games. Turns out my entire system had permission errors all over the place...
As soon as I fixed that, I'll try it again and maybe get rid of this then obsolete INSTALL file

xtraeme: @JayVii this should be fixed upgrading base-files to 0.128.

JayVii:
what the hell -.-" git is trolling me hard. I'll try to fix it

beefcurtains:
Drop the ``do_extract`` function and use ``wrksrc``. And again, remove the architecture requirement.

JayVii:
I did set the arch-restriction before, not because "Lutris" wouldn't run on anything other than x86/x86_64, but because all games on there are x86/x86_64 exclusive, so I thought there was no point in making it available to other architectures.

beefcurtains:
I don't mind overriding ``do_install``, but are you sure ``python setup.py install`` is broken enough not to bother using it?

JayVii:
When i put "python setup.py install" in my do_install() it would place the files in the correct spots ON MY SYSTEM and not inside ${destdir}.

so after running "./xbps-src pkg lutris" i had an EMPTY "lutris-0.3.6.3-1.xbps" file and lutris was installed on my system.
not sure, how that would happen... Doesn't make much sense, does it?

JayVii:
which parent would I chose for that? I'm guessing the first?
-> 532eb6c

pullmoll:
No, the first _before_ your first commit, i.e. the one in `git log` that is below your first change.

JayVii:
that should be d2600d1 then. I'll get right to it :P

beefcurtains:
The build-style passes ``--root=${DESTDIR}``, that should work. Try building without a ``do_install`` and see if there's anything still missing.

JayVii:
@beefcurtains wow, this does indeed give me a functional package this time.
All that changed from last time i tried it without do_install() is the wrksrc=lutris...
this could have been **SO MUCH** easier than it has been.

beefcurtains:
Now it starts to look like a quality template. The only thing left to do is making sure there are no redundant dependencies, or dependencies that are automatically generated. ``python3`` is probably not a real dependency. ``glib-networking`` and ``gsettings-desktop-schemas`` are dependencies of ``gvfs``. Flesh out your dependencies and it'll get my official stamp of approval :smile_cat:

JayVii:
Working for me without these deps. therefor they are not necessary :)
Thanks for the handholding!

beefcurtains:
:+1: looks ok to me

silvernode:
Thanks for taking the time to package this, I've not tried it out yet. I am waiting for the merge :)

srcpkgs/lutris/template
```diff
@@ -0,0 +1,15 @@
...
+version=0.3.6.3
+revision=1
+build_style=python-module
+short_desc="Open gaming platform for managing games in a unified way"
+maintainer="Jan Wey. <janwey.git@gmail.com>"
+homepage="https://lutris.net/"
+license="GPL-3"
+distfiles="http://lutris.net/releases/${pkgname}_${version}.tar.gz"
+checksum=a9061403ef8844b2c1443b1c5f15fe98af224d652d66268b47ae8e215e289058
+depends="python gir-freedesktop python-yaml python-xdg gvfs"
```
xtraeme: gir-freedesktop -> python-gobject

xtraeme:
This needs to set `pycompile_module="lutris"`

JayVii:
I should have asked **BEFORE** rebase'ing:
anything else to do?

xtraeme:
Looks ok now

JayVii:
Alright. Thanks :)

PR #1774, opened 2015-06-05
Correcting display of icons in xombrero

diogoleal:
Correcting display of icons in xombrero
Without the Adwaita-icon-themes the icons in Xombrero were not displayed.

Screenshot without adwaita-icon-theme
http://diogoleal.com/tmp/xombrero_no_icons.png

With adwaita-icon-theme:
http://diogoleal.com/tmp/xombrero_with_icons.png

Also changed the desc field to have 72 columns.


tonkazoid:
Midori has the same problem. Wanted to fix it but forgot about it... (using firefox)

diogoleal:
You're right, I forgot to change revision-number. :(

I will change and re-commit.

I'll take the test with Midori and if it works, do a pull request.

PR #1775, opened 2015-06-05
xombrero: Correcting display of icons

diogoleal:
Correcting display of icons in xombrero
Without the Adwaita-icon-themes the icons in Xombrero were not displayed.
Also changed the desc field to have 72 columns.


beefcurtains:
Your commit message is too long.

cavalier911:
Let the user decide what icon themes they have installed. Xombrero looks fine with  lxde-icon-theme.
Why should a user have to install a 25MB icon theme in order to have xombrero on there system?

diogoleal:
I agree that the font size is pretty big (25mb). The problem is that if you look, there are some icons that are missing from this screenshot you sent.

See the difference with mine.
screenshot


I use the awesome that is a dynamic window manager and it has no icon by default when I use the xombrero first thought it was a packaging bug

beefcurtains:
Perhaps there could be a virtual package, e.g. ``gtk-icon-theme``, which is satisfied by either icon theme? Some solution where they don't conflict would be great. This would also be a good case for optional depends, imho. The user could install her own icon theme and never bother with any of the packages.

xtraeme:
This is not the solution, a specific icon theme shouldn't be a forced dependency of a GTK app.

PR #1776, opened 2015-06-05
midori: Correcting display of icons in Midori

diogoleal:
Gtk3 uses Advaita-icon-theme as default icon


beefcurtains:
Your commit message should follow [our guidelines](https://github.com/voidlinux/void-packages/blob/master/CONTRIBUTING.md#user-content-committing-your-changes).

cavalier911:
Let the user decide what icon themes they have installed. Midori looks fine with lxde-icon-theme.
Why should a user have to install a 25MB icon theme in order to have xombrero on there system?

xtraeme:
This is not the solution, a specific icon theme should not be a dependency of a GTK application.

PR #1785, opened 2015-06-07
DO NOT MERGE New package: git-imerge-0.7.0 

thypon:


dominikh:
Maybe install the bash completion as well?

thypon:
It's not in the 0.7.0, i could install the latest since there is no update in an year.

PS: ops i changed your commend instead of creating a new one :-)

dominikh:
Maybe convince the maintainer to make a new release. Unless there's no useful/important changes (other than bash completion), then just package 0.7.0 and be happy :-)

P.S. addressing your P.S.S.: You're a collaborator, you can edit issues and comments. Apparently that includes those of other collaborators. Why? Because GitHub.

thypon:
The stable version isn't working as expected, closing for now.

PR #1786, opened 2015-06-07
dcron: fix path of run-cron in root's crontab

CMB:


xtraeme:
This shouldn't be necessary after the /usr/sbin merge... upgrade your system and runit-void will make the `/usr/sbin -> /usr/bin` symlink if `/usr/sbin` is a real directory and does contain only symlinks.

PR #1787, opened 2015-06-07
   qt5: update to 5.4.2  

pullmoll:
+ remove obsolete qtwebkit patch
+ don't use -reduce-relocations with musl (for a try)
+ add qt5-examples package (uncompiled, noarch)

I also reordered some functions so that the *-pkg-devel is always right before *-pkg.


xtraeme:
Looks ok :+1:

PR #1789, opened 2015-06-07
libvirt: add libvirt group

thypon:
/usr/share/polkit-1/rules.d/50-libvirt.rules:

```
// Allow any user in the 'libvirt' group to connect to system libvirtd
// without entering a password.

polkit.addRule(function(action, subject) {
    if (action.id == "org.libvirt.unix.manage" &&
        subject.isInGroup("libvirt")) {
        return polkit.Result.YES;
    }
});
```

### DIFF
--- a/srcpkgs/libvirt/template
+++ b/srcpkgs/libvirt/template
@@ -1,7 +1,7 @@
 # Template file for 'libvirt'
 pkgname=libvirt
 version=1.2.16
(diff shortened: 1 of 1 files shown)


beefcurtains:
Does this change anything for sasl users?

thypon:
Isn't sasl orthogonal to polkit?

PR #1792, opened 2015-06-07
Cinnamon 2.6

Gottox:
Tested x86_64 in virtualbox and native

Not tested on musl though.


Gottox:
screenshot from 2015-06-07 23-09-50

xtraeme:
Looks ok :+1:

PR #1793, opened 2015-06-07
diffutils: revive po/Makefile

pullmoll:
Since autoreconf is not run, mkdir_p remains undefined (@mkdir_p@).
Patch it to get the translations installed.


xtraeme:
Looks ok :+1:

PR #1794, opened 2015-06-08
nodejs: use shared http-parser by default

beefcurtains:
I fixed the ``http-parser-devel`` package, this build_option isn't broken anymore :space_invader: :space_invader:


xtraeme:
Looks ok :+1:

PR #1802, opened 2015-06-09
New Package: lmms-1.1.3

logenkain:
closes #1749 

This is a bit of a test.  I notice that 
"./xbps-src -a i686 install lmms"
Fails on my machine without nocross (can't find qt4 for 32bit), but I also noticed on my last pull request that travis built it just fine.  Using that same template, without nocross, I was unable to build it locally.  So this should be interesting.

### DIFF
--- a/srcpkgs/ca-certificates/files/remove-expired-certs.sh
+++ b/srcpkgs/ca-certificates/files/remove-expired-certs.sh
@@ -1,0 +1,51 @@
+#!/bin/sh
+# Begin remove-expired-certs.sh
+#
+# Version 20120211
+
+# Make sure the date is parsed correctly on all systems
+mydate()
+{
+  local y=$( echo $1 | cut -d" " -f4 )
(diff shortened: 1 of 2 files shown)


beefcurtains:
You can enable travis on your own repository.

logenkain:
On Tue, Jun 09, 2015 at 10:54:31AM -0700, beefcurtains wrote:
> You can enable travis on your own repository.
I should probably learn how to do that.  I'm sure it's not hard.

logenkain:
Alright, well this was my last idea to get the VST plugins working anyway.  I'll rebase and revert everything to be normal without VST functionality.

I just need one answer, how do I prevent the musl?

logenkain:
Oh, and "/bin/sh: WINE_CXX-NOTFOUND: command not found" seems to be the primary issue if someone has an idea before I just remove the VST support.

PR #1805, opened 2015-06-10
New package: darkhttpd-1.11

thypon:


srcpkgs/darkhttpd/template
```diff
@@ -0,0 +1,15 @@
...
+build_style=gnu-makefile
+short_desc="A small and secure static webserver"
+maintainer="Andrea Brancaleoni <miwaxe@gmail.com>"
+license="BSD"
+homepage="https://dmr.ath.cx/net/darkhttpd/"
+distfiles="$homepage/$pkgname-$version.tar.bz2"
+checksum=afad244add43c53010b36ada6ccdf1a854885cb37d6aeff103ad0aa4aed96c51
+
+do_install() {
+	vbin darkhttpd
```
xtraeme: I'm missing:
- a manpage
- a runit service
- the license file

thypon: https://unix4lyfe.org/gitweb/darkhttpd/tree no manpage or license file. 

```
head -n 18 darkhttpd.c > "$pkgdir/usr/share/licenses/$pkgname/LICENSE"
```

This is how the archlinux license is generated.

xtraeme:
Running an HTTP server in a privileged port needs root perms... so that I would expect it to have a runit service.

thypon:
It starts at :8080 when it's not run with root perms. I use it as a `python -m SimpleHttpServer` replacement. But it could also have a runit service if i know what's the default directory to serve. 

`mongoose` that does more or less the same thing, doesn't have runit service.

xtraeme:
Ok then.

xtraeme:
:+1:

Issue #1672, opened 2015-05-28
Package request: fcron


thypon:
"Fcron is a periodical command scheduler which aims at replacing Vixie Cron, so it implements most of its functionalities.

But fcron makes no assumptions on whether your system is running all the time or regularly : you can, for instance, tell fcron to execute tasks every x hours y minutes of system up time or to do a job only once in a specified interval of time.

Fcron has also much more functionalities : you can also set a nice value to a job, run it depending on the system load average and much more !"

Particularly pursued for notebook usage where the system isn't always on.

chneukirchen:
(cronie with anacron works well for this usecase already, btw.)

thypon:
I used on my previous installation and it works quite well / has a nice user interface.

PR #1807, opened 2015-06-10
New package: fcron-3.2.0

thypon:
#1672

### DIFF
--- a/srcpkgs/fcron/INSTALL
+++ b/srcpkgs/fcron/INSTALL
@@ -1,0 +1,18 @@
+# *-*-shell-*-*
+#
+case ${ACTION} in
+post)
+	# fix permissions and owners
+	chown fcron:fcron usr/bin/fcrontab usr/bin/fcrondyn
+	chown root:fcron usr/bin/fcronsighup
+	chmod  110 usr/bin/fcron
+	chmod 4110 usr/bin/fcronsighup
+	chmod 6111 usr/bin/fcrontab usr/bin/fcrondyn
+
+	chown root:fcron etc/fcron/fcron.conf etc/fcron/fcron.allow etc/fcron/fcron.deny
+	chmod  640 etc/fcron/fcron.conf etc/fcron/fcron.allow etc/fcron/fcron.deny
+
+	chown fcron:fcron var/spool/fcron
+	chmod  770 var/spool/fcron
+	;;
+esac
--- a/srcpkgs/fcron/template
+++ b/srcpkgs/fcron/template
@@ -1,0 +1,66 @@
+# Template file for 'fcron'
+pkgname=fcron
+version=3.2.0
+revision=1
+build_style=gnu-configure
+configure_args="
+	--sysconfdir=/etc/fcron
+	--with-answer-all=no
+	--with-boot-install=no
+	--datarootdir=/usr/share
+	--datadir=/usr/share
+	--docdir=/usr/share/doc
+	--localstatedir=/var
+	--with-piddir=/run
+	--with-sendmail=sendmail
+	--with-editor=vi
+	--with-shell=/bin/sh
+	--disable-checks"
+make_dirs="
+	/etc/cron.daily 0755 root root
+	/etc/cron.hourly 0755 root root
+	/etc/cron.monthly 0755 root root
+	/etc/cron.weekly 0755 root root"
+hostmakedepends="autoconf perl"
+makedepends="pam-devel readline-devel"
+depends="run-parts"
+short_desc="Feature-rich cron implementation"
+maintainer="Andrea Brancaleoni <miwaxe@gmail.com>"
+license="GPL-2"
+homepage="http://fcron.free.fr"
+distfiles="$homepage/archives/$pkgname-$version.src.tar.gz"
+checksum=eb13b86ed96805d117e19847f1da30b834f1990f7ef4c38f5007234d44679564
+
+system_accounts="fcron"
+fcron_homedir="/var/spool/fcron"
+fcron_shell="/bin/sh"
+
+conf_files="
+	/etc/fcron/fcron.conf
+	/etc/fcron/fcron.allow
+	/etc/fcron/fcron.deny
+	/etc/fcron/pam.conf
+	/var/spool/fcron/systab.orig"
+
+## me: fcron installer is not the king. I'm.
+## fcron: Wine is coming.
+pre_configure() {
+	autoupdate
+	autoconf -f -Wall
+}
+pre_install() {
+	sed -i 's/chgrp/true/g' Makefile doc/Makefile
+	sed -i 's/chown/true/g' Makefile doc/Makefile
+}
+
+post_install() {
+	chmod +rx $DESTDIR/usr/bin/*
+
+	vinstall files/fcron.pam 644 etc/pam.d fcron
+	vinstall files/fcrontab.pam 644 etc/pam.d fcrontab
+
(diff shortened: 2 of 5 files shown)


srcpkgs/fcron/INSTALL
```diff
@@ -0,0 +1,21 @@
...
+	chown root:fcron /usr/bin/fcronsighup
+
+	for _f in /etc/fcron/fcron.{conf,allow,deny}; do
+		[[ -e $_f ]] || continue
+		chown root:fcron "$_f"
+		chmod 640 "$_f"
+	done
+
+	chown fcron:fcron /var/spool/fcron
+	chmod 770 /var/spool/fcron
```
beefcurtains: Why not do all of this in post_install()?

srcpkgs/fcron/INSTALL
```diff
@@ -0,0 +1,21 @@
...
+post)
+	# fix permissions and owners
+	chmod  110 /usr/bin/fcron
+	chmod 4110 /usr/bin/fcronsighup
+	chmod 6111 /usr/bin/fcrontab /usr/bin/fcrondyn
+	chown fcron:fcron /usr/bin/fcrontab /usr/bin/fcrondyn
+	chown root:fcron /usr/bin/fcronsighup
+
+	for _f in /etc/fcron/fcron.{conf,allow,deny}; do
+		[[ -e $_f ]] || continue
```
xtraeme: INSTALL/REMOVE scripts are meant to be run by POSIX shell... please remove those bashisms.

srcpkgs/fcron/template
```diff
@@ -0,0 +1,64 @@
...
+	/etc/fcron/fcron.deny
+	/var/spool/fcron/systab.orig"
+
+## me: fcron installer is not the king. I'm.
+## fcron: Wine is coming.
+pre_configure() {
+	touch /usr/bin/vi /usr/bin/sendmail
+	chmod 755 /usr/bin/vi /usr/bin/sendmail
+	echo "fcron:x:23:" >> /etc/group
+	echo "fcron:fcron:23:23:FCron User:/var/spool/fcron:/bin/sh" >> /etc/passwd
```
beefcurtains: Radical! I'll look into patching the configure.ac instead... :smile::hammer:

srcpkgs/fcron/INSTALL
thypon: seems that using chgrp,chown launch an error. I prefer to mantain these fixes in the same place.

srcpkgs/fcron/template
```diff
@@ -0,0 +1,66 @@
...
+	--sysconfdir=/etc/fcron
+	--with-answer-all=no
+	--with-boot-install=no
+	--datarootdir=/usr/share
+	--datadir=/usr/share
+	--docdir=/usr/share/doc
+	--localstatedir=/var
+	--with-piddir=/run
+	--with-sendmail=/usr/bin/sendmail
+	--with-editor=/usr/bin/vi
```
beefcurtains: Just ``sendmail`` and ``vi`` here, please. And drop the ``nvi`` depends.

srcpkgs/fcron/template
```diff
@@ -0,0 +1,66 @@
...
+	/etc/fcron/fcron.conf
+	/etc/fcron/fcron.allow
+	/etc/fcron/fcron.deny
+	/var/spool/fcron/systab.orig"
+
+## me: fcron installer is not the king. I'm.
+## fcron: Wine is coming.
+pre_configure() {
+	touch /usr/bin/vi /usr/bin/sendmail
+	chmod 755 /usr/bin/vi /usr/bin/sendmail
```
beefcurtains: ``vi`` is not needed due to the ``--disable-checks`` argument above. Sendmail could be added as a host dep, or patched out be removing the relevant ``AC_MSG_ERROR`` line in ``configure.in``

srcpkgs/fcron/template
thypon: A vi implementation is in the unix spec. Right now nvi is also included in hour base-system. If you want to use another you could use EDITOR env variable.

beefcurtains:
[Here's a patch](https://gist.github.com/beefcurtains/81940a5b0176abc08c8a), testing now...

thypon:
the pre_install part is necessary to pass the make phase.

beefcurtains:
That was just a test to see what happens. The ``chmod``s should be done in ``post_install``, is there any particular reason for doing it in ``INSTALL``?

thypon:
Yep, it's a xbps-src workaround. As you see the perms are -r. However without read permission xbps-src can't create the hashes to construct the package. The `make install` leave, more or less, these permission on files but before packaging i need to leave +r, so i fix this thing on extraction.

beefcurtains:
There's no reason to remove the read bit after installation, the Makefile is just being stupid here.

thypon:
Yep and not. 3 out of 4 executable(s) are suid. Being not readable leave no possibility to an attacker that wants to create a ropchain to see the application layout without other information disclousure vector.

beefcurtains:
That's stupid, an attacker can just download the xbps file and take a look.

thypon:
True. Not so usefull in a rolling release + not knowing if the package is custom built or updated. That argument is less more valid in debian/ubuntu/fedora.

beefcurtains:
None of the other setuid binaries are -r, and no one has a problem with that. This reeks of security through obscurity.

thypon:
It's a little overkill. But it's exactly the reason i want this priviledge separated cron server in voidlinux. It's obvious it's not a great security feature if you run the same version in current repository, but it's quite good to avoid run anywhere attacks where a script tries to read the application layout + tries to find gadget to use rops and avoid ASLR.

xtraeme:
You can't run chown/chgrp in templates because xbps-src runs as a regular user. Permissions must be set thru INSTALL scripts.

srcpkgs/fcron/INSTALL
```diff
@@ -0,0 +1,18 @@
+# *-*-shell-*-*
+#
+case ${ACTION} in
+post)
+	# fix permissions and owners
+	chown fcron:fcron /usr/bin/fcrontab /usr/bin/fcrondyn
```
xtraeme: Do not use absolute paths, use relative... i.e `chown ... usr/bin/foo`

srcpkgs/fcron/template
```diff
@@ -0,0 +1,65 @@
...
+makedepends="pam-devel readline-devel"
+depends="run-parts"
+short_desc="Feature-rich cron implementation"
+maintainer="Andrea Brancaleoni <miwaxe@gmail.com>"
+license="GPL-2"
+homepage="http://fcron.free.fr"
+distfiles="$homepage/archives/$pkgname-$version.src.tar.gz"
+checksum=eb13b86ed96805d117e19847f1da30b834f1990f7ef4c38f5007234d44679564
+
+system_accounts="fcron:23"
```
xtraeme: Why this needs to set `uid 23` specifically?

srcpkgs/fcron/template
```diff
@@ -0,0 +1,65 @@
...
+fcron_homedir="/var/spool/fcron"
+fcron_shell="/bin/sh"
+
+conf_files="
+	/etc/fcron/fcron.conf
+	/etc/fcron/fcron.allow
+	/etc/fcron/fcron.deny
+	/var/spool/fcron/systab.orig"
```
xtraeme: PAM files are missing from `conf_files`.

xtraeme:
Looks ok :+1:

PR #1812, opened 2015-06-11
New package: processing-2.2.1

diogoleal:


beefcurtains:
If it's not built from source, it should go into ``nonfree``. This isn't AUR.

xtraeme:
Packages that use binaries by default should have the `-bin` postifx in pkgname.

Issue #1207, opened 2015-03-27
Missing gnome-3.16 packages


Gottox:
These packages are part of the gnome-3.16 release but not in void right now. See https://github.com/Gottox/void-packages/tree/gnome-3.16 for ongoing process. Help is welcome, too ;)

* [ ] accerciser
* [x] aisleriot
* [x] five-or-more
* [x] four-in-a-row
* [ ] gitg
* [ ] gnome-calendar
* [ ] gnome-characters
* [x] gnome-chess
* [ ] gnome-devel-docs
* [x] gnome-klotski
* [x] gnome-mahjongg
* [x] gnome-mines
* [x] gnome-nibbles
* [x] gnome-robots
* [ ] gnome-software
* [x] gnome-sudoku
* [ ] gnome-taquin
* [x] gnome-tetravex
* [ ] gnome-todo
* [x] hitori
* [x] iagno
* [x] lightsoff
* [x] orca
* [ ] polari
* [x] quadrapassel
* [x] swell-foop
* [x] tali

pullmoll:
Updated the list with the games which are there now.

PR #1815, opened 2015-06-11
New package: libgit2-glib-0.22.8

beefcurtains:
This is for ``gitg`` as requested by #1207. It doesn't want to build without gir.


xtraeme:
Looks ok :+1:

PR #1816, opened 2015-06-11
opencv: update to 3.0.0

beefcurtains:
- [ ] musl support
- [x] ~~missing cv.py~~ (shouldn't be a problem)


beefcurtains:
I gave up on musl :crying_cat_face:

xtraeme:
Note that due to those soname bumps, its revdeps must also be revbumped to be rebuilt against this new version.

beefcurtains:
What revdeps?

thypon:
```
$ xbps-query -RX libopencv
libopencv-devel-2.4.11_1
libopencv-python-2.4.11_1
nomacs-2.4.4_1
nomacs-qt5-2.4.4_1
opencv-2.4.11_1
openimageio-1.5.13_4
qtox-20150325_2
```

thypon:
How is going with that? I really need the performance gain i'd have with opencl or gomp. I run a opencv "script" during system startup.

beefcurtains:
Alright, feel free to pick up where I left. If you can live without -musl, it's ready to go.

PR #1820, opened 2015-06-11
New package: OpenXcom-1.0

beefcurtains:
Won't build on ARM because it uses GL.

Requires [data files from the original game](http://www.myabandonware.com/game/x-com-ufo-defense-1pl).


xtraeme:
Looks ok :+1:

PR #1826, opened 2015-06-12
New package: thermald-1.4.2

thypon:


srcpkgs/thermald/template
```diff
@@ -0,0 +1,28 @@
+# Template file for 'thermald'
+pkgname=thermald
```
Gottox: why not call it thermal_daemon like the package is actually named?

thypon: Bacause the (only) binary and the service are called thermald: plus the name is thermald in ubuntu/debian/arch/fedora.

xtraeme:
Looks ok :+1:

srcpkgs/thermald/template
```diff
@@ -0,0 +1,28 @@
...
+version=1.4.2
+revision=1
+_pkgname=thermal_daemon
+wrksrc="${_pkgname}-${version}"
+only_for_archs="i686 i686-musl x86_64 x86_64-musl"
```
beefcurtains: Why this? Does Linux not support thermal devices on ARM?

thypon: 01.org it's intel. It's an intel specific piece of software; uses intel technologies to achieve better thermal management.

PR #1827, opened 2015-06-13
New package: gtkspell3

beefcurtains:
This builds gtkspell 3.0.7 for GTK+3 only. The API is incompatible with 2.x versions, packages using that API can continue using the existing gtkspell packages.

With this package, it should be possible to build gnome applications (e.g. evolution) with spell check support.


xtraeme:
:+1:

PR #1834, opened 2015-06-14
New package: hugo-0.14

thypon:


srcpkgs/hugo/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'hugo'
+pkgname=hugo
+version=0.14
+revision=1
+build_style=go
+hostmakedepends="git mercurial"
+depends="python-Pygments"
+go_import_path="github.com/spf13/${pkgname}"
+short_desc="A Fast % Modern Static Website Engine"
```
dominikh: Is the `%` some clever stylistic trick or a typo of `&`?

thypon: Typo :-D

xtraeme:
Another pkg that can be easily installed thru go's pkg manager... 

Seriously I don't see the point of packaging this.

thypon:
Because they are tools i use i'm happy to share with voidlinux users. A lot of C programs are easy to install with the curl/make package manager too but we package them however.

logenkain:
On Sun, Jun 14, 2015 at 10:26:38PM -0700, Juan RP wrote:
> Another pkg that can be easily installed thru go's pkg manager... 

I've noticed that you don't like to have packages in void (except depends) that
can be installed via the various language package managers.

I understand you don't see the point, but other than maintainence,
what are the problems with packaging them?  

I've never used go, but I'd prefer using xbps over most other options.
Even if xbps templates could be written to use the other package
managers, at least then xbps would be able to track them.

But yea, just currious.

xtraeme:
The reason is that commonly xbps needs root to be able to install those programs into the rootfs, and with the specific package manager for that language, those applications can be built and installed independently for any users.

Another reason is its mutability: like those pkgs that depend on specific software versions as dependencies, an update of a dependency might break completely the target software.

That's why I'd prefer to not duplicate this effort in xbps.

dominikh:
Go *doesn't have* a package manager. Go is a compiled language. Users of a static site compiler (hugo) do not *care* about Go, and shouldn't have it installed, either. If gcc had a tool for downloading a tarball and compiling it, would Void not package 99% of its packages?

xtraeme:
You are right, for this specific package I was wrong.

Go software is special because once built there's no possibility of breakage due to external dependencies. I was talking explicitely about ruby gems, python pip, npm and ghc cabal.

This package is ok, and my comments do not apply for Go applications.

xtraeme:
:+1:  then :-)

PR #1836, opened 2015-06-15
Add kpcli package

deathsyn:
Plus dependencies


deathsyn:
I think this failure is due to the musl workers being down

deathsyn:
The patch for this I snagged from the Alpine package.

PR #1841, opened 2015-06-16
New package: processing-bin-2.2.1

diogoleal:


thypon:
I don't know if the package is okay or not, but _please_ learn how to rebase commits instead of opening a new pull request **every** time you do something wrong. You have 12 pull request and 1 package merged. Only processing-bin package has 4 pull request associated.

The command is really simple:
```
EDITOR=$YOUREDITOR git rebase -i HEAD~$NUM
```
where YOUREDITOR is your editor and NUM cotains how many commits you want to rebase/squash/rename. `-i` means interactive mode.

beefcurtains:
I've had problems with that, so I just ``git reset --hard HEAD^``, ``git pull upstream master``, ``git cherry-pick $commit_id`` (can be found in ``git reflog`` if lost). This never fails for me 😍

diogoleal:
I apologize, but I did not know what to do in relation to resend the commit, so I ran ```git reset --hard HEAD~1``` and performed a new commit.

I will not go wrong again about it.

I apologize again.

thypon:
:+1:

srcpkgs/processing-bin/template
```diff
@@ -0,0 +1,49 @@
...
+
+do_install() {
+	cd processing-${version}
+
+	# Remove libraries from other operating systems and architecture.
+	rm -rf modes/java/libraries/serial/library/macosx
+	rm -rf modes/java/libraries/serial/library/windows*
+	rm -rf modes/java/application/launch4j/w32api/
+
+	if [ $XBPS_TARGET_MACHINE == "x86_64" ]; then
```
xtraeme: `==` -> `=`

srcpkgs/processing-bin/template
```diff
@@ -0,0 +1,49 @@
+# Template file for 'processing-bin'
+pkgname=processing-bin
+version=2.2.1
+revision=1
+only_for_archs="i686 x86_64"
+short_desc="Processing is a programming language, IDE and online community"
+maintainer="Diogo Leal <diogo@diogoleal.com>"
+license="GPL-2, LGPL-2.1"
+homepage="http://processing.org"
+repository="nonfree"
```
xtraeme: The software is GPL licensed... there's no need to put it into the `nonfree` repository.

diogoleal: @xtraeme It internally includes java Oracle, so I put in nonfree repository.

beefcurtains: Java isn't a standard part of the operating system, so this can't be considered GPL and thus non-free. It's lawyertasticly logical, but holy shit, why is there a JRE in there?

diogoleal: According to its documentation, it only "works" with java Oracle, so they provide a binary with java in by issues of dependency and compatibility

I started writing a template for Oracle jdk, to compile the processing, but my question is really worth it.

thypon: The oracle JDK is alredy included in voidlinux.

diogoleal: I was not using ```nonfree``` repository, so did not see him.

I will then include a message in ```INSTALL.msg``` stating that this package comes with Oracle's Java not confiltar with openjdk.

diogoleal:
@thypon  Yes, these lines run!

thypon:
I think you did the opposite of what i said. :-P

Remove ${DESTDIR} from first operand. The command shouldn't fail however you remove the ${DESTDIR} but the semantic is wrong.

diogoleal:
Ah! Sorry, I got it wrong. I made the changes you suggested.

thypon:
Ok for me.

diogoleal:
Processing only working with Oracle's Java, so I think create two packages:

processing-bin with Java Oracle's internal. to peoples who do not want install Oracle's Java or who just want to OpenJDK.

When processing want install the Java Oracle's.

In your opinion, I'm thinking correctly?

thypon:
I'm curious if using another version of java is needed. Right now this [class][1] downloads the Oracle standard JDK. The fact that it needs the oracle JDK doesn't mean we could repackage the oracle JDK without paying the Oracle JDK Embedded fee's.

Right now [they state][2] that processing under openjdk is not supported; in the same page they say also they only support ubuntu. If i'm correctly interpretating what they state isn't that processing currently doesn't work under openjdk but that they don't want to address bugs if you run something dissimilar from their particular software configuration.

If we can remove the java oracle jre directory and also the oracle jdk/jre dependency the correct action should be to only use openjdk.

Otherwise the correct action, should be to remove the java directory in the DESTDIR and link to the correct system wide oracle jdk location downloaded through the oracle-jdk/oracle-jre official voidlinux package.

[1]: https://github.com/processing/processing/blob/master/build/jre/src/Downloader.java
[2]: https://github.com/processing/processing/wiki/Supported-Platforms#Linux.html

diogoleal:
The processing seems to me that is attached to this internal java because they use the 1.7.40 version. and when I tried to create a symbolic link from the java directory for Oracle's Java, I received an error message when trying to use the processing.

```Annotation processing got disabled, since it requires a 1.6 compliant JVM```

I tried to link only the bin directory, but there was the same error message.

In my view, the operation of the processing is fully locked into that directory with the JVM.

**Comment by thypon** [comment]
testing m'kay

**Comment by diogoleal** [comment]
I reversed the lines.

**Comment by thypon** [comment]
Please also add these lines:

```
rm -rf java
ln -s /usr/lib/jvm/oracle-jre/ ${DESTDIR}/usr/share/processing/java
```

and `oracle-jre` dependency. I dunno if processing team pay the Java Embedded Fee to include oracle jre in their build. We shouldn't for sure.

**Comment by diogoleal** [comment]
I tested remove the java directory and create a symbolic link and not properly worked to compile the processing code.

I did it without creating packaging, just downloaded the tgz and creating the link manually.

**Comment by beefcurtains** [comment]
IIRC with Java you have to set ``JRE_HOME`` and/or ``JDK_HOME``

**Comment by thypon** [comment]
Give me the code that didn't compile.

**Comment by diogoleal** [comment]
```
float[] coswave; 

void setup() {
  size(640, 360);
  coswave = new float[width];
  for (int i = 0; i < width; i++) {
    float amount = map(i, 0, width, 0, PI);
    coswave[i] = abs(cos(amount));
  }
  background(255);
  noLoop();
}

void draw() {

  int y1 = 0;
  int y2 = height/3;
  for (int i = 0; i < width; i+=2) {
    stroke(coswave[i]*255);
    line(i, y1, i, y2);
  }

  y1 = y2;
  y2 = y1 + y1;
  for (int i = 0; i < width; i+=2) {
    stroke(coswave[i]*255 / 4);
    line(i, y1, i, y2);
  }
  
  y1 = y2;
  y2 = height;
  for (int i = 0; i < width; i+=2) {
    stroke(255 - coswave[i]*255);
    line(i, y1, i, y2);
  }
  
}
```

**Comment by thypon** [comment]
```
....
< JAVA_VERSION="1.7.0"
---
> JAVA_VERSION="1.8.0_45"
```

diffing java and oracle-jdk directory ^^

We should try to compile processing from source to see if there are binary "abi" compatibility.

diogoleal:
Okay, I'm doing it.

skrzyp:
OK… what's up with Processing, where did you stopped?

diogoleal:
I could not continue it because my album had problem of I/O and had to wait goodwill Dell support.

I already have new album and I'm creating the processing compiled.

I believe that by Tuesday it is ready (the processing with the new dependencies.).

That was the mistake that was taking place:
http://pastebin.com/L9y1S30s

diogoleal:
I'm working on a template processing compiled, I think I can close this pull request and work on a new.

PR #1845, opened 2015-06-17
webkit2gtk: unbreak musl

pullmoll:


xtraeme:
Looks ok :+1: 

Thanks!

pullmoll:
It doesn't work for armv6l, though. Perhaps an only_for_archs="..." is appropriate here?

PR #1846, opened 2015-06-17
New package: binfmt-support-2.1.5

thypon:


srcpkgs/wine/INSTALL
```diff
@@ -0,0 +1,6 @@
+# INSTALL
```
xtraeme: I'd like to see a xbps trigger rather than duplicating this into multiple packages.

See `srcpkgs/xbps-triggers/files`.

thypon: Only on wine or also for qemu. It's possible to copy binfmt files automatically with triggers?

xtraeme: The xbps trigger should run at post INSTALL time to register new entries for this package, and at post-remove time to unregister the pkg entries.

You'll have to introduce a new variable that accepts one argument or more, like:

```
binfmt="/usr/bin/qemu"
```

and then xbps-src will have to activate the binfmt trigger to be run at post-install or post-remove time.

Note this needs multiple changes to xbps-src and xbps-triggers.

thypon: I have no time to generalize also the qemu case. I'd need architecture checker helpers. The wine one shouldn't be a problem.

thypon:
@xtraeme better? :-)

xtraeme:
Looks good if it works as expected :+1:

thypon:
I fixed two lines in xbps-trigger and added qemu-user-static binfmts. If it's ok could be merged right now.

PR #1848, opened 2015-06-18
lxrandr: add xrandr to depends

cavalier911:
lxrandr displays this error window after install without xrandr on system.
lxrandr without xrandr error


dominikh:
Please bump the revision so the package gets rebuilt with the correct dependencies.

dominikh:
Merged as 4d1a941

PR #1849, opened 2015-06-18
socat: unbreak musl builds for arm*, i686

CMB:
musl's termios doesn't have c_ispeed.

### DIFF
--- a/common/environment/configure/autoconf_cache/arm-common
+++ b/common/environment/configure/autoconf_cache/arm-common
@@ -170,7 +170,6 @@ slrn_cv_va_val_copy=${slrn_cv_va_val_copy=yes}
 sc_cv_sys_crdly_shift=9
 sc_cv_sys_csize_shift=4
 sc_cv_sys_tabdly_shift=11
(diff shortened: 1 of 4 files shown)


common/environment/configure/autoconf_cache/arm-common
```diff
@@ -205,8 +204,6 @@ sc_cv_type_uint32=yes
 sc_cv_type_uint64=yes
 sc_cv_type_uint8=yes
 sc_cv_typeof_struct_cmsghdr_cmsg_len='4 /* unsigned int */'
-ac_cv_ispeed_offset=${ac_cv_ispeed_offset=13}
-sc_cv_termios_ispeed=${sc_cv_termios_ispeed=yes}
```
xtraeme: These ones you've removed should be added to `autoconf_cache/common-glibc` instead.

CMB: > These ones you've removed should be added to
> autoconf_cache/common-glibc instead.

I wondered about that.  Done and force-pushed.

PR #1852, opened 2015-06-19
New package: cdecl-2.5

beefcurtains:
Can translate c type definitions into English and back.

### DIFF
--- a/srcpkgs/cdecl/template
+++ b/srcpkgs/cdecl/template
@@ -1,0 +1,28 @@
+# Template file for 'cdecl'
+pkgname=cdecl
+version=2.5
+revision=1
+wrksrc=${pkgname}-blocks-${version}
+build_style=gnu-makefile
+make_build_args="LIBS=-lreadline"
+hostmakedepends="flex"
+makedepends="readline-devel"
+short_desc="Compose C and C++ type declarations"
+maintainer="beefcurtains <beefcurtains@voidlinux.eu>"
+license="GPL-3"
+homepage="http://${pkgname}.org/"
+distfiles="${homepage}files/${wrksrc}.tar.gz"
+checksum=9ee6402be7e4f5bb5e6ee60c6b9ea3862935bf070e6cecd0ab0842305406f3ac
+CFLAGS="-DUSE_READLINE"
+
+pre_build() {
+	flex -l -o cdlex.c cdlex.l
+	sed -i 's/getline/cdgetline/' cdecl.c
+}
+
+do_install() {
+	vbin c++decl
+	vbin cdecl
+	vman cdecl.1
+	vman c++decl.1
+}
--- a/srcpkgs/cdecl/update
+++ b/srcpkgs/cdecl/update
@@ -1,0 +1,1 @@
+pattern="${pkgname}-blocks-\K[\d.]+(?=\.tar\.gz)"


srcpkgs/cdecl/template
```diff
@@ -0,0 +1,35 @@
...
+license="$(vopt_if readline GPL-3 'Public Domain')"
+homepage="http://${pkgname}.org/"
+distfiles="${homepage}files/${wrksrc}.tar.gz"
+checksum=9ee6402be7e4f5bb5e6ee60c6b9ea3862935bf070e6cecd0ab0842305406f3ac
+
+if [ "${XBPS_TARGET_MACHINE:0-5}" != "-musl" ]; then
+	build_options_default="readline"
+fi
+
+if [ "x$build_option_readline" != "x" ]; then
```
xtraeme: Please do not use `test x$var != x`... just `test "$var"` is enough.

srcpkgs/cdecl/template
```diff
@@ -0,0 +1,35 @@
+# Template file for 'cdecl'
+pkgname=cdecl
+version=2.5
+revision=1
+wrksrc=${pkgname}-blocks-${version}
+build_options="readline"
```
xtraeme: I don't see any point in making readline a build option... I'd just enable it by default.

srcpkgs/cdecl/template
```diff
@@ -0,0 +1,35 @@
...
+hostmakedepends="flex"
+makedepends="$(vopt_if readline readline-devel)"
+short_desc="Compose C and C++ type declarations"
+maintainer="beefcurtains <beefcurtains@voidlinux.eu>"
+license="$(vopt_if readline GPL-3 'Public Domain')"
+homepage="http://${pkgname}.org/"
+distfiles="${homepage}files/${wrksrc}.tar.gz"
+checksum=9ee6402be7e4f5bb5e6ee60c6b9ea3862935bf070e6cecd0ab0842305406f3ac
+
+if [ "${XBPS_TARGET_MACHINE:0-5}" != "-musl" ]; then
```
chneukirchen: This is a not-allowed bashism/kshism.

srcpkgs/cdecl/template
beefcurtains: It is, just not on *-musl, because it fails to build. (Upstream builds without readline by default.)

srcpkgs/cdecl/template
beefcurtains: What's the preferred way to check for ``!musl``?

chneukirchen: case is POSIX sh.

srcpkgs/cdecl/template
```diff
@@ -0,0 +1,36 @@
...
+makedepends="$(vopt_if readline readline-devel)"
+short_desc="Compose C and C++ type declarations"
+maintainer="beefcurtains <beefcurtains@voidlinux.eu>"
+license="$(vopt_if readline GPL-3 'Public Domain')"
+homepage="http://${pkgname}.org/"
+distfiles="${homepage}files/${wrksrc}.tar.gz"
+checksum=9ee6402be7e4f5bb5e6ee60c6b9ea3862935bf070e6cecd0ab0842305406f3ac
+
+# Enable readline on non-musl only (prevents build error)
+if [ "${XBPS_TARGET_MACHINE##*-}" != "musl" ]; then
```
xtraeme: I don't like this. Either enable it for all archs or disable it globally.

xtraeme:
Looks good now :+1:

PR #1853, opened 2015-06-19
opencv: enable opencl/openmp

thypon:


beefcurtains:
Have you tried compiling it on musl? Maybe you can figure out the recursive mutex initializer :star:

thypon:
Nope, we should upgrade to 3.0 instead, maybe it solves musl bugs upstream.

beefcurtains:
[It doesn't](https://github.com/voidlinux/void-packages/pull/1816).

thypon:
Could i merge this one?

beefcurtains:
I'd rather update first, either or both might be default now.

thypon:
We have to revbump also all the dependencies i think.

This update should only improve performance.

thypon:
opencl is default but it needs the headers/devel files to be installed.

beefcurtains:
We'd have to revbump again after updating to 3.0.0, but I don't see any revdeps.
Let's see if adding ``opencl2-headers opencl-headers libgomp-devel`` on 3.0 adds more shlibs.

thypon:
I think we could close this one continuing with opencv 3.0

PR #1855, opened 2015-06-19
Xpra Fixes

thypon:


beefcurtains:
What's the rush? Aren't we supposed to do peer review in here?

thypon:
Sure, it's only a little lib with no fuss. I wasn't sure if was right to merge directly.

PR #1856, opened 2015-06-19
New package: execline-2.1.2.2

bougyman:


srcpkgs/execline/template
```diff
@@ -0,0 +1,27 @@
...
+maintainer="bougyman <bougyman@voidlinux.eu>"
+license="ISC"
+homepage="http://skarnet.org/software/execline/"
+distfiles="http://skarnet.org/software/${pkgname}/${pkgname}-${version}.tar.gz"
+checksum=aea08c2c1b986c91f300d40737dd43067b91705d2c729859344ec65f989aec06
+
+post_install() {
+	vlicense COPYING
+	vdoc README
+	vcopy doc/* usr/share/doc/${pkgname}
```
beefcurtains: ``"doc/*"``, no?

srcpkgs/execline/template
```diff
@@ -0,0 +1,27 @@
...
+}
+
+execline-devel_package() {
+	short_desc+=" - development files"
+	depends="${sourcepkg}>=${version}_${revision}"
+	pkg_install() {
+		vmove usr/include
+		vmove usr/lib
```
beefcurtains: Pardon me, but sure it's not ``"usr/lib/*.so"``?

bougyman: the only thing in usr/lib/ is usr/lib/execline/libexecline.a. Should I specify the full file path, do you think?

srcpkgs/execline/template
bougyman: It worked without quotes, but I will add.

srcpkgs/execline/template
beefcurtains: In that case, might as well pass ``--disable-static`` to ``configure_args``.

srcpkgs/execline/template
```diff
@@ -0,0 +1,26 @@
+# Template file for 'execline'
+pkgname=execline
+version=2.1.2.2
+revision=1
+build_style=gnu-configure
+makedepends="skalibs-devel"
```
xtraeme: This pkg needs to set `configure_args="--libdir=/usr/lib"` and mv the static libs to -devel.

bougyman: Done, rebased, and squashed.

xtraeme:
Looks ok now :+1:

PR #1858, opened 2015-06-19
gnome-desktop: add build option 'gir'

pullmoll:
Like in other packages disable gobject-introspection for cross builds.


beefcurtains:
:ok: by me

xtraeme:
:+1:

PR #1859, opened 2015-06-19
gnome-control-center: Don't abort on failing check

pullmoll:
The test-datetime-gfx check fails because it is missing a not yet
installed data file. This causes the musl build process to abort,
while the glibc build continues.

This patch makes the Makefile continue on errors in datetime checks.


xtraeme:
:+1:

PR #1860, opened 2015-06-19
libgnomekbd: add build option 'gir'

pullmoll:
Also fix cross building by setting LDFLAGS to the appropriate
${XBPS_CROSS_BASE}/usr/lib

### DIFF
--- a/srcpkgs/libgnomekbd/template
+++ b/srcpkgs/libgnomekbd/template
@@ -1,10 +1,10 @@
 # Template file for 'libgnomekbd'
 pkgname=libgnomekbd
 version=3.6.0
-revision=4
+revision=5
 build_style=gnu-configure
 configure_args="--disable-static"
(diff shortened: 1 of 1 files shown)


beefcurtains:
I think this can be done unconditionally: ``LDFLAGS="-L${XBPS_CROSS_BASE}/usr/lib"`` and there's no ``vopt_enable`` or ``vopt_if`` in ``configure_args`` or ``hostmakedepends``. Why is that?

pullmoll:
True, the LDFLAGS can be set unconditionally, since XBPS_CROSS_BASE is empty or undefined for non cross. I simply forgot the vopt_if.

beefcurtains:
Sorry, didn't mean to be rude! :heart: :heart: It looks ok now.

pullmoll:
You weren't rude but right. Thanks for auditing this change :)

PR #1862, opened 2015-06-20
libgweather: fix cross building

pullmoll:
Fix a possible typo: $(vopt_enable ...) doesn't expect true/false parameters,
but $(vopt_if ...) does.

### DIFF
--- a/srcpkgs/libgweather/template
+++ b/srcpkgs/libgweather/template
@@ -1,10 +1,10 @@
 # Template file for 'libgweather'
 pkgname=libgweather
 version=3.16.1
-revision=1
+revision=2
 build_style=gnu-configure
 build_options="gir"
-configure_args="$(vopt_enable gir '--enable-introspection --enable-vala' '--disable-introspection --disable-vala')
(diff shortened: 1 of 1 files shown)


srcpkgs/libgweather/template
```diff
@@ -1,10 +1,10 @@
 # Template file for 'libgweather'
 pkgname=libgweather
 version=3.16.1
-revision=1
+revision=2
 build_style=gnu-configure
 build_options="gir"
-configure_args="$(vopt_enable gir '--enable-introspection --enable-vala' '--disable-introspection --disable-vala')
+configure_args="$(vopt_if gir '--enable-introspection --enable-vala' '--disable-introspection --disable-vala')
```
xtraeme: Good catch, thanks!

pullmoll:
Perhaps the `common/environment/setup/options.sh` function `vopt_enable()` could be improved to check for excess parameters and issue a warning or error?

xtraeme:
Sounds ok

PR #1863, opened 2015-06-20
vopt_enable: check for excess parameters

pullmoll:
I'm don't know if there is an easy way to obtain the function name for the msg_error.
The hint should suffice to detect erroneous use of `vopt_enable` as in #1862.

### DIFF
--- a/common/environment/setup/options.sh
+++ b/common/environment/setup/options.sh
@@ -16,6 +16,9 @@ vopt_with() {
 
 vopt_enable() {
     local opt="$1" flag="${2:-$1}"
+    if [ "$#" -gt "2" ]; then
+        msg_error "vopt_enable $opt: $(($# - 2)) excess parameter(s)\n"
+    fi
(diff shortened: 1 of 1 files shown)


common/environment/setup/options.sh
```diff
@@ -16,6 +16,9 @@ vopt_with() {
 
 vopt_enable() {
     local opt="$1" flag="${2:-$1}"
+    if [ "$#" -gt "2" ]; then
```
xtraeme: Not an issue, but you can skip quoting here completely because `$#` returns an integer and you can compare directly.

beefcurtains:
Shouldn't be such a problem, it's why we review pulls here *shrug*

xtraeme:
:+1:

pullmoll:
Heh, I thought for myself *Should I omit the double quotes? But what if someone someday replaces the "2" by a variable which might be empty?* ... :)

xtraeme:
Either way is ok :-)

Thanks

PR #1866, opened 2015-06-20
ConsoleKit2: update to 0.9.5.

Gottox:
Works for me on Gnome-3.16

Please recheck with other DEs as I haven't checked it yet.


xtraeme:
Does not work for me.

Neither 0.9.3 nor 0.9.4 worked, there must be a regression somewhere.

xtraeme:
Try to open gnome-terminal and then type `ck-list-sessions` the X session must be ACTIVE to work completely.

Gottox:
```
Session1:
	unix-user = '1000'
	realname = '(null)'
	seat = 'Seat1'
	session-type = ''
	active = TRUE
	x11-display = ':0'
	x11-display-device = '/dev/tty7'
	display-device = ''
	remote-host-name = ''
	is-local = TRUE
	on-since = '2015-06-20T21:27:36.817322Z'
	login-session-id = '1'
```

xtraeme:
Will test again just to verify.

xtraeme:
Verified, does not work when running it via startx.

There seems to be a regression since 0.9.3.

Gottox:
https://github.com/ConsoleKit2/ConsoleKit2/issues/24 could be this bug.

xtraeme:
Regression introduced in 0.9.3, waiting for this bug to be fixed: https://github.com/ConsoleKit2/ConsoleKit2/issues/27

Gottox:
The upstream issue is fixed and they release 0.9.5.

PR #1872, opened 2015-06-21
New package: qmc2-0.52

pullmoll:
It obviously needs some more fixes for i686. Please wait a little.
And it seems using rsync for installation, as the original Makefile does, is a bad idea.


pullmoll:
So travis fails for the sheer number of fly/*.png files.
Other than that I think it's okay now.

srcpkgs/qmc2/template
```diff
@@ -0,0 +1,104 @@
...
+wrksrc=${pkgname}
+build_style=gnu-makefile
+make_build_args="PREFIX=/usr DATADIR=/usr/share MAN_DIR=/usr/share/man TARGET=qmc2-mame"
+make_install_args="${make_build_args}"
+hostmakedepends="rsync unzip"
+makedepends="SDL2-devel qt5-multimedia-devel qt5-script-devel qt5-svg-devel
+ qt5-webkit-devel qt5-xmlpatterns-devel qt5-x11extras-devel"
+depends="rsync ${pkgname}-data mame"
+conf_files="/etc/${pkgname}/${pkgname}.ini"
+mutable_files="/etc/${pkgname}/${pkgname}.ini"
```
xtraeme: mutable_files is wrong here, please remove it.

This should only be used with files that are provided by a package but could be modified by other means. If the file is a configuration file there's no need to add it to mutable_files.

srcpkgs/qmc2/template
```diff
@@ -0,0 +1,104 @@
...
+	sed -e "s;sdl-config;sdl2-config;g" -i ${wrksrc}/scripts/sdl-version.sh
+	# Replace obsolete -sdlmame suffix
+	sed -e "s;-sdlmame;-mame;g" -i ${wrksrc}/Makefile
+	# Replace lzma/CpuArch.c with patched version (http://sourceforge.net/p/sevenzip/bugs/1338/)
+	cp -av ${FILESDIR}/CpuArch.c ${wrksrc}/lzma
+}
+do_build() {
+	# Print configuration info
+	make ${make_build_args} config
+	make ${make_jobs} qchdman
```
xtraeme: make_jobs -> makejobs

xtraeme:
:+1:

PR #1876, opened 2015-06-21
Libvirtd musl

thypon:


beefcurtains:
Looks :ok:

srcpkgs/portablexdr/template
```diff
@@ -0,0 +1,26 @@
+# Template file for 'portablexdr'
+pkgname=portablexdr
+version=4.9.1
+revision=1
+only_for_archs="i686-musl x86_64-musl armv6l-musl armv7l-musl aarch64-musl"
+build_style=gnu-configure
+short_desc="Portable XDR is an XDR (RFC 4506) implementation"
```
xtraeme: "Portable XDR is" is really redundant.

beefcurtains: ``XDR (RFC 4506) implementation`` should suffice. I usually think of newspaper headlines when I write the desc (not baity ones), that works.

PR #1879, opened 2015-06-22
minicom: unbreak musl

pullmoll:
Here xlint wanted me to use $DEBIAN_SITE while the original download was on a `alioth.debian.org`, not `ftp.debian.org`.

I switched to $DEBIAN_SITE's corresponding minicom-2.7-orig.tar.gz file to silence xlint :)


pullmoll:
I'm always baffled at _how fscking fast_ musl builds are finished on travis and on the build hosts. 48 vs. 70 seconds is a whopping 32% speedup...

Perhaps it's about time to move away from trusty, old glibc and give musl a try for my box.

PR #1880, opened 2015-06-22
mesa-demos: exclude broken eglut/opengl code

### LINKED ISSUE #1871: mesa-demos: broken with MesaLib-10.6.0
It seems some macros and structures are no longer (not) defined in /usr/include/EGL/eglext.h
+ macro EGL_SCREEN_BIT_MESA
+ struct EGLScreenMESA

AFAICT these were available in earlier versions.
There's a [bug report](https://bugs.freedesktop.org/show_bug.cgi?id=91020) on freedesktop.org suggesting someone should devise a fix by inserting compile guards. That someone is not me :)

mesa-demos is a dependency of gnome-shell through glxinfo (?). Perhaps it should be removed.

pullmoll:
The mesa-demos code including EGL/eglext.h expects some interfaces
which no longer exist in Mesa-10.6.0. Disable that code.

Closes #1871

### DIFF
--- a/srcpkgs/mesa-demos/template
+++ b/srcpkgs/mesa-demos/template
@@ -1,9 +1,9 @@
 # Template build file for 'MesaLib'.
 pkgname=mesa-demos
 version=8.2.0
-revision=6
+revision=7
 build_style=gnu-configure
-hostmakedepends="pkg-config"
+hostmakedepends="automake libtool pkg-config"
 makedepends="libXext-devel MesaLib-devel glu-devel glew-devel freetype-devel libfreeglut-devel"
 short_desc="Mesa 3D demos and tools"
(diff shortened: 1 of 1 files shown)


srcpkgs/mesa-demos/template
```diff
@@ -1,9 +1,9 @@
 # Template build file for 'MesaLib'.
 pkgname=mesa-demos
 version=8.2.0
-revision=6
+revision=7
 build_style=gnu-configure
-hostmakedepends="pkg-config"
+hostmakedepends="autoconf automake libtool pkg-config"
```
xtraeme: autoconf is a dependency of automake

xtraeme:
:+1: after fixing the redundant dependency.

PR #1884, opened 2015-06-22
gpm: unbreak musl

pullmoll:
This has to be tested on a musl box. I'm not sure the SA_INTERRUPT flag bit for sigaction works and causes no harm.


xtraeme:
Looks good, please also remove FILESDIR/gpm.service.

PR #1887, opened 2015-06-22
apr: add missing make dependency expat-devel

pullmoll:
Without the dependency I saw this:
```
xml/apr_xml.c:35:19: fatal error: expat.h: No such file or directory
 #include <expat.h>
                   ^
compilation terminated.
```
It may have worked before because expat was always there or a dependency of libuuid-devel?


xtraeme:
:+1:

PR #1889, opened 2015-06-22
New package: libtorrent1-1.0.5

pullmoll:
Create a new major version package of libtorrent(-rasterbar)
and switch to the new homepage http://www.libtorrent.org/
The license also changed from GPL-2 to BSD.


xtraeme:
:+1:

PR #1891, opened 2015-06-23
New package: qbittorrent-3.2.0

pullmoll:


xtraeme:
:+1:

srcpkgs/qbittorrent/template
```diff
@@ -0,0 +1,29 @@
...
+ $(vopt_if x11 '--enable-gui --disable-webui' '--disable-gui --enable-webui')"
+hostmakedepends="automake libtool pkg-config"
+makedepends="boost-devel libtorrent1-devel
+ qt5-declarative-devel qt5-devel qt5-tools-devel qt5-qmake"
+short_desc="Free software alternative to µtorrent"
+maintainer="Jürgen Buchmüller <pullmoll@t-online.de>"
+license="GPL-2"
+homepage="http://www.qbittorrent.org/"
+distfiles="${SOURCEFORGE_SITE}/${pkgname}/${pkgname}-${version}.tar.xz
+ http://geolite.maxmind.com/download/geoip/database/GeoLiteCountry/GeoIP.dat.gz"
```
beefcurtains: This file is updated every month (July 7th is next), it may be worthwhile to package it separately because of this. Help offered, if wanted.

srcpkgs/qbittorrent/template
```diff
@@ -0,0 +1,29 @@
+# Template file for 'qbittorrent'
+pkgname=qbittorrent
+version=3.2.0
+revision=1
+build_style=gnu-configure
+configure_args="--disable-systemd --with-qt5 --with-geoip-database-embedded
+ $(vopt_if x11 '--enable-gui --disable-webui' '--disable-gui --enable-webui')"
```
Gottox: I prefer two build options. one for enabling/disabling gui and one for webui.

srcpkgs/qbittorrent/template
pullmoll: I'd rather disable the --with-geoip-database-embedded option in that case and see how to make it use the external, separate package. Thank you for any help.

srcpkgs/qbittorrent/template
pullmoll: Okay, I would call the options just like the configure uses them: `gui` and `webui`.
The question remains if the global `x11` option should still be used to define the build_options_default?

PR #1893, opened 2015-06-23
bash: add systemwide bashrc

lemmi:
this adds:
  - /etc/bash/bashrc - general bashrc which should only contain essentials and should not be touched by the user. i'm thinking about making this read-only
  - /etc/bash/bashrc.d/ - the user directory where one can add configs without touching files owned by the package


chneukirchen:
If bashrc is not meant to be modified, it should not be in conf_files.

beefcurtains:
Everything in /etc/ should be user-modifiable (it's not like there's any support on anything, anyway). Changing permissions to read-only is just silly, it doesn't mean shit to root.

lemmi:
@beefcurtains i know root doesn't care, but vim tells you that you are doing something you should think about first. IIRC sudoers file is also read-only. also at the moment it's still modifyable and we could debate this in another issue, i'd rather have this pulled than delayed because of that.

maybe the documentation lacks some information about conf_files

Issue #1894, opened 2015-06-23
mozjs17: broken library and script names


pullmoll:
Something with configuring mozjs17 version number goes wrong.
+ The dynamic library coming out of the build is `libmozjs-.so`
+ The configure script is just `js-configure`, not `js17-configure` as it is used in the -devel package

I have no real clue which part of the build system could be responsible for this bug.
It is possibly the new perl-5.22 (#1850), because in `configure` the MOZJS_MAJOR_VERSION is derived from MOZILLA_VERSION which in turn is returned by a perl script in the configure line `… $PERL $srcdir/config/milestone.pl -topsrcdir $srcdir`

pullmoll:
It looks like some deprecated warning output of perl-5.22 is coming in the way of the expected output.
I ran the `milestone.pl` script manually and got this:
```shell
perl /tmp/glibc/builddir/mozjs17.0.0/js/src/config/milestone.pl -topsrcdir /tmp/glibc/builddir/mozjs17.0.0/js/src
defined(@array) is deprecated at /tmp/glibc/builddir/mozjs17.0.0/js/src/config/milestone.pl line 58.
	(Maybe you should just omit the defined()?)
17.0.4esrpre
```

I think the `configure` script of mozjs17 expected to see only the line `17.0.4esrpre`. Has this to be solved manually or is it a matter of redirecting perl's deprecated warnings or disabling them by default?

xtraeme:
Fixed by @chneukirchen

PR #1895, opened 2015-06-23
mozjs17: unbreak perl5.22 issue (temporary hack)

pullmoll:
This hack enables one to build mozjs17 by avoiding to run the broken `js/src/config/milestone.pl` script. If anyone can instead fix the perl script, that'd be certainly a preferred solution. I don't speak perl.
See #1894


srcpkgs/mozjs17/patches/perl5_22.patch
```diff
@@ -0,0 +1,16 @@
...
+
+--- js/src/configure	2015-06-23 18:17:31.130237760 +0200
++++ js/src/configure	2015-06-23 18:50:13.735454723 +0200
+@@ -1523,8 +1523,8 @@
+ 
+ fi
+ 
+-MOZILLA_VERSION=`$PERL $srcdir/config/milestone.pl -topsrcdir $srcdir`
+-MOZILLA_UAVERSION=`$PERL $srcdir/config/milestone.pl -topsrcdir $srcdir -uaversion`
++MOZILLA_VERSION=17.0.4esrpre
```
xtraeme: Hmm

chneukirchen:
Does this happen for mozjs24 as well?

PR #1896, opened 2015-06-24
DO NOT MERGE dracut: added crypt-ssh module

thypon:
Here only for testing purposed. Based on a pull request on dracut project.

I didn't manage to made it work. Maybe it's better to fix this one before it goes mainline.

https://github.com/haraldh/dracut
https://github.com/haraldh/dracut/pull/43


xtraeme:
I do not want this as a subpkg of dracut.

Make it an independent pkg.

thypon:
If #1899 is merged i close and remove this PR.

PR #1899, opened 2015-06-25
New package: mkinitcpio-18

thypon:
Before someone asks why:

- it creates smaller images then dracut (5x by now)
- it handles more or less the same hardware as dracut given it's completely written in bash script
- it supports early crypt dropbear unlocking

I love it.


thypon:
lvm2 module right now needs lvmetad enabled plus the encrypt hook needs "cryptdevice=/dev/yourdev:yourvolumegroup"

srcpkgs/mkinitcpio/template
```diff
@@ -0,0 +1,35 @@
+# Template file for 'mkinitcpio'
+pkgname=mkinitcpio
+version=18
+revision=1
+noarch=yes
+build_style=gnu-makefile
+depends="busybox-static libarchive base-system bsdcpio"
```
xtraeme: Depending on libarchive is probably not the thing you thought it does.

srcpkgs/mkinitcpio/template
```diff
@@ -0,0 +1,35 @@
...
+maintainer="Andrea Brancaleoni <miwaxe@gmail.com>"
+license="GPL-2"
+homepage="https://wiki.archlinux.org/index.php/Mkinitcpio"
+distfiles="https://sources.archlinux.org/other/$pkgname/$pkgname-$version.tar.gz"
+checksum=187bdeeade08996010fbff480ccc91e47722d275c22fd6feb4a4b63061e9fc22
+
+conf_files="/etc/mkinitcpio.conf"
+
+provides="dracut-9999_1"
+replaces="dracut>0"
```
xtraeme: hmm nope, this definitely shouldn't replace dracut.

srcpkgs/mkinitcpio/template
```diff
@@ -0,0 +1,35 @@
...
+
+patch_args="-Np1"
+
+_confs="udev encrypt lvm2"
```
xtraeme: Do not add unnecessary private vars to the global namespace unless it's required, move it to post_install().

srcpkgs/mkinitcpio/files/kernel-hook-postrm
```diff
@@ -0,0 +1,13 @@
...
+#
+# Kernel post-remove hook for mkinitcpio.
+#
+# Arguments passed to this script: $1 pkgname, $2 version.
+#
+PKGNAME="$1"
+VERSION="$2"
+
+if [ -f /boot/initramfs-${VERSION}.img ]; then
+	rm -f /boot/initramfs-${VERSION}.img
```
beefcurtains: I think it's better to move this into ``vkpurge``

xtraeme: vkpurge already removes the initramfs for obsolete kernels.

It's good as is, and matches dracut.

beefcurtains: It doesn't, but sure. Having a duplicate hook shouldn't be a problem.

xtraeme: It does because vkpurge runs the kernel hooks like any pkg that uses the kernel-hooks trigger.

beefcurtains: That's not the same as vkpurge itself doing it. Just saying.

xtraeme: That's what kernel hooks are for (/etc/kernel.d/).

srcpkgs/mkinitcpio/template
thypon: I think bsdcpio is enough, have to test.

srcpkgs/mkinitcpio/template
thypon: Do you have ideas. Maybe a virtual package.

srcpkgs/mkinitcpio/template
thypon: fixing

srcpkgs/mkinitcpio/template
beefcurtains: They should co-exist, just like the bootloaders. For example, have the kernel install/remove hooks read a config file that enables it if a certain variable is set. Are there any other conflicts?

thypon: The install/remove hook is the main conflict. The problem is dracut is mandatory in base system while a generic mkinitrd virtual-package should be mandatory and dracut/mkinitcpio/genkernel/... should provide this generic mkinitrd.

thypon:
I splitted the package with subpackages and i had also to include the full lvm.conf in mkinitcpio-lvm2 to include "use_lvmetad = 1" on early boot. If this thing is resolved in lvm2 mainline i remove differentiated lvm.conf file.

Also there is a problem crosscompiling busybox on arm musl.

srcpkgs/mkinitcpio/template
beefcurtains: That can be done later, it's not going to replace dracut overnight.

thypon: It has to remove dracut when it's installed. There is a conflict. And if it has to do overnight, then let it do.

xtraeme:
Just a FYI: even if this pkg is pushed to the repo it won't replace dracut.

Dracut is distro agnostic and saves me to waste LOT of time.

xtraeme:
It's ok to use a virtual pkg to choose between multiple implementations.

But I'd like to VERIFY all changes you are planning to introduce.

thypon:
Also mkinitcpio is distro agnostic. And it creates less bloated initrd by default.

I don't bother making it the default initrd implementation. Voidlinux is all about choice. I need crypto remote unlocking and mkinitcpio has it and it works https://aur.archlinux.org/packages/dropbear_initrd_encrypt/ .

I lost like 8 hours last month trying to crypto remote unlock on dracut without success. Then yesterday i decide to lost 30 more minutes to add this feature with mkinitcpio. If you don't want to use mkinitcpio don't use it, i posted this because i think it could help someone else. I could also put in a different repository if you think it arms voidlinux.

srcpkgs/mkinitcpio/template
beefcurtains: Then remove the hooks until the conflict can be resolved sanely. It's easier to test if both can be invoked manually on the same system. And definitely :-1: to rush this, this should be considered experimental as it hasn't been proven to work as well as dracut yet. Sweeping changes may well break things.

thypon: I'm alredy using when i'm talking to you. I'll repeat it can't coexist with dracut. If you don't want to use it don't do.

xtraeme:
- I'm not against it, I'm simply stating that dracut works well enough for me and for general usage in void.
- As long as it's optional and someone else maintains it I don't really care.
- I don't see any reason to make it conflict with dracut.

thypon:
It installs a kernel.d hooks where dracut does and generates images in the same place. I could put the hook later in the chain to make mkinitcpio override dracut files, but it's a waste to generate the same thing twice. If you are thinking a better solution let me know.

thypon:
I removed the conflict and i override dracut output. A waste, but if you think it's better somehow.

srcpkgs/mkinitcpio/template
beefcurtains: Sure it may work perfectly for you, but your use case may not be the same as my use case. In fact, it isn't.

beefcurtains:
Agreed, it's wasteful to do initrd generation twice. Adding a config file to both so you can enable/disable them individually should resolve this, you could even add a hook that disables dracut (make sure it outputs a warning just in case anything goes awry).

thypon:
Splitted busybox from busybox-ntpd; the former can be crosscompiled for musl.

beefcurtains:
Please rebase, thanks :smile_cat: :smiley_cat:

thypon:
I't alredy rebased. It's a multi package PR.

beefcurtains:
Then why are all these Perl updates in here?

thypon:
No idea. I don't bother.

2015-06-25 19:42 GMT+02:00 beefcurtains <notifications@github.com>:

thypon:
I think mkinitcpio it's ready. I'm alredy using to crypto unlock devices with LVM on LUKS.

srcpkgs/busybox-ntpd/template
```diff
@@ -0,0 +1,46 @@
...
+
+	if [ "$CROSS_BUILD" ]; then
+		sed -e \
+			"s,\(CONFIG_CROSS_COMPILER_PREFIX\).*,\1=\"${XBPS_CROSS_TRIPLET}-\",g" \
+			-i ${t}/.config
+	fi
+}
+do_build() {
+	local t="busybox-ntpd"
+	make -C $t SKIP_STRIP=y CFLAGS="$(_cflags $t)" ${makejobs}
```
beefcurtains: Could be moved to ``make_build_args`` I guess, maybe?

thypon: Maybe. I don't want to break things and i follow the template before split.

srcpkgs/mkinitcpio/template
```diff
@@ -0,0 +1,67 @@
+# Template file for 'mkinitcpio'
+pkgname=mkinitcpio
+version=18
+revision=1
+noarch=yes
+build_style=gnu-makefile
+depends="busybox-static base-system bsdcpio"
```
beefcurtains: Depends on ``base-system``? Isn't that like, every template ever?

srcpkgs/mkinitcpio/files/encrypt_hook
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/ash
```
beefcurtains: Should probably add a depend on ``busybox`` for these scripts (or change them to use /bin/sh?)

srcpkgs/mkinitcpio/template
thypon: fixed

srcpkgs/mkinitcpio/files/encrypt_hook
thypon: the hooks are runned during initialization. `busybox-static` is alredy included.

thypon:
Ok now it builds. `busybox-ntpd` can't be cross build with musl right now but it's not needed by mkinitcpio.

srcpkgs/mkinitcpio/files/encrypt_hook
beefcurtains: Then I guess this is :ok: by me now, pending second peer

srcpkgs/mkinitcpio/template
```diff
@@ -0,0 +1,67 @@
...
+		for RULES in ${FILESDIR}/*.rules; do
+			vinstall $RULES 644 usr/lib/initcpio/udev $(basename $RULES)
+		done
+		vinstall ${FILESDIR}/lvm.conf 644 usr/lib/initcpio/conf
+	}
+}
+
+mkinitcpio-encrypt_package() {
+	depends="${sourcepkg}>=${version}_${revision}
+		${sourcepkg}-lvm2>=${version}_${revision} cryptsetup"
```
pickfire: Actually, why would mkinitcpio-encrypt need to depend on lvm2? People can use LUKS without lvm and this setup requires others to pull in lvm2 as well which is extra.

PR #1901, opened 2015-06-25
New package: tin-2.2.1

allan:


srcpkgs/tin/template
```diff
@@ -0,0 +1,18 @@
+# Template file for 'tin'
+pkgname=tin
+version=2.2.1
+revision=1
+build_style=gnu-configure
+hostmakedepends="flex"
+makedepends="ncurses-devel readline-devel"
```
beefcurtains: ``ncurses-devel`` is a dependency of ``readline-devel``, please remove this redundant makedep

allan: I like it better the way it is.

beefcurtains:
For the record, when is it ok to have a redundant makedep?

PR #1904, opened 2015-06-25
New Package: cuetools-1.4.1

tonkazoid:


srcpkgs/cuetools/template
```diff
@@ -0,0 +1,21 @@
...
+build_style=gnu-configure
+
+distfiles="https://github.com/svend/${pkgname}/archive/${version}.tar.gz"
+checksum=24a2420f100c69a6539a9feeb4130d19532f9f8a0428a8b9b289c6da761eb107
+
+pre_configure() {
+	aclocal
+	autoheader
+	automake --add-missing
+	autoreconf --install
```
xtraeme: `autoreconf -fi` is not enough?

PR #1905, opened 2015-06-25
Linux: update to 4.1.

beefcurtains:
Because why not?

### DIFF
--- a/srcpkgs/linux/template
+++ b/srcpkgs/linux/template
@@ -1,6 +1,6 @@
 # Template file for 'linux'
 pkgname=linux
-version=4.0
+version=4.1
 revision=1
 build_style=meta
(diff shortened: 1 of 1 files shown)


chneukirchen:
Because we usually wait for 4.1.1.

xtraeme:
What @chneukirchen said. 4.1 is not yet considered stable.

GiedriusS:
4.1.1 is out. Could you please update the commit? Thanks

beefcurtains:
No need, this is a metapackage.

beefcurtains:
Perhaps we could do non-pae kernel first, and add a friendly message for Xen users that warns them i686 doesn't have Xen guest support anymore. Just in case anyone is running vanilla i686 void kernels on a Xen host. Will have to rebase that pull, np.

PR #1908, opened 2015-06-25
linux4.1: build non-pae kernel on i386

beefcurtains:
* Void Linux currently fails to boot on older computers, because they often lack physical address extensions. For example, processors based on the [Yonah core](https://en.m.wikipedia.org/wiki/Yonah_%28microprocessor%29) support SSE2 (and SSE3) but not PAE.
* You can test this on a modern computer with: ``qemu-system-x86_64 -enable-kvm -cpu host,-pae -kernel path/to/vmlinuz``
* Unfortunately, it appears Xen guest support does not want to be built in this configuration yet. Could reintroduce this as an alternative kernel package
* This reduces the maximum memory capacity of i386 kernels from 64GiB to 4GiB, I recommend running a 64 bit kernel on systems with more than 4GiB of RAM


thypon:
We really need to support ten years old i686?

beefcurtains:
Computers from before 2009 that still work just fine, but Intel was too cheapskate to include PAE everywhere. Granted, i686 is an old architecture. However, supporting i686 and subsequently failing to run on the majority of desktop/mobile i686 processors doesn't make sense.

chneukirchen:
I agree that when providing i686, they should generally run on i686...

xtraeme:
Makes sense

xtraeme:
:+1:

PR #1909, opened 2015-06-26
New package: obs-0.10.1

lemmi:
Open Broadcaster Software Alpha Release

### DIFF
--- a/srcpkgs/obs/patches/0001-musl.patch
+++ b/srcpkgs/obs/patches/0001-musl.patch
@@ -1,0 +1,30 @@
+From d0c8195c1a863e35f23a4c3f5d130078a22bd651 Mon Sep 17 00:00:00 2001
+From: lemmi <lemmi@nerd2nerd.org>
+Date: Fri, 26 Jun 2015 22:13:26 +0200
+Subject: [PATCH] musl
+
+---
+ libobs/util/threading-posix.c | 8 ++------
+ 1 file changed, 2 insertions(+), 6 deletions(-)
+
+diff --git libobs/util/threading-posix.c libobs/util/threading-posix.c
+index dcf6d28..72df7dc 100644
+--- libobs/util/threading-posix.c
++++ libobs/util/threading-posix.c
+@@ -263,11 +263,7 @@ bool os_atomic_compare_swap_long(volatile long *val, long old_val, long new_val)
+ 
+ void os_set_thread_name(const char *name)
+ {
+-#if defined(__APPLE__)
+-	pthread_setname_np(name);
+-#elif defined(__FreeBSD__)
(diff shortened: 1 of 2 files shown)


srcpkgs/obs/template
```diff
@@ -0,0 +1,16 @@
+# Template file for 'obs'
+pkgname=obs
+version=0.10.1
+revision=1
+wrksrc=obs-studio-${version}
+only_for_archs="i686 x86_64"
```
xtraeme: Please add `i686-musl x86_64-musl` too.

lemmi:
@xtraeme musl build fails because pthread_setname_np is apparently not available on musl. should i revert or leave the only_for_archs variable?

pullmoll:
You could perhaps shield this non-portable (_np) function call with
```
#if defined(__GLIBC__)
    pthread_setname_np(pthread_self(), name);
#endif
```
i.e. create a patch like `srcpkgs/obs/patches/musl-no_pthread_setname_np.patch` and add these two lines to `libobs/util/threading-posix.c`.

Of course only if this functionality is not essential for the package to work right.

lemmi:
this is more or less ripping out everything none glibc/musl related and replacing it with pullmoll's suggestion. this doesn't change the glibc version and lets it build with musl, too. i can't confirm this is a working fix for musl, though.

beefcurtains:
When writing a patch, try to create only minimal delta. For example:
```
@@ -267,7 +267,7 @@
 	pthread_setname_np(name);
 #elif defined(__FreeBSD__)
 	pthread_set_name_np(pthread_self(), name);
-#elif !defined(__MINGW32__)
+#elif !defined(__MINGW32__) && defined(__GLIBC__)
 	pthread_setname_np(pthread_self(), name);
 #endif
 }
```

srcpkgs/obs/patches/0001-musl.patch
```diff
@@ -0,0 +1,30 @@
...
+ void os_set_thread_name(const char *name)
+ {
+-#if defined(__APPLE__)
+-	pthread_setname_np(name);
+-#elif defined(__FreeBSD__)
+-	pthread_set_name_np(pthread_self(), name);
+-#elif !defined(__MINGW32__)
+-	pthread_setname_np(pthread_self(), name);
++#if defined(__GLIBC__)
++    pthread_setname_np(pthread_self(), name);
```
xtraeme: I guess this patch won't be accepted by upstream, you've removed the existing cases...

lemmi:
after talking with some other people about this i posted that upstream https://github.com/jp9000/obs-studio/pull/441
i could change the patch to the same thing, but it won't be of any consequences for us. either this gets accepted upstream and then i have to delete this patch again, or it won't then we don't care about the other stuff

PR #1912, opened 2015-06-26
qtcreator: update to 3.4.1

pullmoll:
+ Split into basic and full dependencies (qtcreator-full)
+ Patch to unbreak musl by removing references to execinfo.h + backtrace()

### DIFF
--- a/srcpkgs/qtcreator/template
+++ b/srcpkgs/qtcreator/template
@@ -1,29 +1,27 @@
 # Template file for 'qtcreator'
 pkgname=qtcreator
-version=3.4.0
-revision=4
+version=3.4.1
+revision=1
 wrksrc=qt-creator-opensource-src-${version}
+build_style=gnu-makefile
+build_pie=yes
+make_build_args="INSTALL_ROOT=\${DESTDIR}/usr"
+make_install_args="${make_build_args}"
 hostmakedepends="perl python pkg-config"
 makedepends="qt5-declarative-devel qt5-script-devel qt5-tools-devel"
-depends="qt5-connectivity-devel qt5-declarative-devel qt5-enginio-devel
- qt5-location-devel qt5-multimedia-devel qt5-quick1-devel qt5-quickcontrols
- qt5-script-devel qt5-sensors-devel qt5-serialport-devel qt5-svg-devel
- qt5-tools-devel qt5-wayland-devel qt5-webchannel-devel qt5-webengine-devel
- qt5-webkit-devel qt5-websockets-devel qt5-x11extras-devel qt5-xmlpatterns-devel
- qt5-declarative-devel qt5-script-devel qt5-tools-devel qt5-doc qt5-plugin-sqlite"
+depends="qt5-declarative-devel qt5-quickcontrols qt5-plugin-sqlite"
 nocross=yes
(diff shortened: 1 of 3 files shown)


srcpkgs/qtcreator/template
```diff
@@ -33,3 +28,17 @@ do_install() {
...
+
+qtcreator-full_package() {
+	depends="${sourcepkg}>=${version}_${revision}"
+	short_desc+=" - full dependencies"
+	depends+="qt5-connectivity-devel qt5-declarative-devel qt5-enginio-devel qt5-graphicaleffects
+ qt5-location-devel qt5-multimedia-devel qt5-quick1-devel qt5-script-devel qt5-sensors-devel
+ qt5-serialport-devel qt5-svg-devel qt5-tools-devel qt5-wayland-devel qt5-webchannel-devel
+ qt5-webengine-devel qt5-webkit-devel qt5-websockets-devel qt5-x11extras-devel qt5-xmlpatterns-devel
+ qt5-declarative-devel qt5-script-devel qt5-tools-devel qt5-doc qt5-plugin-gtk
+ qt5-plugin-mysql qt5-plugin-odbc qt5-plugin-pgsql qt5-plugin-sqlite"
```
beefcurtains: I think you missed ``qtcreator``

pullmoll: There was a blank missing in depends+= which should be fixed now.

beefcurtains: Yeah, that's better. But it's still going to install qt*everything* except qtcreator afaict.

pullmoll: depends="${sourcepkg}>=${version}_${revision}"
This should install qtcreator and the basic dependencies, no?

beefcurtains: Hot damn, didn't catch that. My bad :sob:

beefcurtains:
:+1:

pullmoll:
It looks like do_build() runs as if `$makejobs` was `-j1`. I couldn't find why that is and how to correct it. It takes longer than necessary...

beefcurtains:
That's because it's building in ``do_install``, specify ``build_style=gnu-makefile`` and it should be fine.

pullmoll:
Good catch! I'll try it here and rebase...

pullmoll:
I _think_ this should work now, adding the escaped `\${DESTDIR}` to the `make_build_args` and also `make_install_args`. The ${DESTDIR} is expanded only when used in the hooks. The `do_install()` is gone and additional files moved in `post_install()`.

I tested this with (ccache warmed up) x86_64 and x86_64-musl builds locally.

beefcurtains:
Awesome! Updated and improved a bunch! Didn't have to rebase again, but it's alright. Are all the changes you want in here? Can hardly wait to do the mergaroo.

pullmoll:
I rebased to cancel the pending travis jobs (I hope). I think all changes are there, but why not wait for travis to finish?

beefcurtains:
But of course, let's see if the CI agrees

pullmoll:
E2BIG for travis. Too bad there is no `-silent` option.

beefcurtains:
Good enough for me

PR #1916, opened 2015-06-27
New package: ircii-20141122

diogoleal:


diogoleal:
I added ```only_for_archs="i686 x86_64"```because I could not compile is package for ARM

xtraeme:
This is not the solution if compilation is broken

srcpkgs/ircii/template
```diff
@@ -0,0 +1,17 @@
+# Template file for 'ircii'
+pkgname=ircii
+version=20141122
+revision=1
+wrksrc=${pkgname}-${version}
+build_style=gnu-configure
+only_for_archs="i686 x86_64"
+configure_args="--prefix=/usr --mandir=/usr/share/man"
```
xtraeme: This configure_args line is redundant, those are set by default.

srcpkgs/ircii/template
```diff
@@ -0,0 +1,17 @@
+# Template file for 'ircii'
+pkgname=ircii
+version=20141122
+revision=1
+wrksrc=${pkgname}-${version}
```
xtraeme: This wrksrc line is redundant, wrksrc is set to ${pkgname}-${version} by default.

srcpkgs/ircii/template
```diff
@@ -0,0 +1,17 @@
...
+build_style=gnu-configure
+only_for_archs="i686 x86_64"
+configure_args="--prefix=/usr --mandir=/usr/share/man"
+hostmakedepends="ncurses-devel  libressl-openssl"
+makedepends="ncurses-devel libressl-devel"
+depends="lftp ncurses"
```
xtraeme: This depends line is incorrect, why does it depend on lftp? and depending on ncurses is also probably wrong.

srcpkgs/ircii/template
```diff
@@ -0,0 +1,17 @@
...
+short_desc="The iircII is an IRC and ICB client"
```
xtraeme: `the iircII is` is redundant.

srcpkgs/ircii/template
```diff
@@ -0,0 +1,17 @@
...
+maintainer="Diogo Leal <diogo@diogoleal.com>"
+license="BSD"
```
xtraeme: If the license is BSD you should install the license file with `vlicense`.

srcpkgs/ircii/template
```diff
@@ -0,0 +1,17 @@
...
+homepage="http://www.eterna.com.au/ircii/"
+distfiles="http://ircii.warped.com/ircii-20141122.tar.gz"
```
xtraeme: replace `20141122` with `${version}`

srcpkgs/ircii/template
beefcurtains: Make it ``virtual?smtp-server``.

beefcurtains: Actually, that would be a ``hostmakedepend``, unless you don't ship the ``ircbug`` script. Not sure if you want it as a plain ``depends``, it's only used by said script to send bug reports.

srcpkgs/ircii/template
beefcurtains: Starting ``short_desc`` with "A(n)" is useless, too. Everything is a(n) something.

srcpkgs/ircii/template
beefcurtains: It'll actually compile in support for checking email in ``/var/spool/mail`` when it's present at compile time. Having the ``hostmakedepend`` will enable this feature.

srcpkgs/ircii/template
```diff
@@ -0,0 +1,17 @@
+# Template file for 'ircii'
+pkgname=ircii
+version=20141122
+revision=1
+wrksrc=${pkgname}-${version}
+build_style=gnu-configure
+only_for_archs="i686 x86_64"
```
beefcurtains: It should be possible to cross compile this by [patching out most of the tests](http://sprunge.us/DVRC), running ``autoconf`` (in ``pre_configure``), and then adding the [missing defines](http://sprunge.us/aiZF) to ``defs.h`` (in ``post_configure``). Every test that tries to compile and run a program should be patched out.

xtraeme:
This package is not installing the files into DESTDIR correctly.

diogoleal:
I will fix it and solve the problem of cross-compiller

xtraeme:
Stalled for 3 months... feel free to resubmit it later.

PR #1920, opened 2015-06-28
CONTRIBUTING.md: adding information about changes commits 

diogoleal:


beefcurtains:
Please also include a few words about using this to avoid having to create new pull requests :smile_cat:

diogoleal:
@beefcurtains I add a quick text as you requested.

Ah! Examples of commands were not with the same formatting as README.md and Manual.md, changed the file to maintain a standard.

PR #1925, opened 2015-06-29
libtirpc: fix #1923

### LINKED ISSUE #1923: RPCBind can't be compiled anymore
```
/usr/lib/gcc/x86_64-unknown-linux-gnu/4.9.3/../../../../lib/libtirpc.so: undefined reference to `__rpc_get_default_domain'
Makefile:451: recipe for target 'rpcbind' failed
make: *** [rpcbind] Error 1
make: *** Waiting for unfinished jobs....
```

thypon:


srcpkgs/libtirpc/template
```diff
@@ -1,9 +1,9 @@
 # Template file for 'libtirpc'
 pkgname=libtirpc
 version=0.3.2
-revision=1
+revision=2
 build_style=gnu-configure
-hostmakedepends="pkg-config mit-krb5-devel"
+hostmakedepends="pkg-config mit-krb5-devel automake"
```
xtraeme: If don't run autoconf/automake the patch won't be applied to Makefile.am...

xtraeme: i.e `pre_configure() { autoreconf -fi }`

thypon: I don't think it's needed. Maybe configure script run this for me. During this phase complains if automake isn't installed, then when installed compiles cleanly (also rpcbind do the same).

beefcurtains: ``configure`` should work regardless. But you'll get an error when ``make`` is invoked. Regenerating the ``configure`` script indeed does not appear to be necessary.

PR #1926, opened 2015-06-29
New package: tidy5

beefcurtains:
Run all your HTML through tidy5, the output it produces is really beautiful :yum:


common/build-style/cmake.sh
```diff
@@ -32,13 +33,14 @@ _EOF
...
-	cmake ${configure_args} ..
+	cmake ${configure_args} $(echo ${cmake_builddir}|sed \
+		-e 's|[^/]$|/|' -e 's|[^/]*||g' -e 's|/|../|g')
 }
 
 do_build() {
 	: ${make_cmd:=make}
 
-	cd build
+	cd ${cmake_builddir}
```
xtraeme: `cd ${cmake_builddir:=build}` instead.

common/build-style/cmake.sh
```diff
@@ -2,8 +2,8 @@
 # This helper is for templates using cmake.
 #
 do_configure() {
-	[ ! -d build ] && mkdir build
-	cd build
+	[ ! -d ${cmake_builddir:=build} ] && mkdir -p ${cmake_builddir}
```
xtraeme: you are missing here some more assignments if cmake_builddir is unset, i.e `${cmake_builddir:=build}`

srcpkgs/tidy5/template
```diff
@@ -0,0 +1,54 @@
...
+
+libtidy5_package() {
+	short_desc+=" - library files"
+	pkg_install() {
+		vmove "usr/lib/*.so.*"
+	}
+}
+
+libtidy5-devel_package() {
+	depends="${sourcepkg}>=${version}_${revision}"
```
xtraeme: This pkg really needs to depend on `libtidy5` instead.

common/build-style/cmake.sh
beefcurtains: Where? ``[ ! -d ${cmake_builddir:=build} ]`` should set it, although assignment within an if expression is somewhat bad practice.

xtraeme: Err right, didn't see it.

common/build-style/cmake.sh
```diff
@@ -32,13 +32,14 @@ _EOF
 
 	configure_args+=" -DCMAKE_INSTALL_SBINDIR=bin"
 
-	cmake ${configure_args} ..
+	cmake ${configure_args} $(echo ${cmake_builddir}|sed \
+		-e 's|[^/]$|/|' -e 's|[^/]*||g' -e 's|/|../|g')
```
xtraeme: really? why so complicated?

xtraeme: I can't see why you need to declare a new variable specifically for cmake, just reuse build_wrksrc.

beefcurtains: This should cover all bases, and I couldn't figure out how to do this in a single regex.
It's actually rather simple:
* Add trailing "/" if it's missing
* Strip everything except any "/" characters
* Change "/" into "../"
It would be better if a path with embedded "../" would be handled as well, but at least this way it's derived from `cmake_builddir` and works with any level of subdirectories.

beefcurtains: Reusing build_wrksrc fails in the multiple distfiles case.

xtraeme:
Looks ok then, but would be nice if you can separate changes to xbps-src next time.

beefcurtains:
I did, it's a separate commit. Should it also be a separate pr?

xtraeme:
It's ok, did not see you had a separate commit.

:+1:

PR #1927, opened 2015-06-29
net-tools: unbreak musl

pullmoll:
+ The home page returns _403: Forbidden_ thus switch to sourceforge
+ The most recent update to the repo was on 2015-05-26


xtraeme:
:+1:

beefcurtains:
:ok:

PR #1939, opened 2015-07-01
Remove systemd support

beefcurtains:
- [X] Remove unit files
- [X] Build all modified packages (this is going to take a while)
- [X] ~~polkit is broken~~ moved to #1947 

(thanks whoever added the list items! :3)


xtraeme:
:+1: 

Thanks :-)

srcpkgs/enlightenment/template
```diff
@@ -22,13 +21,6 @@ distfiles="http://download.enlightenment.org/rel/apps/${pkgname}/${pkgname}-${ve
 checksum=9634464281a58823ead37dc34b6818802378b03b5a293f2f0529c1ec68762939
 
 post_install() {
-	if [ -z "$build_option_systemd" ]; then
```
xtraeme: This change is wrong.

xtraeme: You should have only dropped the `if` conditional and keep the `sed` cmds, they are for !systemd.

xtraeme:
Looks good except the enlightment changes.

Please fix this one and you can merge it.

pullmoll:
I think there are some more cases where the `else` branch of an `if [ "$build_option_systemd" ]` statement were dropped. These should most probably be kept.

srcpkgs/enlightenment/template
beefcurtains: Thank you! :heart:

beefcurtains:
@pullmoll else branches were integrated into the templates in each case. ``gdm`` moved to #1947

mid-kid:
May I get some enlightenment as to why this feature was removed? I don't use systemd myself, but I'm just curious.

xtraeme:
Because there are no plans to support systemd in a short or long term.

pullmoll:
To add to @mid-kid 's question: Supporting systemd requires system wide changes and lots of conditionals in many packages. This was, is and will be a maintenance nightmare, especially when more and more packages introduce dependencies on systemd and its baggage. Generally disabling systemd support is easy in most cases (until now).

There are various init systems causing less trouble than systemd, because they do nothing else but handle process id 1. Void Linux defaults to runit, which does its job very well IMHO.

PR #1943, opened 2015-07-01
udisks2: add service directory

bougyman:


beefcurtains:
Should caja depend on this?

xtraeme:
Why? as I said on IRC you don't really need this. dbus services like this one are activated on demand.

xtraeme:
This is not really necessary, dbus services are activated on demand.

PR #1944, opened 2015-07-01
bash: bashrc: bail out in non-interactive shell

lemmi:


srcpkgs/bash/files/bashrc
```diff
@@ -3,6 +3,11 @@
 # Do not edit this file.
 # Place your readable configs in /etc/bash/bashrc.d/*.sh 
 
+if [[ $- != *i* ]] ; then
+	# Shell is non-interactive.  Be done now!
+	return
+fi
+
```
beefcurtains: Not ``[ -z "$PS1" ] && return``?

lemmi: don't think it'll matter in the end. somehow i like it a little bit more to search in the options whether "interactive" is set than to look if the prompt is set in order to find out if this shell is "interactive".

lemmi:
changed the commit message, too, thanks.

PR #1949, opened 2015-07-02
New package: pcsc-ccid-1.4.19

pullmoll:
The generic PC/SC CCID compliant smart card readers driver from the pcsclite author Ludovic Rousseau. Should work with most (read: more than pcsc-acsccid) USB card readers.

### DIFF
--- a/srcpkgs/pcsc-ccid/template
+++ b/srcpkgs/pcsc-ccid/template
@@ -1,0 +1,44 @@
+# Template file for 'pcsc-ccid'
+pkgname=pcsc-ccid
+version=1.4.19
+revision=1
+wrksrc=ccid-${version}
+build_style=gnu-configure
+configure_args="--enable-udev"
+hostmakedepends="pkg-config perl"
+makedepends="pcsclite-devel libudev-devel libusb-compat-devel"
+short_desc="PC/SC driver to support CCID compliant readers"
+maintainer="Jürgen Buchmüller <pullmoll@t-online.de>"
+license="LGPL-2.1"
+homepage="http://pcsclite.alioth.debian.org/ccid.html"
+distfiles="${DEBIAN_SITE}/main/c/ccid/ccid_${version}.orig.tar.bz2"
+checksum=cccb2c2bb4e56689efe34559f713102d92f94735542e58d3e4334e1459e934e1
+
+post_install() {
+	vinstall src/92_pcscd_ccid.rules 644 usr/lib/udev/rules.d/
+	vdoc readers/supported_readers.txt
+	vdoc README
+}
+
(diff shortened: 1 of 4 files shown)


pullmoll:
I am undecided whether or not to install the `contrib/Kobil_mIDentity_switch` and `contrib/RSA_SecurID` binaries and the `examples` stuff as well.
Perhaps additional packages like `pcsc-ccid-contrib` and `pcsc-ccid-examples` would be a *Good Idea (tm)*.

beefcurtains:
:+1:

srcpkgs/pcsc-ccid/template
```diff
@@ -0,0 +1,22 @@
+# Template file for 'pcsc-ccid'
+pkgname=pcsc-ccid
+version=1.4.19
+revision=1
+wrksrc=ccid-${version}
+build_style=gnu-configure
+configure_args="--enable-udev"
+hostmakedepends="pkg-config perl"
+makedepends="pcsclite-devel libudev-devel libusb-compat-devel"
+depends="pcsclite"
```
beefcurtains: If it depends on a .so, this should also be automatic.

pullmoll: Hmm... I copied this from pcsc-acsccid. I can try, though.

pullmoll:
*grrr* the examples path is wrong.. wait

pullmoll:
Now the 3 packages should be ok AFAICT.

beefcurtains:
The GPLv2 is also an example?

pullmoll:
Well, whatshamado? The example is GPL-2 as opposed to the driver, which is LGPL-2.1. Should I ignore this fact and remove the file? Different license for a subpackage? Incorporate in the main package?

Nitpicking was never my main objective.

srcpkgs/pcsc-ccid/template
```diff
@@ -0,0 +1,44 @@
...
+	depends="${sourcpkg}>=${version}_${revision}"
+	pkg_install() {
+		vbin contrib/Kobil_mIDentity_switch/Kobil_mIDentity_switch
+		vman contrib/Kobil_mIDentity_switch/Kobil_mIDentity_switch.8
+		vbin contrib/RSA_SecurID/RSA_SecurID_getpasswd
+		vman contrib/RSA_SecurID/RSA_SecurID_getpasswd.1
+	}
+}
+pcsc-ccid-examples_package() {
+	short_desc+=" - examples"
```
xtraeme: noarch=yes

xtraeme: ignore that, looks like this contains bins.

PR #1950, opened 2015-07-02
libCello: unbreak musl

pullmoll:


beefcurtains:
Might as well drop the case...esac, but other than that :ok:

srcpkgs/libCello/template
```diff
@@ -9,6 +9,14 @@ license="BSD"
 distfiles="http://$pkgname.org/static/$pkgname-$version.tar.gz"
 checksum=2273fe8257109c2dd19054beecd83ddcc780ec565a1ad02721e24efa74082908
 
+post_configure() {
+	case "$XBPS_TARGET_MACHINE" in
+	*-musl)	# There is no include <execinfo.h> in musl libc
+		sed -e "s;defined(__unix__);(defined(__unix__) \&\& defined(__GLIBC__));" \
+			-i ${wrksrc}/src/Exception.c
```
beefcurtains: These hooks are already run from ``${wrksrc}``

PR #1951, opened 2015-07-02
New package: SDL2_gfx-1.0.1

pullmoll:


srcpkgs/SDL2_gfx/template
```diff
@@ -0,0 +1,39 @@
+# Template file for 'SDL2_gfx'
+pkgname=SDL2_gfx
+version=1.0.1
+revision=1
+build_style=gnu-configure
+hostmakedepends="pkg-config"
+makedepends="SDL2-devel"
+depends="SDL2"
```
beefcurtains: This should be automatic, is it not?

xtraeme: Yes, this line should be removed...

pullmoll: Removed the `depends="SDL2"`

PR #1953, opened 2015-07-03
Desystemd

Gottox:
Please recheck my commits, I did them half asleep :sleeping:


pullmoll:
I think in some packages there are systemd service files left in `${FILESDIR}` which could be removed.

Gottox:
Should be fixed now.

pullmoll:
Looks :ok: for me.

xtraeme:
`rm -rf systemd` has been 100% completed, thanks.

PR #1955, opened 2015-07-03
firefox: updated to 39.0

GiedriusS:
Only updated the checksum, not build tested because it has a lot of dependencies and builds for very long. Would change the -i18n package too but it requires a bigger change that I don't have time to do now: so.xpi has been removed and son.xpi now appeared. Need to check if it is the same language and update the symlinks.


GiedriusS:
It seems like some patch has to be fixed.

xtraeme:
If it's incomplete it does not help in any way...

PR #1956, opened 2015-07-03
xbps-src: implement a by_sha256 hash cache

### LINKED ISSUE #1952: xbps-src: create sources/by_sha256/<checksum>
Whenever I see the next package downloading the same `gcc-4.9.3.tar.gz`, e.g. for the cross compilers, or other big files just because the package has a different name, I wonder if it would be easy enough to:

1. for every successfully downloaded distfile (after chksum verification) create a hard link from the file file to `.../sources/by_sha256/<checksum>`.
2. whenever a distfile is to be downloaded, first check for its sha256 hash in `.../sources/by_sha256/<checksum>` and, if it exists, create a hard link from there to the current packages' distfile.

This shouldn't be too difficult to integrate into `common/hooks/do-fetch/00-distfiles.sh`.
It would certainly save some space, time and bandwidth.

I touched the `00-distfiles.sh` for the proposed XBPS_DISTFILES_MIRROR stuff, so I may have to rebase that one afterwards.

If no one beats me to it, I may try to create a pr for this idea sometime soon.

pullmoll:
Closes #1952

Any successfully verified distfile is hard linked into `$XBPS_SRCDISTDIR/by_sha256/$cksum`. Before trying to download a distfile, an exsiting `$XBPS_SRCDISTDIR/by_sha256/$cksum` is hard linked to the `$distfile`.

There is a new xbps-src command `update-hash-cache` which will hard link your existing distfiles to their corresponding cache entries. You may run this once to initially setup the cache, while it isn't mandatory. It will merge distfiles from different directories with identical sha256sum and should thus reduce the space required on disk.

### DIFF
--- a/common/hooks/do-fetch/00-distfiles.sh
+++ b/common/hooks/do-fetch/00-distfiles.sh
@@ -2,22 +2,29 @@
 # the $distfiles variable and then verifies its sha256 checksum comparing
 # its value with the one stored in the $checksum variable.
 
-verify_cksum() {
-	local curfile="$1" distfile="$2" dfcount="$3" filesum ckcount cksum found i
+# Get the checksum for $curfile at index $dfcount
+get_cksum() {
+	local curfile="$1" dfcount="$2" ckcount cksum i
 
 	ckcount=0
+	cksum=0
 	for i in ${checksum}; do
 		if [ $dfcount -eq $ckcount -a -n "$i" ]; then
 			cksum=$i
-			found=yes
-			break
 		fi
-		ckcount=$(($ckcount + 1))
+		ckcount=$((ckcount + 1))
 	done
-	if [ -z $found ]; then
+	if [ -z "$cksum" ]; then
 		msg_error "$pkgver: cannot find checksum for $curfile.\n"
 	fi
+	echo "$cksum"
+}
 
+# Verify the checksum for $curfile stored at $distfile and index $dfcount
+verify_cksum() {
+	local curfile="$1" distfile="$2" dfcount="$3" filesum cksum
+
+	cksum=$(get_cksum $curfile $dfcount)
 	msg_normal "$pkgver: verifying checksum for distfile '$curfile'... "
 	filesum=$(${XBPS_DIGEST_CMD} $distfile)
 	if [ "$cksum" != "$filesum" ]; then
@@ -25,7 +32,22 @@ verify_cksum() {
 		msg_red "SHA256 mismatch for '$curfile:'\n$filesum\n"
 		errors=$(($errors + 1))
 	else
+		if [ ! -f "$XBPS_SRCDISTDIR/by_sha256/$cksum" ]; then
+			mkdir -p "$XBPS_SRCDISTDIR/by_sha256"
+			ln -f "$distfile" "$XBPS_SRCDISTDIR/by_sha256/${cksum}_${curfile}"
+		fi
 		msg_normal_append "OK.\n"
+	fi
+}
+
+# Link an existing cksum $distfile for $curfile at index $dfcount
+link_cksum() {
+	local curfile="$1" distfile="$2" dfcount="$3" filesum cksum
+
+	cksum=$(get_cksum $curfile $dfcount)
+	if [ -n "$cksum" -a -f "$XBPS_SRCDISTDIR/by_sha256/${cksum}_${curfile}" ]; then
+		ln -f "$XBPS_SRCDISTDIR/by_sha256/${cksum}_${curfile}" "$distfile"
+		msg_normal "$pkgver: using known distfile $curfile.\n"
 	fi
 }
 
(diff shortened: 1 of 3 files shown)


pullmoll:
This reminds me: do we have a tool to clean outdated distfiles from sources? I mean something that detects sources which are no longer referenced from any template?

If not, it would be nice to have such a tool and now also to make it check for the inode numbers of distfiles to find unused by_sha256/cksum files, i.e. ones with an exclusive inode number, and offer to purge them as well.

xtraeme:
There's no tool to clean up /sources.

xtraeme:
This is perfect, well done!

xtraeme:
Hmm I noticed that does not work correctly with new distfiles, please fix!

pullmoll:
You mean new distfiles aren't cached? I haven't observed that, but will take a closer look.

xtraeme:
For some reason it detected the existing coreutils-8.23 distfiles valid for 8.24, while it wasn't even downloaded

pullmoll:
Hmm.. but the checksum for the distfile is definitely new? I would not know how this could happen.

I tested with a package I didn't have before: zmap. It downloaded the new file and created the hardlink for ff01f2b139e98a7e64ed2872ae332929c500dd6120888d57900fac82dccfeea8.

pullmoll:
One problematic case is if you change the version but not the checksum! It will then copy the existing (old) file and since it matches the (old) checksum take it for correctly *downloaded* for the new version and with the new distfile name.

xtraeme:
This is exactly the case I experienced.

When updating pkgs I always bump version and then exec `./xbps-src -NI fetch` to know the checksum.

pullmoll:
Yup. I see, but I don't see how to avoid it. Perhaps have a xbps-src switch to disable hard linking?

xtraeme:
This is imho unacceptable :-)

pullmoll:
Well, then we (or you) can't use this pr :sob:

xtraeme:
I think if you change the order of hard links it would work: download, verify, if ok, create hard link.

pullmoll:
Hmm.. I don't see how this could work. By changing the version number, you change the distfile name, nothing else. The checksum is used to see if the file exists, and because it exists (for the old checksum), it hard links the old checksum to the new file name. How would I detect I have to download to make the checksum verification fail?

xtraeme:
if distfile does not exist:
- download it
- verify checksum against template
- if checksum matches create the hardlink

if distfile does exist:
- check if by_sha256/checksum exists, use it
- verify checksum against template

xtraeme:
Not tested, but I think this could work

pullmoll:
This defeats the idea of this pull request :grin:
The idea was to _avoid_ downloading a file if it does not exist for the _current_ package.
If I did it your way, then every not yet existing gcc-4.9.3.tar.gz in the cross packages would be downloaded again...

xtraeme:
Hmm true, will think of a way for this.

pullmoll:
Think of a `force fetch and update checksums` switch for xbps-src. You would only use it if you knew what you do and instead of your `-NI`. It could even set the corresponding flags, i.e. ignore dependencies and disable use of remote repositories.

dominikh:
I must be missing something obvious here, but why do you only check if the checksum matches and don't check if the distfile name also does? If {dist1, sha1} and {dist2, sha1} weren't treated as the same, this problem wouldn't occur.

pullmoll:
Well, you are right. This should work. It would not allow for different names, but that shouldn't occur anyway or only if distifle renaming with `>` is used.

beefcurtains:
Just ignore the cached file if template is newer than the downloaded file, problem solved. No?

pullmoll:
The files usually keep the date from the download location, not the download date/time.
The `${cksum}_${curfile}` change @dominikh suggested is there and should work for >99% of the cases. Only exception is if two packages use the same distfile but different names with `>localname.ext`.

beefcurtains:
If that's the case, a simple ``touch`` after downloading should suffice.

dominikh:
Personally, I don't want a cached distfile to be invalidated just because I changed something unrelated about a template. My suggested solution works and has simpler rules to keep in mind.

beefcurtains:
In that case you'd still have the copy in ``hostdir/sources``, the cache should only be queried when the file is about to be downloaded.

pullmoll:
I think for a first evaluation of its usability this pr is ok as it is now. The file date of the distfile may even be useful to know in specific cases, so destroying it by touching the file is a loss of information. There's of course always room for improvement, though :smiley:

FWIW I have two separate build environments for `musl` and `glibc` on the same mount point. I can now even `ln /mnt/data/void-musl/sources/by_sha256/* /mnt/data/void-glibc/sources/by_sha256` - or the other way round - from time to time to share the distfiles files between both environments.

xtraeme:
Looks good this time, thanks :+1: 

I haven't found any issue.

PR #1972, opened 2015-07-06
xbps-src: Add purge-distfiles command

### LINKED ISSUE #1970: Purging obsolete distfiles
I wrote a bash script which takes advantage of the new `sources/by_sha256` directory to find obsolete distfiles and purge them. You can [find it here](http://distfiles.voidlinux.de/xpurge) for now.

It is to be run from the same path where `xbps-src` is and with at least a parameter for the sources path (e.g. `masterdir/sources`). It runs through the following steps
- scans all `srcpkgs/*/template` files to collect the SHA256 hashes of all packages
- scans the files in `.../sources` and collects their inode numbers
- scans `.../sources/by_sha256` for the hashes in the filenames there and
    - if it finds an undefined hash, removes all files sharing that file's inode number.

It seemed to work for me, but you better run it in dry run `-n` mode first on your sources directory.

I don't know whether I should convert it to an e.g. `common/xbps-src/shutils/purge_distfiles.sh` entry, because it has some options which would then either have to be dropped or added to `xbps-src`.

pullmoll:
Closes #1970

### DIFF
--- a/xbps-src
+++ b/xbps-src
@@ -58,6 +58,9 @@ remove <pkgname>
 remove-autodeps
     Removes all package dependencies that were installed automatically.
 
+purge-distfiles
+    Removes all obsolete distfiles in <hostdir>/sources.
+
 show <pkgname>
     Show information for the specified package.
 
@@ -647,6 +650,9 @@ case "$XBPS_TARGET" in
         ;;
     list)
         $XBPS_QUERY_CMD -l
+        ;;
+    purge-distfiles)
+        purge_distfiles
         ;;
     show)
         read_pkg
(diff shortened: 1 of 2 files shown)


xbps-src
```diff
@@ -58,6 +58,9 @@ remove <pkgname>
 remove-autodeps
     Removes all package dependencies that were installed automatically.
 
+purge-distfiles
+    Removes all obsolete distfiles in <masterdir>/sources.
```
xtraeme: masterdir -> hostdir

xtraeme:
Works great, thanks :+1:

PR #1973, opened 2015-07-06
qupzilla: fixes for Qt5-5.5.0

pullmoll:
The two fixes to build with Qt5-5.5.0 are already included in upstream and will become obsolete with the next version.


pullmoll:
Uh, travis build didn't do anything but download the build environment?

PR #1978, opened 2015-07-07
qca-qt5: Fixes for Qt5-5.5.0

pullmoll:
Since Qt5-5.5.0 it seems to be necessary to explicitly `#include <QIODevice>`
if this class is used. In previous versions this header may have been included as a
side-effect by other Qt headers.

Also silence the exessive warnings regarding the macro _BSD_SOURCE by
replacing it with _DEFAULT_SOURCE.


Duncaen:
:+1:

PR #1984, opened 2015-07-09
msp430-toolchain: update to 3.03.04.00

allan:


srcpkgs/msp430-toolchain/template
```diff
@@ -1,15 +1,15 @@
...
+version=3.03.04.00
 revision=1
 short_desc="Toolchain for the TI MSP430 microprocessor"
 maintainer="allan <mail@may.mooo.com>"
 homepage="http://www.ti.com/tool/msp430-gcc-opensource"
 license="GPL-2"
-distfiles="http://software-dl.ti.com/msp430/msp430_public_sw/mcu/msp430/MSPGCC/latest/exports/msp430-gcc-source.tar.bz2
-	   http://software-dl.ti.com/msp430/msp430_public_sw/mcu/msp430/MSPGCC/latest/exports/msp430-gcc-support-files.zip"
+distfiles="http://software-dl.ti.com/msp430/msp430_public_sw/mcu/msp430/MSPGCC/${version//./_}/exports/msp430-gcc-source.tar.bz2
+	   http://software-dl.ti.com/msp430/msp430_public_sw/mcu/msp430/MSPGCC/${version//./_}/exports/msp430-gcc-support-files.zip"
```
beefcurtains: Looks like a bashism

allan: it is

beefcurtains: Is it allowed?

allan: don't know for sure too, but i thought xtraeme gave up on posix shell.

xtraeme: xbps-src requires bash, there's no reason to not use bash features.

chneukirchen: I think we allow ${VAR//} and VAR+=..., I'd rather refrain from other bashisms.

xtraeme: Those bashisms that @chneukirchen pointed out are already used in lots of templates. But in other hand we do not use arrays and other stuff.

pullmoll: Except for the new file `common/xbps-src/shutils/purge_distfiles.sh` which would require quite a bit of refactoring without bash.

Gottox:
Travis timeout is hit, but the changes look reasonable.

Gottox:
:+1:

PR #1985, opened 2015-07-09
New package: xbmc-adult-1.0.6

beefcurtains:
This is a nice alternative to using Adobe Flash (which RMS doesn't approve) to watch porn. Full-screen, no ads, what's not to like? Enjoy! :point_right::ok_hand:

### DIFF
--- a/srcpkgs/xbmc-adult/template
+++ b/srcpkgs/xbmc-adult/template
@@ -1,0 +1,24 @@
+# Template file for 'xbmc-adult'
+pkgname=xbmc-adult
+version=1.0.6
+revision=1
+wrksrc=repository.xbmcadult
+noarch=yes
+hostmakedepends="unzip"
+depends="kodi"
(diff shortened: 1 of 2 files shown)


Gottox:
I'm not sure, that this should be part of void. Interested users can install this repository in the home directory.

dominikh:
I don't disagree with porn, but I'm not sure we should start packaging arbitrary Kodi repositories.

beefcurtains:
@Gottox: prude

@dominikh  This isn't an arbitrary repository, it's the one and only adult repo

dominikh:
I'll rephrase: I don't think we should package Kodi repositories.

Gottox:
@beefcurtains In some cases yea. ;) But I'm with @dominikh here.

xtraeme:
I agree with @Gottox 

There's no reason to pkg this if the user is able to download it and put it into its home directory.

PR #1987, opened 2015-07-10
fortune-mod: unbreak musl

pullmoll:
librecode expects to link against something which defines a function `void error(int status, int errnum, const char* format, ...);` which gnu libc defines. This is despite the fact that musl libc does not have error.h.

The easiest way around this is to add a function to the fortune-mod object files for musl builds which mimicks what glibc error(3) does.


beefcurtains:
:+1:

PR #1988, opened 2015-07-10
xscreensaver: distfile checksum changed

pullmoll:
The checksum doesn't match anymore. No version change, though.
Perhaps someone who has the old version can find out what changed?

### DIFF
--- a/srcpkgs/xscreensaver/template
+++ b/srcpkgs/xscreensaver/template
@@ -1,7 +1,7 @@
 # Template file for 'xscreensaver'
 pkgname=xscreensaver
 version=5.33
-revision=1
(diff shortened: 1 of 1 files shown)


Gottox:
```
/src/void-packages/srcpkgs/xscreensaver(master ✗) diff -upr xscreensaver-5.33 xscreensaver-5.33.new  
diff: xscreensaver-5.33/android/project/xscreensaver/jni/xscreensaver: recursive directory loop
Only in xscreensaver-5.33/android/project/xscreensaver: libs
Only in xscreensaver-5.33/android/project/xscreensaver: obj
```

PR #1989, opened 2015-07-10
New package: kadu-2.1

### LINKED ISSUE #1969: Package request: Kadu
http://www.kadu.im Kadu Instant Messenger

We don't have a pre-packaged Qt-based IM on Void Linux which would support OTR out of the box.
Kadu is very nice, I would like to recommend it as an alternative to pidgin.

Thanks.

pullmoll:
Closes #1969


pullmoll:
I don't know what I did wrong: it crashes here when main calls `QCA::init(QCA::MemoryMode, int)`. Perhaps somebody else can help? I'm not acquainted with qca-qt5

PR #1991, opened 2015-07-10
libxfce4ui: add gtk+3 and new shlibs

pullmoll:
The gtk+3 libs are required for parole


pullmoll:
Hmm.. perhaps they should go into a subpkg libxfce4ui-2, because it seems parole is looking for that name (pkgconfig?).

Issue #1990, opened 2015-07-10
Package request: hddtemp


7185:
hddtemp
======
hddtemp is a small utility that gives you the temperature of your hard drive by reading S.M.A.R.T. informations (for drives that support this feature).

Website: https://savannah.nongnu.org/projects/hddtemp/

asergi:
https://github.com/voidlinux/void-packages/commit/57c37a9b0287ef715a312f73baf4d8c37d7513f3

PR #1992, opened 2015-07-10
New package: hddtemp-0.3-beta15

tonkazoid:
for https://github.com/voidlinux/void-packages/issues/1990
i didn't add a service file, because you need to add a drive. sda, sdb etc.

edit: that's a patch from debian. http://ftp.debian.org/debian/pool/main/h/hddtemp/ hddtemp_0.3-beta15-52.diff.gz


tonkazoid:
I'll try to get the patch directly from debian without embedding.

thypon:
https://gist.github.com/thypon/5cc13aa94fd74c19809d  @tonkazoid

PR #1995, opened 2015-07-11
New package: iscan-2.30.1 and related

dominikh:
This commit adds support for Epson scanners. It packages the iscan
application, the epkowa SANE backend, iscan-data, as well as the
scanner-specific plugin for the GT S650 (and by extension V19 and V39).
More scanner plugins may be added later when people request them.

iscan is unfree because it ships with a binary blob. sane-epkowa is
free, however, same as iscan-data. Scanner plugins are unfree again.

All unfree components are covered by the AVASYS license. Not only does
it allow distribution, it's also almost a free license. However, it
forbids reverse engineering the binary blobs.

The binary blobs make this x86/amd64-only packages. The epkowa backend
itself could technically be compiled for other architectures, but it's
not evident that it's worth the time. Most modern scanners need a
proprietary plugin.


dominikh:
> srcpkgs/iscan-plugin-gt-s650/template:11: short_desc should be less than 72 chars

Sorry, no can do. Need to include the scanner names so searching for the plugin isn't a pain, and that gets us above 72 characters

> srcpkgs/iscan/template:29: variable set to empty string: 	repository=""

/cc @chneukirchen The main package is nonfree, the subpackage is not. Have fun teaching xlint about that :P

pullmoll:
Could `unset repository` work?

srcpkgs/iscan-data/INSTALL
```diff
@@ -0,0 +1,5 @@
+case "${ACTION}" in
+    post)
+        /usr/lib/iscan-data/make-policy-file -m udev --template /usr/lib/udev/rules.d/*-sane.rules -o /usr/lib/udev/rules.d/99-epkowa.rules
```
xtraeme: Cannot this file be generated via `post_install()` and stored in the pkg?

srcpkgs/iscan/sane-epkowa.REMOVE
```diff
@@ -0,0 +1,3 @@
+if [ "$UPDATE" = "no" -a "$ACTION" = "post" ]; then
+	sudo rm "/var/lib/iscan/interpreter"
```
xtraeme: `sudo` is wrong here

srcpkgs/iscan-data/INSTALL
dominikh: make-policy-file generates a udev rule based on the existing SANE rules. What if the user has a custom SANE rule that for example uses a different group? Then doing it in post_install wouldn't really work. Or do we not support this? (letting the user edit their udev rules)

srcpkgs/iscan/sane-epkowa.REMOVE
dominikh: Nice catch.

chneukirchen:
(It's called xlint, not xthelaw.  It's output is meant to be interpreted by a smart person. ;)
And unset actually sounds good here...

pullmoll:
Fool xlint like this perhaps?
```
short_desc="Epson scanner plugin for GT-S650,"
short_desc+=" Epson Perfection V19 and Epson Perfection V39"
```

dominikh:
`unset` sounds fine. Making the template uglier just to circumvent the short_desc warning doesn't sound acceptable. I think we'll be fine with the xlint warning for now, and maybe @Gottox can be convinced to continue the travis built even if xlint failed. After all, it's not xthelaw ;)

pullmoll:
As soon as travis doesn't error out on xlint issues, people will start to ignore them :)
Another *ugly* template suggestion:
```
_scanners="GT-S650, Epson Perfection V19 and Epson Perfection V39"
short_desc="Epson scanner plugin for ${_scanners}"
```
*lol*

dominikh:
To be fair, if we ever start packaging a large number of those plugins (unlikely to happen), we could have a build_style for them that sets the short_desc based on a scanners variable ;)

dominikh:
> As soon as travis doesn't error out on xlint issues

Run xlint last, not first. Problem solved. The error is fine, but I want to see if it cross-compiled properly, first.

pullmoll:
Sounds reasonable. You could just modify .travis.yml accordingly with this pr and give it a try.

xtraeme:
Looks ok if you have tested it, thanks :+1:

dominikh:
Merged.

PR #1996, opened 2015-07-11
gst-ffmpeg: fix cross compiling for armv7l

pullmoll:


beefcurtains:
If this is easier than changing autoconf.ac and running autogen.sh/autoreconf -fi, fine by me.

PR #1997, opened 2015-07-11
New package: ibus-anthy-1.5.6

7185:
and anthy-9100h

Maybe I should have done 2 PR for this?

### DIFF
--- a/srcpkgs/ibus-anthy/template
+++ b/srcpkgs/ibus-anthy/template
@@ -1,0 +1,27 @@
+# Template file for 'ibus-anthy'
+pkgname=ibus-anthy
+version=1.5.6
+revision=1
+build_style=gnu-configure
+configure_args="--prefix=/usr --libexec=/usr/lib/ibus $(vopt_enable gir introspection)"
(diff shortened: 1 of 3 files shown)


srcpkgs/ibus-anthy/template
```diff
@@ -0,0 +1,32 @@
...
+if [ -z "$CROSS_BUILD" ]; then
+	build_options_default="gir"
+fi
+
+if [ "$build_option_gir" ]; then
+	configure_args+=" --enable-introspection"
+	makedepends+=" gobject-introspection"
+else
+	configure_args+=" --disable-introspection"
+fi
```
Gottox: You can use `$(vopt_enable introspection)` and $(vopt_if gobject-introspection) shortcuts here. See https://github.com/voidlinux/void-packages/blob/master/Manual.md#build-options

srcpkgs/ibus-anthy/template
```diff
@@ -0,0 +1,32 @@
...
+
+pre_configure() {
+	NOCONFIGURE=1 ./autogen.sh --prefix=/usr --libexec=/usr/lib/ibus
+}
```
Gottox: afaics this will run configure twice. use autoreconf -fi instead

Gottox:
Otherwise :+1:

Gottox:
please rebase.

PR #1999, opened 2015-07-11
libcdr: -O0 prevents _FORTIFY_SOURCE

pullmoll:
The `Makefile`s contain a trailing `-O0` in the `CXXFLAGS` which prevents`_FORTIFY_SOURCE`.


xtraeme:
I rebooted the server at that time, sorry :-)

pullmoll:
No problem. If the fix is okay, you could merge it right away :)

I think the -O0 is a result of the the enabled debug option.

pullmoll:
Oops, I removed the branch by accident.

PR #2000, opened 2015-07-11
New package: pev-0.70

diogoleal:


beefcurtains:
:+1: looks good

srcpkgs/pev/template
```diff
@@ -0,0 +1,13 @@
+# Template file for 'pev'
+pkgname=pev
+version=0.70
+revision=1
+wrksrc=${pkgname}
+build_style=gnu-makefile
+makedepends="libressl-devel pcre-devel glibc-devel"
```
beefcurtains: ``glibc-devel`` is redundant, I think

diogoleal: ```glibc-devel``` removed

beefcurtains:
Not sure if ``/usr/lib/libpe.so`` should be in there, if it's possible to create a ``-devel`` with header files it should go in there. Other than that it's good to go.

diogoleal:
this is not necessary. the ```libpe``` is a library part. I'll create the package soon.
https://github.com/merces/libpe

beefcurtains:
Feel free to pull that whenever.

PR #2004, opened 2015-07-11
mu: enable emacs frontend

abuibrahim:


beefcurtains:
It would be better if the frontend is always built, if necessary you can create a subpackage and vmove it in there. Extraneous build options like this aren't very useful, you can't xbps-install it.

beefcurtains:
:+1:

PR #2005, opened 2015-07-12
New package: libpe-git-20150711

diogoleal:


beefcurtains:
No git packages anymore, use a tag if it exists.

Gottox:
Packages without releases won't be accepted anymore.

PR #2007, opened 2015-07-12
slop: unbreak musl

pullmoll:


Gottox:
We should consider reporting those fixes upstream.

pullmoll:
Agreed. This one is actually a non-issue fix, because no execinfo.h defined function seems to be used.

PR #2009, opened 2015-07-12
libtorrent: update to 0.13.5

### LINKED ISSUE #1924: RTorrent/libtorrent links are down
https://github.com/rakshasa/rtorrent/issues/295

Maybe we should move to github releases.

pullmoll:
The homepage does not exist anymore, thus migrate to github
Use the github download location and rename file to include pkgname

Partially closes #1924

### DIFF
--- a/srcpkgs/libtorrent/template
+++ b/srcpkgs/libtorrent/template
@@ -1,7 +1,7 @@
 # Template build file for 'libtorrent'
 pkgname=libtorrent
-version=0.13.4
-revision=4
+version=0.13.5
+revision=1
 build_style=gnu-configure
 configure_args="--enable-static --disable-debug --without-kqueue
  --enable-aligned --with-posix-fallocate"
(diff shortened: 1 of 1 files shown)


pullmoll:
I can't get cross building for arm* working. It always tries to link against `/lib/libstdc++.so` instead of `${XBPS_CROSS_BASE}/lib/libstdc++.so`

Perhaps someone knows how to adapt the patch `srcpkgs/libtorrent/patches/fix-sysroot-handling.patch`?

xtraeme:
This issue should be fixed by forcing a `libtoolize -f` or `autoreconf -fi`.

pullmoll:
`autoreconf -fi` is already there in `pre_configure()` and it runs libtoolize.
Running libtoolize once more with `-f` or `-fi` doesn't change a thing it seems.

xtraeme:
In this case there must be a `-L/usr/lib` somewhere that must be removed.

pullmoll:
Thank you! Several wrong `-L/lib` caused the problem. Replacing them with `-L${XBPS_CROSS_BASE}/lib` solves the issue.

PR #2021, opened 2015-07-14
run xlint parallel

### LINKED ISSUE #2017: travis: run xlint last
1. xlint is great, but it's not perfect. It has false positives. It's a tool intended for human consumption, not for machines to make decisions
2. Even when there's xlint errors, people care more about actual build errors. I don't want to work around xlint just to know if the package cross-compiles correctly.

Thus, xlint should run last, not first. It should still fail the build, so that people don't ignore it, but only after packages were actually built.

/cc @Gottox

Gottox:
This PR adds an extra build job which only runs xlint on changed templates and disables xlint for all others.

Fixes #2017

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ env:
         - PATH=$PATH:$HOME/bin
 
     matrix:
+        - XLINT=1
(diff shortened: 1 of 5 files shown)


xtraeme:
:+1:

PR #2027, opened 2015-07-14
rtorrent: add color and vimkeys option

tonkazoid:


pullmoll:
:ok: for me, since the default is to not patch the source.

tonkazoid:
idk how many people find this useful. But i always add the color patch
and i don't have arrow-keys, that's why the vim-patch ;)
btw, is there a way to show install-messages only for options? Just for a
notice, how to add colors to .rtorrentrc.

pullmoll:
I think when building you would have to patch (or create) INSTALL.msg accordingly, depending on the selected options. But take this with a grain of salt, as there is no example for a pkg doing that.

Gottox:
Strong 👎.

We're not the right place to add features to applications. Please report such patches to the rtorrent developers.

tonkazoid:
I would compare it to mutts trash and sidebar patches https://github.com/voidlinux/void-packages/tree/master/srcpkgs/mutt/files . Around forever and almost everyone is using the trash-patch.
The rtorrent color patch is at least 5 years old

tonkazoid:
@Gottox i totaly get your point. I'll close this. If someone want the patches he can ask me. maybe i'll make a package for https://github.com/pyroscope/rtorrent-ps but it has too much changes for my liking :p

PR #2030, opened 2015-07-14
openjdk: fix building in current environment

pullmoll:
- The hotspot OS version tests supported at most Linux-3.x; add 4 and 4.1
- The -D_FORTIFY_SOURCE=2 conflicts with CXX_O_FLAG_NONE="-O0", thus remove it


chneukirchen:
@pullmoll do you want to adopt openjdk? :P

pullmoll:
@chneukirchen rather no. I don't use Java myself and don't think I have enough insight into how bootstrapping works, or to handle a version update.

The patch was just meant to make it build again on current Void + kernel 4.x.

chneukirchen:
Well, neither do I. :P

Let's hope this merges.

PR #2036, opened 2015-07-15
Poppler 0.34

Gottox:


Gottox:
cups-filter is broken with poppler-0.34. There's a patch that fixes it, but even the maintainer thinks it's crap:

https://sources.gentoo.org/cgi-bin/viewvc.cgi/gentoo-x86/net-print/cups-filters/files/cups-filters-1.0.71-poppler0340.patch?view=markup

Gottox:
cups-filter patch is applied

PR #2037, opened 2015-07-16
New package: hunspell-hr_HR-20030928

gour:
It's not very new, but there is no better option and it works - tested within Weechat.

Sincerely,
Gour


pullmoll:
As you [can see](https://travis-ci.org/voidlinux/void-packages/jobs/71223197) xlint is not happy with your template. You can `xbps-install -S xtools` and run `xlint srcpkgs/hunspell-hr/template` locally to see what to fix.

After that you can `git commit --amend` and `git push -f`.

Gottox:
:+1:

PR #2039, opened 2015-07-17
fix couchdb service

nopjmp:
couchdb -b forks the process which makes runit think the process stopped.


nopjmp:
Fixing the commit to follow the styling of other commits.

PR #2040, opened 2015-07-17
couchdb: fix build and runit file

nopjmp:
Upstream couchdb has a patch to fix building on Erlang 18. Additionally, the service file has couchdb fork to a background process which breaks how runit expects services to run.


Gottox:
Please increase the revison of the couchdb template

nopjmp:
I'll remember that in the future! Thanks for the reminder.

Gottox:
You're welcome :)

Looks okay to me. Still fails with musl.

PR #2043, opened 2015-07-17
audacity: update to 2.1.1

tonkazoid:
... and change distfile/homepage.
http://www.fosshub.com/Audacity.html (the minsrc download) doesn't like direct links it seems.


tonkazoid:
@chneukirchen

PR #2046, opened 2015-07-18
New package: dockstar-base-1

thypon:


srcpkgs/dockstar-base/INSTALL
```diff
@@ -0,0 +1,8 @@
+case "$ACTION" in
+post)
```
beefcurtains: Please also check if ``$UPDATE`` is ``no`` (or not ``yes``)

thypon: It's done following the odroid base. All the things i run inside are idempotent commands, then checking for UPDATE it's not necessary for now.

beefcurtains: It's not that they might fail, it's that these services are enabled whenever the package is updated. But, perhaps this should be fixed in a separate bug/pr. You have my :ok:

srcpkgs/dockstar-base/template
```diff
@@ -0,0 +1,13 @@
+# Template file for 'dockstar-base'
+pkgname=dockstar-base
+version=1
+revision=1
+build_style=meta
+homepage="http://www.voidlinux.eu"
+short_desc="Void Linux Dockstar platform package"
+maintainer="Andrea Brancaleoni <miwaxe@gmail.com>"
+license="Public Domain"
+noarch=yes
```
xtraeme: Please drop `noarch`. This package is only meant to be available for the target platform, and not to be shared by all architectures in the same repository.

PR #2050, opened 2015-07-19
New package: partclone-0.2.79

beefcurtains:
It's generally a good idea to make a backup of your old OS before installing a new OS. This can help tremendously with that.


pullmoll:
:+1:

PR #2051, opened 2015-07-19
fix compile w/ automake-1.15, inline res_randomid on musl

jprjr:
The libasr from http://www.opensmtpd.org/archives/libasr-1.0.2.tar.gz has already had ./bootstrap ran, which seems to use autoconf/automake 1.14.1. This breaks the current build script, it tries to find aclocal-1.14, automake-1.14, etc.

Instead, this pulls the release from Github and runs ./bootstrap and now compiles correctly.

Interesting sidenote: the [libasr-1.0.2 announcement](https://www.opensmtpd.org/announces/libasr-1.0.2.txt) mentions that Github is the primary location for libasr information.

The other fix is for usage on musl - OpenSMTPD fails to get through ./configure on musl, because libasr uses the `res_randomid()` function from glibc/uclibc. I made a patch that implements the workaround from this issue: https://github.com/OpenSMTPD/libasr/issues/8

Note: this does not fix OpenSMTPD compilation entirely.

### DIFF
--- a/srcpkgs/libasr/patches/musl-inline-res_randomid.patch
+++ b/srcpkgs/libasr/patches/musl-inline-res_randomid.patch
@@ -1,0 +1,24 @@
+diff --git configure.ac configure.ac
+index 10aff04..07e149f 100644
+--- configure.ac
++++ configure.ac
+@@ -456,6 +456,7 @@ AC_CHECK_FUNCS([ \
+ 	issetugid \
+ 	memmove \
+ 	reallocarray \
++	res_randomid \
+ 	snprintf \
+ 	strdup \
+ 	strerror \
+diff --git src/asr_private.h src/asr_private.h
+index 0960e37..158e7ff 100644
+--- src/asr_private.h
++++ src/asr_private.h
+@@ -365,3 +365,7 @@ extern FILE * asr_debug;
+ 		asr_statestr((a)->as_state),	\
+ 		asr_statestr((s)));		\
+ 	(a)->as_state = (s); } while (0)
++
++#ifndef HAVE_RES_RANDOMID
++#define res_randomid(x) (0xffff & getpid())
++#endif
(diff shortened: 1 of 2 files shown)


srcpkgs/libasr/template
```diff
@@ -1,15 +1,26 @@
 # Template file for 'libasr'
 pkgname=libasr
 version=1.0.2
-revision=1
+revision=2
 build_style=gnu-configure
+hostmakedepends="automake autoconf libtool"
```
beefcurtains: Just automake and libtool should do.

jprjr:
@beefcurtains fixed

srcpkgs/libasr/patches/musl-inline-res_randomid.patch
```diff
@@ -0,0 +1,24 @@
...
+index 0960e37..254e83e 100644
+--- src/asr_private.h
++++ src/asr_private.h
+@@ -365,3 +365,7 @@ extern FILE * asr_debug;
+ 		asr_statestr((a)->as_state),	\
+ 		asr_statestr((s)));		\
+ 	(a)->as_state = (s); } while (0)
++
++#ifndef HAVE_RES_RANDOMID
++#define res_randomid(x) (0xfff & getpid())
```
beefcurtains: ``0xfff`` is a range of 0...4095 instead of 0...65535 as in the glibc implementation. Of course using ``getpid`` as a random number source is heavily frowned upon, but at least try to match glibc here.

jprjr:
@beefcurtains Good call. I was copying the patch from [here](https://github.com/OpenSMTPD/libasr/issues/8) where they use `0xfff`- probably a typo on their part?

beefcurtains:
:+1:

pullmoll:
:+1: but upstream should take a look at e.g. [this workaround](http://www.openbsd.org/advisories/sni_12_resolverid.txt) or something similar...

PR #2056, opened 2015-07-20
transmission: unbreak musl

pullmoll:
- Add missing includes which happen as side-effects with glibc
- Move usr/lib/locale to usr/share/locale for musl libc (better fix?)


beefcurtains:
:+1:

PR #2060, opened 2015-07-21
libnfnetlink: unbreak musl

pullmoll:


dominikh:
I'm not happy with shortening "library" to "lib".

I'd probably prefer `find` + `sed` on all files, instead of `grep` and the somewhat nasty loop.

pullmoll:
Is it ok now?

dominikh:
:+1:

PR #2064, opened 2015-07-21
xbps-src: support for unstable repo

thypon:
The unstable repo should contain -git packages and things the mantainers considers not ready/unstable. It's the 1337 way to avoid AUR.


xtraeme:
This is what the `-r` flag in xbps-src is for. xbps-src also appends the default branch as primary repository if the git branch != master.

There's no reason for such a special case.

Issue #2070, opened 2015-07-22
Tracking bug: change scm fetch methods to proper distfiles


beefcurtains:
These packages invoke ``git clone``:

- [ ] a10disp: @xtraeme
- [ ] android-tools: @uggedal
- [ ] arm-mem-git: @xtraeme
- [ ] bbswitch: @xtraeme
- [ ] btrfs-progs: @xtraeme
- [ ] checkbashisms: @asergi
- [ ] clementine: @typhon
- [ ] compton-git: orphaned
- [ ] connman-ui-git: @xtraeme
- [ ] cppcheck: @xtraeme
- [ ] cubieboard2-uboot: @xtraeme
- [ ] docker: @xtraeme
- [ ] dssi-vst: @xtraeme
- [ ] dtc: @typhon
- [ ] dzen2: @xtraeme
- [ ] efivar: @xtraeme
- [ ] f2fs-tools: @xtraeme
- [ ] fakechroot: @xtraeme
- [ ] filteraudio-git: @teh-jazzman
- [ ] fs-utils: @xtraeme
- [ ] fuse-usmb: @xtraeme
- [ ] gitflow: @xtraeme
- [ ] cargo: @chneukirchen
- [ ] gcc-go: @xtraeme
- [ ] go: @dominikh
- [x] hugo: @typhon
- [ ] libgo: @xtraeme
- [x] ~~libreoffice-bin-i18n-dgo: @xtraeme~~ (false positive)
- [x] ~~pango: @xtraeme~~ (false positive)
- [x] ~~perl-Pango: orphaned~~ (false positive)
- [x] ~~python-Django: @uggedal~~ (false positive)
- [ ] xlogo: @chneukirchen
- [ ] google-fonts-ttf-git: @asergi
- [ ] gtk-theme-config: @xtraeme
- [ ] ioquake3-rpi: @xtraeme
- [ ] ipe: @chneukirchen
- [ ] libapp: @xtraeme
- [ ] libclc-git: @RealJohnGalt
- [ ] libdri2-git: @xtraeme
- [ ] libosmocore: @Gottox
- [ ] libump-git: @xtraeme
- [ ] libvdpau-va-gl: @xtraeme
- [ ] linux-firmware: @xtraeme
- [ ] linux-user-chroot: @xtraeme
- [ ] ls++-git: @Ypnose
- [ ] ltrace: @xtraeme
- [ ] luakit: @xtraeme
- [ ] monsterwm-git: @Ypnose
[...]

PR #2073, opened 2015-07-22
filteraudio-git: fetching code via proper distfile

teh-jazzman:
In answer to #2070


srcpkgs/filteraudio-git/template
```diff
@@ -1,17 +1,16 @@
 # Template file for 'filteraudio-git'
 pkgname="filteraudio-git"
-version="20150313"
-revision=1
+version="20150722"
+revision=2
```
beefcurtains: Good work, but bumping both version _and_ revision seems redundant to me.

teh-jazzman: That just happend. I first only updated the version and tested it but later made some other changes and increased the revision.

beefcurtains: Please? :smile_cat:

thypon:
Alredy fixed.

PR #2075, opened 2015-07-22
kodi: update to 15.0

beefcurtains:
* TexturePacker was moved and no longer has a Makefile.in
* configure can't find libcec and errors out, even when it's installed just fine
* I haven't tested this. It compiles, therefore it works


Gottox:
@beefcurtains kodi-15.0 needs libcec-3.0

PR #2079, opened 2015-07-24
New package: bogofilter-1.2.4

gour:
I created bogofilter package which uses sqlite3 as database back-end.

Usually, other distros have bogofilter pacakge built with Berkeley DB as default and then create other (sub)packages with other back-ends.

I have old sqlite3 wordlist.db and have need for it; the package works for me, built on x86_64. 

Sincerely,
Gour

### DIFF
--- a/srcpkgs/bogofilter/template
+++ b/srcpkgs/bogofilter/template
@@ -1,0 +1,21 @@
+# Template file for 'bogofilter'
+pkgname=bogofilter
+version=1.2.4
+revision=1
+build_style=gnu-configure
+configure_args="--sysconfdir=/etc/${pkgname} --with-database=sqlite"
+depends="perl"
+makedepends="gsl-devel sqlite-devel"
+short_desc="A fast Bayesian spam filtering tool"
+maintainer="Gour <gour@atmarama.net>"
+license="GPL-3"
+homepage="http://bogofilter.sourceforge.net"
+distfiles="${SOURCEFORGE_SITE}/${pkgname}/${pkgname}-${version}.tar.gz"
+checksum=d2f2598d1300307209b7b1905cc0637d2d053c0a4701a8d44383867299578471
+
+conf_files="/etc/bogofilter/bogofilter.cf"
+
+post_install() {
+	vinstall bogofilter.cf.example 644  etc/bogofilter bogofilter.cf
+	vcopy contrib usr/share/contrib
+}


srcpkgs/bogofilter/template
```diff
@@ -0,0 +1,23 @@
...
+distfiles="${SOURCEFORGE_SITE}/${pkgname}/${pkgname}-${version}.tar.gz"
+checksum=d2f2598d1300307209b7b1905cc0637d2d053c0a4701a8d44383867299578471
+
+conf_files="/etc/bogofilter/bogofilter.cf"
+
+do_install() {
+						 make DESTDIR=${DESTDIR} install
+						 mv "${DESTDIR}/etc/bogofilter/bogofilter.cf.example" "${DESTDIR}/etc/bogofilter/bogofilter.cf"
+						 install -dm755 "${DESTDIR}/usr/share/${pkgname}/contrib"
+						 install -m644 contrib/* "${DESTDIR}/usr/share/${pkgname}/contrib/"
```
beefcurtains: Please use ``vinstall``, and one tab is enough. Thanks

srcpkgs/bogofilter/template
```diff
@@ -0,0 +1,23 @@
...
+short_desc="A fast Bayesian spam filtering tool"
+maintainer="Gour <gour@atmarama.net>"
+license="GPL-3"
+homepage="http://bogofilter.sourceforge.net"
+distfiles="${SOURCEFORGE_SITE}/${pkgname}/${pkgname}-${version}.tar.gz"
+checksum=d2f2598d1300307209b7b1905cc0637d2d053c0a4701a8d44383867299578471
+
+conf_files="/etc/bogofilter/bogofilter.cf"
+
+do_install() {
```
beefcurtains: Should be ``post_install``, you can drop the ``make install`` below.

srcpkgs/bogofilter/template
```diff
@@ -0,0 +1,23 @@
+# Template file for 'bogofilter'
+pkgname=bogofilter
+version=1.2.4
+revision=1
+build_style=gnu-configure
+configure_args="--sysconfdir=/etc/${pkgname} --with-database=sqlite"
+depends="db perl sqlite"
```
beefcurtains: ``sqlite`` should be automatic, it's already in ``makedepends`` and the binary is not needed

srcpkgs/bogofilter/template
gour: Can vinstall copy several files, iow. something like:

``` 
contrib/*
```

As far as tabs are concerned, something is broken with my Emacs. :-(

I use .dir-locals.el in the root of the repo, but somehow it does not work. :-/

Considering there are several file in contrib/* - 20, do you consider it's better to have 20 invocations of vinstall?

srcpkgs/bogofilter/template
xtraeme: Does it really depend on berkeley db utilities?

srcpkgs/bogofilter/template
xtraeme: You can use `vcopy` instead.

srcpkgs/bogofilter/template
gour: > Does it really depend on berkeley db utilities? 

Noppe, that's mistake and it's dep for 'normal' bogofilter package, this one uses sqlite.

beefcurtains:
:+1:

PR #2083, opened 2015-07-25
facter: update to 2.4.4;puppet: update to 3.8.1

mattiasgiese:


beefcurtains:
The build seems broken.
```
Could not load hiera; cannot install
```

mattiasgiese:
That's because the version of hiera has not changed but was not rebuilt for ruby 2.2.0.
I have testet the build locally (explcitly rebuilt hiera) and it worked for me. I have informed @xtraeme in the irc channel.

xtraeme:
Those three packages have been updated to the latest official versions.

Thanks.

PR #2086, opened 2015-07-25
New package: light-locker-1.6.0

LockeAnarchist:


srcpkgs/light-locker/template
```diff
@@ -0,0 +1,15 @@
...
+pkgname="light-locker"
+version="1.6.0"
+revision=1
+build_style=gnu-configure
+short_desc="A simple session-locker for lightdm"
+maintainer="Emanuel Serpa <emanuel@openmailbox.org>"
+hostmakedepends="pkg-config"
+makedepends="intltool libglib-devel libX11-devel gtk+3-devel dbus-glib-devel"
+depends="lightdm"
+license="GPLv3"
```
beefcurtains: ``GPL-3`` (it has to be a file in ``/usr/share/licenses``)

srcpkgs/light-locker/template
```diff
@@ -0,0 +1,15 @@
+# Template file for 'light-locker'
+
+pkgname="light-locker"
+version="1.6.0"
+revision=1
+build_style=gnu-configure
+short_desc="A simple session-locker for lightdm"
+maintainer="Emanuel Serpa <emanuel@openmailbox.org>"
+hostmakedepends="pkg-config"
+makedepends="intltool libglib-devel libX11-devel gtk+3-devel dbus-glib-devel"
```
beefcurtains: ``libglib-devel`` is redundant, dependency of ``dbus-glib-devel``

srcpkgs/light-locker/template
```diff
@@ -0,0 +1,15 @@
+# Template file for 'light-locker'
+
+pkgname="light-locker"
+version="1.6.0"
+revision=1
+build_style=gnu-configure
+short_desc="A simple session-locker for lightdm"
+maintainer="Emanuel Serpa <emanuel@openmailbox.org>"
+hostmakedepends="pkg-config"
+makedepends="intltool libX11-devel gtk+3-devel dbus-glib-devel"
```
xtraeme: - `intltool` must be in `hostmakedepends` instead.

xtraeme:
The build now fails to cross compile because it needs `glib-devel` in `hostmakedepends`.

xtraeme:
musl is broken, but it's not your fault.

Issue #2088, opened 2015-07-25
Package Request: Racket


anarat:
http://racket-lang.org/

PR #2092, opened 2015-07-26
New package: racket-6.2

thypon:
#2088


srcpkgs/racket/template
```diff
@@ -0,0 +1,33 @@
+# Template file for 'racket'
+pkgname=racket
+version=6.2
+revision=1
+build_style=gnu-configure
+build_wrksrc=src
+hostmakedepends="gsfonts"
+makedepends="sqlite-devel gtk+-devel"
+depends="sqlite gtk+"
```
beefcurtains: You sure these aren't automatic shlibs?

thypon: yes

thypon:
I think we can merge this one. Travis is too slow to compile this one before timeout.

PR #2093, opened 2015-07-26
New package: tarsnap-1.0.35

gour:
New package with 'proprietary' license and 'restricted' for distribution, but ready for local builds.

Sincerely,
Gour

### DIFF
--- a/srcpkgs/tarsnap/template
+++ b/srcpkgs/tarsnap/template
@@ -1,0 +1,25 @@
+# Template file for 'tarsnap'
+pkgname=tarsnap
+version=1.0.35
+revision=1
+wrksrc=${pkgname}-autoconf-${version}
+build_style=gnu-configure
+configure_args="--sysconfdir=/etc/${pkgname}"
+makedepends="e2fsprogs-devel libressl-devel"
+depends="bzip2 xz"
+short_desc="An online encrypted backup service (client)"
+maintainer="Gour <gour@atmarama.net>"
+license=proprietary
+homepage="http://www.tarsnap.com/"
+restricted=yes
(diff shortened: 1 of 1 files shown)


thypon:
Shouldn't be in the nonfree repository?

beefcurtains:
Tarsnap doesn't ship the files used to create the ``configure`` script (usually ``configure.ac``). Either you have to carefully audit all that junk, or ask the author to start shipping the files used to build that script. Personally, I consider the script effectively a binary and am therefore unable to fully reproduce tarsnap from the given source code.

So that is in addition to the problem of not being able to distribute binaries. :-1: from me.

gour:
No problem. xtraeme added 
```
XBPS_ALLOW_RESTRICTED=yes
```

option today and told me to submit it with

```
license=proprietary
```

for building it locally.

Since I have need for it, local build is enough for me. ;)

beefcurtains:
I don't like that feature if that means opening the floodgates to stuff we previously would've rejected. Sure, this package is just an oddball and building it locally is a useful workaround. But this shouldn't become an excuse to merge in shit programs. Call me a pessimist, but there was nothing stopping you from building it locally before (either wholly outside of xbps-src, or in a local branch). Also it could lead to support questions of the "can't install package x" variety.

By all means, go for it. But in my book this is declared a misfeature.

gour:
Fair-enough. ;)

xtraeme:
I added `restricted=yes` and `XBPS_ALLOW_RESTRICTED` specifically for cases like this one.

You will be able to build the binary packages if `XBPS_ALLOW_RESTRICTED` is defined in `etc/conf`, but the official build servers will never build a binary package, because redistribution of sources and binaries is restricted.

This is how FreeBSD/ports and pkgsrc works, you do have the build definitions to build and pkg it locally.

PR #2094, opened 2015-07-26
bogofilter: rm sample conf file

gour:
bumping revision due to rm-ing extra sample file.


thypon:
Why removing this file?

gour:
Because the copy is already provided as bogofilter.cf

PR #2099, opened 2015-07-26
New package: regina-rexx-3.9.1

diogoleal:


xtraeme:
- This needs a development subpkg
- The shared library soname must be added to `common/shlibs`

srcpkgs/regina-rexx/template
```diff
@@ -0,0 +1,21 @@
...
+license="GPL-2"
+homepage="http://regina-rexx.sourceforge.net/"
+distfiles="${SOURCEFORGE_SITE}/project/regina-rexx/regina-rexx/3.9.1/Regina-REXX-3.9.1.tar.gz"
+checksum=5d13df26987e27f25e7779a2efa87a5775213beeda449a9efac59b57a5d5f3ee
+
+regina-rexx-devel_package() {
+	short_desc+=" - development files"
+	depends="${sourcepkg}>=${version}_${revision}"
+	pkg_install() {
+		vmove usr/lib/*.so
```
xtraeme: Development pkg must contain headers, static libraries, pkg-config files, etc. i.e files that are only required when compiling against the software, not at runtime.

diogoleal: Sorry. 
I made the requested changed.

pullmoll:
It looks like `common/shlibs` has changed. Please rebase and again add
`libregina.so.3 regina-rexx-3.9.1_1`
to `common/shlibs`, preferably without the empty line :smile:

PR #2103, opened 2015-07-27
New package: mlt-0.9.6

diogoleal:


xtraeme:
- dependencies are wrong for cross compilation
- this needs a development subpkg

diogoleal:
@xtraeme Sorry, but I could not understand about "dependencies are wrong for cross compilation."
Its you give me an example?

thypon:
the -devel packages should stay in makedepends while the binaries executed during compilation phase should stay in hostmakedepends.

srcpkgs/mlt/template
```diff
@@ -0,0 +1,34 @@
+# Template file for 'mlt'
+pkgname=mlt
+version=0.9.6
+revision=1
+build_style=gnu-configure
+configure_args="--prefix=/usr ${XBPS_MAKEJOBS:+--parallel=$XBPS_MAKEJOBS}"
```
xtraeme: This `configure_args` line is completely unnecessary.

pullmoll:
Please rebase and fix the `common/shlibs` addition to not include unrelated lines.

pullmoll:
@diogoleal I took your package and added missing bits and pieces, such as all dependency packages we have in Void.

I'm not sure if the `ladspa-sdk-example-plugins` is all that useful, but without it `melt` complains it can't find any plugins.

PR #2104, opened 2015-07-27
android-tools: add distfiles+checksum

thypon:


xtraeme:
I don't get why this change is even necessary, this makes the template harder to read and there's no real benefit.

thypon:
I'm aware of it, but it helps having something cacheable. I guess if it's possible to use system libselinux/f2fs-tools to avoid multiple copies + to simplify the build script.

thypon:
Close for now

PR #2113, opened 2015-07-28
alsa-lib: Patch pcm.h for musl libc compatibility

### LINKED ISSUE #2109: alsa-lib: problem with musl
alsa-lib has a problem with musl when a package depends from alsa-lib.

for example, my PR(freshplayerplugin) depends from alsa-lib-devel's pcm.h (when building to musl) and that file has u_int.

I found a diff about this in github, hope this helps:
https://github.com/GregorR/musl-pkgsrc-patches/blob/master/audio-alsa-lib.diff

pullmoll:
closes  #2109

### DIFF
--- a/srcpkgs/alsa-lib/patches/stdint_compat.patch
+++ b/srcpkgs/alsa-lib/patches/stdint_compat.patch
@@ -1,0 +1,29 @@
+Use stdlib.h defined types uint8_t, uint16_t, uint32_t and uint64_t
+for compatibility with musl libc
+
+--- include/pcm.h	2015-02-26 13:34:52.000000000 +0100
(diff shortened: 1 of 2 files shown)


Gottox:
Are `u_int8_t` and `uint8_t` guaranteed to be of the same type?

Gottox:
They are, ignore my comment

pullmoll:
Yeah, `u_int*_t` is for the most part a *BSD thing, because they defined those types in some header (sys/types.h?) long before `<stdint.h>` was widely adopted.

Many packages nowadays still use these types, but they all can be `stdint'ized` AFAICT.

In this pr I just had to make sure that stdint.h is actually included, because not all packages including `pcm.h` had included it before.

PR #2128, opened 2015-07-30
New package: cantata-1.5.2

LockeAnarchist:


srcpkgs/cantata/template
```diff
@@ -0,0 +1,14 @@
+# Template file for 'cantata'
+pkgname=cantata
+version=1.5.2
+revision=1
+build_style=cmake
+hostmakedepends="cmake pkg-config"
+makedepends="qt-devel zlib-devel ffmpeg-devel taglib-devel taglib-extras-devel libmtp-devel libmpg123 mpg123-devel
+cdparanoia libcdparanoia libcddb-devel speex-devel eudev-libudev-devel media-player-info udisks2-devel perl-URI"
```
xtraeme: Adding packages without development files in `makedepends` is wrong. That means that `cdparanoia`, `libcdparanoia` `libmpg123` `media-player-info` and `perl-URI` are wrong.

PR #2134, opened 2015-07-30
remove retext4

Gottox:
Not needed anymore, ReText5 works fine :smile:


pullmoll:
:+1:

beefcurtains:
Please add ``replaces=`` to ReText if that makes sense.

PR #2135, opened 2015-07-30
README.md: Improve description of distcc usage

### LINKED ISSUE #1367: Improvement: README.md distcc description
With the given description alone, a build slave's distcc does not accept any connections, since it isn't configured to accept the host's IP address. Editing `/etc/distcc/clients.allow` on the slave and appending e.g. a network mask like `192.168.2.0/24` does not help, perhaps because this file isn't used at all by the service...

The script `/etc/sv/distccd/run` however has an explicit `--allow 127.0.0.1` parameter.

After manually appending another `--allow 192.168.2.0/24` it seems the host can now distribute compiler jobs to that slave. I'm not quite sure what should be improved: the documentation in README.md, the /etc/sv/distccd/run script or even something else.

To me it seems the /etc/sv/distccd/run script should check for and use `/etc/distcc/clients.allow`, just like the `${wrksrc}/packaging/RedHat/init.d/distcc` script does it. Or, alternatively, the clients.allow file should not be installed to avoid confusion.

A clarification if one has to increase XBPS_MAKEJOBS to account for the total number of cores, or if the xbps-src scripts handle that aspect would be helpful as well.

pullmoll:
Closes #1367
Native speakers please proof read :beers:

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,14 +285,29 @@ Setup the slaves (machines that will compile the code):
 
     # xbps-install -Sy distcc
 
+Modify the configuration to allow your local network machines to use distcc (e.g. `192.168.2.0/24`):
+
+    # echo "192.168.2.0/24" >> /etc/distcc/clients.allow
+
 Enable and start the `distccd` service:
 
     # ln -s /etc/sv/distccd /var/service
 
-In the host (machine that executes xbps-src) enable the following settings in the `void-packages/etc/conf` file:
+Install distcc on the host (machine that executes xbps-src) as well.
+Unless you want to use the host as slave from other machines, there is no need
+to modify the configuration.
+
+On the host you can now enable distcc in the `void-packages/etc/conf` file:
 
     XBPS_DISTCC=yes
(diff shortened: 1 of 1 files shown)


beefcurtains:
:beer: :beer: :+1:

pullmoll:
Hmm.. thinking about it and after re-reading I think the localhost part requires installation of distcc on the build machine as well.

README.md
```diff
@@ -285,14 +285,29 @@ Setup the slaves (machines that will compile the code):
...
+Modify the configuration to allow your local network machines to use distcc (e.g. `192.168.2.0/24`):
+
+    # echo "192.168.2.0/24" >> /etc/distcc/clients.allow
+
 Enable and start the `distccd` service:
 
     # ln -s /etc/sv/distccd /var/service
 
-In the host (machine that executes xbps-src) enable the following settings in the `void-packages/etc/conf` file:
+Install distcc on the host (machine that executs xbps-src) as well.
```
xtraeme: tyop: executs -> executes

xtraeme:
:+1:

PR #2137, opened 2015-07-30
boost: fix incorrect include path

pullmoll:
One example is the `openimageio` package which fails even after patching `#include <sys/poll.h>` to `#include <poll.h>` in the source, because it also includes boost's `socket_types.hpp`.

In glibc headers `/usr/include/poll.h` includes `/usr/include/sys/poll.h` so this patch shouldn't affect anyone, but silence or even unbreak some packages for musl libc.


beefcurtains:
:+1:

PR #2138, opened 2015-07-30
New package: quimup-1.4.0

beefcurtains:
:musical_note: A pretty good qt5 mpd client. It currently does have a space in place of ``-`` in the ``wrksrc``, which broke xbps-src. This is fixed in a separate commit. :musical_note:


pullmoll:
:ok: but we should wait for @xtraeme to confirm the xbps-src changes.

xtraeme:
The changes for xbps-src are ok, thanks :-)

PR #2139, opened 2015-07-30
openimageio: unbreak musl

pullmoll:


beefcurtains:
:+1:

srcpkgs/openimageio/template
```diff
@@ -20,6 +20,9 @@ distfiles="https://github.com/OpenImageIO/oiio/archive/Release-${version}.tar.gz
 checksum=497c60bba9525b913935f0af771126338f07c2f78bd84e96fd19df38b6574706
 nocross=yes # -> ilmbase
 
+pre_build() {
+	find ${wrksrc} -type f -exec sed -i "{}" -e "s;<sys/poll\.h>;<poll.h>;g" \;
```
Gottox: whats the ```{}``` good for?

pullmoll: The `{}` is replaced with the current filename with `-exec`.

Gottox: Ah, I'm blind.

PR #2141, opened 2015-07-30
New package: opentyrian-2.1.20130907

pullmoll:


xtraeme:
@pullmoll thanks for this, but I'd like to see the data files packaged in a `noarch` subpkg. This is how other games with data files are packaged currently.

xtraeme:
:+1:

srcpkgs/opentyrian/template
```diff
@@ -0,0 +1,45 @@
...
+	sed -i ${wrksrc}/Makefile -e "s;-g0;$CFLAGS;"
+}
+do_install() {
+	vbin ${pkgname}
+	vman linux/man/${pkgname}.6
+	vinstall linux/${pkgname}.desktop 644 usr/share/applications
+	vinstall ${wrksrc}/linux/icons/tyrian-128.png 644 usr/share/pixmaps ${pkgname}.png
+	cat <<EOT > ${wrksrc}/manual.txt
+******** This is the original manual of the DOS game. The install
+instructions are obsolete, while the story and gameplay instructions
```
beefcurtains: Do you think void users are stupid? :smile_cat:

pullmoll: Not more or less than Slackware users, where this idea to prefix the DOS manual.doc stems from. If you believe it may hurt the self-esteem of potential readers, you can follow up with a pr to remove it :smiling_imp:

PR #2143, opened 2015-07-31
New package: thunar-archive-plugin-0.3.1

LockeAnarchist:


srcpkgs/thunar-archive-plugin/template
```diff
@@ -0,0 +1,16 @@
+# Template file for 'thunar-archive-plugin'
+pkgname=thunar-archive-plugin
+version=0.3.1
+revision=1
+build_style=gnu-configure
+configure_args="--prefix=/usr --sysconfdir=/etc --libexecdir=/usr/lib/xfce4 --localstatedir=/var --disable-static"
```
xtraeme: This line is redundant and wrong, remove it.

srcpkgs/thunar-archive-plugin/template
```diff
@@ -0,0 +1,16 @@
...
+makedepends="Thunar-devel exo-devel"
+hostmakedepends="intltool pkg-config xfce4-dev-tools"
+depends="Thunar"
```
xtraeme: This is also unnecessary and redundant.

srcpkgs/thunar-archive-plugin/template
LockeAnarchist: at least libexecdir is necessary, I think

xtraeme: No, --libexecdir is completely wrong. We *do* libexec unlike Arch.

silvernode:
no wonder I couldn't extract or create archives in Thunar. Thanks for adding this!

PR #2150, opened 2015-08-02
New packages: faenza-icon-theme-1.3.1, faience-icon-theme-0.5.1

MamayAlexander:


skrzyp:
Good… I think that we need more packages for making Void prettier - themes, iconpacks, sounds, cursors, etc. Of ourse, I don't mean overloading our repo with tons of them, but this niche is not fullfilled yet.

Duncaen:
Why not just use themes, iconspacks etc. from your home directory?

LockeAnarchist:
Why not have in the repositories?

PR #2152, opened 2015-08-02
New package: cegui-0.8.4

pullmoll:
Prerequisite for *Secret Maryo Chronicles* package `smc-1.9` (working on it)


pullmoll:
For some reaons it's not building CEGUI-OPENGL which smc-1.9 depends on.
I'm looking into missing dependencies.

pullmoll:
It now builds on all targets, however
+ The libogre-devel is not available on all arm* targets
+ boost-python is not available when cross building, thus also disabled python for arm*
+ I couldn't get gles working with arm* (like SDL2 does), thus I didn't make it an option
+ No libCEGUIOpenGLRenderer-0.so.* is created for arm*, perhaps due to missing gles...

IMHO it could be merged now and perhaps be improved later on.

PR #2153, opened 2015-08-02
New package: octoxbps-0.1.0

beefcurtains:
Qt frontend for xbps

### DIFF
--- a/srcpkgs/octoxbps/template
+++ b/srcpkgs/octoxbps/template
@@ -1,0 +1,21 @@
+# Template file for 'octoxbps'
+pkgname=octoxbps
+version=0.1.0
+revision=1
+build_style=gnu-makefile
+hostmakedepends="qt5-qmake"
+makedepends="qt5-declarative-devel qt5-quick1-devel"
+depends="curl"
+short_desc="Qt-based XBPS front-end"
+maintainer="beefcurtains <beefcurtains@voidlinux.eu>"
+license="GPL-2"
+homepage="https://github.com/aarnt/octoxbps"
+distfiles="${homepage}/archive/v${version}.tar.gz"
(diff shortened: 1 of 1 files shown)


pullmoll:
The result is rather empty, or is it?
The musl build log tells there's a function without return value.

beefcurtains:
It's like Qt went insane overnight. Can't get qtpass and quimup to install, either.

Tried to build locally, but I ran out of device space.

pullmoll:
I think you just need to
```
do_install() {
        make INSTALL_ROOT="${DESTDIR}" install
}
```

beefcurtains:
No amount of extra variables will make it do something. The problem may be in the ``.pro`` file, but I already tried things from other projects and it didn't work right. The error here goes like ``Nothing to do for target ''install''``.

pullmoll:
You can do this in `do_configure()`
```
        sed -i ${pkgname}.pro -e "/DESTDIR += /s;bin;${DESTDIR}/usr/bin;"
```
or alternatively
```
do_install() {
        vbin bin/${pkgname}
}
```

Adding some `octoxbps.desktop` file to `/usr/share/applications` and also adding some icon to `/usr/share/pixmaps` would be helpful for the average GUI user :)
And the package should depend on either `gksu` or `kdesu` as mentioned in the `README`

beefcurtains:
It has occurred to me to just ``vbin`` the binary, but that doesn't cover anything that should go into /usr/share, like icons.

pullmoll:
Icons are compiled in as a resource through the generated `qrc_resources.cpp`.
I tested the app and it seems to work so far. I didn't test it with kdesu or gksu, though.

beefcurtains:
I'm waiting for Qt to compile locally, then I'll build the package myself and test it more thoroughly.

PR #2155, opened 2015-08-02
New package: entr-3.2

thypon:


pullmoll:
Some kind of `compat.h` shielding against `strlcpy()` redifinition?
I looks like configure thinks musl libc doesn't define `strlcpy()`, which it does in `string.h`.

thypon:
I tried patching with the debian patch but other errors comes out. @pullmoll help with musl is welcome.

pullmoll:
Adding these lines to `pre_build()` solves it for me:
```
                *-musl) # No need for _LINUX_PORT compat.h and strlcpy.c
                        sed -i Makefile -e "s;-D_LINUX_PORT;;"
        esac
```

thypon:
Errors persists with crosscompilation.

pullmoll:
What kind of errors?
Does the executable not behave as expected?

thypon:
forget what i said, seems to work.

PR #2173, opened 2015-08-04
bash: source system bashrc on login

lemmi:
This adds /etc/profile.d/bash.sh
```
# Bash login shells only run /etc/profile
# Bash non-login shells run only /etc/bash/bashrc
# We want to source /etc/bash/bashrc in any case
```

### DIFF
--- a/srcpkgs/bash/files/bash.sh
+++ b/srcpkgs/bash/files/bash.sh
@@ -1,0 +1,7 @@
+# Check for interactive bash
+[ -z "$BASH_VERSION" -o -z "$PS1" ] && return
+
+# Bash login shells only run /etc/profile
+# Bash non-login shells run only /etc/bash/bashrc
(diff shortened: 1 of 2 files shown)


uggedal:
-1

Duncaen:
Isn't this what /etc/skel/.bashrc does?

uggedal:
No, login shells don't source `~/.bashrc`

lemmi:
it isn't used and i'm preparing the change to base-files.
the alternative is to remove the systemwide /etc/bash/bashrc sourcing from bash, fix the skels to point to that.
this is close to what gentoo does.

Duncaen:
But ``/etc/skel/.bash_profile`` sources ``~/.bashrc``

lemmi:
exactly, and i'm doing now the same thing system-wide

PR #2183, opened 2015-08-05
New package: freetds-0.95.19

diogoleal:


common/shlibs
```diff
@@ -2107,3 +2107,5 @@ libfacter.so.3.0.2 facter-3.0.2_1
 libOIS-1.3.0.so ois-1.3_1
 libSILLY.so.1 silly-0.1.0_1
 libkeybinder-3.0.so.0 libkeybinder3-0.3.0_1
+libsybdb.so.5 freetds-devel-0.95.19_1
```
xtraeme: This is wrong, shlibs should never be assigned to `-devel` packages.

PR #2189, opened 2015-08-05
deadbeef: unbreak musl

MamayAlexander:


pullmoll:
Seems :ok: to me.
A common patch like `musl_limits_h.patch` collecting the `#include <limits.h>` additions shouldn't hurt glibc builds, though, and perhaps upstream can be convinced to accept it.

xtraeme:
I agree with @pullmoll 

I'm not a fan of adding those `sed` additions but a proper patch is the way to go to submit it upstream.

PR #2193, opened 2015-08-06
[DO NOT MERGE] firejail: update to 0.9.28.

Duncaen:


Duncaen:
@pullmoll could you help me with musl, does this happen somewhere else or do you have quick fix in mind?

pullmoll:
I'll take a closer look at it.

pullmoll:
Here's a very dirty hack to get around the problems. I haven't tested the resulting binary, as I don't have a musl system running, while I believe it should work ok.

`srcpkgs/firejail/files/musl_defs.h`:
```
#ifndef _MUSL_DEFS_H
#define _MUSL_DEFS_H

#ifdef __cplusplus
extern "C" {
#endif

#include <features.h>

#define __NEED_FILE
#define __NEED_dev_t
#define __NEED_ino_t
#define __NEED_mode_t
#define __NEED_nlink_t
#define __NEED_uid_t
#define __NEED_gid_t
#define __NEED_off_t
#define __NEED_time_t
#define __NEED_blksize_t
#define __NEED_blkcnt_t
#define __NEED_struct_timespec

#include <bits/alltypes.h>
#include <bits/stat.h>

#ifdef __cplusplus
#define NULL 0L
#else
#define NULL ((void*)0)
#endif

int printf(const char *format, ...);
int sprintf(char *__restrict, const char *__restrict, ...);
char *fgets(char *buffer, int size, FILE *fp);

#ifdef __cplusplus
}
#endif
#endif
```

And in `srcpkgs/firejail/template` add this:
```
pre_configure() {
        *-musl) # XXX: dirty hack - very limited stdio.h definitions
                # to avoid redefinition of intercepted libc functions
                cp ${FILESDIR}/musl_defs.h ${wrksrc}/src/libtrace
                # Include our header instead of stdio.h and sys/stat.h
                # Use struct stat instead of struct stat64 (they are equal)
                sed -i ${wrksrc}/src/libtrace/libtrace.c \
                        -e 's;#include <stdio.h>;#include "musl_defs.h";' \
                        -e 's;#include <sys/stat.h>;;' \
                        -e 's;struct stat64;struct stat;g'
        esac
}
```

Don't forget to change spaces to tabs :)

pullmoll:
@Duncaen Any news on this?

Duncaen:
@pullmoll FYI i reviewed this and found a clean solution now.
https://github.com/Duncaen/firejail/commit/dc503e3a53b8edf88ebe8da5425506b136fc2b3c

pullmoll:
This looks __way__ better than my hackery :)

PR #2201, opened 2015-08-07
youtube-dl: update to 2015.08.06.1.

MamayAlexander:


uggedal:
The pre-prepared upstream tarball is there for a reason. With it you don't need the pandoc dep since the man pages are prepared. Also your update seems to remove the zsh autocompletion, without any explanation. Closing this and bumping it myself.

srcpkgs/youtube-dl/template
```diff
@@ -1,20 +1,21 @@
...
-version=2015.07.28
+version=2015.08.06.1
 revision=1
 noarch=yes
-wrksrc="${pkgname}"
 depends="python"
+hostmakedepends="pandoc python zip"
+build_style=gnu-makefile
+make_install_args="MANDIR=/usr/share/man"
+nostrip=yes
```
uggedal: Not needed as this is a noarch package and therefore should have no binaries to strip.

PR #2208, opened 2015-08-08
New package: gogs

jprjr:
Gogs is a self-hosted git service written in Go.

http://gogs.io/

I'm not sure if my 'export CGO_ENABLED=1' is redundant or not.


jprjr:
Just fixed some problems with the template file

PR #2210, opened 2015-08-09
jupp: add conflicts=joe

diogoleal:


beefcurtains:
Why? ``joe`` doesn't even exist!

pullmoll:
Now it exists.

PR #2231, opened 2015-08-10
New package: boswars-2.7

diogoleal:


diogoleal:
How I resolve this problem? 
```/usr/bin/arm-linux-gnueabihf-strip:/destdir/arm-linux-gnueabihf/boswars-2.7/opt/boswars/boswars: File format not recognized```

Honestly do not know how I should act, sorry :(

**Comment by pullmoll** [comment]
The template compiles cross arm* packages with just `g++`. It should be e.g. `arm-linux-gnueabihf-g++`, which is in the variable `$CC` when `do_build()` is called.

I don't know if and how `make.py` detects environment or needs options to cross compile.
If you can't fix it, you should set `nocross=yes`.

File: srcpkgs/boswars/template [hunk 1]
```diff
@@ -0,0 +1,32 @@
...
+pkgname=boswars
+version=2.7
+revision=1
+wrksrc=${pkgname}-${version}-src/
+hostmakedepends="SDL-devel pkg-config zlib-devel libpng-devel lua51-devel"
+makedepends="lua51-devel libGL libpng-progs libtheora-devel libogg-devel libpng-devel libvorbis-devel MesaLib-devel libX11-devel SDL-devel"
+short_desc="A futuristic real-time strategy game (RTS)"
+maintainer="Diogo Leal <diogo@diogoleal.com>"
+license="GPL-2"
+homepage="boswars.org"
```
xtraeme: `homepage` expects a complete URL.

srcpkgs/boswars/template
```diff
@@ -0,0 +1,32 @@
...
+distfiles="http://www.boswars.org/dist/releases/boswars-${version}-src.tar.gz"
+checksum=dc3718f531e9ea413cf37e1333b62a4c5e69f1405502d9c59b9e424635135e3e
+do_build() {
+	python make.py
+}
+
+do_install() {
+	vmkdir opt/boswars
+	vcopy fbuild/release/boswars opt/boswars/
```
xtraeme: There's no need to install files to `/opt` because it's not a binary build.

And I believe we should not use `/opt` anymore.

PR #2232, opened 2015-08-10
unbound: disable build_pie

allan:
since 1.5.4 segfaults with build_pie because of
```
@@ -789,8 +789,8 @@ int ub_openssl_lock_init(void)
 {
 #if defined(HAVE_SSL) && defined(OPENSSL_THREADS) && !defined(THREADS_DISABLED)
        int i;
-       ub_openssl_locks = (lock_basic_t*)malloc(
-               sizeof(lock_basic_t)*CRYPTO_num_locks());
+       ub_openssl_locks = (lock_basic_t*)reallocarray(
+               NULL, (size_t)CRYPTO_num_locks(), sizeof(lock_basic_t));
```


chneukirchen:
NACK.

What's the relation between pie and that diff?

chneukirchen:
The actual crash happens at

```
Program received signal SIGSEGV, Segmentation fault.
__GI___pthread_mutex_init (mutex=0x55827060, mutexattr=<optimized out>)
    at pthread_mutex_init.c:91
91	pthread_mutex_init.c: No such file or directory.
(gdb) bt
#0  __GI___pthread_mutex_init (mutex=0x55827060, mutexattr=<optimized out>)
    at pthread_mutex_init.c:91
#1  0x0000555555580391 in ub_openssl_lock_init () at util/net_help.c:797
#2  0x00005555555b4e17 in daemon_init () at daemon/daemon.c:222
#3  0x000055555555fe30 in run_daemon (debug_mode=0, cmdline_verbose=0, 
    cfgfile=0x5555555e2b2d "/etc/unbound/unbound.conf") at daemon/unbound.c:639
#4  main (argc=0, argv=<optimized out>) at daemon/unbound.c:766
```

#1 is
```
        for(i=0; i<CRYPTO_num_locks(); i++) {
                lock_basic_init(&ub_openssl_locks[i]);
        }
```

allan:
the diff is where it segfaults and what has changed from 1.5.3 to 1.5.4.

chneukirchen:
Ok, there happens an adress truncation in reallocarray:

```
Breakpoint 4, __GI___libc_realloc (oldmem=0x0, bytes=1640) at malloc.c:2968
2968	malloc.c: No such file or directory.
(gdb) fin
Run till exit from #0  __GI___libc_realloc (oldmem=0x0, bytes=1640)
    at malloc.c:2968
0x000055555558035c in ub_openssl_lock_init () at util/net_help.c:792
792	util/net_help.c: No such file or directory.
Value returned is $22 = (void *) 0x555555827060
...
(gdb) p ub_openssl_locks 
$23 = (lock_basic_t *) 0x55827060
```

Note how the latter value is only the lower 32-bit of the realloced/malloced address.

chneukirchen:
Ok, this could happen when reallocarray is detected but called without prototype.

chneukirchen:
Indeed, building unbound:

```
checking for LibreSSL... no
...
checking for reallocarray... yes
...
util/config_file.c: In function 'config_set_option':  
util/config_file.c:479:3: warning: cast to pointer from integer of different size [-Wint-to-pointer-cast]
   (char**)reallocarray(NULL, (size_t)cfg->num_out_ifs+1, sizeof(char*));
   ^  
...
util/net_help.c: In function 'ub_openssl_lock_init':  
util/net_help.c:792:21: warning: cast to pointer from integer of different size [-Wint-to-pointer-cast]
  ub_openssl_locks = (lock_basic_t*)reallocarray(
```

etc.

For some reason HAVE_LIBRESSL is not set, which would define the proper prototypes then.

chneukirchen:
Ok, this check now fails:

```
if grep OPENSSL_VERSION_TEXT $ssldir/include/openssl/opensslv.h | grep "LibreSSL" >/dev/null; then
```

due to

```
#define LIBRESSL_VERSION_NUMBER 0x20020002L
#define LIBRESSL_VERSION_TEXT   "LibreSSL 2.2.2"

/* These will never change */
#define OPENSSL_VERSION_NUMBER  0x20000000L
#define OPENSSL_VERSION_TEXT    LIBRESSL_VERSION_TEXT
#define OPENSSL_VERSION_PTEXT   " part of " OPENSSL_VERSION_TEXT
```

allan:
nice, thanks for fixing it more properly!

PR #2245, opened 2015-08-13
sslh: add pcre support (fixes musl)

jprjr:
This adds a patch to sslh - it enables using PCRE as the regex library instead of using POSIX regex.

On musl, sslh doesn't compile w/ regex enabled because sslh makes use of a non-POSIX regex feature.

I have a PR open upstream for this fix - https://github.com/yrutschle/sslh/pull/62

### DIFF
--- a/srcpkgs/sslh/template
+++ b/srcpkgs/sslh/template
@@ -1,10 +1,10 @@
 # Template file for 'sslh'
 pkgname=sslh
 version=1.17
-revision=1
+revision=2
 build_pie=yes
 hostmakedepends="perl"
-makedepends="libconfig-devel libcap-devel"
+makedepends="pcre-devel libconfig-devel libcap-devel"
 short_desc="Applicative Protocol Multiplexer"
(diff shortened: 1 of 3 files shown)


srcpkgs/sslh/template
```diff
@@ -1,10 +1,11 @@
 # Template file for 'sslh'
 pkgname=sslh
 version=1.17
-revision=1
+revision=2
 build_pie=yes
 hostmakedepends="perl"
-makedepends="libconfig-devel libcap-devel"
+depends="libpcre"
```
xtraeme: This depends line is wrong... xbps-src will detect dependencies on ELF files when linking against shared libraries.

jprjr: Good to know! Got a corrected package coming in shortly.

jprjr:
Pushed a corrected template file.

There's also a small fix to the run script - it turns out sslh doesn't expect any spaces between an opt flag and its argument - I learned this in trying to setup a second sslh instance.

PR #2246, opened 2015-08-13
libldap: rename to openldap and make libldap a subpkg

jprjr:
Question - is there a good way to state "this packages requires a syslog daemon to be running" ?

I'll have to double-check (I might be wrong!), but in my testing it seemed like slapd wouldn't start without a syslog daemon running.


srcpkgs/openldap/template
```diff
@@ -0,0 +1,66 @@
...
+	short_desc+=" library - development files"
+	pkg_install() {
+		vmove usr/include
+		vmove "usr/lib/*.so"
+		vmove usr/share/man/man3
+	}
+}
+
+libldap_package() {
+	depends="libsasl libressl"
```
xtraeme: This depends line is redundant.

srcpkgs/openldap/template
```diff
@@ -0,0 +1,66 @@
...
+makedepends="libressl-devel libsasl-devel db-devel"
+depends="libldap"
+conf_files="/etc/openldap/ldap.conf" 
+short_desc="OpenLDAP (Lightweight Directory Access Protocol)"
+license="OpenLDAP License v2.8 - BSD alike"
+maintainer="Juan RP <xtraeme@voidlinux.eu>"
+homepage="http://www.openldap.org"
+distfiles="ftp://ftp.openldap.org/pub/OpenLDAP/openldap-release/openldap-$version.tgz"
+checksum=27856bb4a8b44feca2b326c309000e16a9dadd52362c8ab6eec6c67a43737f6e
+subpackages="libldap-devel libldap"
```
xtraeme: This is redundant too, the default subpkg order is fine.

srcpkgs/openldap/template
```diff
@@ -0,0 +1,66 @@
...
+ --localstatedir=/var/lib/openldap
+ --sbindir=/usr/bin
+ --enable-dynamic --enable-proctitle --with-tls
+ --enable-ipv6 --enable-local --with-cyrus-sasl --with-threads
+ --enable-syslog --enable-slapd --enable-crypt --enable-spasswd
+ --enable-bdb --enable-hdb --enable-overlays=mod
+ --disable-static --with-yielding_select=yes"
+hostmakedepends="pkg-config groff automake libtool"
+makedepends="libressl-devel libsasl-devel db-devel"
+depends="libldap"
```
xtraeme: This is redundant.

srcpkgs/openldap/files/slapd/conf
```diff
@@ -0,0 +1,2 @@
+LDAPUSER=ldap
```
xtraeme: This file shouldn't be supplied by the package, should only be created by the administrator.

jprjr:
Pushed a corrected commit

srcpkgs/openldap/template
```diff
@@ -0,0 +1,63 @@
...
+homepage="http://www.openldap.org"
+distfiles="ftp://ftp.openldap.org/pub/OpenLDAP/openldap-release/openldap-$version.tgz"
+checksum=27856bb4a8b44feca2b326c309000e16a9dadd52362c8ab6eec6c67a43737f6e
+
+conf_files="/etc/openldap/ldap.conf"
+system_accounts="ldap"
+ldap_homedir="/var/lib/openldap"
+make_dirs="
+	/var/lib/openldap 0755 ldap ldap
+	/run/openldap 0755 ldap ldap"
```
xtraeme: `/run/openldap` should not be created via `make_dirs`, because `/run` is tmpfs.

This directory should be created by the runit service.

srcpkgs/openldap/template
```diff
@@ -0,0 +1,63 @@
...
+makedepends="libressl-devel libsasl-devel db-devel"
+conf_files="/etc/openldap/ldap.conf" 
+short_desc="OpenLDAP (Lightweight Directory Access Protocol)"
+license="OpenLDAP License v2.8 - BSD alike"
+maintainer="Juan RP <xtraeme@voidlinux.eu>"
+homepage="http://www.openldap.org"
+distfiles="ftp://ftp.openldap.org/pub/OpenLDAP/openldap-release/openldap-$version.tgz"
+checksum=27856bb4a8b44feca2b326c309000e16a9dadd52362c8ab6eec6c67a43737f6e
+
+conf_files="/etc/openldap/ldap.conf"
```
xtraeme: Duplicated line, see above.

jprjr:
Pushed another corrected commit

PR #2249, opened 2015-08-14
cross-arm-linux-musleabihf: fix unresolved extern

pullmoll:
When linking against libgcc.a, there is an unresolved (in musl libc) external
__write() referenced from libgcc/config/arm/linux-atomic-64bit.c,
so make the patch replace it with libc write(2).

See: http://build.voidlinux.eu/builders/armv6l-musl_builder/builds/2991/steps/shell_3/logs/stdio

### DIFF
--- a/srcpkgs/cross-aarch64-linux-musl/template
+++ b/srcpkgs/cross-aarch64-linux-musl/template
@@ -11,7 +11,7 @@ _sysroot="/usr/${_triplet}"
 
 pkgname=cross-${_triplet}
 version=0.20
-revision=1
+revision=2
 short_desc="Cross toolchain for ARM64 LE target (musl)"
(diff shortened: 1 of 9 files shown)


xtraeme:
Is this also necessary for `cross-armv7l-linux-musl` ?

pullmoll:
As far as I've seen errors, no, for unknown reasons. IMO it can't hurt to also update the other pkgs depending on gcc/files/gcc-4.9.2-musl.diff, because write(2) should always be available.

xtraeme:
@pullmoll agreed, feel free to make the changes.

pullmoll:
@xtraeme there you go, every pkg using gcc-4.9.2-musl.diff revbumped.

PR #2250, opened 2015-08-14
Adding saltstack into the void:

niamtokik:
Hi everyone,

Here my first pull request for adding saltstack into void linux. I've tested it on my own laptop. If you have any remarks, please, say it! :)

  - get package from pypi
  - package salt (full)
    - build   => ok
    - package => ok
    - salt-master:
      - standalone     => work
      - starter script => work
      - configuration  => work
    - salt-minion:
      - standalone     => work
      - starter script => work
      - configuration  => work

  - subpackage salt-ssh:
    - build   => ok
    - package => ok

  - subpackage salt-master:
    - build   => ok
    - package => ok

  - subpackage salt-minion:
    - build   => ok
    - package => ok

  - notes:
    - currently all subpackage are built
      with full salt library.


srcpkgs/salt/template
```diff
@@ -0,0 +1,151 @@
+######################################################################
+# xbps-src template for salt, salt-master, salt-minion, salt-ssh     #
+######################################################################
+pkgname=salt
+short_desc="Portable, distributed, remote execution and configuration management system."
+version=2015.5.3
+revision=0
```
Ypnose: Should be <code>revision=1</code>

srcpkgs/salt/template
```diff
@@ -0,0 +1,151 @@
+######################################################################
+# xbps-src template for salt, salt-master, salt-minion, salt-ssh     #
+######################################################################
```
Ypnose: You should remove those comments blocks, IMHO

srcpkgs/salt/template
```diff
@@ -0,0 +1,151 @@
...
+pkgname=salt
+short_desc="Portable, distributed, remote execution and configuration management system."
+version=2015.5.3
+revision=0
+homepage="http://saltstack.com/"
+license="Apache-2.0"
+wrksrc="salt-${version}"
```
Ypnose: Not needed. <code>wrksrc</code> is <code>${pkgname}-${version}</code> by default.

srcpkgs/salt/template
```diff
@@ -0,0 +1,151 @@
...
+build_style=python-module
+maintainer="Mathieu Kerjouan <mkvoid@steepath.eu>"
+
+# define dependencies
+depends="python-Jinja2 python-msgpack>0.3 python-yaml python-MarkupSafe python-requests>=1.0.0"
+makedepends="python-Jinja2 python-msgpack python-yaml python-MarkupSafe python-requests"
+hostmakedepends="python-Jinja2 python-msgpack python-yaml python-MarkupSafe python-requests"
+subpackages="salt-master salt-minion salt-ssh"
```
Ypnose: Unknown property. Using symlinks is enough.

pullmoll:
You should use `short_desc+=" - xyz"` i.e. use a single dash. It's nothing written in stone anywhere, it's just that every other package adheres to this rule.

You can simplify the vmkdir commands, since it creates parents (uses install -d)

niamtokik:
> If you're done, please squash the pull request using `git rebase -i ...`

Do you have documentation about bests practices with `git` and void-linux package repository? Maybe some non-standard methods or some kind of habits? I'm not a daily user of git... So, small helps will be good. :)

Thanks in advance ;)

JayVii:
@pullmoll wrote a decent [HOW-TO on using git](https://github.com/voidlinux/documentation/wiki/How-to-use-git,-by-@pullmoll)

niamtokik:
Oh! Really nice! Thanks :D

srcpkgs/salt/template
```diff
@@ -0,0 +1,124 @@
...
+	post_install () {
+		vsv salt-minion
+	}
+}
+
+salt-ssh_package () {
+	short_desc+=" - ssh"
+	depends="python-Jinja2 python-msgpack>0.3 python-yaml python-MarkupSafe "
+	depends+="python-requests>=1.0.0 salt-devel"
+	makedepends+=" salt"
```
xtraeme: subpackages don't need build time dependencies, just run time dependencies. Therefore, `hostmakedepends` and `makedepends` is wrong.

srcpkgs/salt/template
```diff
@@ -0,0 +1,124 @@
...
+	hostmakedepends+=" salt"
+	pkg_install () {
+		vmkdir "usr/lib/python2.7/site-packages/"
+		vcopy  "build-2.7/lib/salt" "usr/lib/python2.7/site-packages/"
+	}
+}
+
+salt-master_package () {
+	short_desc+=" - master"
+	depends="python-Jinja2 python-msgpack>0.3 python-yaml python-MarkupSafe "
```
xtraeme: I don't get the point of splitting into subpackages if those pkgs depend on the same set of dependencies.

srcpkgs/salt/template
```diff
@@ -0,0 +1,124 @@
...
+}
+
+salt-devel_package () {
+	short_desc+=" - development files"
+	depends="python-Jinja2 python-msgpack>0.3 python-yaml python-MarkupSafe "
+	depends+="python-requests>=1.0.0"
+	makedepends+=" salt"
+	hostmakedepends+=" salt"
+	pkg_install () {
+		vmkdir "usr/lib/python2.7/site-packages/"
```
xtraeme: There's nothing of `development` in this package.

srcpkgs/salt/template
```diff
@@ -0,0 +1,124 @@
...
+makedepends="python-Jinja2 python-msgpack python-yaml python-MarkupSafe python-requests"
+hostmakedepends="python-Jinja2 python-msgpack python-yaml python-MarkupSafe python-requests"
+
+# package informations
+pycompile_module="salt"
+distfiles="${PYPI_SITE}/s/salt/salt-${version}.tar.gz"
+checksum=f8b04dc8bd4b00ce803d31ce5fba033c2f20fa38cdc5a3bf54b4c47362fbe853
+
+# configuration
+conf_files="/etc/salt/master
```
xtraeme: `/etc/salt/master` is a configuration file in at least two packages: `salt` and `salt-master`.

IMHO those subpackages are just redundant and unnecessary.

srcpkgs/salt/files/salt-master/run
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
+######################################################################
+# salt-master starter for runit.                                     #
+######################################################################
+
+salt_cache="/var/cache/salt"
```
xtraeme: These `run` scripts are unnecessarily complicated, this simpler script should accept custom settings via `service/conf`:

```
#!/bin/sh
[ -r conf ] && . ./conf
: ${SALT_LOGDIR:=/var/log/salt}
: ${SALT_CACHEDIR:=/var/cache/salt}
[ ! -d $SALT_LOGDIR ] && mkdir -p $SALT_LOGDIR
[ ! -d $SALT_CACHEDIR ] && mkdir -p $SALT_CACHEDIR
exec salt-master
```

srcpkgs/salt/template
niamtokik: salt is a management tool:
  * server side -> `salt-master`
  * client side -> `salt-minion`
  * remote control via ssh -> `salt-ssh`

I don't think its a good idea to install salt-master on each minion, but I don't have really worked on separate totally `salt-minion` and `salt-master` in python library path.

The main package (`salt`) install all executable, but, if someone want to only install `salt-master` without a minion or a minion without `salt-master` (its my case), he can install only `salt-master` or `salt-minion` package. I think its the good way to do, but, maybe I'm wrong.

srcpkgs/salt/template
niamtokik: hum, okay, I will remove that, during my test I was encoutered some error without these too variables set. I will recheck that.

srcpkgs/salt/template
niamtokik: I'm not agree with you. We can use salt only for some funtionality, without using executable (like any python modules/libraries).

srcpkgs/salt/files/salt-master/run
niamtokik: ok, I will merge your script, seems good :)

srcpkgs/salt/template
```diff
@@ -0,0 +1,122 @@
...
+		vman   "doc/man/salt-cp.1"
+		vman   "doc/man/salt-cloud.1"
+		vman   "doc/man/salt-syndic.1"
+		vman   "doc/man/salt-key.1"
+		vman   "doc/man/salt-api.1"
+		vman   "doc/man/salt.7"
+		vlicense LICENSE
+	}
+
+	post_install () {
```
xtraeme: subpackages don't accept any other function than `pkg_install()`, so that `post_install()` is wrong.

srcpkgs/salt/template
```diff
@@ -0,0 +1,122 @@
...
+		vsv salt-minion
+	}
+}
+
+salt-ssh_package () {
+	short_desc+=" - ssh"
+	depends="python-Jinja2 python-msgpack>0.3 python-yaml python-MarkupSafe "
+	depends+="python-requests>=1.0.0 salt-devel"
+
```
xtraeme: This subpackage is installing `/etc/salt/roster` but it's not declared via `conf_files`.

srcpkgs/salt/template
```diff
@@ -0,0 +1,122 @@
+pkgname=salt
+short_desc="Remote execution management system - full"
```
xtraeme: The main package is installing the same files than its subpackages, so that there will be conflicting files.

If you want to make `salt` a full featured pkg, you can make it depend on all its subpackages instead.

niamtokik:
ok, I can't do a "full package" with full option, and other minor package for selected function (master, minion, ssh)? 

If I want, I need to remove each subpackage, and build only one main package or, build one package with shared files and lot of others subpackage for each useful scripts?

yopito:
IMHO, should have the following packages: `salt-core`, `salt-master`, `salt-minion`, `salt-ssh`, etc.
with package `salt-core` containing all the python libs of SaltStack; this is the way it's usually packaging on other distro.
"full package" does not make really sense ? In real life, you will install `salt-minion` on every nodes, from time to time `salt-ssh`, and `salt-master` only once or two, no ?

@niamtokik: we can work together on this packaging if you want (even in French ;) I have some WIP I can share, including XBPS support for SaltStack.

Gottox:
What's the state of this PR? At least the commit history needs to be cleaned up using `git rebase`.

niamtokik:
hi everyone,

sorry for the late... I was working on another field, and, not  have really time for working/updating this current commit. FYI, this package work well on my servers (3*arm-void+2*amd64-void).

@yopito i've see your mail today, you have more information on "why" i've not updated yet this package and yes, we can work on same package! Could yoy give me a link or something to check your work - and perhaps merge our works?

Thanks

yopito:
@niamtokik : glad to here from you.
my own WIP packaging for SaltStack is on branch saltstack of repo https://github.com/yopito/void-packages

niamtokik:
I close this PR and give all right to my code to @yopito. I don't have currently the time to work on this PR.

@yopito you have my mail if you have some issue with this code. ;)

May the force be with you!

PR #2251, opened 2015-08-15
New package: cava-0.3.3

JayVii:
[Console-based Audio Visualizer for ALSA](https://github.com/karlstav/cava)
build_styles failed because, there's neither "configure", nor "make" needed. "make install" is all.


JayVii:
well, no.

PR #2255, opened 2015-08-16
clucene: unbreak musl + cross builds

pullmoll:


pullmoll:
It looks like sourceforge.net is becoming unreliable again.
Perhaps it's time to reconsider using a distfiles mirror as a primary download location...

PR #2261, opened 2015-08-16
New package: aisleriot-3.16.2

pullmoll:


pullmoll:
There's a curious failure on travis-ci which does not happen when compiling locally.
From the error messages it may be that some environment setting like `LANG`, `LANGUAGE` or `LC_ALL` is missing?

PR #2274, opened 2015-08-16
New package: cava-0.3.3

JayVii:
"do_configure()" not necessary/possible, so no build_style used.

arm-architectures might fail building.


JayVii:
I emailed the developer and he told me, that he tested the application on armv7l and it worked for him. If anyone knows how to fix this, I would greatly appreciate it.

pullmoll:
To me it seems as if the `Makefile` is trying to build `iniparser` using the `HOST_CC` while on the other hand it uses the target's `CFLAGS`. This follows from the message `gcc: error: unrecognized command line option '-mfpu=vfp'`.

If the Makefile should be cross-compilation aware, perhaps you just have to set some `HOST_CFLAGS` environment when invoking `make`?

Otherwise you would have to patch the Makefile to host compile `iniparser` and cross compile the other files.

JayVii:
In my email-conversation with the developer, he suggests to try an earlier version of cava:
>I'm not an expert on gcc and cross compiling, but '-mfpu=vfp', and '-mfpu=vfp' looks like normal gcc options for arm
>(https://gcc.gnu.org/onlinedocs/gcc/ARM-Options.html)

>It fails already on the iniparser <https://github.com/ndevilla/iniparser> (which is a bundled software to read config files), you could check if an earlier version of cava without inipraser also fails (https://github.com/karlstav/cava/archive/0.3.0.tar.gz). Maybe iniparsers makefile is incompatible with how travis-ci handles arm.

This is most likely a rather unsatisfying solution, so how would I patch the makefile to treat `iniparser` differently than the rest of the build?

srcpkgs/cava/template
```diff
@@ -0,0 +1,17 @@
...
+pkgname=cava
+version=0.3.3
+revision=1
+short_desc="Console-based Audio Visualizer for ALSA"
+maintainer="Jan Wey. <janwey.git@gmail.com>"
+homepage="https://github.com/karlstav/cava"
+license="MIT"
+distfiles="https://github.com/karlstav/${pkgname}/archive/${version}.tar.gz"
+checksum=188efe7f7bb4903cae3ef8e5fd6af6660c7115492f39315a97bdb9e3c95b4f77
+depends="alsa-lib"
```
xtraeme: This depends line is unnecessary and wrong.

srcpkgs/cava/template
```diff
@@ -0,0 +1,17 @@
...
+makedepends="alsa-lib-devel fftw-devel ncurses-devel"
+make_install_args="PREFIX=/usr"
```
xtraeme: If you are using a custom `do_install()` there's no need to set `make_install_args`, add the vars directly into the function, and note that probably you are installing the package files into masterdir/usr, not `DESTDIR/usr`.

xtraeme: Also because you are not setting `build_style`, this package is missing a `do_build()` function.

srcpkgs/cava/template
```diff
@@ -0,0 +1,17 @@
+# Template file for 'Cava'
+pkgname=cava
+version=0.3.3
+revision=1
+short_desc="Console-based Audio Visualizer for ALSA"
+maintainer="Jan Wey. <janwey.git@gmail.com>"
+homepage="https://github.com/karlstav/cava"
+license="MIT"
```
xtraeme: Non GPL licenses, including BSD, ISC and MIT require the license file to be supplied with the binary package.

Use `vlicense file` to install it.

pullmoll:
I can take a look and package iniparser for you. We have similar problems in other packages, where tools are not build for the host (native) architecture, but for the target and can then not be run on the host.

One example is ilmbase, where I wrote some shell scripts to do the same thing as compiled C code to generate tables for cross builds. It doesn't help openexr, which depends on ilmbase, though, because there even more tables are generated.

A proper Makefile (or Makefile.in) would detect a cross compile situation and build intermediate helper binary using the host's CC/CXX and CFLAGS/CXXFLAGS.

pullmoll:
Ok, so I looked at iniparser and it's a library, not an executable

Therefore the problem is not using a system iniparser or not, but the other way round, in that iniparser does not have a configure script and defaults to `CC     ?= gcc` if no compiler is specified, thus failing to cross-compile.

It tries to compile with `gcc` where it should use e.g. `arm-linux-gnueabihf-gcc` or in general the template defined `$CC`.

You could try to simply prepend your `make` with `CC=$CC`.

Another fix I can think of is to patch iniparser/Makefile to use `$CC` instead of `gcc`:
```
        sed -i iniparser/Makefile -e "s;= gcc;= $CC;"
        make …
```

srcpkgs/cava/template
JayVii: which lines would be in that `do_build()`?

pullmoll: That would be in `do_install()`, i.e.
```
do_install() {
        make PREFIX=/usr install
}
```

JayVii: @pullmoll yes, i already added a similar measure in my local template
```
do_install() {
    make SYSTEM_INIPARSER=1 PREFIX=${DESTDIR}/usr install
    vinstall LICENSE
}
```
However, what I **don't** know, is what goes into `do_build()`, as this package to my knowledge doesn't require a preparation of that sort.

pullmoll: If you don't need to do anything in `do_build()`, but the template needs that function, just do
```
do_build() {
        :
}
```

pullmoll:
Looks :ok: to me now. I'll leave it to @xtraeme to merge.

JayVii:
@pullmoll `make CC=$CC ` did the trick. It seems to be building prtoperly on ARM now.

@xtraeme I removed the unnecessary dependency, included the license and added a `do_build()` function.

srcpkgs/cava/template
```diff
@@ -0,0 +1,21 @@
...
+checksum=188efe7f7bb4903cae3ef8e5fd6af6660c7115492f39315a97bdb9e3c95b4f77
+makedepends="alsa-lib-devel fftw-devel ncurses-devel iniparser-devel"
+
+do_build() {
+	:
+}
+
+do_install() {
+	SYSTEM_INIPARSER=1
+	make CC=$CC PREFIX=${DESTDIR}/usr install
```
xtraeme: This is wrong... do_build() is there for the compilation stage, do_install() is only meant to install the files into the package destdir.

JayVii: so i should move the `make install` & `SYSTEM_INIPARSER` lines into the `do_build()` function?

i'll get to that ASAP.

srcpkgs/cava/template
```diff
@@ -0,0 +1,20 @@
...
+maintainer="Jan Wey. <janwey.git@gmail.com>"
+homepage="https://github.com/karlstav/cava"
+license="MIT"
+distfiles="https://github.com/karlstav/${pkgname}/archive/${version}.tar.gz"
+checksum=188efe7f7bb4903cae3ef8e5fd6af6660c7115492f39315a97bdb9e3c95b4f77
+makedepends="alsa-lib-devel fftw-devel ncurses-devel iniparser-devel"
+
+do_build() {
+	SYSTEM_INIPARSER=1
+	make CC=$CC PREFIX=${DESTDIR}/usr install
```
xtraeme: The install stage must be executed in `do_install()`, so that what I was telling you is:

```
do_build() {
    make CC=$CC SYSTEM_INIPARSER=1 PREFIX=${DESTDIR}/usr
}
do_install() {
    make PREFIX=${DESTDIR}/usr install
}
```

PR #2295, opened 2015-08-18
hooks/post-install: set the mtime in packages to the corresponding commit date.

Gottox:


common/hooks/post-install/90-set-timestamps.sh
```diff
@@ -0,0 +1,11 @@
+# hook to set the mtime of any file to the mtime of the template
+hook() {
+	local timestamp= timedef= tmpl=$XBPS_SRCPKGDIR/$pkgname/template
+	timestamp=$(git log -n 1 --format=%at "$tmpl" 2> /dev/null)
+	if [ $? -ne 0 ]; then
+		timestamp=$(stat --printf='%X\n' "$tmpl")
+	fi
+	timedef=$(date +%Y%m%d%H%M.%S --date @$timestamp)
+
+	find $PKGDESTDIR -print0 | xargs touch -t $timedef
```
uggedal: Don't `xargs` need a `-0` arg with `find -print0`?

xtraeme: Even after changing the package files mtime, noarch packages won't be fully reproducable because `xbps-create(1)` will still set a diffferent `install-date` object to the package metadata.

xtraeme: If you really need a way to make package builds 100% reproducable you'll also have to modify `xbps-create(1)` to change the `build-date` object.

dominikh:
I'm not sure how this behaves with packages like Go, which uses mtimes to see if object files are older than source files and need recompilation (I don't know if the check uses `<` or `<=`).

common/hooks/post-install/90-set-timestamps.sh
Gottox: Yea, I will work on that later.

Gottox:
@dominikh I'm not into Go. Do we ship any packages, that build at runtime?

Gottox:
https://github.com/voidlinux/xbps/issues/110

pullmoll:
Just a thought:

In addition to creating and storing the SHA256 of the tarball, one could create an SHA256 hash over all the SHA256 hashes of the contents of an xbps archive. This way it would be easy to verify if two consecutive builds generated identical output.

This method would of course fail for any file which stores the current date/time in some way. It would also fail for packages which create some kinde of keys. I think nss would fall into this category. Also the linux kernel generates some key(s) IIRC?

dominikh:
> Do we ship any packages, that build at runtime?

That wasn't the point. The go package contains the Go compiler. It comes with object files, and it comes with source files. If source files are newer than object files, they will be recompiled whenever any packages from the standard library are used (resulting in increased compilation times for users). Furthermore, it breaks `go install`, because the user doesn't have write permissions to the Go installation.

The mtimes of files in our packages don't just encode the point in time at which the package was built; they also encode an order of events. Which files existed in which order. This may be relevant. It is to Go. it may be to other languages, too. Just setting all files to the same mtime throws that information away.

Gottox:
So, what should happen with this PR? Is it obsolete?

xtraeme:
This is not applicable anymore.

PR #2297, opened 2015-08-19
ca-certificates: remove expired certificates

pullmoll:
This is inspired by (beyond) linux from scratch ca-certificates build description.
The script is slightly modified for this pr.
Currently there are 3 expired certificates removed.


pullmoll:
It looks like there is something wrong with glibc locale or iconv, because travis-ci fails to translate t filename `./usr/share/ca-certificates/mozilla/AC_Raíz_Certicámara_S.A..crt` which works fine with musl libc.

PR #2300, opened 2015-08-19
Implement XBPS_DISTFILES_MIRROR (second try)

pullmoll:
The idea is to remove some burden from the original download
locations for distfiles by specifying one or more mirror locations.

+ Works with `http://` or `ftp://` mirrors for all supported `$XBPS_CHROOT_CMD` variants
+ Works with `file://` mirrors together with `uchroot` and `proot` (fixed a typo and it works now)

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,27 @@ To enable it:
     $ cd void-packages
     $ echo XBPS_CHROOT_CMD=proot >> etc/conf
 
+### Distfiles mirror(s)
+
+In etc/conf you may optionally define a mirror or a list of mirrors to search for distfiles.
+    $ echo 'XBPS_DISTFILES_MIRROR="ftp://192.168.100.5/gentoo/distfiles"' >> etc/conf
+
+If more than one mirror is to be searched, you can either specify multiple URLs separated
+with blanks, or add to the variable like this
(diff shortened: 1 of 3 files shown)


pullmoll:
It works good so far with `proot` and a `file://some/where/distfiles` URL pointing to a collection of distfiles. With `uchroot` or `unshare` it works with non-file URLs.

It now also supports `$pkgname-$vesion/$curfile` download locations, if the URL contains `*voidlinux*`. For testing purposes I set up http://distfiles.voidlinux.de/distfiles/  which contains a tree just like your `masterdir/sources`.

If the URL does not contain `*voidlinux*`, the script assumes a flat directory and looks for the `$basefile` of the original URL and also for the renamed `$curfile`, if it differs from `$basefile`. This works with mirrors like http://mirror.netcologne.de/gentoo/distfiles for many packages.

All this happens only if the user defines `XBPS_DISTFILES_MIRROR` in their `etc/conf`, thus I think it could be merged without causing trouble.

Why `uchroot -b path1:path2` doesn't work the same as `proot -b path1:path2` is to be solved and, if fixed, the xbps-src script can be changed to support `file://` URLs for `uchroot` as well.

PR #2306, opened 2015-08-19
libtirpc: fix musl.

Gottox:
fixes musl build, but I'm not 100% sure if it has any side effects as I'm a autoconf noob.


xtraeme:
Looks good :+1:

PR #2307, opened 2015-08-20
gogs: update to 0.6.5

jprjr:
The gogs build process relies on a tool called "gopm" - it allows for specifying specific versions of dependencies. Without it, "go get" pulls down the latest code from the needed dependencies and the build can break.

Instead of installing and using gopm, I wrote a short script (`misc/generate-distfiles`) to parse the .gopm file and generate distfiles lines, checksum lines, and the do_extract() function.

That's why the distfiles is so huge - it captures all the dependencies from the .gopm file in the source repo.

### DIFF
--- a/srcpkgs/gogs/files/gopm-lite
+++ b/srcpkgs/gogs/files/gopm-lite
@@ -1,0 +1,134 @@
+#!/bin/sh
+
+# gopm-lite
+# a shell script to download, extract, and link
+# dependencies from a .gopmfile
+# requirements: curl, coreutils
+
+# Needed environment variables:
+# GOPATH - packages will be linked to $GOPATH/src/
+# GOPMSTORE - where to store downloaded packages
+
+# Optional environment variables
+# MAXDEPTH - Maximum amount of attempts to recurse
+#            and download all packages
+# Defaults to 5
+
+# Needed arguments:
+# 1 - path to a .gopmfile
+
+if [ -z "$GOPATH" -o -z "$GOPMSTORE" ]; then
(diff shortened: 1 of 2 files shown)


silvernode:
I think I want gopm anyway since it seems useful for other things. Also, why not have gopm just to have it? After all it's a tool right?

jprjr:
You know what, that's a good call. I'll see about making a gopm package 

On August 21, 2015 12:04:48 AM CDT, silvernode <notifications@github.com> wrote:
>I think I want gopm anyway since it seems useful for other things.
>Also, why not have gopm just to have it? After all it's a tool right?
>
>---
>https://github.com/voidlinux/void-packages/pull/2307#issuecomment-133287242

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

jprjr:
@silvernode I made a `gopm-git` package then tried building `gogs` with that. For the life of me, I can't get it to work correctly - it's failing to pull down the memcache and redis dependencies, so when I try to build it's getting an error. Some operations seem to not work correctly. I tried to run `gopm get -d` to only download gogs + its dependencies (and not try to install) and it wound up downloading multiple commits from every single dependency for some reason.

I think at some point it might be usable, but manually pulling down the dependencies seems to work better than `gopm` - plus by placing the dependent go modules into `distfiles`, you get chucksumming etc.

jprjr:
Modified the commit - I wrote a short `gopm-lite` script, it just downloads + extracts + links dependencies from the `.gopmfile`

The `template` file is closer to the original `template` file, instead of having a ton of distfiles

PR #2311, opened 2015-08-20
xbps-src: New build-style qmake

pullmoll:
Tested it with qtcreator and qupzilla which looked good.
I'll adapt the templates later on when this pr got merged.

### DIFF
--- a/common/build-style/qmake.sh
+++ b/common/build-style/qmake.sh
@@ -1,0 +1,28 @@
+#
+# This helper is for templates using Qt4/Qt5 qmake.
+#
+do_configure() {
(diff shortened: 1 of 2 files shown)


common/build-style/qmake.sh
```diff
@@ -0,0 +1,28 @@
+#
+# This helper is for templates using Qt4/Qt5 qmake.
+#
+do_configure() {
+	: ${qmake_cmd:=qmake}
```
xtraeme: There's no reason to add new vars specifically for this build style, please use `configure_script` rather than `qmake_cmd`.

PR #2317, opened 2015-08-21
ulatencyd: unbreak + build pie

thypon:


pullmoll:
It looks like `arm*-musl` misdetects `HAVE_ERROR_H` or similar. There is no `<error.h>` in  `*-musl`, and for `x86_64-musl` detecting it seems to be working, which is odd IMO.

I'm not too good at fixing CMake things, so I can't help too much in this case.

thypon:
Thank you. The only thing that includes it unconditionally is `src/ulatencyd.c`.

Do you know what to include instead?

pullmoll:
I don't think you can include anything, because e.g. `error_at_line(3)` is not available.
It's rather the code using `<error.h>` defined functions has to be disabled.

thypon:
Wasn't error.h related. At least non on ulatencyd.

thypon:
Wasn't error.h related. At least non on ulatencyd.

PR #2318, opened 2015-08-21
New package: wemux-3.2.0

diogoleal:


pullmoll:
A line `vlicense MIT-LICENSE` is missing

What's the purpose of `make_dirs="/etc/wemux/"`?

I think since you install the example configuration as `/etc/wmux.conf`, you should just remove the `make_dirs="/etc/wemux/"` and change the `sed -i 's#/usr/local/etc/#/etc/wemux/#g' wemux` to `sed -i 's#/usr/local/etc/#/etc/#g' wemux`.

As far as I understand the script it just needs to locate wmux.conf and no additional configuration files?

diogoleal:
@pullmoll  Fix the flaws that you pointed out.

There is no need to make any extra settings.

pullmoll:
I overlooked that at the first glance:
Usually the `distfiles=…` line refers to `$version` (and sometimes `$pkgname`) where applicable, as otherwise you will always have to change these numbers whenever a new version comes out.

pullmoll:
...and now remove the leading blank in `short_desc=" …` :grinning:

diogoleal:
Ops. Sorry about the lack of attention :-(

pullmoll:
No problem. You can take look at what catastrophic things I did when I began contributing to Void a couple of months ago. It's all in the history. :smile_cat:  I think you will get used to the common pitfalls sooner or later...

diogoleal:
@pullmoll Thanks a lot for the help and consolation.
The next I will try to be more careful ;)

pullmoll:
BTW If you take a look at the `mlt` package, you can perhaps see how I handled some things.
I don't claim this template is perfect or even good. However you can perhaps look at it and see how I did the -data, -docs and -devel subpackage splitting.

diogoleal:
I saw, and it was with open flap here to thank you for helping him. just looking at him, he took some doubts.

PR #2319, opened 2015-08-21
New package: gopm

jprjr:


thypon:
Please rebase on top of master. There is a problem with go build_style without.

jprjr:
@thypon rebased, fixed, and pushed!

thypon:
Almost done: now rename commit 'New package: gopm-git' to 'New package: gopm' and remove go_get line.

jprjr:
Done and done

jprjr:
Changed the order of version/revision

thypon:
Just to know: the stable version isn't functional?

jprjr:
The last released version is from about a year ago, I can't get it to compile

thypon:
btw a version in the form $laststableversion.$dateoflastcommit should also rule out update-check related errors. Maybe it's better to wait for @xtraeme opinion.

@dominikh could you include a `: ${go_cmd:=go}` instead of plain go to made possible to switch to gopm?

srcpkgs/gopm/template
```diff
@@ -0,0 +1,14 @@
+# Template build file for 'gopm'
+pkgname=gopm
+version=20150713
```
xtraeme: Please version it as `0.0.date`, this way we can upgrade when an official version is released.

jprjr:
Hi sorry, I've been out of town for about a week - expect some fixes soon!

jprjr:
Fixed and rebased

PR #2325, opened 2015-08-21
New Package: pdf2htmlEX-0.14.6

### LINKED ISSUE #1801: Package request: pdf2htmlEX
https://coolwanglu.github.io/pdf2htmlEX/

logenkain:
closes #1801

### DIFF
--- a/srcpkgs/pdf2htmlEX/template
+++ b/srcpkgs/pdf2htmlEX/template
@@ -1,0 +1,14 @@
+# Template file for 'pdf2htmlEX'
+pkgname=pdf2htmlEX
+version=0.14.6
+revision=1
+build_style=cmake
+only_for_archs="x86_64 i686 armv6l armv7l"
+hostmakedepends="cmake pkg-config"
+makedepends="poppler-devel cairo-devel fontforge-devel python-devel pango-devel libspiro-devel"
+short_desc="Renders PDF files in HTML"
+maintainer="Logen K <logen@sudotask.com>"
+license="GPL-3"
+homepage="http://coolwanglu.github.io/pdf2htmlEX/"
+distfiles="https://github.com/coolwanglu/${pkgname}/archive/v${version}.tar.gz"
+checksum="320ac2e1c2ea4a2972970f52809d90073ee00a6c42ef6d9833fb48436222f0e5"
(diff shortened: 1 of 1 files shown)


pullmoll:
Instead of disabling *-musl by `only_for_archs="…"` I think the problem (it requires `<execinfo.h>`) should be fixed by patching the source to include and use e.g. backtrace(3) only, if `<execinfo.h>` is available.

This can usually be done by shielding the relevant code with `#if defined(__GLIBC__)`.

logenkain:
On Fri, Aug 21, 2015 at 11:48:21PM -0700, Jürgen Buchmüller wrote:
> Instead of disabling *-musl by `only_for_archs="…"` I think the problem (it requires `<execinfo.h>`) should be fixed by patching the source to include and use e.g. backtrace(3) only, if `<execinfo.h>` is available.
> 
> This can usually be done by shielding the relevant code with `#if defined(__GLIBC__)`.

I'd love to do stuff like that, but it's beyond my understanding.  The
most actuall programming I've done is high-low games...  Hell, I'm still
not even sure what musl is for.

If someone else wants to make the patch I'll be happy to include it, but
beyond that I really don't know what I'm doing.

pullmoll:
Ok, no problem. Here's a short explanation:
The musl libc is an alternative to the GNU libc (glibc), is not as big and also notably faster.
Some people prefer to run their system with musl libc and so (important) packages need to be linked against it to work on such systems. We are always trying to make every package work on as many architectures as possible.

After #2333 is merged, you can remove the `only_for_archs=` from your template and pdf2htmlEX should then build for musl libc. The cause for the problem with musl libc was not in your part of the game.

logenkain:
On Sat, Aug 22, 2015 at 09:15:29AM -0700, Jürgen Buchmüller wrote:
> Ok, no problem. Here's a short explanation:
> The musl libc is an alternative to the GNU libc (glibc), is not as big and also notably faster.
> Some people prefer to run their system with musl libc and so (important) packages need to be linked against it to work on such systems. We are always trying to make every package work on as many architectures as possible.
> 
> After #2333 is merged, you can remove the `only_for_archs=` from your template and pdf2htmlEX should then build for musl libc. The cause for the problem with musl libc was not in your part of the game.
> 
Yea I knew it was an alternative library, I just didn't know why it
mattered, thanks for the info.  I'll be sure to take out the arch
restriction.  Thanks again.

PR #2326, opened 2015-08-21
ulatencyd: fix musl

thypon:
@pullmoll this should fix also musl problems.


pullmoll:
Fine! The patch to add the BSD `fts.c/h` implementation is clever. In other packages requiring fts.h I disabled the corresponding code by adding `#if defined(__GLIBC__)` or similar.

Perhaps we can generalize the implementation and create some packages like `musl-fts` and `musl-fts-devel`? That way more packages could be musl-fixed in a better and more consistent way.

pullmoll:
I created a github.com project musl-fts here https://github.com/pullmoll/musl-fts
So far I tested that bootstrap.sh generates the required configure and should work with Void's `make_style=gnu-configure` and a `pre_configure()` which runs `./bootstrap.sh`.
I will now create a Void package pr for a `musl-fts` package using this so we can use the same implementation for all packages requiring fts.

PR #2327, opened 2015-08-22
New package: libtermkey

steski:


pullmoll:
I think the libtermkey template needs `vlicense LICENSE` in `do_install()` because its license is `MIT`.

pullmoll:
:+1:

PR #2328, opened 2015-08-22
New package: texmaker, closes #2147

### LINKED ISSUE #2147: Package Request: Texmaker
http://www.xm1math.net/texmaker/
LaTeX editor

thypon:


pullmoll:
IMO `INSTALL_ROOT=${DESTDIR}/usr` is right. Look into qtcreator, qupzilla (not yet converted), octoxbps or other qmake packages. Usually the executable target is `${INSTALL_ROOT}/bin`, and thus `INSTALL_ROOT=${DESTDIR}/usr` is correct for them.

If you need another `INSTALL_ROOT` because texmaker installs to `${INSTALL_ROOT}/usr/bin`, you should rather add `make_install_args="INSTALL_ROOT=\${DESTDIR}"` to this template.

pullmoll:
Hmm.. actually qupzilla expects `INSTALL_ROOT=${DESTDIR}` as well.
So the question is *what should be the default*?

I think we could change `common/build-style/qmake.sh` to simply replace `/usr/usr/` with `/usr/` since certainly nobody ever will want to install to `/usr/usr/...` :grimacing:

thypon:
For the 3/4 packages inside our repository i tried to port to build_style=qmake `INSTALL_ROOT=${DESTDIR}` is correct.

pullmoll:
Ok, then qtcreator is wrong.
I'll change its template to modify INSTALL_ROOT to append /usr  and change qmake.sh accordingly.

PR #2332, opened 2015-08-22
New package: musl-fts-1.0

pullmoll:
This is intended to standardize handling of packages for musl libc
which require an fts(3) implementation. Adding musl-fts to the
makedepends and linking /usr/lib/libfts.a to the target (i.e.
`LIBS="-lfts"`) should suffice.

### DIFF
--- a/srcpkgs/musl-fts/template
+++ b/srcpkgs/musl-fts/template
@@ -1,0 +1,21 @@
+# Template file for 'musl-fts'
+pkgname=musl-fts
+version=1.0
+revision=1
+build_style=gnu-configure
+hostmakedepends="automake"
+only_for_archs="aarch64-musl armv6l-musl armv7l-musl i686-musl mips-musl mipsel-musl mipselhf-musl x86_64-musl"
+short_desc="Implementation of fts(3) for musl libc"
+maintainer="Jürgen Buchmüller <pullmoll@t-online.de>"
+license="BSD"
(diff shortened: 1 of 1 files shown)


Gottox:
+1

chneukirchen:
I started a similar thing once... https://github.com/chneukirchen/libegacy

Didn't find enough time to finish it yet tho.

pullmoll:
Well, I saw the pr #2326 by @thypon and thought it was good to generalize it. It was a quick hack...

And of course I didn't know you already had something WIP in that regard.

chneukirchen:
No, feel free to pick it up or do your own... I just guess we need more than fts (e.g. glob too) often.

thypon:
Tecnically my "work" predates the libegacy fts implementation. I think we could merge the force on libegacy inside a voidlinux subproject.

pullmoll:
I agree to merge the force, while I think not merging everything into a single libegacy.a makes sense:
+ We can update just one of e.g. musl-fts, musl-glob, musl-base64 should the need arise.
+ Only packages linking to one of the updated archives need to be rebuilt.
+ There may be different licenses on the code snippets we use.

Oh and BTW the configure.ac and Makefile.in I created could use some tweaking to actually support the symbols (`HAVE_...`) used in the code :grin:

As a starting point and if you see fit @thypon you could merge this and try to use it in ulatencyd?

PR #2345, opened 2015-08-23
New package: perl-DBD-Pg

jprjr:


pullmoll:
I tried to make it cross compile to no avail.
Suggest `nocross=yes`.

jprjr:
Done!

PR #2349, opened 2015-08-24
New package: icecat-31.8.0

thypon:
The generated icecat binary works, but:
+ It displays an error on every page at the bottom
+ This is while trying to display the source file
  /builddir/icecat-31.8.0/extensions/gnu/spyblock@gnu.org/chrome/content/ui/overlay.xul
  which is included by the javascript file
  /builddir/icecat-31.8.0/extensions/gnu/spyblock@gnu.org/lib/ui.js

One solution would be to disable spyblock@gnu.org, but that is not desirable.


pullmoll:
I would not yet have created the pr, because I wan't finished with fixing issues.
Anyway, I'll try to fix it up now.

pullmoll:
Closing until the remaining issues are fixed to avoid travis-ci overload.

PR #2351, opened 2015-08-24
base-system-busybox+runit-void: Using relative symbolic link to for /sbin/init

moul:
Fixing "initrd + busybox' switch_root" based boots

### DIFF
--- a/srcpkgs/base-system-busybox/template
+++ b/srcpkgs/base-system-busybox/template
@@ -1,7 +1,7 @@
 # Build template for 'base-system-busybox'.
 pkgname=base-system-busybox
 version=1.23.2
-revision=6
+revision=7
 wrksrc="busybox-${version}"
 homepage="http://www.busybox.net"
 hostmakedepends="perl"
@@ -59,5 +59,5 @@ do_install() {
 	mv ${DESTDIR}/usr/sbin/* ${DESTDIR}/usr/bin
 	rm -f ${DESTDIR}/{,s}bin
 	# Provide /sbin/init symlink to runit.
-	ln -s /usr/bin/runit-init ${DESTDIR}/usr/bin/init
+	ln -s runit-init ${DESTDIR}/usr/bin/init
 }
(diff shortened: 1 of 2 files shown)


Gottox:
please increase the revision of the package :+1:

moul:
@Gottox done + rebased upstream/master

Gottox:
Some other opinions?

I don't like to merge stuff that can break boot without a second review.

pullmoll:
I would like to understand the reasoning for this change. It doesn't seem necessary at the first glance.

moul:
/sbin/init is often the entry point when working outside of the rootfs (using a chroot or in my case, from an initrd), testing existence of file using test -x, or test -f, or even readlink will fail to resolve the absolute symlink, and I need to check it before `pivot_root`int inside the rootfs

I saw this issue before in other init systems:

* [systemd](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=750360)
* [upstart](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=590744)
* [ueventd (android)](https://gitlab.com/pbeeler/system_core/commit/33680b106dcb7025de0c37340869e32db61ad190)

Gottox:
It seems reasonable to me. /sbin/init links are broken if you build your own early userspace.

moul:
a rebase is needed

moul:
rebased

srcpkgs/base-system-busybox/template
```diff
@@ -59,5 +59,5 @@ do_install() {
 	mv ${DESTDIR}/usr/sbin/* ${DESTDIR}/usr/bin
 	rm -f ${DESTDIR}/{,s}bin
 	# Provide /sbin/init symlink to runit.
-	ln -s /usr/bin/runit-init ${DESTDIR}/usr/bin/init
+	ln -s ./runit-init ${DESTDIR}/usr/bin/init
```
xtraeme: You can drop `./` completely.

moul: done

PR #2362, opened 2015-08-25
New package: musca-0.9.24

diogoleal:


pullmoll:
+ `homepage="ihttp://aerosuidae.net/musca.html"` remove the `i` before http
+ `distfiles="http://aerosuidae.net/musca-0.9.24.tgz"` use `${version}` and perhaps also `${pkgname}`
+ add a newline to the last line

pullmoll:
:smile_cat: I meant a line break after the last line, not an extra empty line... well, doesn't matter :)

PR #2366, opened 2015-08-25
base-files: add a default prompt

moul:
Not sure it will pass the 'minimalism' requirement :)

Seems to have been removed in: https://github.com/voidlinux/void-packages/commit/7c3fb28abbedafe818da4f4a6feb9be922c28465

### DIFF
--- a/srcpkgs/base-files/files/profile
+++ b/srcpkgs/base-files/files/profile
@@ -9,6 +9,10 @@ export PATH
 # Set default umask
 umask 022
 
+# Set default prompt
+PS1='\h:\w\$'
+export PS1
+
 # Load profiles from /etc/profile.d
 if [ -d /etc/profile.d/ ]; then
 	for f in /etc/profile.d/*.sh; do


chneukirchen:
The prompt is set in `/etc/skel/.bashrc` for new users, so for reasons of consistency we probably should use the same here.

(Also, /etc/profile is read by all sh-ish shells, that prompt doesn't work in dash nor mksh.)

moul:
Ok, good point, I think a bash dedicated config file will be better than `/etc/profile` too

From what I see `/root` is not based on `/etc/skel`, do you think we should try to populate `/root` from `/etc/skel` or do you think I need to update `/etc/skel/.bashrc` for all users and another file for `/root` ?

chneukirchen:
IMO a file in /etc/bash/bashrc.d would be appropriate, but I don't use bash and couldn't care less. :)

Pinging @lemmi @xtraeme

xtraeme:
I agree with @chneukirchen, setting `PS1` via `/etc/profile` this way won't work with our default `/bin/sh` (dash).

If you want to have a reasonable `PS1` switch your user to the `bash` shell. Once the user is populated with the contents of `/etc/skel` it will DTRT.

```
[juan@leysa void-packages]$ cat /etc/skel/.bashrc 
# .bashrc

# If not running interactively, don't do anything
[[ $- != *i* ]] && return

alias ls='ls --color=auto'
PS1='[\u@\h \W]\$ '
[juan@leysa void-packages]$
```

moul:
:+1:

moul:
It seems that the /etc/skel directory is not copied for root user

```console
Last login: Thu Sep  3 10:50:23 2015 from x.x.x.x
-bash-4.3# echo $SHELL
/bin/bash
-bash-4.3# cat /etc/skel/.bashrc
# .bashrc

# If not running interactively, don't do anything
[[ $- != *i* ]] && return

# Source global definitions
[ -f /etc/bashrc ] && . /etc/bashrc

alias ls='ls --color=auto'
PS1='[\u@\h \W]\$ '
-bash-4.3# pwd
/root
-bash-4.3# ls -la
total 20
drwx------  3 root root 4096 Sep  3 10:51 .
drwxr-xr-x 17 root root 4096 Sep  3 10:11 ..
-rw-------  1 root root   40 Sep  3 10:51 .bash_history
drwx------  2 root root 4096 Sep  3 10:11 .ssh
-bash-4.3#
```

skrzyp:
But still, I think that "user@host$" prompt instead of "$" prompt, even for _da_sh will be a better option for base system, at least more reasonable…
Some of my silly research shown that users will assume that system is "broken" or "incomplete" when they see only the "$" mark. So, please reconsider changing this if you can :)

PR #2367, opened 2015-08-25
scaleway-base: Initial version

moul:
Means to be run on Scaleway C1 servers (https://www.scaleway.com).

Actually, we are working on the rpi2 base image:
http://repo.voidlinux.eu/live/current/void-rpi2-rootfs-20150713.tar.xz

https://github.com/scaleway/image-voidlinux/blob/master/Dockerfile

### DIFF
--- a/srcpkgs/scaleway-base/template
+++ b/srcpkgs/scaleway-base/template
@@ -1,0 +1,15 @@
+# Template file for 'scaleway-base'
+pkgname=scaleway-base
+version=1
+revision=1
+homepage="http://www.scaleway.com/"
+short_desc="Void Linux Scaleway base files"
+maintainer="Manfred Touron <mtouron@scaleway.com>"
+license="Public Domain"
+
+only_for_archs="armv6l armv6l-musl armv7l armv7l-musl"
+depends="virtual?ntp-daemon"
+
+do_install() {
+	:
+}
--- a/srcpkgs/scaleway-base/INSTALL
+++ b/srcpkgs/scaleway-base/INSTALL
@@ -1,0 +1,13 @@
+case "$ACTION" in
+post)
+	# enable sshd, ntpdate, ntpd and dhcpcd services.
+	if [ -x /usr/bin/systemctl ]; then
+		systemctl enable sshd.service ntpdate.service ntpd.service dhcpcd.service
+	else
+		mkdir -p etc/runit/runsvdir/default/
+		ln -sf /etc/sv/sshd etc/runit/runsvdir/default/
+		ln -sf /etc/sv/ntpd etc/runit/runsvdir/default/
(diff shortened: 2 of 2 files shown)


Ypnose:
Why post this "empty" pull request, where a simple email would suffice. I'm not sure to understand the goal…

srcpkgs/scaleway-base/template
```diff
@@ -0,0 +1,15 @@
...
+pkgname=scaleway-base
+version=1
+revision=1
+homepage="http://www.scaleway.com/"
+short_desc="Void Linux Scaleway base files"
+maintainer="Manfred Touron <mtouron@scaleway.com>"
+license="Public Domain"
+
+only_for_archs="armv6l armv6l-musl armv7l armv7l-musl"
+depends="virtual?ntp-daemon"
```
xtraeme: Why do you need such package, this only depends on `ntp-daemon`.

I'd suggest you to use `base-voidstrap` which contains dependencies for `chroot` -alike environments.

moul: I need a base without kmod nor kbd
Running void-linux on Scaleway is kind of similar with running it using Docker (no kernel embedded).

Ypnose: I'm a bit curious, what do you mean by no kernel embedded? Kernel will be still needed for bare metal, AFAIK.

moul: We are using u-boot, the kernel is dissociated with rootfs, YES we still need a kernel :p

Ypnose: Oh I see. You use your own kernel and only change the rootfs. Then, almost every distros could be supported (and sorry @xtraeme for the spam notifications).

moul: exact :)

xtraeme: @moul in this case I'd suggest you to create a `base-system-scaleway` package instead, with the dependencies set in `base-system`. If you don't need `kbd` nor `kmod`, feel free to drop them.

moul: Ok, I'm not sure to see the difference, do you have a link for some documentation ?

Ypnose: You could take a look at the <code>base-system</code> [template](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/base-system/template). As @xtraeme said, having a <code>base-system-scaleway</code> with only the needed pkgs, seems a good approach.

moul: Thank you @xtraeme and @Ypnose 

I'm closing this PR and will add a reference to #2367 when I will create a new PR for `base-system-scaleway`

Ypnose: Looking forward to see what you will do.

PR #2371, opened 2015-08-25
New package: c-client (imap library)

jprjr:


jprjr:
Fixed + rebased to pass lint, added `nocross="yes"` until I get cross-compilation working

srcpkgs/c-client/template
```diff
@@ -0,0 +1,45 @@
...
+maintainer="John Regan <john@jrjrtech.com>"
+license="Apache"
+homepage="http://www.washington.edu/imap"
+distfiles="ftp://ftp.cac.washington.edu/imap/imap-${version}.tar.gz"
+checksum="53e15a2b5c1bc80161d42e9f69792a3fa18332b7b771910131004eb520004a28"
+wrksrc="imap-${version}"
+makedepends="pam-devel libressl-devel e2fsprogs-devel"
+nocross="yes"
+
+post_extract() {
```
Gottox: This should be done in the configure phase, so use do_configure instead.

jprjr:
Fixed + rebased

PR #2372, opened 2015-08-25
memcached: fix build in musl, add -devel subpkg

jprjr:
The current build fails on musl - `#include <sys/signal.h>` generates a compiler warning and the build fails.

The warning is just that the correct include is `#include <signal.h>` - same for glibc.

The included patch has been submitted to upstream mailing list: https://groups.google.com/forum/#!topic/memcached/ThyxJORhxh0

### DIFF
--- a/srcpkgs/memcached-devel
+++ b/srcpkgs/memcached-devel
@@ -1,0 +1,1 @@
+memcached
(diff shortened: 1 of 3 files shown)


jprjr:
Fixed + rebased to pass lint

PR #2373, opened 2015-08-25
runit: fix build PIE

thypon:


chneukirchen:
I'm against dropping `-static`, by default it only applies to runit and runit-init, which have virtually no attack surface.

thypon:
We have to measure the performance impact. By the way there is the static build option that permits to have a full static runit binary. I don't get when a mixed static/dynamic runit is usefull.

Gottox:
[diff ignoring whitespaces](https://github.com/voidlinux/void-packages/pull/2373/files?w=1)

thypon:
@Gottox thank you, i rebased.

xtraeme:
Looks good to me.

Please revbump it.

xtraeme:
I agree with @chneukirchen 

It's good enough as is already, there's no need to chang ethis.

PR #2378, opened 2015-08-26
New package: libmemcached

jprjr:
This depends on pull #2372

### DIFF
--- a/srcpkgs/libmemcached-devel
+++ b/srcpkgs/libmemcached-devel
@@ -1,0 +1,1 @@
+libmemcached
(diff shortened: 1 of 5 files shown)


Gottox:
pls rebase so we can see what travis thinks of that PR ;)

jprjr:
Done and done!

Gottox:
cross build is broken.

jprjr:
I'm noticing that 'gnu-configure' seems to work correctly for cross-builds over plain 'configure' - just changed to use that, let's see what Travis says.

jprjr:
Looks like cross-build is definitely broken - I also had the wrong license in there (GPL when it should be BSD), plus spaces instead of tabs - fixed all those + rebased.

PR #2379, opened 2015-08-26
gperftools: unbreak musl

pullmoll:


chneukirchen:
If it works, +1. Thanks for porting.

pullmoll:
I haven't tried if it works. Chances are good, as the patch set is more or less cosmetics (standard type names) and excluding things that musl libc does not have (__sbrk, same as uClibc).

PR #2380, opened 2015-08-26
New package: oniguruma (multi-charset regex library)

jprjr:


pullmoll:
You should add
```
post_install() {
        vlicense COPYING
}
```
for the BSD license.

Also `nocross=yes` seems advised, because arm* glibc does not build and arm*-musl doesn't detect the word size (32) right and thus there are warnings when expanding the `__bswap64` macro.

I tried running `autoreconf -if` on the package but it fails.

jprjr:
Added both `nocross=yes` And the license command

PR #2382, opened 2015-08-26
New package: ocaml-findlib

jprjr:

PR #2381, opened 2015-08-26
New package: camlp4

jprjr:

PR #2383, opened 2015-08-26
New package: pfff

jprjr:
This depends on  #2381 and  #2382

### DIFF
--- a/srcpkgs/pfff/template
+++ b/srcpkgs/pfff/template
@@ -1,0 +1,29 @@
+# Template build file for 'pfff'
+pkgname=pfff
+version=0.29
+revision=1
+build_style=configure
+configure_args="--prefix=/usr"
+short_desc="Tools and APIs for code analysis, visualization and transformation"
+maintainer="John Regan <john@jrjrtech.com>"
+license="LGPL-2.1"
+homepage="https://github.com/facebook/pfff"
+distfiles="$homepage/archive/v${version}.tar.gz"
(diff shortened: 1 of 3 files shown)


chneukirchen:
I use opam to install all ocaml tools. pfff is there too, ymmv.

srcpkgs/pfff/template
```diff
@@ -0,0 +1,29 @@
...
+checksum="d31c68f1ebb1770297eb26441d21d83728c0fbdb9c7bb3be27625db45fae72d1"
+makedepends="perl ocaml ocaml-findlib camlp4 zlib-devel ncurses-devel"
+depends="ocaml"
+nocross="yes"
+
+do_build() {
+	make clean
+	make depend
+	make
+	make opt
```
Gottox: Why 4 make calls?

jprjr: You're right, the 'make clean' isn't needed. Their official docs list three make calls, though: https://github.com/facebook/pfff/wiki/Main

Vaelatern:
Please make some change and re-push since it seems at the time of last push, Travis was unable to find `ocaml-findlib` and thus failed on non-arm builds. This is, I believe, no longer an issue.

jprjr:
Done, I just rebased from master and pushed.

xtraeme:
AFAIK ocaml binaries must not be stripped, probably you should set `nostrip=yes`.

PR #2384, opened 2015-08-26
New package: python-greenlet

steski:


pullmoll:
The MIT license needs `vlicense LICENSE` in post_install() and pkg_install()

PR #2386, opened 2015-08-26
mongodb: unbreak musl

pullmoll:


pullmoll:
Hm.. that error for x86_64-musl doesn't happen here.

PR #2389, opened 2015-08-27
New package: ccrypt-1.10

evandeaubl:


evandeaubl:
Looks like the last build failed because Sourceforge hiccuped trying to download the source package. Any way to force a rebuild?

Duncaen:
change the last commit maybe with ``git commit --amend`` and then push with ``--force``

PR #2393, opened 2015-08-28
pam: fix i686 build + cross

pullmoll:
Builds with binary-bootstrap i686 again. It failed when I tried to build it locally in a `./xbps-src binary-bootstrap i686` environment.


pullmoll:
Hmm.. I'll look into fixing cross as well.

pullmoll:
This may be too much hackery for @xtraeme 's taste. Perhaps some of you have a better fix using exsting `$XBPS_...` variables?

xtraeme:
I'm not sure about these changes... can you provide a build log?

pullmoll:
I guess you need a build log with failing for i686. I'll create one - brb.

pullmoll:
Here is [a build log](http://sprunge.us/fgaO). It fails when trying to link `padout`.

pullmoll:
Ouch! Sorry, this patch is probably not necessary.
I tried without ccache and distcc and now it built correctly

pullmoll:
Works after disabling ccache + distcc thus closing.

PR #2394, opened 2015-08-28
base-files: vkpurge ignore running kernel version

uggedal:


uggedal:
+1 from @chneukirchen on IRC

PR #2396, opened 2015-08-28
New package: pdfjam-2.05

r4yan2:


chneukirchen:
Why not simply `tlmgr install pdfjam`?

r4yan2:
yeah, the @chneukirchen method it's better, if only were written somewhere that pdfjam could be installed that way :-P

pullmoll:
We have a Wiki... https://github.com/voidlinux/documentation/wiki

PR #2397, opened 2015-08-28
picocom: update to 2.0

allan:


chneukirchen:
Using `sed -i` is fine for templates.

allan:
and `ed` is fine for inplace editing everywhere:)

chneukirchen:
Is it in base-chroot?

pullmoll:
It ain't, thus the `hostmakedepends=ed`. I'd prefer to see `sed -i` as well.. the `ed` syntax is..umm..awkward.

dominikh:
We use sed in all our other templates, please use sed :)

allan:
There is some ed in other templates like e0b08294e94465711372b2aeb10c3ba72356af8c.  I vote for ed in base-chroot!!

Gottox:
I vote for a real patch ;)

Gottox:
... and no, ed shouldn't be in base. There's absolutely no reason for that.

xtraeme:
Please create a real patch and submit it upstream, there's no need for `ed(1)` nor `sed(1)`.

Thanks

srcpkgs/picocom/template
```diff
@@ -1,18 +1,19 @@
...
+version=2.0
 revision=1
 short_desc="Minimal dumb-terminal emulation program like minicom"
 maintainer="allan <mail@may.mooo.com>"
 license="GPL-2"
 homepage="https://github.com/npat-efault/picocom"
 distfiles="https://github.com/npat-efault/picocom/archive/${version}.tar.gz"
-checksum=6a3e6ba74eec3dd111d5c6f8245b8f08ee54db783e5d319a8589e8f7a4a108f0
+checksum=e0f7095c118f43c1c09787a16ef5c7d4c4ad4457dd3b1281efcc4cdb8906af9a
+hostmakedepends=ed
```
xtraeme: `hostmakedepends` is not necessary anymore.

PR #2400, opened 2015-08-28
sandy: checksum changed w/o version bump

pullmoll:
Someone who has the old archive should perhaps look
for the difference?


pullmoll:
WTF this seems to be completely broken, as everyone gets a different checksum!?

dominikh:
Snapshots are computed on the fly when they're being requested. As the timestamp is part of the archive, no two downloads will have an identical hash.

pullmoll:
Ah, ok. Then it doesn't make sense at all to try to fix this template.

PR #2401, opened 2015-08-28
New package: synapse-0.2.99.1

steski:


Gottox:
0.2.99 is an unstable release. We should wait for 0.3 before merging this.

steski:
Yeah, I overlooked that.

PR #2407, opened 2015-08-30
zile: fix cross compiling

pullmoll:
Also fix *-musl by disabling the regeneration of the zile.1.in file.
The shipped version should be recent enough AFAICT.
@chneukirchen please merge if you agree.


pullmoll:
Hmmm.. it seems help2man is broken, because zile compiled locally for x86_64-musl.
Perhaps just a revbump for help2man?

xtraeme:
This pkg can now be updated after fixing musl+help2man.

PR #2408, opened 2015-08-30
New package: mclinker

thypon:


pullmoll:
I've seen that kind of errors `#include <gnu/stub-32.h>` in other packages as well.
And I haven't found a solution. Not even `autoreconf -if` helps.
Therefore :+1: for `noarch=yes`

srcpkgs/mclinker/template
```diff
@@ -0,0 +1,32 @@
...
+hostmakedepends="automake libtool flex llvm"
+makedepends="zlib-devel libffi-devel libedit-devel ncurses-devel"
+short_desc="LLVM Linker for Mobile Computing"
+maintainer="Andrea Brancaleoni <miwaxe@gmail.com>"
+license="BSD"
+homepage="https://github.com/$pkgname/$pkgname"
+distfiles="$homepage/archive/release_${_version}.tar.gz"
+checksum=af09905283840d3c3f29a202889711eec7fafd6090e09d3d71c9fdba5e83f4a4
+patch_args="-Np1"
+nocross=y
```
Gottox: please use ```yes```

PR #2410, opened 2015-08-30
New package: massren-1.3.0

thypon:
@dominikh go-sqlite3 is a problem. nocross is necessary?


dominikh:
Yes, nocross for now.

PR #2412, opened 2015-08-30
New package: livewallpaper-0.4.1

thypon:


srcpkgs/livewallpaper/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'livewallpaper'
+pkgname=livewallpaper
+version=0.4.1
+revision=1
+build_style=cmake
+hostmakedepends="cmake pkg-config gobject-introspection intltool"
```
xtraeme: `gobject-introspection` should be a build option to allow cross compilation.

thypon: seems to be mandatory:

```
# check dependencies
set(LIVEWALLPAPER_REQUIRES
        gobject-2.0
        gobject-introspection-1.0
        glib-2.0
        gio-2.0
        gdk-3.0
        gdk-pixbuf-2.0
        libpeas-1.0
        glew
)
```

xtraeme: You are right, thanks :-1:

PR #2415, opened 2015-08-30
New package: bish-0.1

diogoleal:


srcpkgs/bish/template
```diff
@@ -0,0 +1,17 @@
...
+revision=1
+build_style=gnu-makefile
+short_desc="A language that compiles to Bash"
+maintainer="Diogo Leal <diogo@diogoleal.com>"
+license="MIT"
+homepage="https://github.com/tdenniston/bish"
+distfiles="https://github.com/tdenniston/bish/archive/v${version}.tar.gz"
+checksum=796d1efcbc9e8c7ea604881860aa8d857aad089c4eeb766283c21c210687942b
+
+do_install(){
```
xtraeme: `do_install` -> `post_install`

xtraeme:
This package is not installing any executable to DESTDIR.

srcpkgs/bish/template
```diff
@@ -0,0 +1,18 @@
...
+revision=1
+build_style=gnu-makefile
+short_desc="A language that compiles to Bash"
+maintainer="Diogo Leal <diogo@diogoleal.com>"
+license="MIT"
+homepage="https://github.com/tdenniston/bish"
+distfiles="https://github.com/tdenniston/bish/archive/v${version}.tar.gz"
+checksum=796d1efcbc9e8c7ea604881860aa8d857aad089c4eeb766283c21c210687942b
+
+post_install(){
```
xtraeme: If you are installing everything via `post_install`, there's no need to use it, move this to `do_install` instead.

PR #2417, opened 2015-08-31
New package: tigervnc-1.5.0

evandeaubl:


srcpkgs/tigervnc/template
```diff
@@ -0,0 +1,17 @@
...
+pkgname=tigervnc
+version=1.5.0
+revision=1
+build_style=cmake
+short_desc="VNC client forked from TightVNC"
+maintainer="Evan Deaubl <evan@deaubl.name>"
+license="GPL-2"
+hostmakedepends="cmake"
+makedepends="fltk-devel zlib-devel libXtst-devel libjpeg-turbo-devel $(vopt_if gnutls gnutls-devel)"
+depends="fltk zlib libXtst libjpeg-turbo $(vopt_if gnutls gnutls)"
```
xtraeme: This depends line is wrong, dependencies are guessed automagically by xbps-src for ELF files.

pullmoll:
If you replace
```
distfiles="https://github.com/TigerVNC/tigervnc/archive/v1.5.0.tar.gz"
```
with
```
distfiles="https://github.com/TigerVNC/tigervnc/archive/v${version}.tar.gz"
```
you save modifying one line on the next update :smiley: 

And if you change it to
```
distfiles="https://github.com/TigerVNC/tigervnc/archive/v${version}.tar.gz`>tigervnc-${version}.tar.gz"
```
you increase the chances of this specific distfile being found on mirror sites.

evandeaubl:
Handy, missed the rename logic for distfiles in the manual.

pullmoll:
Sorry, I didn't see you need to add a `conflicts="turbovnc>=0"` to this package, because the existing package `turbovnc` also installs `/usr/bin/vnc{server,passwd,viewer}`. This package then requires a `conflicts="tigervnc>=0"`.

evandeaubl:
I modified the turbovnc template as requested, and it doesn't build on x86_64-musl. I find it hard to believe it built before my change (it looks like the failing file in question is checking for glibc specifically, and so I'm not surprised musl fails). Do I need to fix the musl build I don't think I broke to get this accepted?

pullmoll:
This may be a temporary problem, perhaps because of parallelism. I'll try to build it locally to see what goes on there.

PR #2424, opened 2015-08-31
New package: libbind

jprjr:
ISC's libbind provides resolver functions not found in musl, particularly the thread-safe versions of functions (like `res_nsearch` and `res_nquery`).

Note: this isn't the same as what's in bind-libs, libbind was split off from the bind package some time ago.

### DIFF
--- a/srcpkgs/libbind-docs
+++ b/srcpkgs/libbind-docs
@@ -1,0 +1,1 @@
+libbind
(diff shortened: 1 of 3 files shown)


jprjr:
Actually this probably isn't a good idea - I found out that musl's resolve functions are already thread-safe, so I can just patch programs to use them.

PR #2426, opened 2015-08-31
[DON'T MERGE] perl: Add musl libc encoding UTF-8-CODE-UNITS

### LINKED ISSUE #2425: help2man (perl): failure with musl libc
With musl libc help2man tries to encode a document using an `$user_encoding` value of `UTF-8-CODE-UNITS` in line 56 of `/usr/bin/help2man`.

The exact error message is:
```
Unknown encoding 'UTF-8-CODE-UNITS' at /usr/bin/help2man line 56.
```
This happens when e.g. trying to build `bison` for musl libc.

The value for this encoding is reported to be contained in `/usr/lib/libc.so` and `/usr/lib/libc.a` in a `x86_64-musl` masterdir, so it seems to be legitimate.

I think perl::Encode needs to handle this specific value for musl libc in some way. Perhaps add a patch for `cpan/Encode/lib/Encode/Alias.pm` for additional valid encodings for musl libc? i don't know too much perl, so perhaps @xtraeme or whoever knows perl can help here.

pullmoll:
Add an alias for UTF-8-CODE-UNITS => UTF-8 to the file
cpan/Encode/lib/Encode/Alias.pm.

Closes #2425 

I don't know if it is necessary to rebuild perl modules which depend on perl::Encode!?


pullmoll:
The issue should be solved on the side of musl-1.1.11

pullmoll:
Closing this, because it's the wrong way to go.

PR #2430, opened 2015-09-01
tbb: fix building for musl

jprjr:


thypon:
Could you add nocross where only_for_archs was defined?

thypon:
fixed upstream

jprjr:
Oh man they were fast on fixing that. Sweet!

PR #2432, opened 2015-09-01
sandy: use content cksum (new feature)

pullmoll:
To overcome a situation as apparent in the sandy package, where
every download generates a new checksum, this is a try introduce a new
semantics for the `checksum=` variable in templates:

If a checksum starts with an '@' (at), it refers to the _contents_ of
the tarball as opposed to regular checksums, which refer to the digest
of the container tarball, zip, etc.

This is achieved by temporarily extracting the tarball's contents
to a temporary file with `tar xOf $distfile > /tmp/$$`, running the
defined ${XBPS_DIGEST_CMD} on `/tmp/$$` and finally removing the
`/tmp/$$` again.

The '@' semantics for now only works with tarballs.

Files with (suggested new term) `contents checksum` are not hard linked into
the /host/sources/by_sha256 tree, because there only container checksum links
should exist.

A developer wishing to use the new feature will have to manually determine the
cksum of a template's tarball by running e.g. `tar xOf file.tar.gz | sha256sum`.


chneukirchen:
I'd rather use Git fetch that tree... besides you only get a 4 year old release from an outdated cgit anyway (I think recent cgit releases have stable tarballs).

pullmoll:
If the `sandy` package is the only one where unstable tarballs happen, then this makes no sense at all.

pullmoll:
Close because of uselessness.

PR #2433, opened 2015-09-01
[testing/feedback requested] New package: hhvm

jprjr:
This is my attempt at packaging HHVM for musl, off the bat there's a few things I still need to do:

* create a service directory + run script
* setup a default config file

Also it looks like upstream's had a new version released in the time it's taken me to do this.

I'd love it if people could look over this! As far as I can tell, this is the first (public) attempt at packaging HHVM for a musl distro.


Gottox:
That's a quite massive patch and I never worked with hhvm. Does hhvm provide some kind of unit testing so that we can verify that we haven't broke something?

jprjr:
@Gottox Honestly I'm not sure - I'm only just now getting into HHVM myself so I'm not *too* familiar with it yet

Vaelatern:
Has anything changed in the past half year for this PR?

jprjr:
I'll go ahead and close this for now, and re-open when I've got more time to dedicate to it.

PR #2439, opened 2015-09-02
New package: mimms-3.2

beefcurtains:
Downloads mms(h) streams to a wmv file


xtraeme:
This package does not seem to build any python C module:

- missing `noarch=yes`
- `makedepends` is unnecessary

beefcurtains:
Thanks, fixed!

PR #2440, opened 2015-09-02
New package: surfraw-2.2.9

evandeaubl:


pullmoll:
It looks like this package is just (perl) scripts? If this is true, it should be `noarch=yes`.

evandeaubl:
It's pretty much all shell scripts actually, but perl is a dependency because a number of the scripts use it for parsing. I was going to check the manual for a way to specify noarch, and then promptly forgot. :stuck_out_tongue:

pullmoll:
You're quite good at writing templates actually. It looks :ok: to me now.

evandeaubl:
Thanks, appreciate the comment. Compared to writing RPM spec files, xbps templates are heaven. :smiley:

PR #2446, opened 2015-09-03
New package: xfce4-verve-plugin-1.1.0

o1ux:


srcpkgs/xfce4-verve-plugin/template
```diff
@@ -0,0 +1,13 @@
+# Template file for 'xfce4-verve-plugin'
+pkgname=xfce4-verve-plugin
+version=1.1.0
+revision=1
+build_style=gnu-configure
+hostmakedepends="pkg-config intltool glib-devel dbus-glib-devel"
+makedepends="exo-devel xfce4-dev-tools libxfce4ui-devel libxfce4util-devel xfce4-panel-devel"
```
xtraeme: `xfce4-dev-tools` must be in `hostmakedepends` instead.

xtraeme: And looks like this is not even required, you can remove it.

PR #2447, opened 2015-09-03
New package: reiserfsprogs-3.6.24

o1ux:


Gottox:
Would be great, if you manage to fix reiserfsprogs on musl too.

pullmoll:
The first fix for musl libc is rather easy. Add a patch like `musl-__compar_fn_t.patch`
```
Define _GNU_SOURCE for the declataion of loff_t in fcntl.h
Add missing definition of __compar_fn_t and compare_fn_t for
non __GLIBC__ case.

--- include/misc.h      2013-08-26 22:55:43.000000000 +0200
+++ include/misc.h      2015-09-03 19:44:24.636873860 +0200
@@ -12,6 +12,7 @@
 #  include "config.h"
 #endif

+#define _GNU_SOURCE
 #include <stdio.h>
 #include <unistd.h>
 #include <sys/types.h>
@@ -34,6 +35,11 @@

 #define INVAL_PTR      (void *)-1

+#if !defined(__GLIBC__)
+typedef int (*__compar_fn_t) (const void*, const void*);
+typedef __compar_fn_t comparison_fn_t;
+#endif
+
 void check_memory_msg(void);
 void die (char * fmt, ...) __attribute__ ((format (printf, 1, 2)));
 void * getmem (int size);
```

The second thing however is `<printf.h>` is not available for musl libc.
Thus either a `musl-printf` package containing the headers and source for the printf stuff would have to be written, or alternatively the entire `reiserfscore/prints.c` needs a rewrite for non `__GLIBC__`, which is tough!

o1ux:
I can not fix reiserfsprogs on musl

pullmoll:
I now packaged obstack (another missing dependency) in a standalone `musl-obstack` package for Void. My try to implement `reiserfscore/prints.c` without `/usr/include/printf.h` is not yet verified, though. It's just that it compiles without error :smile: 

Making reiserfsprogs build with musl libc required a rewrite of the handling of the custom format specifiers and delegating the regular specifiers to musl libc's fprintf. I may have written complete bullshit in there...

Thus [my branch](https://github.com/voidlinux/void-packages/compare/master...pullmoll:reiserfsprogs?expand=1) is really only a first step <b>which needs thorough testing</b>.

@xtraeme or @Gottox: IMHO this pr could be merged now and we can try to make it build and work with musl libc later on.

xtraeme:
@pullmoll :+1: thanks

PR #2448, opened 2015-09-03
Update libreoffice-bin to 5.0.1

mid-kid:


mid-kid:
I have no idea what's wrong with travis, but it works just fine here.

pullmoll:
The travis-ci error is just because of the size. See the last lines of the raw log file:
```
Done. Your build exited with 0.


The log length has exceeded the limit of 4 Megabytes (this usually means that test suite is raising the same exception over and over).

The build has been terminated.
```

mid-kid:
I thought that meant something happened, but it wasn't logged.

PR #2451, opened 2015-09-04
New package: myrepos

Grauwolf:


srcpkgs/myrepos/template
```diff
@@ -0,0 +1,13 @@
+pkgname=myrepos
+version=1.20150503
+revision=1
+noarch=yes
+build_style=gnu-makefile
+makedepends="perl"
```
xtraeme: `makedepends` -> `hostmakedepends` to unbreak cross compilation.

PR #2459, opened 2015-09-05
grub: Add subpackage build grub-i386-efi

evandeaubl:
I tested on a Bay Trail stick PC, and it worked without incident. It should build on either i386 or x86_64, so I put it outside the _GRUBX64 block.


evandeaubl:
It looks like the two x86_64 builds are failing because the build logs are exceeding 4MB (unfortunately for reason, because it's doing 3 configure/make cycles for each build). Any suggestions on quieting down the logs without losing important information?

pullmoll:
Just leave it as is. If it worked for you, it should work on http://build.voidlinux.eu/
Travis-CI is limited and we should not care.

PR #2460, opened 2015-09-05
distcc: can't distribute cpp; add cross-gcc

pullmoll:
Distributing cpp is not possible (see man page example for
symbolic links), thus the symlink from cpp to distcc was wrong.

On the other hand it is very well possible to distribute cross
compiler jobs. For this to work the cross compiler must be
available on the slave host(s) as well.

### DIFF
--- a/srcpkgs/chroot-distcc/template
+++ b/srcpkgs/chroot-distcc/template
@@ -1,7 +1,7 @@
 # Template file for 'chroot-distcc'
 pkgname=chroot-distcc
 version=3.2rc1
-revision=3
+revision=4
 wrksrc="distcc-${version}"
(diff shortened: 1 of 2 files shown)


pullmoll:
The failure on travis-ci is probably due to not cleaning the `/builddir/distcc-3.2rc1` directory and `distcc` being compiled with `makedepends="…popt-devel…"` but `chroot-distcc` with `configure_args=`…-with-included-popt…`.
Compiling `chroot-distcc` worked on my box.

xtraeme:
Yes, this is the reason. packages built on travis use `xbps-src -C`, this does not clean `wrksrc`.

PR #2463, opened 2015-09-05
Libedit

thypon:
Shameless enable libedit instead of readline.


pullmoll:
`s;SIGCLD;SIGCHLD;` for musl

pullmoll:
BTW: SIGCHLD is ok for glibc as well, so no need to do it just for musl.
Ideally a patch to suggest for upstream would be fine :*)

xtraeme:
And the reason to use libedit is?

Gottox:
Maybe it's a war on GPL/GNU software?

thypon:
no

chneukirchen:
NACK on rc, it breaks filename completion iirc.

xtraeme:
There's no need to change this for a gratuitous breakage just because readline has some issues.

PR #2464, opened 2015-09-06
New package: perl-DBD-mysql-4.032

diogoleal:


srcpkgs/perl-DBD-mysql/template
```diff
@@ -0,0 +1,16 @@
+# Template build file for 'perl-DBD-mysql'.
+pkgname=perl-DBD-mysql
+version=4.032
+revision=1
+wrksrc="DBD-mysql-${version}"
+build_style=perl-module
+hostmakedepends="perl perl-DBI zlib-devel libressl-devel"
+makedepends="${hostmakedepends} libmariadbclient-devel"
```
xtraeme: `libmariadbclient-devel` -> `libmysqclient-devel`

pullmoll:
I think [this template](http://sprunge.us/cdAT) could be ok. It compiles native and cross, where cross shows similar warnings to `perl-DBD-SQLite` regarding integer constants and shift amounts... not sure if they are really important.

PR #2465, opened 2015-09-06
New package: strdup-0.0.1

diogoleal:
I think that ```do_install()``` is not good.


chneukirchen:
What the hell is this good for?

Gottox:
strdup is part of all supported libcs. No need for that package.

diogoleal:
This strdup is for replace the strdup from libc. 
I'm sending this template because [clib](https://github.com/clibs/clib) need this strdup.

PR #2466, opened 2015-09-06
ClusterSSH + perl module dependencies.

enguerrand:
Some of the perl modules turned out to actually not be required, but maybe they are still useful to others...


xtraeme:
Thanks.

I really doubt some packages are `noarch`, and most likely if you are including `-devel` packages via `makedepends` those dependencies in `depends` are also unnecessary.

Please recheck!

enguerrand:
I got an additional package (gthumb) that I'd like to submit. Should I just push to master to update this pull request or should I wait until this one is processed?

xtraeme:
Create a branch in your fork and then push off it.

PR #2467, opened 2015-09-06
profile-sync-daemon: bump to v6.00

graysky2:
Note that I need some input from your guys to force a stop of the currently running psd BEFORE version 6 gets installed and again if the user removes it from the system.

Here is what I do for Arch Linux:
```
pre_upgrade() {
 if [ $(vercmp $2 6.00) -lt 0 ]; then
   echo 'ATTENTION: MAJOR CHANGES TO PSD WITH VERSION 6.00+'
   echo '-> 1. A global /etc/psd.conf is no longer used. $HOME/.psd/psd.conf will be'
   echo '->    created when psd is invoked the first time.'
   echo '-> 2. A system service is no longer used. A user service is provided and can be'
   echo '->    used like this:  systemctl --user start psd.service'
   echo '-> 3. Users wanting to use overlayfs mode MUST have sudo access to both mount'
   echo '->    and umount. See the man page for an example configured with visudo.'

  systemctl is-active psd.service &>/dev/null
  if [ $? -eq 0 ]; then
   systemctl stop psd.service &>/dev/null
  fi
 fi

}

pre_remove() {
 for i in $(users); do
  running="$(su $i -c 'XDG_RUNTIME_DIR=/run/user/$UID systemctl --user is-active psd')"
  if [[ "$running" = "active" ]]; then
   echo "--> In order to preserve your profiles, pacman will now stop your psd service."
   echo "--> Any running and managed browsers will be exited."
   su $i -c 'XDG_RUNTIME_DIR=/run/user/$UID systemctl --user stop psd.service'
  fi
 done
}
```


xtraeme:
We do not start/stop services on package installations/upgrades.

srcpkgs/profile-sync-daemon/template
```diff
@@ -1,15 +1,15 @@
 # Template file for 'profile-sync-daemon'
 pkgname=profile-sync-daemon
-version=5.75
+version=6.00
 revision=1
 noarch=yes
-depends="rsync"
+depends="rsync systemd"
```
Gottox: there is no systemd in voidlinux ;)

graysky2:
Without stopping the service manually, profiles will be renamed on the filesystem and unless the user knows about it and to rename them back, they will get angry.

xtraeme:
In that case I'd use something like this in `INSTALL`:

```
case "$ACTION" in
post)
    if [ -x usr/bin/sv ]; then
      sv check psd >/dev/null && sv restart psd;;
    fi
esac
```

That means that if the `psd` service is running it will be restarted.

graysky2:
...your comment that void does not user systemd means psd will not work on void.  Version 6 has been redesigned to use systemd --user features.

xtraeme:
```
29.07.2014
runit enabled by default

The runit init system is now enabled by default, replacing systemd in the base system.
```

So this PR is really useless and we should just keep this version.

graysky2:
Yes, without systemd I believe you should keep version 5.75.

PR #2468, opened 2015-09-06
New package: mon-1.2.3

diogoleal:


xtraeme:
Compilation does not use `$CFLAGS` nor `$LDFLAGS`

xtraeme:
Still not enough...

PR #2469, opened 2015-09-06
mcomix: Added python-setuptools to depends

mid-kid:
pkg_resources, a module imported from /usr/bin/mcomix, is from python-setuptools.


xtraeme:
revision must be bumped.

mid-kid:
Done.

PR #2471, opened 2015-09-07
New package: git-annex-5.20150824

evandeaubl:


chneukirchen:
This is ok now, but in the future, please use the latest version of `lts`, which would be `3.4` right now.

PR #2472, opened 2015-09-07
vpnc: revert part of musl patch

abuibrahim:
commit 1779dabd7d1ac181f2fee102f55e9f8360c33006 causes:

vpnc: can't initialise tunnel interface: No such file or directory


Gottox:
@abuibrahim and I talked about that issue this morning. I'm okay with this change. :+1: 

vpnc has different platform dependant implementations of tun_open, which cares about creating a tun/tap device. To determine if the current platform is linux, it uses `... #elif defined(IFF_TUN) ...`. IFF_TUN is only defined in /usr/include/linux.

I suspect, that musl suffers from the same issue. Can someone verify this?

PR #2477, opened 2015-09-07
soulseekqt: update to 2015625

jarbowski:


jarbowski:
Hi,
soulseekqt did not update, it's still the 20141130 version in the repository.
what happened?

pullmoll:
You did use a version number with a 1 digit month, while the previous had 2 digits. Now `2015625 <  20141130` and thus the version did not increase. I'll fix it accordingly.

PR #2481, opened 2015-09-08
gflags: added arguments for configure

diogoleal:


Gottox:
Please add a -devel subpackage containing the header files and the static libraries.

diogoleal:
added!

pullmoll:
Not everything that belongs into the -devel subpackage is moved there. Add these lines:
```
    vmove /usr/lib/cmake
    vmove /usr/include
```

PR #2485, opened 2015-09-08
Llvm 3.7.0

thypon:
I'll rebase in a single commit.

Btw lld seems not to be compilable anymore in tree. I have to test if can be compiled outside without problems.


thypon:
Broken on musl and armv7hf

xtraeme:
Not enough to merge, an update must build in the same architectures than the previous pkg was built for.

thypon:
Working on it. Unneeded since mesa-11.0.

thypon:
x86_64-musl bug was trivial. i686 and armv7 seems subtle.

- i686 can't find libcompiler-rt.a
- armv7 loops forever when it tries to compile the Verifier

thypon:
Ready for merge. Armv7hf fails on travis-ci but succeeds on a real buildserver (the problem is in an interactive timeout with travis-ci).

xtraeme:
Great! well done :+1: 

Thanks

PR #2486, opened 2015-09-09
acpid: generate user env variables

thypon:


xtraeme:
What's this useful for?

thypon:
```
gattuso@PensieroProfondo:~ $ cat /etc/zzz.d/suspend/lock.sh                                                                                                   
#!/bin/sh
su $USER -c xtrlock &
```

It permits to execute lock screen or graphical application on behalf the current xorg server. Wouldn't arm non interactive users.

xtraeme:
:+1:

PR #2487, opened 2015-09-09
Use proper release flags to build Rust

radare:
Related https://github.com/rust-lang/rust/issues/28317


radare:
Done in https://github.com/voidlinux/void-packages/pull/2490

PR #2490, opened 2015-09-09
Use --release-channel=stable for rust

radare:


radare:
Should be fixed in next release

radare:
https://github.com/rust-lang/rust/issues/28322#issuecomment-139614600

PR #2491, opened 2015-09-09
conky: update to 1.10.0

steski:


steski:
Looks like it's failing because of locale. All is good over here. Is this just a travis thing?

pullmoll:
Yes, the failure seems to be due to locale. We should probably set `LC_ALL=C` for travis-ci.

xtraeme:
Can you please rebase?

srcpkgs/conky/template
```diff
@@ -1,22 +1,42 @@
...
-distfiles="${SOURCEFORGE_SITE}/$pkgname/$pkgname-$version.tar.gz"
-checksum=1de754d77b2671f68795fdb9fc506328a1668dd6ca00984c3f49790a66d368ca
+distfiles="https://github.com/brndnmtthws/${pkgname}/archive/v${version}.tar.gz"
+checksum=f392f3f91fcc37f04ff65d53957512e4d81b7b9b36394acb5ecaefe92623f8eb
 
-pre_configure() {
-	libtoolize -f
+post_extract() {
+	case "$XBPS_TARGET_MACHINE" in
+	*-musl)
```
xtraeme: I'd like to avoid the need to use `$FILESDIR` for patches.

Patches should be applied in all variants, so that those patches should do the right thing with glibc and musl.

This is what the `patches` dir is for.

srcpkgs/conky/template
```diff
@@ -1,22 +1,42 @@
...
- --enable-imlib2 --enable-lua --enable-wlan"
-hostmakedepends="libtool pkg-config docbook2x"
+version=1.10.0
+revision=1
+build_style=cmake
+configure_args="
+ -DCMAKE_BUILD_TYPE=Release -DMAINTAINER_MODE=ON -DDOC_PATH=share/${pkgname}
+ -DBUILD_X11=ON -DBUILD_CURL=ON -DBUILD_XDBE=ON -DBUILD_RSS=ON -DBUILD_WEATHER_METAR=ON
+ -DBUILD_WEATHER_XOAP=ON -DBUILD_IMLIB2=ON -DBUILD_WLAN=ON"
+hostmakedepends="cmake libtool pkg-config docbook2x man-db"
```
xtraeme: Does it really need `man-db` to build?

srcpkgs/conky/template
steski: Right.. I'll make a proper patch.

srcpkgs/conky/template
steski: If IIRC, configure complains without it. I'll look into that.

PR #2492, opened 2015-09-10
Adding xdebug PHP debugger and profiler

asafo:


srcpkgs/xdebug/template
```diff
@@ -0,0 +1,25 @@
+# Template file for 'xdebug'
+pkgname=xdebug
+version=2.3.3
+revision=1
+build_style=gnu-configure
+hostmakedepends="php-devel autoconf"
```
xtraeme: Move `php-devel` to `makedepends` and then remove `php`.

srcpkgs/xdebug/template
```diff
@@ -0,0 +1,25 @@
...
+
+
+pre_configure()
+{
+	phpize
+}
+
+do_install()
+{
+	cd ${wrksrc}
```
xtraeme: You can remove this line, this is done already.

srcpkgs/xdebug/template
```diff
@@ -0,0 +1,25 @@
+# Template file for 'xdebug'
+pkgname=xdebug
+version=2.3.3
+revision=1
+build_style=gnu-configure
+hostmakedepends="php-devel autoconf"
+makedepends="php"
+short_desc="PHP Debugger"
+maintainer="Asaf Ohayon <asaf.tvn@gmail.com>"
+license="The Xdebug License, version 1.01"
```
xtraeme: Custom license but you are not installing it with `vlicense`, please fix.

srcpkgs/xdebug/template
```diff
@@ -0,0 +1,25 @@
...
+homepage="http://xdebug.org"
+distfiles="http://xdebug.org/files/$pkgname-$version.tgz"
+checksum="b27bd09b23136d242dbc94f4503c98f012a521d5597002c9d463a63c6b0cdfe3"
+
+
+pre_configure()
```
xtraeme: Please change functions to use our common style:

```
func() {
   ...
}
```

srcpkgs/xdebug/template
asafo: Thanks xtraeme, don't understand this comment, what is vlicense ?, how to fix

pullmoll: See https://github.com/voidlinux/void-packages/blob/master/Manual.md#global_funcs at the end. `vlicense <file>` is used to install something like `COPYING`, `LICENSE` or similar. You have to determine the actual `<file>` which contains the license, or even have to extract it from some source file.

PR #2493, opened 2015-09-10
reiserfsprogs: try to fix musl libc

pullmoll:
The modified source based on reiserfscore/prints.c is included in
files/prints.c to make it easily possible to apply fixes and create
a new diff against the original.

First test on a Void Linux musl box seem to indicate success.
There is still a chance of the prints.c conversion missing some aspects.

### DIFF
--- a/srcpkgs/reiserfsprogs/files/musl-configure_ac.patch
+++ b/srcpkgs/reiserfsprogs/files/musl-configure_ac.patch
@@ -1,0 +1,16 @@
+Add a library check for musl-obstack package
+
+--- configure.ac	2013-08-26 22:55:43.000000000 +0200
++++ configure.ac	2015-09-10 09:21:53.809280964 +0200
+@@ -29,6 +29,11 @@
+ UUID_LIBS="$LIBS"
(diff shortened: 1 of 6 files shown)


pullmoll:
In the repo http://muslrepo.voidlinux.de/current/reiserfsprogs there's a `reiserfsprogs` package for a Void Linux musl system for testing purposes. If someone wants to give it a try, please do and report problems and bugs.

pullmoll:
As it is now, mkreiserfs segfaults and the output is not what is to be expected:
```
[root@box ~]# mkreiserfs /dev/loop0
mkreiserfs 3.6.24

Guessing about desired format.. Kernel s is running.
Format 3.6 with standard journal
Count of blocks on the device: u
Number of blocks consumed by mkreiserfs formatting process: u
Blocksize: d
Hash function used to sort names: s
Journal Size u blocks (first block u)
Journal Max transaction length u
inode generation number: u
Segmentation fault
```
The `prints.c` code is obviously crap :grimacing: Need to fix, refactor, rethink... whatever.

PR #2496, opened 2015-09-10
c-client: correct library symlink

jprjr:


srcpkgs/c-client/template
```diff
@@ -26,6 +26,7 @@ do_build() {
 
 do_install() {
 	install -D -m755 c-client/libc-client.so.1.0.0 $DESTDIR/usr/lib/libc-client.so.1.0.0
+	ln -sf /usr/lib/libc-client.so.1.0.0 $PKGDESTDIR/usr/lib/libc-client.so.1
```
xtraeme: - Please use relative symlinks.
- License must be set to `Apache-2.0` instead.

Thanks

jprjr:
Fixed + rebased

Issue #2461, opened 2015-09-05
bug: faad2 + firefox/icecat on musl


thypon:
On facebook pages containing a/v firefox/icecat segfaults when they are in libfaad2:

This is the gdb exit location. With firefox:
```
Program received signal SIGSEGV, Segmentation fault.
[Switching to LWP 2179]
ps_decode (ps=0x55555d356c80, X_left=X_left@entry=0x7fffdbe63c10, X_right=X_right@entry=0x7fffdbe68810) at ps_dec.c:1965
```

With icecat:
```
Program received signal SIGSEGV, Segmentation fault.
[Switching to LWP 1817]
ps_decode (ps=0x95cf080, X_left=X_left@entry=0x7fffdd917c00, X_right=X_right@entry=0x7fffdd91c800) at ps_dec.c:1965
```

pullmoll:
I didn't even know you could specifiy `C` function parameters with two dimensions. Line 1965 of `libfaad/ps_dec.c`
```
uint8_t ps_decode(ps_info *ps, qmf_t X_left[38][64], qmf_t X_right[38][64])
```

Perhaps it's simply the stack which overflows or is corrupted. The allocation size of X_left and X_right is each `38*64*sizeof(real_t)` or even `38*64*sizeof(complext_t)`.
If `real_t` is float (4 bytes), which seems to be the case, this is 9728 or even 19456 bytes.
For both arrays that`s 19456 or 38912 bytes...

You might try to statically allocate `X_left` and `X_right` in `libfaad/sbr_dec.c` lines 605, 606, i.e:
```
     static ALIGN qmf_t X_left[38][64];
     static ALIGN qmf_t X_right[38][64];
     /* must set this to 0 */
     memset(X_left, 0, sizeof(X_left));
     memset(X_right, 0, sizeof(X_right));
```

And ps_decode() allocates another chunk of 4096 or 8192 bytes stack. Perhaps it may be changed similarly — always under the assuption those arrays don't have to be reentrency safe.
```
    static qmf_t X_hybrid_left[32][32] = {{0}};
    static qmf_t X_hybrid_right[32][32] = {{0}};
    memset(X_hybrid_left, 0, sizeof(X_hybrid_left));
    memset(X_hybrid_right, 0, sizeof(X_hybrid_right));
```

pullmoll:
Do you have an example download which causes the crash?

thypon:
[...]

PR #2500, opened 2015-09-11
faad2: try to fix crash in icecat/firefox

pullmoll:
Try to fix issue #2461


thypon:
wow fixes the bugs.

pullmoll:
Fine! So my guess was right.

thypon:
i actually don't get why this bug presents only on musl.

pullmoll:
Perhaps the code is actually buggy and writes beyond the margins of the array? On the stack this easily causes faults, in static memory it may write in other others which goes unnoticed.

I'd try to run firefox under valgrind with the debug binaries to see if it reports anything in libfaad2.

PR #2501, opened 2015-09-11
mopidy: update to 1.1.0

diogoleal:


pullmoll:
Please rebase to see if it builds for arm*-musl now that gst-python is fixed.

PR #2502, opened 2015-09-12
New package: frei0r-plugins-1.4

diogoleal:


srcpkgs/frei0r-plugins/template
```diff
@@ -0,0 +1,21 @@
...
+build_style=gnu-configure
+makedepends="autoconf"
+short_desc="Minimalistic API for a collection of free video effects plugins"
+maintainer="Diogo Leal <diogo@diogoleal.com>"
+license="GPL-3"
+homepage="http://frei0r.dyne.org/"
+distfiles="https://files.dyne.org/frei0r/releases/frei0r-plugins-${version}.tar.gz"
+checksum=8470fcabde9f341b729be3be16385ffc8383d6f3328213907a43851b6e83be57
+
+frei0r-plugins-devel_package() {
```
xtraeme: A package that **does not** contain headers and **static/shared library** is not a development package.

That means that this `-devel` subpkg is unnecessary.

diogoleal:
-devel subpkg removed.

PR #2515, opened 2015-09-13
New package: elixir-1.1.0beta

manveru:


srcpkgs/elixir/template
```diff
@@ -0,0 +1,24 @@
+# Template file for 'elixir'
+homepage="http://elixir-lang.org"
+license="Apache-2.0"
+maintainer="Michael 'manveru' Fellinger <mf@seitenschmied.at>"
+pkgname=elixir
```
xtraeme: `pkgname`, `version` and `revision` must be the first three variables appearing in a template.

srcpkgs/elixir/template
```diff
@@ -0,0 +1,24 @@
...
+version=1.1.0beta
+revision=1
+short_desc="Dynamic, functional language for scalable and maintainable applications"
+
+hostmakedepends="erlang git"
+depends="erlang>=17.0"
+distfiles="https://github.com/elixir-lang/elixir/archive/v1.1.0-beta.tar.gz"
+checksum="a506907137dc2432bf40f3c6a86ca807af5bf9d7f3a3efd05fed14e5267beb79"
+wrksrc="elixir-1.1.0-beta"
+only_for_archs="x86_x64 i686 armv6hf armv7hf"
```
xtraeme: There's no reason to force `only_for_archs`, remove this.

srcpkgs/elixir/template
```diff
@@ -0,0 +1,24 @@
...
+
+do_build() {
```
xtraeme: Remove this `LC_ALL` export, it's unnecessary.

srcpkgs/elixir/template
manveru: Then the build fails on the other architectures, I already tried without when I first made the PR.

srcpkgs/elixir/template
manveru: Then the build will complain, since LC_ALL is C in the chroot.

srcpkgs/elixir/template
manveru: How come none of the docs or even xlint mention that requirement?

srcpkgs/elixir/template
xtraeme: If cross compilation fails there's no reason to force a specific list of architectures, use `nocross=yes` instead.

srcpkgs/elixir/template
xtraeme: That was fixed long ago, it's unnecessary.

xtraeme:
Duplicate of #2839

PR #2519, opened 2015-09-13
New package: burp 1.4.40

yopito:
backup and restore program with a client/server architecture. 
Storage is disk based, simplier than bacula.
Home Page: http://burp.grke.net

### DIFF
--- a/srcpkgs/burp-server/template
+++ b/srcpkgs/burp-server/template
@@ -1,0 +1,66 @@
+# Template file for 'burp-server'
+_desc="A network-based backup and restore program"
+
+pkgname=burp-server
+version=1.4.40
+revision=1
+short_desc="${_desc} - Server"
+maintainer="Pierre Bourgin <pierre.bourgin@free.fr>"
+license="AGPL-3, BSD, GPL-2.1 and LGPL-2.1"
+homepage="http://burp.grke.org/"
+wrksrc="burp-${version}"
+distfiles="https://github.com/grke/burp/archive/${version}.tar.gz>${pkgname}-${version}.tar.gz"
+checksum=2e6a9a28453a11f3e36d0beefa185e72e7781a8718b55d3101144c9900752d6f
+
+hostmakedepends="automake libtool perl"
+# note: burp provides its internal copy of "uthash-devel"
+makedepends="
+ libressl-devel zlib-devel librsync-devel ncurses-devel
+ acl-devel yajl-devel"
+
+# unique binary for both client and server: server depends on client package.
+depends="burp-client-${version}_${revision}"
+
+build_style=gnu-configure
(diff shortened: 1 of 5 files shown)


skrzyp:
Please rebase :)

srcpkgs/burp-server/template
```diff
@@ -0,0 +1,90 @@
...
+build_style=gnu-configure
+configure_args="--sysconfdir=/etc/burp
+--sbindir=/usr/bin"
+conf_files="/etc/burp/burp-server.conf"
+
+post_install() {
+	vsv ${pkgname}
+
+	# same files than client package
+	vdoc "README"
```
xtraeme: These files, except `README` and `UPGRADING` are just irrelevant for us, please drop them.

srcpkgs/burp-server/template
```diff
@@ -0,0 +1,90 @@
...
+	sed -e "s,/usr/sbin/burp_ca,/usr/bin/burp_ca,g" -i ${DESTDIR}/etc/burp/burp-server.conf
+
+	# storage folder
+	touch -f ${PKGDESTDIR}/var/spool/burp/.empty_on_purpose
+}
+
+burp-client_package() {
+	short_desc="A network-based backup and restore program - client"
+	# openssl binary needed by burp_ca script
+	depends="libressl-openssl"
```
xtraeme: `libressl-openssl` -> `openssl` instead (it's a virtual package).

srcpkgs/burp-server/template
```diff
@@ -0,0 +1,90 @@
...
+	conf_files="/etc/burp/burp.conf"
+	pkg_install() {
+		vmove "usr/bin/burp"
+		vmove "usr/bin/burp_ca"
+		vmove "usr/share/man/man8/burp.8.gz"
+		vmove "usr/share/man/man8/burp_ca.8.gz"
+		vmove "etc/burp/CA-client"
+		vmove "etc/burp/burp.conf"
+		vdoc "README"
```
xtraeme: Ditto, do not install useless files unnecessarily.

yopito:
Thanks for your inputs.
Regarding "useless files": documentation is sometime difficult to find with burp, that's why I push hard to keep it into packages ...
Alternative: create a dedicated sub-package -doc ?

xtraeme:
Seriously there's no need for a `-doc` subpkg to contain these files... they should not be shipped by the binary packages.

yopito:
OK ... I will strip down my package and rebase it as requested.

yopito:
On 09/17/15 07:16, Juan RP wrote:
> Merged #2519 <https://github.com/voidlinux/void-packages/pull/2519>.

Thanks !
My first package ;)
Pierre

PR #2525, opened 2015-09-15
New package: font-hack-otf-2.013

diogoleal:


srcpkgs/font-hack-otf/template
```diff
@@ -0,0 +1,24 @@
+# Template file for 'font-hack-otf'
+pkgname=font-hack-otf
+version=2.013
+revision=1
+wrksrc=Hack-${version}
+noarch="yes"
+short_desc="A typeface designed for source code"
```
xtraeme: This has the same description than `font-hack-ttf`, maybe explain what's the difference briefly?

PR #2527, opened 2015-09-15
New package: trollock-4.0

thypon:


xtraeme:
Would be nice to have a manpage :-)

PR #2530, opened 2015-09-15
New package: davfs2-1.5.2

o1ux:


srcpkgs/davfs2/template
```diff
@@ -0,0 +1,28 @@
...
+homepage="http://savannah.nongnu.org/projects/davfs2"
+distfiles="${NONGNU_SITE}/${pkgname}/${pkgname}-${version}.tar.gz"
+checksum=be34a19ab57a6ea77ecb82083e9e4c1882e12b2de64257de567ad5ee7a17b358
+patch_args="-p1"
+
+pre_configure() {
+	aclocal -I config
+	autoheader
+	automake --add-missing --force-missing --copy
+	autoconf
```
xtraeme: You can probably replace these lines with just `autoreconf -fi`.

o1ux:
```
pre_configure() {
	autoreconf -fi
}
```

```
=> davfs2-1.5.2_1: running pre_configure ...
/usr/bin/autopoint: line 454: /usr/share/gettext/archive.dir.tar.xz: No such file or directory
tar: This does not look like a tar archive
tar: gettext-0.18.1: Not found in archive
tar: Exiting with failure status due to previous errors
autopoint: *** infrastructure files for version 0.18.1 not found; this is autopoint from GNU gettext-tools 0.19.5.1
autopoint: *** Stop.
autoreconf: autopoint failed with exit status: 1
```

xtraeme: That means that you need to add `gettext-devel` to `hostmakedepends`.

srcpkgs/davfs2/template
```diff
@@ -0,0 +1,25 @@
+# Template file for 'davfs2'
+pkgname=davfs2
+version=1.5.2
+revision=1
+build_style=gnu-configure
+configure_args="dav_user=nobody dav_group=network --enable-largefile"
+hostmakedepends="autoconf automake gettext-devel"
```
Gottox: add `neon-devel` to hostmakedepends. This should fix cross-build :+1:

PR #2539, opened 2015-09-16
New package: gthumb-3.4.0

enguerrand:


srcpkgs/gthumb/template
```diff
@@ -0,0 +1,13 @@
+# Template build file for 'GThumb'.
+pkgname=gthumb
+version=3.4.0
+revision=1
+wrksrc="${pkgname}-${version}"
```
xtraeme: This `wrksrc` is redundant, this is what it's set to by default.

srcpkgs/gthumb/template
```diff
@@ -0,0 +1,13 @@
...
+build_style=gnu-configure
+makedepends="glib-devel pkg-config intltool itstool gnome-desktop-devel"
```
xtraeme: Change this `makedepends` to:

```
hostmakedepends="pkg-config intltool itstool glib-devel"
makedepends="gnome-desktop-devel"
```

srcpkgs/gthumb/template
```diff
@@ -0,0 +1,13 @@
...
+short_desc='An image viewer and browser for the GNOME Desktop'
+maintainer="Enguerrand de Rochefort <voidlinux@rochefort.de>"
+homepage="http://live.gnome.org/gthumb"
+license="GPL-3"
+distfiles="$GNOME_SITE/${pkgname}/3.4/${pkgname}-${version}.tar.xz"
```
xtraeme: Change `3.4` to `${version%.*}`.

enguerrand:
Updated as per your comments. (Sorry I could have figured out some of this myself)

xtraeme:
- This needs a `-dev` subpackage with headers, static libs and pkg-config files.
- This is missing additional dependencies for more features: exiv2, clutter, etc.

dominikh:
@enguerrand Please address xtraeme's last comment if you're still interested in adding this package. Otherwise I'll have to close the PR due to lack of progress.

enguerrand:
Hi Dominik

I apologize for the slow progress. I have been working on this, but at some point I got stuck and was reluctant to reach out to you guys for help because I felt I had not come very far. 
I have given it another try today and I now managed to get the optional features to build. However, the resulting package won't install with a message saying
"ERROR: gthumb-3.4.0_1: SHA256 hash is not valid: Numerical result out of range"
so this is still something I need to debug. Any pointers are welcome.

Anyway. I updated my PR with what I currently have. You'll find that the dev-subpackage is also still missing and I do not expect to solve this quickly. 

Long story short: I am still interested in submitting gthumb but progress will continue to be slow due to my lack of knowledge.

If under these circumstances you prefer to simply close the PR that is fine with me... I guess I can always just open a new one if I manage to figure this out later down the road.

enguerrand:
Please discount the comment on the package not installing due to a checksum error. I don't know what it was but I was able to get rid of this problem by setting up void-packages from scratch again. I must have messed it up somehow. 
I'll now look into the development subpackage...

enguerrand:
Ok. So here is what I have. Let me know what you think...

srcpkgs/gthumb/template
```diff
@@ -0,0 +1,36 @@
...
+desc_option_gnomekeyring="Enable use of the gnome keyring"
+desc_option_brasero="Enable burning discs"
+build_options_default="exiv tiff clutter gstreamer gnomekeyring brasero"
+
+gthumb-devel_package() {
+	short_desc+=" - development files"
+	depends="${sourcepkg}>=${version}_${revision}"
+	pkg_install() {
+		vmove usr/lib/gthumb/extensions/*.a
+		vmove usr/lib/gthumb/extensions/*.so
```
dominikh: I suspect that extensions/*.so should be in the main package and that gthumb will need those to actually load the extensions.

Files that still need to be moved from the main package to -devel are:

- usr/include/

srcpkgs/gthumb/template
```diff
@@ -0,0 +1,36 @@
...
+maintainer="Enguerrand de Rochefort <voidlinux@rochefort.de>"
+homepage="http://live.gnome.org/gthumb"
+license="GPL-3"
+distfiles="$GNOME_SITE/${pkgname}/${version%.*}/${pkgname}-${version}.tar.xz"
+checksum=abd1afe457c65289af31c6db7037452faa3838356480857d8cdc19f0ecde4933
+#
+## Package build options
+build_options="exiv tiff clutter gstreamer soup gnomekeyring brasero"
+desc_option_exiv="Enable support for image meta data manipulation"
+desc_option_tiff="Enable support for tiff"
```
dominikh: desc_option_tiff and desc_option_gstreamer already exist in common/options.descriptions – as such, you don't need to include these two in the template.

srcpkgs/gthumb/template
```diff
@@ -0,0 +1,36 @@
+# Template build file for 'GThumb'.
+pkgname=gthumb
+version=3.4.0
+revision=1
+build_style=gnu-configure
+LDFLAGS+=" -fPIC "
+configure_args="$(vopt_with exiv) $(vopt_with tiff) $(vopt_with clutter) $(vopt_with gstreamer)  $(vopt_with soup)  $(vopt_with gnomekeyring)  $(vopt_with brasero)"
```
dominikh: Please wrap long lines. Consider 80-90 characters per line the soft limit.

dominikh:
It's looking fine overall. I commented on the parts that need improvement. I'll defer to @xtraeme for a final look over the template, but you're getting close.

enguerrand:
Hi Dominik,

thanks a lot for taking the time to review again and for the helpful comments, it is much appreciated. 
I updated the PR accordingly...

PR #2540, opened 2015-09-16
new package: font-awesome

meowstars:
Install the font font awesome, an iconic font.

Hi,

This is my first package. I got the inspiration from font-sourcecodepro package.
I hope it is good. If not I am ready to change my template to make it perfect :)

Thank you!

### DIFF
--- a/srcpkgs/font-awesome/template
+++ b/srcpkgs/font-awesome/template
@@ -1,0 +1,19 @@
+# Template file for 'font-awesome'
+pkgname=font-awesome
+version=4.4.0
+revision=1
+noarch=yes
+wrksrc=Font-Awesome-${version}
+depends="font-util xbps-triggers>=0.58"
+font_dirs="/usr/share/fonts/OTF"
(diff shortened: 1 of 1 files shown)


steski:
Looks good. I would use the tags from github for the <code>distfiles</code>. For example:
 
<code>distfiles="https://github.com/FortAwesome/Font-Awesome/archive/v${version}.tar.gz"</code>

 This would remove <code>unzip</code> for <code>hostmakedepends</code> and <code>xbps-src update-check font-fontawesome</code> will be able to find an update.

diogoleal:
I think the package name could be ```font-awesome``` and not ```font-fontawesome```

meowstars:
@diogoleal actually, I think that the name of the font is font awesome that's why I named the package like this.

@steski OK, I will change my source for the github tags :)

meowstars:
I have changed the distfiles source for github tags, as suggested.

xtraeme:
Please rename it to `font-awesome`.

PR #2543, opened 2015-09-17
mariadb: fix musl

steski:


steski:
Well, AFAICT, <code>x86_64-musl</code> and <code>armv7hf</code> are both working. I'm gonna do some more testing. <code>armv*-musl</code> on the otherhand... I don't know WTF is going on there.

PR #2549, opened 2015-09-17
New packages: pps-tools-20120407 / gpsd-3.15

uriah0:
pps-tools is required for gpsd's pulse per second capabilities.


uriah0:
So, for some reason gpsd builds natively but can't seem to locate python-devel header files when cross-compiling... Any hints?

srcpkgs/gpsd/template
```diff
@@ -0,0 +1,108 @@
...
+				vmove usr/lib/libQgpsmm.so.* ;;
+		esac
+	}
+}
+
+gpsd-qt-devel_package() {
+	short_desc+=" - Qt development files"
+	depends="qt-devel ${sourcepkg}-devel>=${version}_${revision}"
+	pkg_install() {
+		case "$XBPS_TARGET_MACHINE" in
```
xtraeme: This is not the proper way to disable building certain subpackages for musl.

`qt` cannot be cross compiled so that you'd need to disable `qt` support and the `subpackage` like this:

```
subpackages="gpsd-devel gpsd-python ..."
    subpackages+=" gpsd-qt gpsd-qt-devel"
fi
...
```

srcpkgs/gpsd/template
```diff
@@ -0,0 +1,108 @@
...
+
+pre_build() {
+	unset CPPFLAGS
+}
+
+pre_install() {
+	export DESTDIR=${DESTDIR}
+}
+
+do_install() {
```
xtraeme: There's no need to duplicate `do_install` just to use the `udev-install` target.

Remove `do_install` and add `udev-install` to `$make_install_target`

srcpkgs/gpsd/template
```diff
@@ -0,0 +1,108 @@
...
+patch_args=-p1
+make_build_args="dbus_export=0 gpsd_user=gpsd gpsd_group=gpsd sbindir=/usr/bin udevdir=/usr/lib/udev"
+make_install_args="${make_build_args}"
+short_desc="A GPS/AIS receiver monitoring daemon"
+maintainer="uriahheep <uriahheep@gmail.com>"
+license="BSD"
+homepage="http://www.catb.org/gpsd/"
+distfiles="http://download-mirror.savannah.gnu.org/releases/gpsd/gpsd-${version}.tar.gz"
+checksum="81c89e271ae112313e68655ab30d227bc38fe7841ffbff0f1860b12a9d7696ea"
+system_groups="gpsd"
```
xtraeme: This is completely redundant, simply using `system_accounts=gpsd` should be really enough, because by default a group with the same name is also set as its primary group.

srcpkgs/gpsd/template
```diff
@@ -0,0 +1,99 @@
...
+homepage="http://www.catb.org/gpsd/"
+distfiles="http://download-mirror.savannah.gnu.org/releases/gpsd/gpsd-${version}.tar.gz"
+checksum="81c89e271ae112313e68655ab30d227bc38fe7841ffbff0f1860b12a9d7696ea"
+system_accounts="gpsd"
+
+depends="virtual?ntp-daemon ncurses libusb libcap libbluetooth pps-tools"
+hostmakedepends="scons pkg-config bc python libxslt xmlto asciidoc"
+makedepends="libudev-devel ntp libusb-devel python-devel pygtk-devel ncurses-devel pps-tools-devel libcap-devel libbluetooth-devel"
+
+case "$XBPS_TARGET_MACHINE" in
```
xtraeme: This is still wrong, you should set `make_build_args` and `makedepends` in the !cross conditional below.

srcpkgs/gpsd/template
```diff
@@ -0,0 +1,97 @@
...
+maintainer="uriahheep <uriahheep@gmail.com>"
+license="BSD"
+homepage="http://www.catb.org/gpsd/"
+distfiles="http://download-mirror.savannah.gnu.org/releases/gpsd/gpsd-${version}.tar.gz"
+checksum="81c89e271ae112313e68655ab30d227bc38fe7841ffbff0f1860b12a9d7696ea"
+system_accounts="gpsd"
+
+depends="virtual?ntp-daemon ncurses libusb libcap libbluetooth pps-tools"
+hostmakedepends="scons pkg-config bc python libxslt xmlto asciidoc"
+makedepends="libudev-devel ntp libusb-devel python-devel pygtk-devel ncurses-devel pps-tools-devel libcap-devel libbluetooth-devel"
```
xtraeme: - You can remove `ntp`, it's really unnecessary.
- You can remove `python-devel` and `pygtk-devel`.
- Is `pps-tools-devel` really necessary?

srcpkgs/gpsd/template
```diff
@@ -0,0 +1,97 @@
...
+make_install_target="udev-install"
+short_desc="A GPS/AIS receiver monitoring daemon"
+maintainer="uriahheep <uriahheep@gmail.com>"
+license="BSD"
+homepage="http://www.catb.org/gpsd/"
+distfiles="http://download-mirror.savannah.gnu.org/releases/gpsd/gpsd-${version}.tar.gz"
+checksum="81c89e271ae112313e68655ab30d227bc38fe7841ffbff0f1860b12a9d7696ea"
+system_accounts="gpsd"
+
+depends="virtual?ntp-daemon ncurses libusb libcap libbluetooth pps-tools"
```
xtraeme: You don't need to specify dependencies manually via `depends` for shared libraries, it's detected by `xbps-src` automagically.

Change this line to:

```
depends="virtual?ntp-daemon pps-tools"
```

uriah0: Oh, in that case I'm not even certain ntp / pps-tools are necessary deps. pps-tools-dev definitely is for building, though, as timepps.h is required to activate pulse-per-second capabilities in gpsd (needed to run a Stratum-1 ntp server)

srcpkgs/gpsd/template
```diff
@@ -0,0 +1,96 @@
...
+	pkg_install() {
+		vmove usr/bin/xgps
+		vmove usr/bin/xgpsspeed
+		vmove usr/share/man/man1/xgps*
+	}
+}
+
+gpsd-qt_package() {
+	short_desc+=" - Qt bindings"
+	depends="qt ${sourcepkg}>=${version}_${revision}"
```
xtraeme: There's no need to depend on qt because the shared libraries are already linked to qt.

srcpkgs/gpsd/template
```diff
@@ -0,0 +1,96 @@
...
+
+hostmakedepends="scons pkg-config bc python libxslt xmlto asciidoc"
+makedepends="libudev-devel ntp libusb-devel ncurses-devel pps-tools-devel libcap-devel libbluetooth-devel"
+
+subpackages="gpsd-devel"
+if [ -z "$CROSS_BUILD" ]; then
+	subpackages+=" gpsd-python gpsd-xgps gpsd-qt gpsd-qt-devel"
+	makedepends+=" qt-devel python-devel"
+else
+	make_build_args+=" python=0"
```
xtraeme: You can also disable `qt` here with `qt=0`.

PR #2552, opened 2015-09-18
New Package: xinput_calibrator 0.5.7

the-maldridge:
Very useful package for calibrating touchscreen devices.

### DIFF
--- a/srcpkgs/xinput_calibrator/template
+++ b/srcpkgs/xinput_calibrator/template
@@ -1,0 +1,17 @@
+# Template build file for 'xinput_calibrator'.
+pkgname=xinput_calibrator
+version=0.7.5
+revision=1
+build_style=gnu-configure
+hostmakedepends="pkg-config"
+makedepends="libXi-devel"
+short_desc="Utility to calibrate X input devices"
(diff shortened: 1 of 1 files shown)


srcpkgs/xinput_calibrator/template
```diff
@@ -0,0 +1,19 @@
+# Template build file for 'xinput_calibrator'.
+pkgname=xinput_calibrator
+version=0.7.5
+revision=2
```
xtraeme: Please set revision=1 for new packages and updates.

srcpkgs/xinput_calibrator/template
```diff
@@ -0,0 +1,19 @@
...
+build_style=gnu-configure
+configure_args="--prefix=/usr"
```
xtraeme: `--prefix=/usr` is set by default with `build_style=gnu-configure`

srcpkgs/xinput_calibrator/template
```diff
@@ -0,0 +1,19 @@
...
+hostmakedepends="pkg-config"
+makedepends="libXi-devel"
+depends="xinput"
```
xtraeme: Does it really depend on the `xinput` command line utility?

PR #2554, opened 2015-09-18
ntp: add pps-tools-devel makedepends to try and get ATOM PPS working

uriah0:
Needed for setting up a Stratum-1 ntp server...


xtraeme:
Please bump revision

PR #2565, opened 2015-09-19
New package: arpack-ng-3.2.0

diogoleal:


Gottox:
crossbuild is broken, add `nocross=yes`.

PR #2566, opened 2015-09-20
New package: csync-0.50.0

diogoleal:


srcpkgs/csync/template
```diff
@@ -0,0 +1,22 @@
...
+license="GPL-2"
+homepage="http://csync.org"
+distfiles="http://git.csync.org/projects/csync.git/snapshot/csync-0.50.0.tar.bz2"
+checksum=72e95f6924ca01783afcfc9c68d416a50e0bec979cf4e1fc414df352210d6a0f
+
+csync-devel_package() {
+	short_desc+=" - development files"
+	depends="${sourcepkg}>=${version}_${revision}"
+	pkg_install() {
+		vmove usr/lib/*.so
```
xtraeme: A development package without headers does not make sense at all...

common/shlibs
```diff
@@ -2150,3 +2153,17 @@ libnewt.so.0.52 newt-0.52.18_1
...
+libmemcachedutil.so.2 libmemcached-1.0.18_1
+libc-client.so.1 c-client-2007f_1
+libonig.so.2 oniguruma-5.9.6_1
+liblo10k1.so.0 alsa-tools-1.0.29_1
+libgflags.so.2 gflags-2.1.2_1
+libgflags_nothreads.so.2 gflags-2.1.2_1
+libswmhack.so.0.0 spectrwm-2.7.2_1
+libchardet.so.1 libchardet-1.0.4_1
+libcollectdclient.so.1 libcollectdclient-5.5.0_1
+libcsync.so.0 csync-0.50.0_1
```
xtraeme: No development pkg, no need to modify `common/shlibs`.

xtraeme:
You've added just the shared library symlink previously to the development package (*.so), but that was incomplete.

As I said in the outdated comments, a development package without headers is not useful.

In this case, you'd have to add a development package containing:

- headers (usr/include)
- static libs (usr/lib/*.a)
- shared library symlinks (usr/lib/*.so)
- pkg-config files (usr/lib/pkgconfig)

And this development pkg must **depends** on **csync** because this is what provides the real **shared library** to link against.

xtraeme:
sysconfdir is not set correctly:

/home/travis/build/voidlinux/void-packages/masterdir/destdir//csync-0.50.0/usr/etc/csync/csync.conf

/home/travis/build/voidlinux/void-packages/masterdir/destdir//csync-0.50.0/usr/etc/csync/csync_exclude.conf

PR #2568, opened 2015-09-20
New package: rclone-1.20

diogoleal:


xtraeme:
No executables are installed to DESTDIR...

PR #2570, opened 2015-09-20
New package: goomwwm-1.0.0

diogoleal:


srcpkgs/goomwwm/template
```diff
@@ -0,0 +1,21 @@
+# Template file for 'goomwwm'
+pkgname=goomwwm
+version=1.0.0
+revision=1
+build_style=gnu-makefile
+hostmakedepends="pkg-config"
+makedepends="xorg-server-devel libXft-devel libXinerama-devel"
```
xtraeme: `xorg-server-devel` is not the right dependency to use.

diogoleal:
```xorg-server-devel``` removed from makedepends

**Comment by diogoleal** [comment]
```xorg-server-devel```removed from makedepends

File: srcpkgs/goomwwm/template [hunk 2]
```diff
@@ -0,0 +1,21 @@
...
+homepage="https://github.com/seanpringle/goomwwm"
+distfiles="https://github.com/seanpringle/goomwwm/archive/${version}.tar.gz"
+checksum=952be1f588e01850175387d534faa39ea27e1196cd7473d8e4c2ea5e038a6b72
+
+do_install(){
+	vbin goomwwm
+	vdoc goomwwm.md
+	vman goomwwm.1
+	vlicense license.txt
+	vinstall goomwwm.desktop 644 usr/share/applications/
```
xtraeme: This `.desktop` file is for an X session... and shall be installed to `/usr/share/xsessions` instead.

diogoleal:
Changed for /usr/share/xsessions

PR #2571, opened 2015-09-20
hedgewars: add libfreeglut as depend to fix hwengine

skrzyp:
`hwengine` is not dynamically linked against libfreeglut, but depends on it.


xtraeme:
revision must be bumped...

PR #2573, opened 2015-09-20
Armv5 musl

thypon:
base-system built with these mods.


srcpkgs/cross-arm-linux-musleabi/template
```diff
@@ -0,0 +1,223 @@
...
+_triplet=arm-linux-musleabi
+_fpuflags="--with-arch=armv5te --without-fp --with-float=soft"
+_archflags="-march=armv5te -msoft-float -mfloat-abi=soft"
+_sysroot="/usr/${_triplet}"
+
+
+pkgname=cross-${_triplet}
+version=0.20
+revision=1
+short_desc="Cross toolchain for ARMv6 LE Hard Float target (musl)"
```
xtraeme: Please update the description.

common/cross-profiles/armv5te-musl.sh
```diff
@@ -0,0 +1,6 @@
+# Cross build profile for ARM GNU EABI5 Soft Float.
```
xtraeme: Please update this comment.

srcpkgs/musl/template
```diff
@@ -1,7 +1,7 @@
 # Template file for 'musl'.
 pkgname=musl
 version=1.1.11
-revision=2
+revision=3
```
xtraeme: Please do not bump those packages that were modified just for `only_for_archs`.

srcpkgs/cross-arm-linux-musleabi/template
```diff
@@ -0,0 +1,223 @@
...
+
+	# install musl libc for target
+	cd ${wrksrc}/musl-${_musl_version}
+	make DESTDIR=${DESTDIR}/${_sysroot} install
+
+	# Remove useless headers.
+	rm -rf ${DESTDIR}/usr/lib/gcc/${_triplet}/*/include-fixed/
+
+	# Make ld-musl-armhf.so.1 symlink relative.
+	ln -sf libc.so ${DESTDIR}/${_sysroot}/usr/lib/ld-musl-armhf.so.1
```
xtraeme: I suspect this symlink is also wrong, because it's soft float.

xtraeme: Note the comments of the gcc-4.9.2-musl.patch:

```+/* For ARM musl currently supports four dynamic linkers:
+   - ld-musl-arm.so.1 - for the EABI-derived soft-float ABI
+   - ld-musl-armhf.so.1 - for the EABI-derived hard-float ABI
+   - ld-musl-armeb.so.1 - for the EABI-derived soft-float ABI, EB
+   - ld-musl-armebhf.so.1 - for the EABI-derived hard-float ABI, EB
```

xtraeme: The target symlink should be `ld-musl-arm.so.1` instead for EABI soft-float.

srcpkgs/gcc/template
```diff
@@ -2,7 +2,7 @@
 pkgname=gcc
 _majorver=4.9
 version=${_majorver}.3
-revision=2
+revision=3
```
xtraeme: Please undo this revision bump.

xtraeme:
:+1:

PR #2578, opened 2015-09-21
New package: wiggle-1.0

okapia:
wiggle is like patch but will apply patches that patch would reject.

### DIFF
--- a/srcpkgs/wiggle/template
+++ b/srcpkgs/wiggle/template
@@ -1,0 +1,16 @@
+# Template file for 'wiggle'
+pkgname=wiggle
+version=1.0
+revision=1
+wrksrc="${pkgname}-v${version}-b4a87cb"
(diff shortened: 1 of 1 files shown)


chneukirchen:
Please disable running of the test suite, it will not work for cross-builds. (make_build_args="... wiggle" should do.)

Also the STRIP= is superfluous, we strip ourselves anyway.

okapia:
Thanks. I had wondered what was wrong with the ARM builds.

Assigning something to STRIP is necessary because xbps-src provides STRIP=yes, overriding the value in wiggle's Makefile which was passed on to the install command. "yes" is obviously not a valid option to install. I've changed it from -s to the empty string.

chneukirchen:
Good, please squash this together, then we can merge.

okapia:
Now squashed.

PR #2582, opened 2015-09-22
wvstreams: add musl support

uriah0:
I'm pretty sure that some of these patches are inadequate, but it builds, so I'd like to know how I can make them better please. Thanks.

### DIFF
--- a/srcpkgs/wvstreams/patches/wvstreams-4.6.1-argp-fpic.patch
+++ b/srcpkgs/wvstreams/patches/wvstreams-4.6.1-argp-fpic.patch
@@ -1,0 +1,28 @@
+Build argp object files with -fPIC.
+
+The argp bundled dependency is linked to the libwvutils.so shared library.
+MIPS will refuce to link a non-PIC library with a shared one.
+
+We fix this problem by building the argp source files using -fPIC since
+the libargp static library will only be used to link to libwvutils.so
+as instructed by the following rule in the root Makefile:
+
+ifeq ($(USE_WVSTREAMS_ARGP),1)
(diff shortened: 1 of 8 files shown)


pullmoll:
I think `xplc/modulemgr.cc` should unconditionally `#include <limits.h>` which defines `PATH_MAX`. This should work with glibc also.

The execinfo.h / backtrace() patches look ok to me, especially since they rely on `HAVE_EXECINFO_H`.

Not sure about the getcontext(3) and setcontext(3) hackery, which seems to disable everything context related. There is an include file `/usr/include/ucontext.h` in musl libc. Did you try to explicitly include it in `utils/wvtask.cc`? As long as the internals of `struct __ucontext` are not touched - which is always a bad idea and doesn't seem to be the case - including ucontext.h might be sufficient.

uriah0:
Thanks, I'll try #include <limits.h> for PATH_MAX.

As for ucontext.h, I did try to include it, to no avail. I still got the same errors about getcontext/setcontext being undefined. I read somewhere (probably here: https://github.com/9fans/plan9port/issues/36 ) that those functions aren't actually implemented, which is why I did what I did. Any thoughts?

pullmoll:
I didn't know the function declarations are not implemented anywhere. The discussion you linked to talks about implementing the functions should be easy, so perhaps we can conceive a `musl-context` package, like I did in similar cases `musl-fts` and `musl-obstack`.

For now I think disabling the context stuff should be ok. I of course have no idea on the impact of this on the application itself.

uriah0:
pullmoll, made the suggested change to xplc/modulemgr.cc, it succeeds on all tests.

pullmoll:
:+1:

PR #2587, opened 2015-09-23
New package: synfig-1.0.1

diogoleal:


xtraeme:
- This needs a development package.
- Missing build dependencies: libmng-devel, libjpeg-turbo-devel.

pullmoll:
Hmm.. it doesn't build with the current dependencies environment anymore: [build log for x86_64](http://sprunge.us/RaGW). I am not sure where to look for the cause, though.

Perhaps it's due to `libsigc++` not putting a blank between the nested, double `>>` for template type casts, which would make the declarations less ambiguous. The non-spaced `>>` may perhaps be misinterpreted as right shift `>>` by the C++ parser!?

Our `libgsigc++` was updated to `2.6.1` on `2015-09-30`, after the latest travis-ci build of synfig-1.0.1.

PR #2588, opened 2015-09-23
New package: flare-engine-0.20

diogoleal:


pullmoll:
You should set the binary install directory to `/usr/bin`, not `/usr/games` as it is now.
The package files should go to `/usr/share/flare` or perhaps `/usr/share/flare-engine`, not `/usr/share/games/flare`.

It is probably just a matter of specifying the corresponding cmake defines in `configure_args="..."`.

PR #2599, opened 2015-09-25
New Package: frotz 2.43

the-maldridge:
frotz plays games written for the Z-Machine, this includes games such as Zork and others.


srcpkgs/frotz/template
```diff
@@ -0,0 +1,14 @@
+# Template build file for 'xinput_calibrator'.
+pkgname=frotz
+version=2.43
+revision=1
+build_style=gnu-makefile
+make_install_args="-e \"PREFIX=/usr\""
+makedepends="ncurses-devel"
+depends="ncurses"
```
xtraeme: This line is really unnecessary

xtraeme:
Package files are being installed directly to the masterdir, rather than ${DESTDIR}.

Please fix!

xtraeme:
The build still has a wrong path for configuration files:

```
-DCONFIG_DIR="\"/usr/local/etc\""
```

PR #2614, opened 2015-09-27
emacs: update to 24.99

zamnedix:
Updated to musl-compatible Emacs build based on Alpine Linux port [http://git.alpinelinux.org/cgit/aports/plain/testing/emacs]


srcpkgs/emacs/template
```diff
@@ -1,9 +1,11 @@
 # Template file for 'emacs'
 pkgname=emacs
-version=24.5
+version=24.99
 revision=1
+_commit="7eb2077f9dd14ae9d70a12c621151afd50ea6e96"
 nocross=yes
-hostmakedepends="pkg-config"
+noarch=yes
```
xtraeme: why noarch?

chneukirchen:
-1 for updating to a development snapshot.

We could make an emacs-beta package perhaps.

zamnedix:
I thought "noarch" meant it could be built by all supported architectures.
Development version because the code for musl compatibility isn't in a
stable release yet.

srcpkgs/emacs/template
zamnedix: It was a misunderstanding, I'll correct it soon.

Gottox:
@chneukirchen: I don't think we should provide "-beta" packages either.

zamnedix:
The Travis builds seem to be failing due to an issue with exec shield. The GNU emacs build process is very fickle. The way around this in the emacs documentation [https://github.com/emacs-mirror/emacs/blob/master/etc/PROBLEMS#L2552] is to modify kernel parameters to disable exec shield and ASLR during the build process. This solution doesn't seem satisfactory to me but I'm not sure what else to do. I've been able to build the package successfully on my x86_64-musl Void install. You can see it here: http://sleepermud.net/~zamnedix/emacs-unstable/

dominikh:
Should we close this PR? Consensus seems to be that we don't want a dev snapshot nor a beta package. 

The musl patches could be ported to the current emacs release, though.

chneukirchen:
The musl patches wont work on latest stable relase, but freeze for 25.1 will be on Nov 13, so I expect a release this year.

Gottox:
@chneukirchen Is this obsolete?

zamnedix:
Looks like it 
https://forum.voidlinux.eu/t/emacs-available-soon-with-musl-its-now-possible/299

PR #2619, opened 2015-09-27
Gnome 3.18

Gottox:
Some cleanups needs to be done, but anyway it's ready for review.


Gottox:
I'm using Gnome-3.18 for two days now. Works fine so far.

Also I checked the build for i686, works there too.

xtraeme:
:+1:

Gottox:
I found a minor issue with gnome-shell. Please wait till I sort this out.

pullmoll:
gobject-introspection doesn't build anymore with musl libc.

I checked and also the previous version 1.44.0 won't build anymore, so the origin of the problem is elsewhere. I am clueless where to look, though.

PR #2624, opened 2015-09-27
New package: setterm-0.8

diogoleal:


xtraeme:
```
[juan@leysa void-packages]$ xbps-query -o /usr/bin/setterm
util-linux-2.27_1: /usr/bin/setterm (regular file)
```

already provided by util-linux... why do you need this?

diogoleal:
A friend on work use this package. but really, this is not necessary. because exist a aplication with this name. 

Sorry!

PR #2627, opened 2015-09-28
New package: avidemux-2.6.10

### LINKED ISSUE #2626: Package Request: Avidemux
http://www.avidemux.org/

pullmoll:
Closes #2626

Disabled MesaLib, vdpau and libva for now.

I couldn't get vdpau/libva to work right. I could make avidemux link
against the packages, but there was no video frame image displayed in
the UI when the output was set to va or vdpau.

Perhaps I was just missing some additional dependency.

### DIFF
--- a/srcpkgs/avidemux/template
+++ b/srcpkgs/avidemux/template
@@ -1,0 +1,36 @@
+# Template file for 'avidemux'
+pkgname=avidemux
+version=2.6.10
+revision=1
+wrksrc="${pkgname}_${version}"
+hostmakedepends="cmake pkg-config qt5-qmake yasm"
+# XXX: MesaLib-devel, libva-devel, libvdpau-devel, etc. and make display output work
+makedepends="faac-devel faad2-devel gettext-devel jack-devel
+ lame-devel libvorbis-devel libXv-devel sqlite-devel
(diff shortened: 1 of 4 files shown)


srcpkgs/avidemux/template
```diff
@@ -0,0 +1,36 @@
+# Template file for 'avidemux'
+pkgname=avidemux
+version=2.6.10
+revision=1
+wrksrc="${pkgname}_${version}"
+hostmakedepends="bash cmake pkg-config qt5-qmake yasm"
```
xtraeme: bash is unnecessary... the masterdir uses bash by default.

pullmoll:
How would I avoid the plugins being stripped? It seems there could be some copyright and version info which the plugins display expects to be there. The Help -> Plugins widget displays strange results as it is now.

xtraeme:
`nostrip` (globally) or `nostrip_files` (per file, expects basename).

pullmoll:
Okay, I'll add this later today, so please don't merge yet.

pullmoll:
`nostrip=yes` doesn't make a difference, so it looks like the copyright holder display for the plugins is just plain broken. Every plugin has `(c) Mean` and some year(s) in the info. WTF.

PR #2628, opened 2015-09-28
manual: descriptions for -data and -doc pkgs

pullmoll:


xtraeme:
Looks good, thanks.

My only concer is the confusion with `-doc` and `-docs`. I believe that we should just choose one and fix the existing ones.

I vote for `-doc`.

pullmoll:
I agree. Should I go ahead and change *-docs to *-doc in the voidlinux/void-packages repo?

xtraeme:
Go ahead :+1:

pullmoll:
What about `gnome-gettings-started-docs` and `gnome-user-docs`? They are standalone packages, not subpackages.

xtraeme:
Those pkgs are fine as is.

xtraeme:
:+1:

PR #2630, opened 2015-09-28
New package: python-setuptools_scm-1.8.0

okapia:


srcpkgs/python-setuptools_scm/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'python-setuptools_scm'
+pkgname=python-setuptools_scm
+version=1.8.0
+revision=1
+noarch=yes
+wrksrc=setuptools_scm-${version}
+build_style=python-module
+hostmakedepends="python python-setuptools"
+depends="python"
```
xtraeme: This is missing `pycompile_module="${pkgname#python-}"`

xtraeme:
Why do you need this one?

@asergi removed it few days ago:

2015-09-13 13:00 Alessio Sergi         ●  python-setuptools_scm: removed; not needed anymore

Assigning to @asergi

okapia:
Ultimately, I want to have vdirsyncer and khal working:
vdirsyncer appears to require it - see https://github.com/untitaker/vdirsyncer/blob/master/setup.py

asergi:
`setuptools_scm` aims to "simplify" the life of the developer, but it has no value for us.
Anyway, you have two choices here:
- do as I did for python-path: https://github.com/voidlinux/void-packages/blob/master/srcpkgs/python-path/template, or
- if you're lazy enough, here's the correct template: https://github.com/voidlinux/void-packages/commit/db26adc3947f97176e5d21e28a5523c392fa6388. Just update it accordingly.

Make your choice and enjoy.

okapia:
Thanks. I've been able to get the first of the two choices to work for vdirsyncer. It seems there are further Python dependencies to resolve before I have it working however so no pull-request for that just yet.

PR #2643, opened 2015-09-30
New package: trinity-1.5

diogoleal:
When I use ```build_style=gnu-makefile```, occurs this errors:

```
=> trinity-1.5_1: running pre-build hook: 02-script-wrapper ... 
=> trinity-1.5_1: running do_build ... CC  arg-decoder.o arg-decoder.c:5:30: fatal error: arch.h: No such file or directory #include "arch.h" //PAGE_MASK ^ compilation terminated. Makefile:97: recipe for target 'arg-decoder.o' failed make: *** 
[arg-decoder.o] Error 1
```

### DIFF
--- a/srcpkgs/trinity/patches/arch.h.patch
+++ b/srcpkgs/trinity/patches/arch.h.patch
@@ -1,0 +1,6 @@
+--- include/arch.h	2015-10-02 13:44:46.653183687 -0300
++++ include/arch.h.new	2015-10-02 13:44:38.297184059 -0300
+@@ -1,3 +1,4 @@
++#include <stdint.h>
+ #pragma once
+ #include "types.h"
--- a/srcpkgs/trinity/template
+++ b/srcpkgs/trinity/template
@@ -1,0 +1,20 @@
+# Template file for 'trinity'
+pkgname=trinity
+version=1.5
+revision=1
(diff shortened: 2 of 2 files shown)


pullmoll:
You should probably use `build_style=gnu-configure` and `configure_script=configure.sh` if this is anywhere near to a classic configure script (i.e. if it understands the standard `configure_args`).

For musl libc you could patch `include/arch.h` to use `uint32_t` and `uint64_t` for the typedefs of `u32` and `u64` and most probably `#include <stdint.h>` at the top.

pullmoll:
I tried to fix musl support myself, but besides the types `__uint32_t` and `__uint64_t`, which can be replaced with the `<stdint.h>` types, there are some more things I could not fix:
+ Use of GNU libc specific ftw.h enumeration values (FTW_CONTINUE, FTW_STOP, FTW_SKIP_SUBTREE) which musl libc's `ftw(3P)` doesn't implement.
+ Lack of a `<bits/sockaddr.h>` header (replacing with `<sys/socket.h>` *may* work)

I'd say this package really needs `only_for_archs="armv6l armv7l i686 x86_64"`. Not sure about `aarch64`, `armv5te` and the `mips*` architectures...

PR #2646, opened 2015-10-01
lua: update to 5.3.1

pullmoll:
+ Rename lua and lua-devel to lua52 and lua52-devel
+ Modify packages depending on lua52 accordingly:
+ Many packages auto-detect lua52 files
+ Some packages need hints where / how to find lua52
+ One package (xmoto) downgraded to lua51, because it wouldn't find lua52
+ One package (vim) upgraded to lua (5.3.1)

There may be more packages upgradable without problems.
I didn't want to change too much in this move without the
respective package maintainers first taking a look.

Most notably conky-cli wouldn't work with lua52. I upgraded it
to 1.10.0 (same as conky) and tried to keep the flags the same
for the change from gnu-configure to cmake.


Duncaen:
```
nmap-6.47_7: broken, unresolvable shlib `liblua.so.5.2'
vim-huge-7.4.865_1: broken, unresolvable shlib `liblua.so.5.2'
```

xtraeme:
```
$ xbps-install -un
libvlc-2.2.1_9: broken, unresolvable shlib `liblua.so.5.2'
Transaction aborted due to unresolved shlibs.
```

Why this PR was merged even being incomplete?

pullmoll:
Sorry. I checked it built for x86_64 locally.
After not reading a comment after ~12 hours I thought it were okay for you, because you @xtraeme once told "no comment = ok".

I guess I should have waited longer.

xtraeme:
No problem... I've been afk for some hours and couldn't review it.

Looks like vlc is the only pkg that needs to be rebuilt.

pullmoll:
*phew* /me wipes sweat off his forehead.

xtraeme:
lua52 also installs conflicting files, the manpages need to be renamed.

xtraeme:
Please fix this and bump both pkgs.

pullmoll:
Done. I think the nocross=yes from conky-cli can be removed now. I'll try locally and see if it works.

asergi:
My 2 cents:
- lua packages could have been renamed to lua5.x-foo
- the lua package itself should always be versioned: lua -> lua5.3
- lua 5.3 as default version: it could be ok, but to date very few packages are yet compatible with it (w/o patching). Imo, keep version 5.2 as default would have been a more wise choise.

In short, a missed opportunity.

xtraeme:
I agree with @asergi 

But @pullmoll was too fast to have any comment... next time please give a bit more time :-)

pullmoll:
Sorry again. I should have waited longer. I don't really know why I was so impatient. Perhaps because after several hours of working on the changes, trying and fixing and finally creating the pr, the no comment for half a day, my common sense went on vacation ;-P

PR #2647, opened 2015-10-01
New package: lsw-0.3

diogoleal:


Duncaen:
Same suggestions as in #2648

PR #2648, opened 2015-10-01
slock: fixed template 

diogoleal:
Fixed for use vlicense, make_build_args and make_install and build_style.


xtraeme:
Not sure what you've tried to achieve with these changes, but are really unnecessary.

Thanks

PR #2653, opened 2015-10-02
New package: python-atomicwrites-0.1.8

okapia:
This is one of the dependencies of vdirsyncer. I don't really need the Python 3.4 variant but I guess it is good form to build both, right?

### DIFF
--- a/srcpkgs/python-atomicwrites/template
+++ b/srcpkgs/python-atomicwrites/template
@@ -1,0 +1,32 @@
+# Template file for 'python-atomicwrites'
+pkgname=python-atomicwrites
+version=0.1.8
+revision=1
+noarch=yes
+wrksrc="atomicwrites-${version}"
+build_style="python-module"
+python_versions="2.7 3.4"
+hostmakedepends="python-setuptools python3.4-setuptools"
+depends="python"
+pycompile_module="atomicwrites"
+short_desc="Simple Python2 API for atomic file writes"
+maintainer="Oliver Kiddle <okiddle@yahoo.co.uk>"
+license="MIT"
+homepage="https://pypi.python.org/pypi/atomicwrites/"
+distfiles="${PYPI_SITE}/a/atomicwrites/atomicwrites-${version}.tar.gz"
+checksum=3274adb52bd3ae91d87ba923e0f21fd86c05b71bbaefe2ec0d6679c01f98dc8e
+
+post_install() {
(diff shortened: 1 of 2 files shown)


pullmoll:
The template is missing
```
post_install() {
        vinstall COPYING
}
```
or whatever the MIT license file name is.

And you need to create a symlink from `python-atomicwrites` to `python3.4-atomicwrites` and add the latter to the pr, too.

You should probably install the `xtools` package and check your templates locally like
```
$ xlint srcpkgs/python-atomicwrites/template
```
to see what's missing or wrong.

okapia:
Thanks. I've updated the pr including the symlink and license part.

srcpkgs/python-atomicwrites/template
```diff
@@ -0,0 +1,32 @@
+# Template file for 'python-atomicwrites'
+pkgname=python-atomicwrites
+version=0.1.8
+revision=1
+noarch=yes
+wrksrc="atomicwrites-${version}"
+build_style="python-module"
+python_versions="2.7 3.4"
+hostmakedepends="python-setuptools python3.4-setuptools"
+makedepends="${hostmakedepends}"
```
asergi: `makedepends` is unnecessary

srcpkgs/python-atomicwrites/template
```diff
@@ -0,0 +1,32 @@
...
+depends="python"
+pycompile_module="atomicwrites"
+short_desc="Simple Python2 API for atomic file writes"
+maintainer="Oliver Kiddle <okiddle@yahoo.co.uk>"
+license="MIT"
+homepage="https://pypi.python.org/pypi/atomicwrites/0.1.8"
```
asergi: homepage="https://github.com/untitaker/python-atomicwrites"

srcpkgs/python-atomicwrites/template
```diff
@@ -0,0 +1,32 @@
...
+	vlicense LICENSE
+}
+
+python3.4-atomicwrites_package() {
+	noarch=yes
+	pycompile_version="3.4"
+	pycompile_module="atomicwrites"
+	short_desc="${short_desc/Python2/Python3.4}"
+	pkg_install() {
+		vmove usr/lib/python3.4
```
asergi: vlicense LICENSE

okapia:
Thanks. I've corrected those points.

PR #2654, opened 2015-10-02
New package: radicale-1.0.1

lemmi:
first time creating system account and service files, maybe someone should have a look before merge :)


xtraeme:
Looks good :+1: 

This is only missing `pycompile_module=radicale`

PR #2662, opened 2015-10-03
New package: zynaddsubfx-2.5.1

### LINKED ISSUE #2656: Package Request: zynaddsubfx
Would be great to have this packaged if possible - http://zynaddsubfx.sourceforge.net

pullmoll:
+ Add dependency mxml-2.9 (Minimal XML library)

Closes #2656


pullmoll:
To fix arm* I would have to get rid of some compiler flags (`-msse`, `-msse2`, `-mfpmath=sse`) or replace them with arm hard-float equivalents. It's just that I don't see how to achieve that.

PR #2664, opened 2015-10-03
tint2: update to 0.12.2

Spencer-H:
update the tint2 package to 0.12.2

Switch to tint2 on gitlabs - Google Code shut down.
Remove patch, didn't compile with it.


dominikh:
Is the patch not needed anymore, or did you simply remove it so you wouldn't have to port it to the new version?

Spencer-H:
I could attempt to port the patch, but I am not familiar with what it was fixing. And looking at the tint2 code, the battery seems to have been almost completely rewritten. If I knew what this was fixing (and if it was still an issue0, I would try to port it.

I will add that the battery monitor works perfectly fine on my computer, but I as I don't know what this was fixing I don't even know if it would apply to me.

dominikh:
Looking at the code, the patch shouldn't be required anymore. At one point in time, it was part of upstream. Some time after that, they restructured the code entirely.

dominikh:
Can't build reliably because gitlab isn't generating the archive in time. Apparently they're over capacity, and the archive generation ends in a redirect loop. Will merge the PR once gitlab is more stable.

dominikh:
Merged as 1fb3dda.

PR #2666, opened 2015-10-03
New package: txr-118

okapia:


chneukirchen:
Fix for musl:

```--- configure-    2015-10-03 23:20:36.976105846 +0200
+++ configure     2015-10-03 23:21:38.708814989 +0200
@@ -2176,7 +2176,7 @@
   if conftest ; then
     printf "yes\n"
     printf "#define HAVE_ALLOCA 1\n" >> $config_h
-    printf "#define ALLOCA_H <%s>\n" $try_header >> $config_h
+    printf '#define ALLOCA_H "%s"\n' $try_header >> $config_h
     have_alloca=y
     break;
   fi
```

okapia:
Thanks!

I'm still rather puzzled as to that error with musl. This doesn't look like a patch that should be fed back upstream, right?

What does this patch change in terms of what gets included? The only alloca.h I can find is the glibc-devel one in /usr/include

chneukirchen:
Actually the patch should be upstreamed because `alloca.h` can be transformed by macros on other platforms as well, so the quoted version should be used.

For sanity, I'd rewrite the configure script to define HAVE_ALLOCA_H etc and then #ifdef appropriately if I was upstream.

PR #2669, opened 2015-10-03
New package: apache-maven-bin-3.3.3

Spencer-H:


srcpkgs/apache-maven/template
```diff
@@ -0,0 +1,21 @@
+# Template file for 'apache-maven'
+pkgname=apache-maven
+version=3.3.3
+revision=1
+depends="openjdk"
```
xtraeme: This is wrong... should be `depends="java-environment"` if this needs the JDK, or `depends="java-runtime"` if it needs JRE.

Spencer-H:
When I try to package it with either of those selected it simply spits out: 
`   [runtime] java-runtime>=0: not found. ` and then
`=> ERROR: xbps-src: unexistent file: /void-packages/srcpkgs/java-runtime/template`
I'm probably just missing something, but I'm not sure what.

thypon:
This is the apache-maven bin version. Should be named apache-maven-bin instead?

Spencer-H:
Sure that makes sense. Though It's beyond my ability to package the source one, so it would be the only one in here for now.

pullmoll:
`vmkdir usr/bin` without the leading slash is the correct syntax.

Spencer-H:
The syntax for vmkdir should now be correct.

PR #2670, opened 2015-10-04
skype: use 32 bit Qt plugins

dominikh:
Without this, Qt looks for plugins in /usr/lib/qt/plugins, which are 64
bit. In turn, Skype can't display profile pictures.

@xtraeme, is this the right fix? The same issue would apply to all multilib Qt applications. I don't think there are any at the moment, but is this something we can fix in the qt-32bit package itself?

Please don't merge until that's decided.


dominikh:
Apparently we'd need to set `-plugindir /usr/lib32/qt/plugins` for `configure` of Qt when building the -32bit package.

xtraeme:
This change is good until qt is built with `-plugindir /usr/lib32/...`.

Thanks :+1:

dominikh:
Merged as 5d5f66fed06f4c68cfc1d9805571287e3073f519. #2678 is tracking the Qt issue.

PR #2673, opened 2015-10-04
Audio Workstation Programs

thypon:
Right now ardour and guitarix yields errors about CXX11 compatibility. Maybe they expect GCC >= 5.0.


pullmoll:
Regarding cxx11 compatibility: this can be set by simply defining `CXXFLAGS="-std=c++11"` in the template. I did this for `gtkmm2` and `inkscape` which both include `libsigc++` header files and require the c++11 mode as it seems.

thypon:
Alredy done that. Thank you.

logenkain:
Perhaps it's just me, but except for chuck, pd, and sratom I can't install any of these via xbps-install.

Well I just decided to check the void package search. (http://www.voidlinux.eu/packages/)
Only checked yoshimi and guitarx.  yoshimi only showed up only as arm, guitarx didn't even show...

Of course I can build via the templates (tested this with yoshimi), but the binaries seem to be missing in the main repo.

PR #2680, opened 2015-10-04
New package: csync2-2.0

diogoleal:


pullmoll:
For musl there's probably an `#else` or `#elif` which tries to (re-)define strlcpy. If not, disable the static version by wrapping it in `#if defined(__GLIBC__)`.

PR #2692, opened 2015-10-05
New package: otto-0.1.1

diogoleal:


pullmoll:
`short_desc="…"` reminds me of song titles, where everyone uses upper case first letters for every word. And this despite the fact that english writing people don't even write any noun, except the one referring to their own person, in uppercase in a sentence `8*)`
Technically speaking I see no problem in this pr, but I'm a `*no go*` _lol_!? :+1:

diogoleal:
:dancers: 
Yep, you're right!
It was the habit of copying and pasting the description.
:dancers:

PR #2708, opened 2015-10-08
Update lsb_release

the-maldridge:
Certain tools expect the output of lsb_release to be in a specific
format.  This commit alters the output of lsb_release to use this
format such that those tools can correctly identify the system as
Void Linux.


xtraeme:
Looks good; please bump version.

PR #2710, opened 2015-10-08
New package: khal-0.6.0

okapia:
Command-line calendar that works with vdirsyncer for accessing caldav servers.

### DIFF
--- a/srcpkgs/khal/template
+++ b/srcpkgs/khal/template
@@ -1,0 +1,27 @@
+# Template file for 'khal'
+pkgname=khal
+version=0.6.0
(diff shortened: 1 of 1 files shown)


okapia:
The CI failure looks to actually be python-setproctitle. That's not an essential dependency so if it can't be fixed, it could be removed from the dependency list.

pullmoll:
Please rebase and push -f again to see if it works now that python-setproctitle is fixed for musl libc.

okapia:
That seems to have fixed it. Thanks!

PR #2712, opened 2015-10-08
freshplayerplugin: update to 0.3.3

steski:
Just FYI, I haven't tested this (couldn't get the previous versions to work).


xtraeme:
:+1:

PR #2713, opened 2015-10-08
thunderbird[-i18n]: updated to 38.3.0

yopito:
including backported patches from Firefox ESR 38.


srcpkgs/thunderbird/patches/fix-netwerk.patch
```diff
@@ -0,0 +1,38 @@
+Patch from Firefox ESR 38 Void packaging.
```
Gottox: filename wrong? s/netwerk/network/

yopito: no, patch on source file `mozilla/netwerk/sctp/src/netinet/sctp_os_userspace.h` ...

Gottox: :+1:

Gottox:
Looks good so far.

yopito:
still WIP, I'm working on it:
- small fixes in thunderbird-i18n stuff are needed: xlint, missing sv-SE language
- build with musl libc fails (mozilla/memory/mozjemalloc/)
... stay tuned :)

BTW, travis-ci stops build (and fails) since output log is 4MB+ (like x86_64/glibc) ... how to handle it ?
My personal test env is quite constrainted and slow, so I rely on it as QA tests ...

Gottox:
Yea, common error. You can ignore that.

Feel free to merge my WIP branch: https://github.com/Gottox/void-packages/blob/thunderbird-cross/srcpkgs/thunderbird/template

Gottox:
Compiling works, but I'm unable to download any messages after the update.

yopito:
Thanks @enno for your help and tests.
I will merge your WIP branch as much as I can.
Might take me several days to achieve.

TB38 binary: Any trace/output to see what's happening ?

Gottox:
@yopito: No useful output to stdout. I haven't traced it with wireshark or similiar yet. Does it work for you?

Gottox:
I experienced this bug with my branch too.

yopito:
building and running is fine for me (i686 built on x86_84), build with x86_64-musl is fine too.
before pushing my changes: 
- what kind of packaging style: like firefox-esr (with a mozconfig client) or like firefox 40, where all configure stuff is shell-scripted ?
- also need a working sub-dir as firefox 40 ?

pullmoll:
If there are chances that updating thunderbird from 31.7.0 to 38.3.0 breaks user's setup, I'd suggest to create a new package thunderbird38 instead and only after enough people tested it replace the 31.7.0 version.

For the packaging style: if you create a new package, it's your choice; otherwise you should probably keep it as it is.

I don't get the meaning of the question `also need a working sub-dir as firefox 40 ?`. Are you talking about `/usr/lib/firefox` ? If you create a new thunderbird38 package, you should probably use `/usr/lib/thunderbird38` as the plugin etc. install path to not collide with the current `/usr/lib/thunderbird`.

Just my 2 ¢

yopito:
@pullmoll : I was talking about the subdir for build (`xbps-build` stuff in firefox/template)

I've just pushed an updated version for this packaging.
my tests: built is fine for i686 from x86_64 and for x86_64-musl, runtime is ok on my i686, both with new profile or my regular profile (from TB31) so far.

@pullmoll, @Gottox : could you test ? 
if needed, I can provide a thunderbird38 package instead of an update of the thunderbird packaging.

pullmoll:
@yopito I don't use thunderbird and didn't test it, sorry.
It looks like the x86_64-musl build failed on travis-ci because of parallelism. Probably nothing to (really) worry about, though.

I'd still vote for a thunderbird38 package, unless everybody else thinks that upgrading everyone else from 31.7.0 to 38.3.0 is ok.

xtraeme:
I can't see any reason to make a new package.

Gottox:
Let me recheck if it works with a fresh account.

Gottox:
Building... -.-

Gottox:
Works if I delete my profile. I guess that's some plugin-fuckup on my local box. Anyway, my problem. Okay to merge :+1:

yopito:
@pullmoll : x86_64-musl build fails on ``UnifiedBindings21.o`` with weird error about assembler and ``g++: internal compiler error: Killed (program cc1plus)``. Can't reproduce it, probably related to Travis-ci constrained environment ...enough RAM ? Other builds fails because log size exceed 4MB ...

@Gottox : thanks for the test. On my side, upgraded with my TB profile was successfull, including the following  TB add-ons: DisplayQuota, FoxClocks, Lightgning, LookOut, Provider for Google calendar, QuoteCollapse, ToggleFolderView

PR #2717, opened 2015-10-09
New package: python-vobject-0.8.1c

okapia:
Needed for khard which is similarish to khal but for carddav rather than caldav.


okapia:
updated to use ${version} in distfiles.

srcpkgs/python-vobject/template
```diff
@@ -0,0 +1,15 @@
+# Template file for 'python-vobject'
+pkgname=python-vobject
+version=0.8.1c
+revision=1
```
stpx: `noarch=yes`

PR #2729, opened 2015-10-11
lua-MessagePack: rename to lua52-MessagePack; fix lua5.2 dependencies

stpx:
Maybe this should be renamed to `lua52-MessagePack`? Or update it to use use `lua5.3` and add a `lua5.2` subpackage, because I *do* need a `lua5.2` version.

### DIFF
--- a/srcpkgs/lua52-MessagePack/template
+++ b/srcpkgs/lua52-MessagePack/template
@@ -1,20 +1,23 @@
-# Template file for 'lua-MessagePack'
-pkgname=lua-MessagePack
+# Template file for 'lua52-MessagePack'
+pkgname=lua52-MessagePack
 version=0.3.3
-revision=1
-hostmakedepends="lua"
+revision=2
 noarch=yes
-makedepends="lua"
-depends="lua>=5.2"
+wrksrc="${pkgname/52/}-${version}"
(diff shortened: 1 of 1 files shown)


pullmoll:
IMHO renaming would make sense if this package is somehow tied to 5.2 and no update or fix for 5.3 is soon to be expected. I see no package depending on this one, so it's probably no big deal anyway.

stpx:
From http://fperrad.github.io/lua-MessagePack/:

```
It's developed for Lua 5.1, 5.2 & 5.3. There are 2 variants:

* one compatible with all interpreters since Lua 5.1
* another which uses the Lua 5.3 features
```

pullmoll:
Probably renaming to lua52-MessagePack makes sense then and is the easiest way to go for now.

stpx:
I agree

pullmoll:
I think despite the fact that no other package uses this one, it may be necessary to add a `replaces="lua-MessagePack>=0"`. I'm not entirely sure, though, so perhaps someone else can confirm?

xtraeme:
@pullmoll yes, the `replaces` line is still necessary.

stpx:
Ok. Everything else is good?

xtraeme:
Looks ok

PR #2735, opened 2015-10-12
New package: clsync-0.4.1

diogoleal:


pullmoll:
I think there should also be `makedepends="glib-devel"`. For `musl` and `fts` you can use the `musl-fts` package and add something similar to the lines in `srcpkgs/opensmtpd/template`.

Alternatively you could try to patch `configure.ac` to add a pkg-config search for musl-fts, as it has a `musl-fts.pc` file. I don't have the syntax for this ready, though.

diogoleal:
I'm moving to a new house and because of that I'm without internet access. As soon as I get to solve, I will correct these problems you indicated and synfig template.

Sorry for inconvenience

dominikh:
glib-devel should most likely be a makedepends. What's the overall state of this PR?

diogoleal:
I added ```only for archs="x86_64 i686" ```, because this does not compile on ARM architecture.

PR #2741, opened 2015-10-13
picocom: reverts to 1.8

thypon:
version 2.0 can't establish connection with devices.

Revert "picocom: use patch instead of ed call."

This reverts commit 662405406e2d4b1d415e95e3ca53be6590b3de57.

Revert "picocom: update to 2.0"

This reverts commit 313951d823609e52ab4e5aeea30f4e3dafc2454f.


xtraeme:
Would be nice to know why this does not work rather than reverting?

Also `reverts` must appear before `version` and `revision`.

thypon:
Yep, would be; the problem i can't even reproduce a working environment since 2.0 major release, neither i have the time to bisect right now. Leaving a broken package for the sake of having the latest isn't a nice thing either.

thypon:
@xtraeme maybe we should start applying musl only patches to musl build. A filter by filename should be enough.

Gottox:
I'd rather prefer clean patches that fix musl and won't break glibc.

thypon:
The place for clean patches is upstream. Btw i don't understand why we have to worry glibc battle-tested/working code with musl hacks.

PR #2743, opened 2015-10-13
New package: boinc-7.6.12

### LINKED ISSUE #2065: Package Request: BOINC
https://boinc.berkeley.edu/

pullmoll:
Closes #2065

This certainly needs review because of the system account,
config and service files in the template.

I'm not sure the idea of `/etc/sv/boinc/conf` looking for the `/etc/default/boinc-client` and then `/etc/boinc-client.conf` file is appropriate. Perhaps it'd be better to just install the boinc-client.conf as `/etc/sv/boinc/conf`?

I expect several flaws in this pr, because it is my first try to add a service.


pullmoll:
It looks like `xmindep` telling me `libjpeg-turbo-devel` was redundant isn't really true.

xtraeme:
:+1:

PR #2747, opened 2015-10-13
New package: jhead-3.00

okapia:
simple program for showing and editing JPEG EXIF data


pullmoll:
You should perhaps remove the line `	vinstall jhead.1 644 /usr/share/man/man1` and instead add
```
post_install() {
	vman jhead.1
}
```
for conformingness' sake.

okapia:
Done. Thanks.

PR #2750, opened 2015-10-14
aucatctl: fix strlcpy bsd compat

Duncaen:


xtraeme:
Note that `musl` already provides `strlcpy()`.

chneukirchen:
I think this is just an upstream packaging error.

xtraeme:
Stalled for some months, and doubtful.

PR #2755, opened 2015-10-15
[NO-MERGE] Sndio

Duncaen:
With this changes sndio works with mpv and gst-play-1.0 (and other gstreamer utils).
``
gst-play-1.0 --audiosink=sndiosink /tmp/test.wav
``
It also works without --audiosink by settings sndiosink as default with:
``
gconftool-2 -t string --set /system/gstreamer/1.0/default/audiosink sndiosink
``

Except for firefox :angry: I'm not sure where the problem is and how I can debug it.

The `gst-plugins-base1` `sndio.patch` is copied from openbsd without changes.

Update:
tried gstreamer 1.6 with the patch it works too, but firefox doesnt want to play audio with sndio.


Gottox:
status? still needed? @Duncaen

PR #2758, opened 2015-10-16
base-files: implement lsb_release options

pullmoll:
Some packages rely on FSG compliant output of lsb_release.
Implement according to http://man.cx/lsb_release for v1.4


xtraeme:
:+1:

PR #2762, opened 2015-10-17
New package: perl-AnyEvent-I3-0.16

okapia:
This is required for i3-save-tree to work which is already included as part of i3.


xtraeme:
Missing `noarch=yes`

PR #2764, opened 2015-10-18
gtk-vnc: unbreak cross

pullmoll:
Remove call to AM_CHECK_PYTHON_HEADERS from configure.ac
and instead patch Makefiles to replace @PYTHON_INCLUDES@
with the known location ${XBPS_CROSS_BASE}/usr/include/python2.7


xtraeme:
I'd change `configure.ac` instead and `autoreconf`, otherwise you'll have to update the patch on every update.

pullmoll:
I couldn't get autoreconf to work with this configure.ac and Makefile.am. I can try again and see if I can fix the problems.

PR #2765, opened 2015-10-18
rspamd: fix cross armv[67]l

pullmoll:
This fixes the [linker error](http://build.voidlinux.eu/builders/armv7l_builder/builds/18562/steps/shell_3/logs/stdio) (missing scalarmult_donna symbol),
but I don't know if it also works.


Gottox:
:+1:

pullmoll:
Hehe.. too quick for the `vlicense LICENSE` I just wanted to push :)

PR #2769, opened 2015-10-18
New package: cdm-20140805

thypon:


srcpkgs/cdm/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'cdm'
+pkgname=cdm
+version=20140805
+revision=1
+_commit=7b6c395e91aa6c43b94a77994bffd4d9c9e2cffa
+wrksrc="${pkgname}-${_commit}"
+noarch=yes
+short_desc="Console Display Manager"
+maintainer="Andrea Brancaleoni <miwaxe@gmail.com>"
+dependency="bash"
```
xtraeme: `dependency` -> `depends`

srcpkgs/cdm/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'cdm'
+pkgname=cdm
+version=20140805
```
xtraeme: Please use `version=0.0.<date>` instead.

PR #2771, opened 2015-10-19
steam: add xterm as a dependency

bougyman:


bougyman:
@xtraeme was too fast :)

PR #2779, opened 2015-10-20
New package: skroll-0.6

diogoleal:


pullmoll:
Old cgit will never twice return the same checksum.
I fetched one copy and put it on http://repo.voidlinux.de/distfiles/skroll-0.6/skroll-0.6.tar.bz2
checksum="32221fef51f6814a7bf3cf4207b8a2466624544a2fd1e5b110ee2af3580d2959"

xtraeme:
Again, this package is not installing files into DESTDIR:

```
GZ skroll.1

install -D -m 0755 skroll /usr/bin/skroll

install -D -m 0644 skroll.1.gz /usr/share/man/man1/skroll.1.gz
```

srcpkgs/skroll/template
```diff
@@ -0,0 +1,20 @@
...
+pkgname=skroll
+version=0.6
+revision=1
+short_desc="Scroll text to stdout"
+maintainer="Diogo Leal <diogo@diogoleal.com>"
+license="WTFPL"
+homepage="http://git.z3bra.org/cgit.cgi/skroll/"
+distfiles="http://repo.voidlinux.de/distfiles/skroll-${version}/skroll-${version}.tar.bz2"
+checksum=32221fef51f6814a7bf3cf4207b8a2466624544a2fd1e5b110ee2af3580d2959
+nostrip="yes"
```
xtraeme: Setting `nostrip` is hardly the correct solution.

pullmoll:
I fixed the shipped `Makefile` to allow for configurable CC and LD. This makes cross compiling etc. work as expected.

@diogoleal  You should generally try to use the build mechanisms provided by a source package. Using custom `do_build()` and `do_install()` functions is necessary only in case there is no Makefile at all. Usually it's much better to use the `build_style=gnu-makefile` and, only if required, patch missing bits and pieces.

PR #2780, opened 2015-10-20
New package: wrk-4.0.1

diogoleal:


pullmoll:
This package doesn't actually use `LuaJIT-devel` but compiles its own static `libluajit.a`.
To reduce the amount of warnings you could add
```
        sed -i Makefile -e"s;_BSD_SOURCE;_DEFAULT_SOURCE;"
```
before calling make.

diogoleal:
Hi pullmoll 

I removed Lua-JIT from makedepends. But, the archs armv7hf and armv6hf, occours this error:

```
CC src/ae.c
CC src/zmalloc.c
CC src/http_parser.c
LUAJIT src/wrk.lua
LINK wrk
obj/bytecode.o: file not recognized: File format not recognized
Makefile:39: recipe for target 'wrk' failed
make: *** [wrk] Error 1
=> ERROR: wrk-4.0.1_1: failed to run do_build() at line 19.
```

You have sugestion for solution this problem?

pullmoll:
Nope, sorry. I think we currently have the same or a similar problem in the LuaJIT package itself.
You could take a look at that package and see what variables are passed to make to correctly cross build it. Something similar may work with the static build in `wrk`, i.e. except the patches to create the dynamic LuaJIT libraries.

pullmoll:
I think I can give some more hints after looking at the source...

+ The `do_build()` is not necessary, as there is a Makefile.

    This Makefile should be patched to not force `CFLAGS :=`, but rather add to `CFLAGS +=`.

+ The `do_install()` seems to be necessary because there is no `install:` target.

+ The local luajit binary seems to be required just to compile `$(ODIR)/bytecode.o: src/wrk.lua`

    Adding the Void `LuaJIT` package in the hostmakedepends and patching the Makefile to use `luajit` instead of `./lujait` should be possible. You could then rip out of `Makefile` anything that refers to `LDIR` and `$(LDIR)/libluajit`.

I would suggest to create a patch for Makefile or even write a new one and install it from `srcpkgs/wrk/files/Makefile`, based upon the existing, to create a package whoch does not build nor require its own luajit binary.

dominikh:
What is the state of this PR?

dominikh:
Closing this PR due to a lack of progress. If you still wish to contribute this package, please address the issues and open a new PR. Thanks.

PR #2781, opened 2015-10-20
New package: chromium-widevine-1.0.0.0

TheDoctorsLife:
Since I basically used the pepper flash package as a template for this, partial credit goes to them.


srcpkgs/chromium-widevine/template
```diff
@@ -0,0 +1,55 @@
...
+	sed \
+		-e "s|%CHROME_URL%|${_chromeUrl}|" \
+		-e "s|%CHROME_CHECKSUM%|${_chromeChecksum}|" \
+			${FILESDIR}/chromium-widevine-vars.sh.in > chromium-widevine-vars.sh
+}
+
+do_install() {
+	vmkdir "usr/lib/chromium-widevine"
+
+	vinstall "${FILESDIR}/LICENSE" 644 "usr/share/licenses/chromium-widevine"
```
xtraeme: You can replace the two lines in `do_install` with just `vlicense ${FILESDIR}/LICENSE` instead.

TheDoctorsLife: Thanks, fixing that now

PR #2784, opened 2015-10-20
[NOMERGE/WIP] zsh/bash command-not-found handler

stpx:
A command-not-found handler for XBPS and a utitlity, similar to debian's `apt-file(1)`  and archlinux's `pkgfile(1)`, really cleverly named `xpkgfile`.

### DIFF
--- a/srcpkgs/xpkgfile/template
+++ b/srcpkgs/xpkgfile/template
@@ -1,0 +1,23 @@
+# Template file for 'xpkgfile'
+pkgname=xpkgfile
+version=0.1
+revision=1
+_commit=f9e6876bc66f413f1edf058ce6892aa0dc09cfff
+wrksrc="${pkgname}-${_commit}"
+makedepends="libxbps-devel"
+short_desc="XBPS package files searching utility"
+maintainer="Steve Prybylski <sa.prybylx@gmail.com>"
+license="2-clause-BSD"
+homepage="https://github.com/stpx/xpkgfile"
+distfiles="https://github.com/stpx/${pkgname}/archive/${_commit}.tar.gz"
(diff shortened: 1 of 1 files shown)


Gottox:
Nice!

iirc, @xtraeme has planned such a feature for `xbps-query -Ro`

xtraeme:
Looks good for now, thanks :-)

There are some minor memleaks I found in a quick glance:

`xbps_{array,dictionary}_get_cstring()` returns a `char *` that must be `freed`. Switch to `xbps_{array,dictionary}_get_cstring_nocopy()` instead that returns a `const char *`.

stpx:
@xtraeme @Gottox Thanks for the feeback! 
I kinda just whipped this up over the weekend before opening this PR, before doing some real testing and what not. And it has some inefficiencies, 1) Creating the initial cache consumes quite a lot of memory. I was going to work around this by shipping a pre-made cache. 2) It's *substantially* slower than `apt-file` and `pkgfile`,  which is not really ideal as a `command_not_found_handler`.
So, I've been trying a minor rewrite. 
@xtraeme Did you have anything planned in the short term?

xtraeme:
@stpx `xbps-query -Ro` is fast enough with local repos, I was thinking to generate a file in the server side and supply this into the repo. With this approach `xbps-query -Ro` would look at this file remotely (on-the-fly) without the need to download and store the file on disk.

But this won't be implemented for 0.48, maybe for 0.49.

TheDoctorsLife:
Any updates on this?

Gottox:
It's not implemented yet.

the-maldridge:
This seems like this could quite easily hook into xlocate rather than needing to maintain its own cache, perhaps that would be a decent solution?

the-maldridge:
Original maintainer has become inactive.

PR #2798, opened 2015-10-21
Soundtouch update to 1.9.2 (soname bump)

stpx:


pullmoll:
So `dolphin-emu` is trying to build a shipped version of wxWidgets. Perhaps it works with 3.0.2 from the packages now?

xtraeme:
Last time I tried it didn't build, I think they have additional changes.

stpx:
I don't even think the problem is in wxWidgets. I'm investigating right now.

pullmoll:
According to travis-ci it is wxWidgets. The non-portable `pthread_setname_np(pthread_self(), szThreadName);` is the cause. I think you can safely enclose it in `#if defined(__GLIBC__)`.

stpx:
The file with that is ./Source/Core/Common/Src/Thread.cpp

pullmoll:
Yup. Not having a name on a thread is probably no show stopper for the application.

pullmoll:
Hehe.. mame also used it in its shipped `bx` library. Upstream will fix it there, i.e. disable for non-glibc.

asergi:
Tip: to build `dolphin-emu` against our wxWidgets, just add `wxWidgets-devel` to `makedepends` and `-DwxWidgets_CONFIG_EXECUTABLE=/usr/bin/wx-config-3.0` to `configure_args`.

stpx:
```
effects/VST/VSTEffect.cpp:2579:90: error: 'RTLD_DEEPBIND' was not declared in this scope
    void *lib = dlopen((const char *)wxString(realPath).ToUTF8(), RTLD_NOW | RTLD_LOCAL | RTLD_DEEPBIND);
```

From ./src/effects/VST/VSTEffect.cpp
```c
   // Attempt to load it
   //
   // Spent a few days trying to figure out why some VSTs where running okay and
   // others were hit or miss.  The cause was that we export all of Audacity's
   // symbols and some of the loaded libraries were picking up Audacity's and 
   // not their own.
   //
   // So far, I've only seen this issue on Linux, but we might just be getting
   // lucky on the Mac and Windows.  The sooner we stop exporting everything
   // the better.
   //
   // To get around the problem, I just added the RTLD_DEEPBIND flag to the load
   // and that "basically" puts Audacity last when the loader needs to resolve
   // symbols.
   //
   // Once we define a proper external API, the flags can be removed.
   void *lib = dlopen((const char *)wxString(realPath).ToUTF8(), RTLD_NOW | RTLD_LOCAL | RTLD_DEEPBIND);
   if (!lib) 
   {
      return false;
   }
```

I'm not sure of the effects of removing this for musl builds.

stpx:
@asergi Thanks.

xtraeme:
:+1:

PR #2799, opened 2015-10-21
New package: python-lz4-0.7.0

okapia:


srcpkgs/python-lz4/template
```diff
@@ -0,0 +1,33 @@
+# Template file for 'python-lz4'
+pkgname=python-lz4
+version=0.7.0
+revision=1
+wrksrc=lz4-${version}
+build_style=python-module
+python_versions="2.7 3.4"
+hostmakedepends="python-setuptools python-devel python3.4-setuptools python3.4-devel"
+short_desc="LZ4 bindings for Python2"
```
xtraeme: You should move `python{,3.4}-devel` to `makedepends` to fix cross compilation.

srcpkgs/python-lz4/template
```diff
@@ -0,0 +1,34 @@
...
+	sed '/nose/ d' -i setup.py
+	sed -n 2,27p src/lz4.c > LICENSE
+}
+
+post_install() {
+	vlicense LICENSE
+}
+
+python3.4-lz4_package() {
+	depends="python3.4"
```
xtraeme: This depends line is unnecessary because the shared library is already linked to libpython3.4.

PR #2812, opened 2015-10-23
New package: taskd-1.1.0

stpx:


pullmoll:
For musl libc you could add [`srcpkgs/taskd/patches/musl-limits_h.patch`](http://sprunge.us/gNZF).

srcpkgs/taskd/files/taskd/run
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+[ -r conf ] && . ./conf
+TASKDDATA=${TASKDDATA:-/var/lib/taskd}
```
xtraeme: You can simplify this default assignments like this:

```
: ${TASKDDATA:-/var/lib/taskd}
...
```

xtraeme:
Looks ok, thanks :+1:

pullmoll:
I'm a bit late, as most of the times :+1:

PR #2813, opened 2015-10-23
New package: sickbeard-507

ananteris:


diogoleal:
I advise you to create branches for each new template and read this before here
https://github.com/voidlinux/void-packages/blob/master/CONTRIBUTING.md
https://github.com/voidlinux/void-packages/blob/master/Manual.md
and 
https://github.com/voidlinux/documentation/wiki/How-to-use-git,-by-@pullmoll

It helped me to understand how the question of the pull request the Void.

ananteris:
Thanks for the links. I've read these and will use branches for development from here on out. It looks like yeahwm was on my master before I locally branched and resetting it would have closed my yeahwm PR (again) which is mentioned as being a bad thing.

This should never happen again, and I can go ahead and do that if needed.

srcpkgs/sickbeard/template
```diff
@@ -0,0 +1,36 @@
+# Template build file for 'sickbeard'.
+pkgname=sickbeard
+version=507
+revision=1
+makedepends="git"
+depends="python-cheetah"
+short_desc="Usenet personal video recorder"
+maintainer="Ananteris <ananteris@mailinator.com>"
+system_groups="sickbeard"
```
xtraeme: `system_groups` is redundant because `system_accounts` already creates a group with the same name and is set as primary group.

srcpkgs/sickbeard/template
```diff
@@ -0,0 +1,36 @@
+# Template build file for 'sickbeard'.
+pkgname=sickbeard
+version=507
+revision=1
+makedepends="git"
```
xtraeme: Why `git` is required for the target? I really doubt this is necessary.

srcpkgs/sickbeard/template
```diff
@@ -0,0 +1,36 @@
...
+do_build() {
+	sed -i "1s/build Nonea/${version}/" ${wrksrc}/sickbeard/version.py
+}
+
+do_install() {
+	vmkdir usr/share/sickbeard
+	mv * ${DESTDIR}/usr/share/sickbeard
+}
+
+post_install() {
```
xtraeme: If you are using `do_install` already there's no point in using `post_install`...

ananteris:
please hold off on merging this commit.

srcpkgs/sickbeard/template
```diff
@@ -0,0 +1,31 @@
...
+depends="python-cheetah"
+short_desc="Usenet personal video recorder"
+maintainer="Ananteris <ananteris@mailinator.com>"
+system_accounts="sickbeard"
+sickbeard_homedir="/var/lib/sickbeard"
+distfiles="https://github.com/midgetspy/Sick-Beard/archive/build-${version}.tar.gz"
+checksum=eaf95ac78e065f6dd8128098158b38674479b721d95d937fe7adb892932e9101
+wrksrc="Sick-Beard-build-${version}"
+make_dirs="
+ /var/lib/sickbeard 0755 sickbeard sickbeard"
```
xtraeme: `/var/lib/sickbeard` should be created in the runit script instead.

xtraeme:
This package is still missing:

- `noarch=yes`
- `pycompile_dirs="/usr/share/$pkgname"`

PR #2821, opened 2015-10-25
New package: MultiMC-0.4.7

Spencer-H:


srcpkgs/MultiMC/template
```diff
@@ -0,0 +1,27 @@
...
+maintainer="Spencer H <spencernh77@gmail.com>"
+license="Apache 2.0"
+homepage="http://multimc.org"
+distfiles="https://github.com/MultiMC/MultiMC5/archive/0.4.7.tar.gz"
+checksum=961ece29094931b0638244c4a2614623d6df3193ee1a9e156ed36c01325ebae0
+wrksrc=${pkgname}5-$version
+
+pre_configure() {
+export JAVA_HOME=/usr/lib/jvm/openjdk
+export PATH=$PATH:$JAVA_HOME/bin
```
Gottox: indent

Spencer-H:
Fixed.

xtraeme:
why `/opt`? please use `/usr/share/$pkgname` instead.

Spencer-H:
Actually thanks, that just reminded me that I forgot something when I
uploaded this. The executable has to be in a folder that the user has write
access to. Would /usr/share be the right place for that, or is there
something else I should use.

dominikh:
Considering our discussion on IRC, please ping when this PR is ready for review again.

(Summary of the discussion: write-access for all users to where MultiMC is installed isn't acceptable, MultiMC should store its files in the user's home directory. I have no idea if the current version of the template implements that)

Spencer-H:
@dominikh This version actually does have that. I should have mentioned that when I updated the pull request. But this version moves MultiMC and all of the files it uses to /usr/share/multimc and it makes sure that the program puts anything it needs to write to into a .multimc folder in the home of the user executing it. After our discussion I went through the repo for MultiMC looking for where I could change it to treat a different directory as home. It turns out there is a flag for it you can pass to the program, which I hadn't discovered before. This version makes use of that in the script that it places into /usr/bin for executing the program.

Spencer-H:
Added a patch that forces translations to respect the directory selected for install files. I'm going to make a pull request on the MultiMC repo for this.

pullmoll:
You can remove the repeated `vmkdir /usr/share/multimc`.
The line `vmkdir usr/share/multimc/jars` will already create the parent.
The issues with downloading from https should be fixed now.

Spencer-H:
This is up to head and has been ready to be merged for a while. The patch fixes it so that MultiMC runs without needing write access to anything besides the users home ( .multimc ). The patch will also be merged in the next release of MultiMC so that it does not need to be maintained.

srcpkgs/MultiMC/template
```diff
@@ -0,0 +1,26 @@
...
+pkgname=MultiMC
+version=0.4.7
+revision=1
+build_style=cmake
+hostmakedepends="openjdk openjdk-jre cmake qt5-devel qt5-x11extras-devel gtk+-devel xxd"
+short_desc="An instanced Minecraft Client"
+maintainer="Spencer H <spencernh77@gmail.com>"
+license="Apache 2.0"
+homepage="http://multimc.org"
+distfiles="https://github.com/MultiMC/MultiMC5/archive/0.4.7.tar.gz"
```
dominikh: Please use ${version} in the distfiles URL

srcpkgs/MultiMC/template
```diff
@@ -0,0 +1,26 @@
+# Template file for 'MultiMC'
+pkgname=MultiMC
+version=0.4.7
+revision=1
+build_style=cmake
+hostmakedepends="openjdk openjdk-jre cmake qt5-devel qt5-x11extras-devel gtk+-devel xxd"
```
dominikh: The -devel packages should probably be makedepends, not hostmakedepends. That is, please fix the cross-compile failures for ARM that travis is reporting. (You can ignore the musl errors.)

Spencer-H:
And fixed.

dominikh:
I'll make the final fixes myself before merging it.

dominikh:
Merged as 3a967c25cb26473c1d8177ef45e9bc5e910fd420

PR #2822, opened 2015-10-25
common/wrapper/cc: rewrite to use standard posix.

Gottox:


xtraeme:
The issue is that now this needs grep(1).

xtraeme:
xbps-src requires bash anyway so that I don't see any issue with the wrapper using bash specific features.

xtraeme:
This version does not process `-I` and `-L` with blanks, i.e `"-I  /usr/include"`.

Gottox:
That's wrong, it does.

pullmoll:
I agree with @Gottox. It does, though it's pretty non obviously hidden in the fact that the first printf has no newline in the format string. As long as we *God beware!* don't have filenames containing a newline  this should work as I grok it.

Gottox:
Anyway, This script runs into a bash bug:

```
...
CC slcp.c
LDFLAGS  = -Wl,--as-needed  -specs=/void-packages/common/environment/configure/gccspecs/hardened-ld -Wl,-z,relro -Wl,-z,now -L/usr/arm-linux-musleabihf/usr/lib -g -L/usr/lib -L/usr/local/lib -lgit2
/builddir/.xbps-slcp/wrappers/arm-linux-musleabihf-gcc: line 8: syntax error near unexpected token `;;'
/builddir/.xbps-slcp/wrappers/arm-linux-musleabihf-gcc: line 8: `			-L|-I) printf '%s' "$arg" ;;'
Makefile:18: recipe for target 'slcp.o' failed
make: *** [slcp.o] Error 2
make: *** Waiting for unfinished jobs....
CC       = arm-linux-musleabihf-gcc
```

Gottox:
It seems to be fixed with a more recent version of chroot-bash.

Gottox:
closing as discussed in IRC.

PR #2824, opened 2015-10-25
mu: update to 0.9.15.

k00mi:


Gottox:
Would be great if you could fix musl too. But it's okay for now :)

k00mi:
Doesn't build on musl because emacs fails and I'm not going to fix that.

PR #2825, opened 2015-10-26
qutebrowser: update to 0.4.1.

The-Compiler:
I'm not the maintainer of the package, but the upstream author :wink:

v0.4.0 breaks with later versions of Python 3.4 (and with 3.5) - just got a crash report because of that, so I thought I'd quickly update the package.

cc @uggedal


xtraeme:
@The-Compiler thanks :-)

PR #2828, opened 2015-10-27
New package: font-fontin-0.0.20151027

thypon:


thypon:
```
restricted=yes
```

Not compiled by default.

PR #2830, opened 2015-10-27
New package: couchpotato-3.0.1

ananteris:


srcpkgs/couchpotato/files/couchpotato/run
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+exec chpst -ucouchpotato:couchpotato /usr/share/couchpotato/CouchPotato.py --data_dir=/var/lib/couchpotato 2>&1 > /dev/null
```
dominikh: Why are we discarding stderr?

srcpkgs/couchpotato/template
```diff
@@ -0,0 +1,27 @@
...
+system_accounts="couchpotato"
+couchpotato_homedir="/var/lib/couchpotato"
+distfiles="https://github.com/RuudBurger/CouchPotatoServer/archive/build/${version}.tar.gz>${pkgname}-${version}.tar.gz"
+checksum=f08f9c6ac02f66c6667f17ded1eea4c051a62bbcbadd2a8673394019878e92f7
+make_dirs="
+ /var/lib/couchpotato 0755 couchpotato couchpotato"
+homepage="http://couchpota.to"
+license="GPL-3"
+
+post_extract() {
```
dominikh: Instead of this entire function, just do

```
wrksrc=CouchPotatoServer-build-${version}
```

srcpkgs/couchpotato/files/couchpotato/run
ananteris: This has been switched to a /var/log log incase it has usable output (AFAIK it does not and I did not want to risk spamming a tty with log messages). Couchpotato manages its own logs, accessible through the UI.

srcpkgs/couchpotato/INSTALL
```diff
@@ -0,0 +1,8 @@
+# INSTALL
+case "$ACTION" in
```
xtraeme: I'd prefer to create this file in the runit script instead.

xtraeme:
This package is still missing:

- `noarch=yes`
- `pycompile_dirs="/usr/share/$pkgname"`

PR #2849, opened 2015-10-30
linux3.18: update to 3.18.23

kylesusername:
Update package linux3.18 from 3.18.22 to 3.18.23


kylesusername:
Travis is lying. The build was successful, but the log was longer than 4 MiB.

PR #2850, opened 2015-10-30
tpm

allan:


srcpkgs/trousers/files/tcsd/run
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+exec 2>&1
```
xtraeme: I'd prefer to create `/var/lib/tpm` in the runit script instead.

srcpkgs/trousers/template
```diff
@@ -0,0 +1,29 @@
+# Template file for 'trousers'
+pkgname=trousers
+version=0.3.13
+revision=1
+build_style=gnu-configure
+conf_files="/etc/tcsd.conf"
+makedepends="libressl-devel"
+system_accounts=tss
```
xtraeme: The homedir for the `tss` user probably needs to be set to `/var/lib/tpm`:

`tss_homedir=/var/lib/tpm`

allan:
adjusted and added some musl patches

srcpkgs/trousers/files/tcsd/run
```diff
@@ -0,0 +1,5 @@
...
+chown tss:tss /etc/tcsd.conf
```
Gottox: ... but changing owners of config files is none of the `run` scripts business. :)

Gottox:
Please fix the above comments, otherwise this PR will be closed.

Vaelatern:
@Gottox, would you mind if I grabbed this, fixed the above issues, and re-PRed from my account? @allan, with your permission of course, since you seem to be no longer interested.

Gottox:
Feel free. I don't care who does the PR. ;)

srcpkgs/trousers/template
Vaelatern: This was taken care of in the line below this note.

xtraeme:
Merged in #3334

PR #2851, opened 2015-10-30
vboot-utils: fix destdir

allan:


xtraeme:
revision must be bumped

allan:
grr, fixed

PR #2854, opened 2015-10-30
collectd: enable build_pie; unbreak cross build.

stpx:


Duncaen:
@stpx fyi the buildserver currently fails to build packages if they are downloaded over https.
http://build.voidlinux.eu/waterfall
http://build.voidlinux.eu/builders/armv6l-musl_builder/builds/5570/steps/shell_3/logs/stdio

PR #2856, opened 2015-10-31
New package: flightgear-3.4.0

pullmoll:


srcpkgs/simgear/template
```diff
@@ -0,0 +1,21 @@
+# Template file for 'simgear'
+pkgname=simgear
+version=3.4.0	# XXX: must be in sync with flightgear!
```
xtraeme: Please move this comment to another line, `xbps-checkvers(1)` is not able to parse the value correctly.

pullmoll:
*grrr* the md5.c/h patch is killing me with its missing LF at EOF... :)

PR #2862, opened 2015-11-01
New package: katarakt

k00mi:


pullmoll:
The `nocross=yes` isn't really necessary.
It may be wrong if one day we can get qt5 cross compiled.

k00mi:
Removed the nocross=yes.

I have no idea why certificate verification fails for the Travis builds, it works for me locally. If you want I can download over plain HTTP, but I'd prefer not to.

pullmoll:
The problem with ca-certificates in the chroot should have been solved by now. I don't know what may be wrong here. 

Pinging @xtraeme to the rescue :)

xtraeme:
I'm still trying to understand what's wrong with the travis setup. AFAIK it should work everywhere.

xtraeme:
The certificate is known and verified correctly. For some reason in the travis environment it's failing.

pullmoll:
Perhaps travis-ci has a real issue with the Telekom CA. Can we somehow get more detailed info on the failure other just "Certificate verification failed"? It might be the Telekom CA itself or one in the chain from `cip.cs.fau.de` to the root CA, if I understand that concept correctly.

As a last resort we may try to set XBPS_DISTFILES_MIRROR in travis-ci etc/conf and fetch distfiles from elsewhere. The SHA256 should be sufficient to verify its integrity.

xtraeme:
You can disable cert verification in xbps (libfetch) setting the SSL_NO_VERIFY_PEER env var.

I'm investigating how to get more information on the reported openssl issue.
See https://github.com/voidlinux/xbps/blob/master/lib/fetch/common.c#L827

xtraeme:
The travis issue has been fixed, please revert the changes to HTTP.

PR #2869, opened 2015-11-01
New package: neovim-0.1.0.

stpx:


pullmoll:
I think there should perhaps be a FindLUA.cmake script in this package which checks the common versions and their varying locations, no?

IIRC I've seen, and patched, such a script in some other package to also look for lua5.2 or lua52.

xtraeme:
The alternatives framework can be useful for this case, the lua pkgs should set bin/lua and bin/luac as alternatives.

pullmoll:
xbps-alternatives works for binaries only, or does it? Otherwise `lua52.pc` could be an alternative for `lua.pc` and both `pkg-config` and `cmake` should find the installed version.

xtraeme:
With alternatives you can symlink anything, any file or directory.

See srcpkgs/ex-vi as example.

pullmoll:
I really should take the time to RTFS and understand xbps-alternatives. *sigh*

stpx:
Adding `-DLUA_PRG=/usr/bin/lua5.2` to `configure_args` appears to work.... I know very little about lua.

stpx:
Cross-compile is broken... And the last time I tried to compile for `x86_64-musl` it failed, but the was a few months ago.

stpx:
I think `armv*` builds are getting hung up from: building `lua*-lpeg` for the target arch, then for host arch. That shouldn't be a problem on the actual build server, right?

stpx:
The `armv*-musl` builds are failing because of `LuaJIT`.  Good enough to merge?

pullmoll:
:+1:

Issue #2829, opened 2015-10-27
cava segfaults on startup


gregf:
```
uname -a
Linux betsy 4.1.12_1 #1 SMP PREEMPT Tue Oct 27 06:38:21 UTC 2015 x86_64 GNU/Linux
```

```
cava
*** Error in `cava': free(): invalid next size (fast): 0x0000000001f35e30 ***
======= Backtrace: =========
/usr/lib/libc.so.6(+0x71e76)[0x7f0cc44cae76]
/usr/lib/libc.so.6(+0x776c6)[0x7f0cc44d06c6]
/usr/lib/libc.so.6(+0x77eae)[0x7f0cc44d0eae]
/usr/lib/libiniparser.so.0(iniparser_load+0x208)[0x7f0cc562aea8]
cava[0x4047bd]
cava[0x40204d]
/usr/lib/libc.so.6(__libc_start_main+0xf0)[0x7f0cc4479680]
cava[0x403119]
======= Memory map: ========
00400000-00407000 r-xp 00000000 08:04 1983935                            /usr/bin/cava
00606000-00607000 r--p 00006000 08:04 1983935                            /usr/bin/cava
00607000-00608000 rw-p 00007000 08:04 1983935                            /usr/bin/cava
00608000-0060d000 rw-p 00000000 00:00 0
01f35000-01f56000 rw-p 00000000 00:00 0                                  [heap]
7f0cbfde9000-7f0cbfdff000 r-xp 00000000 08:04 1972893                    /usr/lib/libgcc_s.so.1
01f35000-01f56000 rw-p 00000000 00:00 0                                  [heap]                                                                                                                            [0/59]
7f0cbfde9000-7f0cbfdff000 r-xp 00000000 08:04 1972893                    /usr/lib/libgcc_s.so.1
7f0cbfdff000-7f0cbfffe000 ---p 00016000 08:04 1972893                    /usr/lib/libgcc_s.so.1
7f0cbfffe000-7f0cbffff000 rw-p 00015000 08:04 1972893                    /usr/lib/libgcc_s.so.1
7f0cc0000000-7f0cc0021000 rw-p 00000000 00:00 0
7f0cc0021000-7f0cc4000000 ---p 00000000 00:00 0
7f0cc4049000-7f0cc4050000 r-xp 00000000 08:04 1972815                    /usr/lib/librt-2.22.so
[...]

PR #2871, opened 2015-11-01
cava: use bundled iniparser to fix segfault

JayVii:
**Workaround** for #2829 

This pullrequest can be rejected, if someone gets void's [iniparser-package](https://github.com/voidlinux/void-packages/tree/f4fd6b1d9883470e12b314d1de8942a03d3be20e/srcpkgs/iniparser) to work correctly with cava again. 
@pullmoll is the maintainer of that package, so he may have an idea on that.


JayVii:
Travis failed getting the download-file from github:
```
Certificate verification failed for /C=US/O=DigiCert Inc/OU=www.digicert.com/CN=DigiCert SHA2 Extended Validation Server CA
SSL_connect returned 1
https://github.com/karlstav/cava/archive/0.3.5.tar.gz: Operation not permitted

=> ERROR: cava-0.3.5_2: failed to fetch 0.3.5.tar.gz.
```

pullmoll:
We've seen failures with travis-ci and certificate chains in other pull requests already. It is being worked on.

For iniparser I'll check if there are any patches around which could or should be applied. Other than that iniparser is a plain vanilla 4.0 with no frills. And it's just __2__ source files where a bug could hide.

pullmoll:
__OMFG__ - the first function I looked at is like this:
```
/*-------------------------------------------------------------------------*/
/**
  @brief    Copy string in a newly mallocced area
  @param    str  String to copy.
  @return   str  Copied version of the given string allocated with malloc

  Original strdup is not portable, need to implement our own
 */
/*--------------------------------------------------------------------------*/
static char * _strdup(const char *s)
{
    char * copy = (char*) malloc(strlen(s));
    strcpy(copy, s);
    return copy ;
}
```
This is certainly going to write beyond the allocated memory...

If the rest of the code is of the same quality we'd better ditch the iniparser package altogether.

pullmoll:
Please try with iniparser-4.0_3

gregf:
@pullmoll iniparser-4.0_3 fixes the problem I had.

JayVii:
Yes, it fixes the issue. Thank you, @pullmoll :)

PR #2874, opened 2015-11-01
hedgewars: update to 0.9.22

skrzyp:


pullmoll:
It looks like you need to squash the two commits into one because they both modify `srcpkgs/hedgewars/template`. Thus the `This branch has conflicts that must be resolved`.

skrzyp:
wtf…

PR #2878, opened 2015-11-02
[WIP] xbps-src: implement obsolete_checksum handling

pullmoll:
Whenever a distfile changes its checksum without also changing its
version number, someone would have to manually clean the obsoleted
distfile(s) from the build farm machines' file caches.

The variable `obsolete_checksum="abcdef..."` allows a package maintainer
to explicitly purge one or more obsolete checksums and their hard linked
files from the cache. Afterwards the distfile(s) will be downloaded again
and verified against the new checksum(s).


xtraeme:
:-1: 

This should be automatic and transparent to the users.

PR #2887, opened 2015-11-02
lua*: convert for xbps-alternatives

pullmoll:
First try to make use of the xbps-alternatives for the three versions
of lua we have now. As an example how this simplifies packages the
xmoto package patches for detecting lua52 were removed.

Also revbump or update the packages depending on lua53 to cope
for the soname change (liblua.so.5.3 to liblua5.3.so.5.3)
+ imapfilter (rev 3)
+ vim (rev 2)
+ mame (rev 2)
+ xmoto (rev 6)
+ glu (rev 6) - rebuild for musl


srcpkgs/lua52/template
```diff
@@ -12,6 +12,18 @@ license="MIT"
...
 
+alternatives="
+ lua:lua:/usr/bin/lua5.2
+ lua:luac:/usr/bin/luac5.2
+ lua:lua.1:/usr/share/man/man1/lua5.2.1
+ lua:luac.1:/usr/share/man/man1/luac5.2.1
+ lua:lua:/usr/include/lua5.2
+ lua:lua.pc:/usr/lib/pkgconfig/lua5.2.pc
+ lua:liblua.a:/usr/lib/liblua5.2.a
+ lua:liblua.so:/usr/lib/liblua5.2.so
```
xtraeme: But this won't work if the `devel` pkg is not installed, because those files do not belong to the `lua` pkg.

pullmoll: So can I have alternatives in the -devel subpackage?

xtraeme: Hmm thinking about this, your changes are ok. If the devel pkg is not installed it will create dangling symlinks.

pullmoll:
Just for the record:
Dangling symlinks won't work yet, because they may also contain missing path parts.
No merge until this works.

muddmaker:
I'm not sure if this is a good idea. Lua 5.1, 5.2 and 5.3 are not backwards compatible with each other, and many packages will depend on the behavior of a specific version. Specifically, Lua5.2 changed the way that setting function environment works, and 5.3 added (and replaced certain operations with) integer operations. This will cause more problems than it is worth.

On the other hand, it might make sense to add an alternatives group with Lua5.1 and LuaJIT, because they are compatible with each other.

stpx:
This will just add the ability to symlink `/usr/bin/lua => /usr/bin/lua5.x`, `/usr/bin/luac => /usr/bin/luac5.x` and so on... `x` being whatever version you configure `xbps-alternatives` for. `/usr/*/lua5.1`, `/usr/*/lua5.2` and `/usr/*/lua5.3` will still exist.

chneukirchen:
I fail to see the benefit of these alternatives?

pullmoll:
Benefit: simplification of templates for packages depending on lua5.1, lua5.2 or even newer (more to come).

Deficiency: to be evidenced.

pullmoll:
Let me clarify some things which seem to be misunderstood:

This setup for xbps-alternatives is mostly useful for the build (compile) stage.
It sets some symlinks for the one lua version to be used when building a packge.
These links make header files and pkg-config info appear with names that are known to (older) packages and so they can pick them up without the packager having to patch paths or filenames.

The resulting package is in no way depending on the later on selected lua version. It is linked against one of the shared or static libraries of lua5.1, lua5.2 or lua5.3.

For a user it is perfectly possible to have all three lua versions installed and they won't collide. You, as a user of applications, don't ever have to select lua versions. The packages you use are linked against one or another. The only case where you would select alternatives is, if you were developing lua applications and wanted to quickly switch between the luas.

chneukirchen:
I just smell weird bugs when building in a non-cleaned masterdir...

pullmoll:
Hmm... why would you not clean before build another package? xbps-src does this for you, except if you manage to ^C before it has a chance to clean up. It also cleans up before building a package.

Please note that the idea to use xbps-alternatives for the lua versions came from @xtraeme. I just went and actually created the patches and associated revbumps/updates. Perhaps Juan can chime in to tell if and why nothing bad will happen with this pr being merge. :)

xtraeme:
I can't see any issue with lua using alternatives, to me the changes are acceptable and useful enough.

muddmaker:
> For a user it is perfectly possible to have all three lua versions installed and they won't collide. You, as a user of applications, don't ever have to select lua versions. The packages you use are linked against one or another. The only case where you would select alternatives is, if you were developing lua applications and wanted to quickly switch between the luas.
That seems pefectly acceptable in that case. Before this is merged, should LuaJIT be added to the alternatives group?

pullmoll:
@muddmaker I'd like to do this after the first step, i.e. this pr as is. LuaJIT currently has some issues when cross compiling AFAIR and needs attention.

xtraeme:
Is this now ready to merge with xbps-0.50? please test.

pullmoll:
Seems to be working for me. I can install and switch between lua, lua51 and lua52.

PR #2889, opened 2015-11-02
New package: pdfgrep

k00mi:


xtraeme:
travis should work now with https, please repush to verify.

PR #2891, opened 2015-11-02
Update gradle version in executable

WKuipers:
This commit makes the version consistent with the version actually shipped with this package, which fixes gradle exiting with an error on launch.


xtraeme:
revision must be bumped and a proper commit message must be used.

See CONTRIBUTING.md for details.

pullmoll:
Done

PR #2892, opened 2015-11-02
New package: workrave-1.10.8.

### LINKED ISSUE #2877: Package Request: workrave
> Workrave is a program that assists in the recovery and prevention of Repetitive Strain Injury (RSI). The program frequently alerts you to take micro-pauses, rest breaks and restricts you to your daily limit. The program runs on GNU/Linux and Microsoft Windows. 

http://www.workrave.org/

stpx:
Closes #2877.

### DIFF
--- a/srcpkgs/workrave/template
+++ b/srcpkgs/workrave/template
@@ -1,0 +1,25 @@
+# Template file for 'workrave'
+pkgname=workrave
+version=1.10.8
+revision=1
+_realversion="${version//./_}"
+wrksrc="${pkgname}-${_realversion}"
+nocross=yes # gobject-introspection
+build_style=gnu-configure
+configure_args="--disable-static --disable-gnome $(vopt_enable pulseaudio pulse)"
+hostmakedepends="automake autoconf-archive gettext-devel intltool pkg-config
+ gobject-introspection python-cheetah"
+makedepends="gtkmm-devel libSM-devel libXtst-devel libXScrnSaver-devel
+ $(vopt_if pulseaudio pulseaudio-devel)"
+short_desc="Program that assists in the recovery and prevention of RSI"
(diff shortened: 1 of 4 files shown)


pullmoll:
I think `$(vopt_if pulseaudio pulseaudio-devel)` belongs into `makedepends`.
If `gobject-introspection` is the only thing blocking cross builds, you can add `$(vopt_if gir gobject-introspection)`, add to the `build_options="gir pulseaudio"` and finally add this to the template's body:
```
        build_options_default="gir"
fi
```

stpx:
@pullmoll You're right about the pulseaudio, that's a mistake. It wouldn't build without `gobject-intospection`, though. I'll take another look today. I was doing a couple other things when putting this together.

pullmoll:
If it won't build that's another story. We have quite a few packages where introspection is optional. Usually there's also a `--enable-instrospection` configure parameter.

srcpkgs/workrave/template
```diff
@@ -0,0 +1,25 @@
...
+version=1.10.8
+revision=1
+_realversion="${version//./_}"
+wrksrc="${pkgname}-${_realversion}"
+nocross=yes # gobject-introspection
+build_style=gnu-configure
+configure_args="--disable-static --disable-gnome $(vopt_enable pulseaudio pulse)"
+hostmakedepends="automake autoconf-archive gettext-devel intltool pkg-config
+ gobject-introspection python-cheetah"
+makedepends="gtkmm-devel glibmm-devel libSM-devel libICE-devel libXtst-devel
```
xtraeme: - gtkmm-devel depends on glibmm-devel, so that glibmm-devel is redundant.
- libSM-devel depends on libICE-devel, so that libICE-devel is redundant.
-

PR #2897, opened 2015-11-03
00-distfiles.sh: add code to purge wrong checksum

pullmoll:
If a distfile changed its checksum (SHA256) without at the same time
also changing its version, the distfile cache needs to be purged.

This is done by verifying only for already existing distfiles that the
specified checksum matches. If it does not, the file and its entry in
/host/sources/by_sha256 are purged using the inode number of the file.

The second loop then as usual fetches the distfiles and verifies their
checksum against the list entry.

### DIFF
--- a/common/hooks/do-fetch/00-distfiles.sh
+++ b/common/hooks/do-fetch/00-distfiles.sh
@@ -26,7 +26,7 @@ verify_cksum() {
 
 	cksum=$(get_cksum $curfile $dfcount)
 	msg_normal "$pkgver: verifying checksum for distfile '$curfile'... "
-	filesum=$(${XBPS_DIGEST_CMD} $distfile)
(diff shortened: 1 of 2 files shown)


xtraeme:
Looks good at a first glance :+1:

pullmoll:
I tried with some small packages like `mmv`.. just change one digit and the file is removed from the cache.

BTW sometimes the error message showing the checksum found is not printed.. the prompt appears at the end of the line before. bash trap issue?

pullmoll:
It took me some time to get `drbd-utils` cross compile again. I wonder why it worked before without the tweaks that were now necessary, i.e. installing the host's `drbd-utils` for `drbdsetup{,-83,84}`.

Before merging the result should be double checked.

PR #2899, opened 2015-11-03
w3m: patch to fix w3mhelp.cgi and w3mman2html.cgi.

stpx:
Some patches to fix:
- w3m help menu
- w3mman
- form field issues


xtraeme:
:+1:

PR #2909, opened 2015-11-05
neovim: add vi and vim alternatives

Duncaen:


stpx:
:+1: 
But just FYI, `neovim` wont run ex-mode when run as `ex`.

Duncaen:
Thanks, didn't know that https://github.com/neovim/neovim/issues/1089

PR #2910, opened 2015-11-05
Revert "zsh: register 'sh' alternatives group."

Gottox:
zsh is no POSIX compatible shell. They are working on it, but I think, that it will have many bad implications to use it as an sh alternative.


Duncaen:
Probably the best in this case, i think he didn't regenerate his initrd and this would break more.

PR #2932, opened 2015-11-10
New package: spim-8.0

Vaelatern:
This is a zip file from the creator's homepage, not the project homepage. The code for spim has not changed in a while. It is still often used in university classes, as the creator acknowledges. I took a hint from of Gentoo and FreeBSD's ports, and just built spim from [here](http://pages.cs.wisc.edu/~larus/SPIM/) instead of fiddling with SVN checkouts. Since the page in question belongs to the author (Dr. James Larus), I feel it is an acceptable substitute to the official project page (where there are not zip files or tarballs for easy download).

### DIFF
--- a/srcpkgs/spim/template
+++ b/srcpkgs/spim/template
@@ -1,0 +1,32 @@
+# Template file for spim
+
+pkgname="spim"
+version="8.0"
+revision=1
+homepage="http://spimsimulator.sourceforge.net/"
+license="BSD"
+distfiles="http://pages.cs.wisc.edu/~larus/SPIM/spim-${version}.tar.gz"
+checksum="6f205776cb9fa112729507008843b289012190ed3131cbd426c610a58387ee4b"
+maintainer="Toyam Cox <vaelatern@gmail.com>"
+hostmakedepends="flex"
+short_desc="Self-contained simulator/debugger that runs MIPS32 programs"
+do_configure() {
+	cd spim
+	sed '/Copyright (c) 1990-2010, James R. Larus./,+26!d' README > LICENSE
+	sed -i Makefile \
+	-e 's:EXCEPTION_DIR = /usr/local/lib/spim:EXCEPTION_DIR = /usr/share/spim:' \
+	-e 's:CFLAGS =:CFLAGS +=:' \
+	-e 's:LDFLAGS =:LDFLAGS +=:' \
+	-e 's:CC = gcc:CC ?= gcc:'
+	CC="$CC" CFLAGS="$CFLAGS" LDFLAGS="$LDFLAGS" ./Configure
+}
+do_install() {
+	vbin spim/spim
+	vman Documentation/spim.man spim.1
+	vlicense spim/LICENSE LICENSE.BSD
+	vinstall CPU/exceptions.s 755 /usr/share/spim/
+}
+do_build() {
+	cd spim
+	make
+}


Vaelatern:
Don't waste your time right now. Still working things out with upstream.

xtraeme:
The first three variables of a template must be:

```
pkgname
version
revision
```

Vaelatern:
@xtraeme: Fixed
Changed source of tarball, works reliably on amd64.

srcpkgs/spim/template
```diff
@@ -0,0 +1,29 @@
...
+
+pkgname="spim"
+version="8.0"
+revision=1
+homepage="http://spimsimulator.sourceforge.net/"
+license="BSD"
+distfiles="http://pages.cs.wisc.edu/~larus/SPIM/spim-8.0.tar.gz"
+checksum="6f205776cb9fa112729507008843b289012190ed3131cbd426c610a58387ee4b"
+maintainer="Toyam Cox <vaelatern@gmail.com>"
+makedepends="bison flex"
```
xtraeme: Change this line to `hostmakedepends="flex"`

srcpkgs/spim/template
```diff
@@ -0,0 +1,29 @@
...
+only_for_archs="i686 x86_64"
```
xtraeme: This is bullshit, there's no need to force a specific architecture.

srcpkgs/spim/template
```diff
@@ -0,0 +1,29 @@
...
+short_desc="Spim is a self-contained simulator/debugger that runs MIPS32 programs"
+wrksrc="spim-8.0"
```
xtraeme: This is redundant, wrksrc defaults to `${pkgname}-${version}`

srcpkgs/spim/template
```diff
@@ -0,0 +1,29 @@
...
+do_configure() {
+	cd "${wrksrc}/spim"
```
xtraeme: `cd spim`

srcpkgs/spim/template
```diff
@@ -0,0 +1,29 @@
...
+	sed '/Copyright (c) 1990-2010, James R. Larus./,+26!d' README > LICENSE
+	./Configure
+}
+do_install() {
+	vbin "${wrksrc}/spim/spim"
```
xtraeme: There's no need to use `${wrksrc}` nor quoting the arguments.

srcpkgs/spim/template
```diff
@@ -0,0 +1,29 @@
...
+	vman "${wrksrc}/Documentation/spim.man" "spim.1"
+	vlicense "${wrksrc}/spim/LICENSE" "LICENSE.BSD"
+}
+do_build() {
+	cd "${wrksrc}/spim"
```
xtraeme: `cd spim`

srcpkgs/spim/template
```diff
@@ -0,0 +1,29 @@
+# Template file for spim
+# This is not qtspim, although they are distributed together.
+
+pkgname="spim"
+version="8.0"
+revision=1
+homepage="http://spimsimulator.sourceforge.net/"
+license="BSD"
+distfiles="http://pages.cs.wisc.edu/~larus/SPIM/spim-8.0.tar.gz"
```
xtraeme: `distfiles=".../spim-${version}.tar.gz"`

srcpkgs/spim/template
Vaelatern: Build fails on arm. Will push changes to test again.

Vaelatern:
Travis build fails on arm builds without only_archs. Can't find an option to just disable arm6 and arm7.

srcpkgs/spim/template
Vaelatern: Done.

srcpkgs/spim/template
```diff
@@ -0,0 +1,27 @@
...
+
+pkgname="spim"
+version="8.0"
+revision=1
+homepage="http://spimsimulator.sourceforge.net/"
+license="BSD"
+distfiles="http://pages.cs.wisc.edu/~larus/SPIM/spim-${version}.tar.gz"
+checksum="6f205776cb9fa112729507008843b289012190ed3131cbd426c610a58387ee4b"
+maintainer="Toyam Cox <vaelatern@gmail.com>"
+makedepends="flex"
```
xtraeme: `flex` needs to be run in the host, so that it should be added to `hostmakedepends` instead.

srcpkgs/spim/template
```diff
@@ -0,0 +1,27 @@
...
+only_for_archs="i686 x86_64"
+short_desc="Spim is a self-contained simulator/debugger that runs MIPS32 programs"
```
xtraeme: `Spim is a` is redundant in description.

Vaelatern:
Tested and known working, used it for a school assignment.

pullmoll:
Try with [this template](http://sprunge.us/QPXj). It has some necessary changes to enable cross compiling and use the `xbps-src` defined settings for `$CC`, `$CFLAGS` and `$LDFLAGS`.

Vaelatern:
It also compiles three times as fast this way

Issue #2943, opened 2015-11-11
Package Request: Yandex Disk


Astrosurge:
https://yandex.com/support/disk/cli-clients.xml

Yandex Disk is a cloud storage service similar to Dropbox and Google Drive.

Thanks!

xtraeme:
#2947 has been merged

PR #2947, opened 2015-11-11
New package: yandex-disk-0.1.5.940

o1ux:
https://github.com/voidlinux/void-packages/issues/2943


xtraeme:
The license explicitely states that it's non redistributable, see https://yandex.com/support/legal/desktop_software_agreement/ 4.2.

This package needs to set `restricted=yes`.

PR #2948, opened 2015-11-11
New package: nodm-0.7

Duncaen:


srcpkgs/nodm/files/nodm/run
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+[ -r conf ] && . ./conf
+export NODM_USER NODM_X_OPTIONS NODM_MIN_SESSION_TIME NODM_XINIT NODM_XSESSION
```
xtraeme: The `conf` should not be provided by the package.

The `run` file should set defaults if unset, i.e:

```
: ${NODM_X_OPTIONS:="vt7 -nolisten tcp"}
...
```

Duncaen:
Removed the `conf` file and pushed it in 8e5745d, the defaults are actually not needed, nodm has built in defaults.

PR #2954, opened 2015-11-12
firefox: enable gtk3 ui

skrzyp:
This patch enables GTK3 UI in Firefox, since it's started to be finally supported in Firefox 42.
While I was always against GTK3, I found this UI faster and lewss glitched than GTK2, which might be deprecated soon.

I don't want to transfer maintainership of `firefox` to myself, so I want maintainer to review this patch, or another person more knowledgeable in Firefox's internals and build system _(why we are not using mozconfig?)_.

Depending on final decision, it might be merged directly to `firefox` package, or as separately `firefox-gtk3` one, but I'd prefer the first variant, because it can help us to break from old GTK2 dependencies.

### DIFF
--- a/srcpkgs/firefox/template
+++ b/srcpkgs/firefox/template
@@ -1,7 +1,7 @@
 # Template build file for 'firefox'.
 pkgname=firefox
 version=42.0
-revision=1
+revision=2
 short_desc="Lightweight gecko-based web browser"
 maintainer="Juan RP <xtraeme@voidlinux.eu>"
 homepage="https://www.mozilla.org/firefox/"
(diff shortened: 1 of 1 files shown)


Gottox:
> (why we are not using mozconfig?).
cross compiling.

Gottox:
Providing both firefox and firefox-gtk3 is senseless. If gtk3 isn't going default, we should introduce a build option instead.

skrzyp:
@Gottox: GTK3 will be default in 43.0, according to the latest news - it was delayed for one version, probably for testing purposes.

@xtraeme said that it's not acceptable where Firefox builds against *two* toolkits at once. I might be wrong on this, but I think it's the configuration fault and can be fixed in some way. But, like I said before, we need some Mozilla's "expert" here :)

skrzyp:
Related issues on Bugzilla: [624422](https://bugzilla.mozilla.org/show_bug.cgi?id=624422)

PR #2974, opened 2015-11-15
New Package: gmusicbrowser-1.1.15

logenkain:
Couldn't get arm to work, but... here we go.


pullmoll:
`homepage="http://gmusicbrowser.org"`

srcpkgs/gmusicbrowser/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'gmusicbrowser'
+pkgname=gmusicbrowser
+version=1.1.15
+revision=1
+noarch=yes
+only_for_archs="x86_64 i686 x86_64-musl"
```
xtraeme: Seriously, there's nothing x86 specific in this software. It's a perl script that works anywhere if the dependencies are there.

Just remove this line.

srcpkgs/gmusicbrowser/template
```diff
@@ -0,0 +1,19 @@
...
+hostmakedepends="perl"
+makedepends="perl"
+depends="perl gtk+ perl-Gtk2 mpv desktop-file-utils webkitgtk2"
```
xtraeme: There's no need to depend on `perl, gtk+ and mpv`. Why?

- perl-Gtk2 already depends on perl and gtk+
- mpv is an optional dependency, and this software can also use mpg123 and others.
- I really doubt this needs webkitgtk2.

logenkain:
>> * to consult wikipedia pages and search google for lyrics, one of :
>>          o gtk2-mozembed perl bindings (package named libgtk2-mozembed-perl or perl-Gtk2->> MozEmbed) (see this page)
>>          o gtk2-webkit perl bindings (package named perl-Gtk2-WebKit)

That's why I threw webkit in there.  I'll get rid of the extra depends (I don't know what I was thinking).

I put mpv there because it's the only thing that seemed to work on void, but I suppose that's not a great reason to do that.

And arm.  I have no idea what to do about that.

I get this:

Can't locate ExtUtils/Depends.pm in @INC (you may need to install the ExtUtils::Depends module) (@INC contains: /usr/lib/perl5/site_perl /usr/share/perl5/site_perl /usr/lib/perl5/vendor_perl /usr/share/perl5/vendor_perl /usr/lib/perl5/core_perl /usr/share/perl5/core_perl .) at (eval 8) line 1.
BEGIN failed--compilation aborted at (eval 8) line 1.

MakeMaker FATAL: prerequisites not found.
    ExtUtils::Depends not installed
    ExtUtils::PkgConfig not installed

Please install these modules first and rerun 'perl Makefile.PL'.

Gottox:
Arm crossbuild shouldn't matter if the package is noarch.

dominikh:
`hostmakedepends="perl-ExtUtils-Depends perl-ExtUtils-PkgConfig`

logenkain:
On Thu, Nov 19, 2015 at 09:01:20AM -0800, Dominik Honnef wrote:
> `hostmakedepends="perl-ExtUtils-Depends perl-ExtUtils-PkgConfig`

I'm pretty sure I tried that, but I'll give it another shot.

dominikh:
Either hostmakedepends or makedepends (or possibly both). I'm no perl expert.

dominikh:
Oh... disregard me entirely, it's failing trying to build  perl-Glib, which is expected. Disregard the issue and carry on.

logenkain:
On Thu, Nov 19, 2015 at 10:06:57AM -0800, Dominik Honnef wrote:
> Oh... disregard me entirely, it's failing trying to build  perl-Glib, which is expected. Disregard the issue and carry on.

Eh I tried again anyway, didn't seem to work.  If anyone else has any ideas or instructions let me know.  I will, of course, rebase at that point.

logenkain:
Well, with no ideas how to fix it for arm, I reset to using only_for.  I'd like to have it work on ARM, but I have no idea how to do that.  At least with this people can still enjoy it where it's used most often.

pullmoll:
As @Gottox already said: this package is __`noarch=yes`__ which is why it is irrelevant, if the package can be cross built or not. And as @xtraeme said using `only_for_arch` is definitely wrong, because there's nothing x86 specific in the package.

Just remove the only_for_arch. The failing cross builds don't hurt anyone.

logenkain:
On Sun, Nov 22, 2015 at 10:16:02AM -0800, Jürgen Buchmüller wrote:
> Just remove the only_for_arch. The failing cross builds don't hurt anyone.

I thought I had it that way for a couple of days with no reaction.

I'll put it back.

pullmoll:
I did add perl-GStreamer and perl-GStreamer-Interfaces packages, so gmusicbrowser can make use of this interfaces and does not have to rely on mplayer or mpv. In order to play mp3s it needs mpg123, though.

PR #2976, opened 2015-11-15
New package: mlocate-0.26.

stpx:


Duncaen:
adding `#include <paths.h>` to `src/updatedb.c` fixes arm*-musl, i have no idea why x86_64-musl worked.

stpx:
@Duncaen Thanks!

xtraeme:
- The commit message is wrong `coreutils vs findutils`.
- Please use the `g` prefix for `locate`,  `updatedb` and others in findutils.

xtraeme:
:+1:

Duncaen:
I think you breake `/usr/bin/updatedb` in `finutils` if you just rename the files afterwards.
```
: ${frcode:=${LIBEXECDIR}/@frcode@}
: ${bigram:=${LIBEXECDIR}/@bigram@}
: ${code:=${LIBEXECDIR}/@code@}
```

stpx:
```sh
$ xbps-alternatives --set mlocate -g locate
```
Breaks `xlocate`.

cc @chneukirchen

chneukirchen:
xlocate only works with findutils (as by deps), but if the paths change we can fix the script of course.

stpx:
Right... I just wanted to give you a heads up before we merge this.

chneukirchen:
This is missing a config file and a daily cronjob?

chneukirchen:
mlocate should be setgid or something?

chneukirchen:
setgid is not enough here tho?

chneukirchen:
You need a hook to set /usr/bin/mlocate root:mlocate mode 2755

dicktyr:
locate and mlocate each have their own cron.daily script
and `/usr/sbin/run-cron /etc/cron.daily` runs both scripts
instead of only the chosen alternative's script

not really a problem for me but I thought I should mention it

in any case
thank you for the package!

Issue #2970, opened 2015-11-15
99-pkglint: meta package issues


pullmoll:
Some templates now generate an error when dependency only parts are to be packaged:
```
=> ERROR: elfutils-0.161_3: PKGDESTDIR is empty and build_style != meta
=> ERROR: elfutils-0.161_3: cannot continue with installation!
```
The same happens for `git-all` where I could fake a `build_style=meta` in the `git-all_package()` function to continue with the install.

xtraeme:
This is how it's expected to work, packages that don't install any files must set `build_style=meta`.

PR #2977, opened 2015-11-15
elfutils: change to build_style=meta

pullmoll:
The main package is empty and requires build style meta.
See #2970


xtraeme:
Nope... elfutils should not be an empty package, it should provide the `eu-xxx` utils (eu-strip, etc).

pullmoll:
Missed the point.

PR #2978, opened 2015-11-16
New package: z-1.9

diogoleal:


xtraeme:
- `noarch=yes`
- `vbin z.sh z`

dominikh:
- Is this actually supposed to be installed in /usr/bin? The install instructions indicate that you're supposed to source the script in your shell. Depending on the shell, it sets up tab completion and registers a shell function "z".
- z depends on awk

xtraeme:
Indeed, this package is really useless.

PR #2991, opened 2015-11-16
urxvt-perls: remove xsel dependency

rolf1304:
The clipboard script does not depend upon xsel. It can be configured to use any command.


Gottox:
increase the revision

xtraeme:
Please follow the commit guidelines explained in `CONTRIBUTING.md`, and then submit a proper pull request.

PR #2995, opened 2015-11-17
New package: strongswan-5.3.4

Vaelatern:


srcpkgs/strongswan/template
```diff
@@ -0,0 +1,13 @@
+# Template file for 'xl2tpd'
```
Gottox: change this to strongswan

Gottox:
Good so far. I think this will need an sv file too.

xtraeme:
I really doubt this is ok... shared libraries without development packages.

xtraeme:
Adding just shared libraries in `makedepends` without `development packages` is also completely wrong.

Vaelatern:
Can you explain what needs to be done for `development packages`? Just vmove the files as specified in the manual?

Tried looking at the systemd file, but can't find the executables that it calls. Was hoping for some help in writing the sv file.

Gottox:
See: https://github.com/voidlinux/void-packages/blob/master/Manual.md#libraries

Vaelatern:
I believe the provided library files belong in the main package, not a -devel package.

pullmoll:
The unversioned `/usr/lib/*.so` files and any static `/usr/lib/*.a` libraries belong into -devel, as well as header files, pkgconfig files, cmake helpers etc.

The versioned `/usr/lib/*.so.*` files should stay in a main package, or in a `lib<packge>` subpackage which main depends upon.

Gottox:
Nevertheless, I believe that strongswan is not a Library at all. Rather remove those files from common/shlibs.

Vaelatern:
Went with Gottox's thought, as I find it hard to think strongswan is a library. I simply got warnings during `./xbps-src install strongswan` and blindly tried to fix them.

pullmoll:
True. I see no other distribution defining a strongswan-dev(el) package.

xtraeme:
- Please disable static libraries, they are useless if you can't use them i.e `--disable-static` in `configure_args`.
- Set `build_pie=yes` to build it with hardening features (PIE, relro, etc).

Gottox:
Please update to 5.3.4. See https://www.strongswan.org/blog/2015/11/16/strongswan-vulnerability-%28cve-2015-8023%29.html

Vaelatern:
`build_pie` added then removed when it became default.

Issue #2994, opened 2015-11-17
umask setting is crazy


ayghor:
It recently came to my attention that my default umask value was
magically changed to 022, when permissions got messed up and another
user couldn't group-write my files.

I tried to change login.defs, but It didn't work, because it was being
overriden by /etc/profile. Running `git log --all --grep='umask'`, I
noticed it was changed a lot and that it's state is confusing (not to
say inconsistent):

 * /etc/login.defs	-> UMASK 077
 * /etc/profile		-> umask 022
 * /etc/pam.d/su	-> umask=022

IMHO, I think that default umask should be set by, and only by, PAM:

 * usergroup setting
 * shell independence
 * consistency
 * works on non-login shells

and that it should be 077, as suggested by pancake, since home dirs are
world-readable.

Output of `git log --all --grep='umask'`:

    commit 915e2fa0af97aa2403f4cb2753db002b22f39a41
    Author: Juan RP <xtraeme@voidlinux.eu>
    Date:   Thu Jul 2 08:06:52 2015 +0200
    
        util-linux: set su(1) default umask via pam.
        
        This should override umask setting from /etc/login.defs to 022.
        
        Should help #1941
    
    commit 83b38fc69fbf1812038689aedc7c75b848d97ebe
    Author: Juan RP <xtraeme@voidlinux.eu>
    Date:   Sat Jun 6 10:58:35 2015 +0200
    
        shadow: set umask to 077 via login.defs for sane home user dirs.
    
    commit 64c332096b245420531f4321509a2bcb0aee1473
    Author: Juan RP <xtraeme@voidlinux.eu>
    Date:   Sat Jun 6 10:57:30 2015 +0200
    
        base-files: set umask 022 via /etc/profile for sane defaults.
    
    commit a7b6662df99be1f39fa3bc1ef29479bc8183da0a
    Author: Juan RP <xtraeme@voidlinux.eu>
    Date:   Sat Jun 6 08:38:42 2015 +0200
    
        shadow: set default umask in login.defs... fixes weird perms introduced yesterday.
    
    commit 284b8a98d60c76c4d984b06f0f5315272c190370
    Author: Juan RP <xtraeme@voidlinux.eu>
    Date:   Fri Jun 5 11:26:55 2015 +0200
    
        shadow: set umask 077 in /etc/login.defs to avoid world-readable home dirs.
        
        Thanks to pancake for finding this.
    
    commit 9eb2562b3be1140d3c838561903ffb226623952a
[...]

PR #3000, opened 2015-11-17
Fix default umask maddness.

ayghor:
Default umask 022, set by login.defs, which applies to, but not only

 * tty logins
 * `su` sessions
 * `su -l` sessions

issue #2994


xtraeme:
I don't get what are you solving here. The current behaviour works just fine:

- home dirs are created with 077
- umask is set to 022 for all users
- umask 022 is also set via pam in required modules

The `umask 022` setting in `/etc/profile` can be overrided via your `~/.profile` or similar for your shell.

Can you explain why your changes are required?

ayghor:
1 - usergroups setting
2 - scattered umask all over teh place

xtraeme:
And can you explain why your changes are better than what's there currently? to me this just adds another stupid indirection.

ayghor:
Usergroups:

Users get `umask 002` when their primary group has the same name. This
is useful when you have `setgid` directories.

Consistency:

This way you set 022 in one place, and 077 in another place. Teh current
state is setting 022 for each shell and on /etc/pam.d{su,su-l}.

I have left some comments about what is being done too.

ayghor:
If people must set in `~/.profile`, it is not default and it won't work on non-login shells.

ayghor:
any thoughts on this?

ayghor:
Once more, I updated teh system and got my users umask broken. It should be 002, not 022. This is a bug.

xtraeme:
#3480 is already there.

PR #2934, opened 2015-11-10
hooks/do-pkg: clean up 00-gen-pkg.sh

Gottox:

PR #3001, opened 2015-11-17
globs files defined in conf_files=

Gottox:
Depends on https://github.com/voidlinux/void-packages/pull/2934

### DIFF
--- a/common/hooks/pre-pkg/99-pkglint.sh
+++ b/common/hooks/pre-pkg/99-pkglint.sh
@@ -1,5 +1,7 @@
 # This hook checks for common issues related to void.
 
+source ${XBPS_COMMONDIR}/environment/pkg/extglob.sh
+
 hook() {
 	local error=0 filename= rev= libname= conflictPkg= conflictFile=
 	local conflictRev= ignore= found= mapshlibs=$XBPS_COMMONDIR/shlibs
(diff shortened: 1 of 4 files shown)


Vaelatern:
Doesn't work.
Tested with xinetd.
When I tried with `{}` globbing, `=> ERROR: xinetd-2.3.15_6: '/etc/xinetd.{conf,d/serv{ers,ices}}' configuration file not in PKGDESTDIR!`
When I tried with `/etc/xinetd.*` I say a notice, but not warning or error, `/void-packages/common/hooks/pre-pkg/99-pkglint.sh: line 29: [: /destdir//xinetd-2.3.15//etc/xinetd.conf: binary operator expected`

Gottox:
Please recheck. brace expansion is not supported.

common/hooks/pre-pkg/99-pkglint.sh
```diff
@@ -1,5 +1,7 @@
 # This hook checks for common issues related to void.
 
+source ${XBPS_COMMONDIR}/environment/pkg/extglob.sh
```
xtraeme: Hmm those both two lines sourcing the files shouldn't be required.

Gottox: You're right for ``00-gen-pkg``. But ``99-pkglint`` does not source this file.

Vaelatern:
Checked with lightdm this time.

    conf_files= "
         /etc/lightdm/*.conf                                                     
         /etc/pam.d/lightdm*"                                                    

and the necessary files were created, listed below.

    conf_files="
	    /etc/lightdm/keys.conf
	    /etc/lightdm/lightdm.conf
	    /etc/lightdm/users.conf
	    /etc/pam.d/lightdm
	    /etc/pam.d/lightdm-greeter
	    /etc/pam.d/lightdm-autologin"

I can't do `conf_files="/etc/**/*"`, the regex needs to be a little precise about files.

Gottox:
All this bash madness is not supported at the moment. Just plain simple * and ? expansions.

Let's focus on that and extend it later if needed.

Gottox:
Any further comments?

PR #3004, opened 2015-11-18
urxvt-perls: remove xsel dependency

rolf1304:
The clipboard script does not depend upon xsel. It can be configured to use any command.

Sorry, this time with just one commit.


Gottox:
You can rearrange/rename commits without reopening the PR: `git rebase -i HEAD~10; git push -f`

PR #3006, opened 2015-11-18
New package: enigma-1.21

pullmoll:
Also add required dependency xerces-c-3.1.2


srcpkgs/enigma/files/remove-SDL_ttf-check.patch
```diff
@@ -0,0 +1,81 @@
...
+-int
+-main ()
+-{
+-return main ();
+-  ;
+-  return 0;
+-}
+-_ACEOF
+-if ac_fn_c_try_link "$LINENO"; then :
+-  ac_cv_lib_SDL_ttf_main=yes
```
xtraeme: You can get rid of this patch completely (which is bogus because it won't be valid if it's regenerated) and simply set `ac_cv_lib_SDL_ttf_main=yes` in `configure_args`.

pullmoll: Hmm... that doesn't work for me. I still get
```
checking for SDL_ttf >=2.0.6... configure: error: in `/builddir/enigma-1.21':
configure: error: cannot run test program while cross compiling
See `config.log' for more details
=> ERROR: enigma-1.21_1: failed to run do_configure() at line 7.
```
I'll try to find out what else is missing to disable the check.

pullmoll:
Patch `configure.ac` and run `autoreconf -if` to get rid of the `SDL_ttf` check which won't work for cross compiling.

srcpkgs/enigma/patches/musl-redirect.patch
```diff
@@ -0,0 +1,31 @@
...
+            "    --showfps      Show the framerate (FPS) during the Game\n"
+            "    --version      Print the executable's version number\n"
+@@ -298,6 +300,7 @@
+ #ifdef WIN32
+     ap.redirect = true;
+ #endif
++#if defined(__GLIBC__)
+     if (ap.redirect) {
+         FILE *newfp;
+         newfp = std::freopen((userStdPath + "/Output.log").c_str(), "w", stdout);
```
chneukirchen: freopen stdout should work with musl, just stdout = fopen doesnt.

pullmoll: Thanks. I'm going to try to modfiy the patch accordingly.

chneukirchen:
I'm not sure why the build on i686 breaks, it works fine locally... can anyone reproduce?

chneukirchen:
It builds here with i686 and pie, do you have any error logs?

pullmoll:
You can see the failure here http://build.voidlinux.eu/builders/i686_builder/builds/18572/steps/shell_3/logs/stdio

I used `configure_args="--disable-pretty-make"` to see what was going on and it just failed without error. I'll create a build log and upload it...

pullmoll:
DUH! Disabling `XBPS_CCACHE=yes` in my `etc/conf` also gets me rid of this error! So it looks like it is some kind of ccache corruption.

Perhaps the file `src/xercesc/util/PlatformUtils.cpp` which was compiled for `x86_64` gets into the way of the `i686` build!? This is rather strange.

We could try to revert the change to xbps-src from a couple of months ago which made all compilers share just one ccache directory and see if that fixes the build.

PR #3007, opened 2015-11-18
libwebsockets: update to 1.5

raedwulf:
Update libwebsockets to 1.5 - note that /usr/lib/cmake is no longer installed due to upstream change:
http://git.libwebsockets.org/cgi-bin/cgit/libwebsockets/commit/CMakeLists.txt?id=3ae1badae7a05e0982e0dfbcb078da3d4b92a81d


raedwulf:
The Travis-ci check is failing because travis couldn't download from the upstream server in the last build platform...

Gottox:
restarting the failing build.

PR #3008, opened 2015-11-19
lua: fix documentation

tonkazoid:


Gottox:
Can you explain this a little further? Why was it broken?

Duncaen:
```
$ ls -lsa /usr/share/doc/lua
0 lrwxrwxrwx 1 root root 6 12. Nov 21:21 /usr/share/doc/lua -> lua5.3
```
The alternative for docs is in `lua` and the docs itself in `lua-devel`.

tonkazoid:
https://github.com/tonkazoid/void-packages/blob/lua/srcpkgs/lua/template#L55 this is currently useless, because -devel get built first and vmoves the docs.
`lrwxrwxrwx 1 root root 6 Nov 19 07:31 /usr/share/doc/lua -> lua5.2` for me, but the 5.2 directory does not exist.

Duncaen:
If you install `lua-devel` you have the docs, I'm not if the docs should be part of the main package or better in `-devel` or a new `-doc` package.

tonkazoid:
Personally i would expect them to be in the main package, but it's up to you/maintainer. I just wanted to fix a broken symlink :p

pullmoll:
IMHO a developer _usually_ needs the docs and a user _may_ need them, thus the best choice seems to be to create lua-doc, lua51-doc and lua52-doc subpackages so both can have them, if they want.

Gottox:
I would put it in -devel, but -doc is okay too. @tonkazoid can you update the PR accordingly and rebase?

tonkazoid:
I'm fine with -devel, i wanted to close this. I just was irritated by the dangling/(broken) symlink
and thought it would be ok in the main-package (since -devel depends on it).

PR #3009, opened 2015-11-19
New Package: cbp2make v1.4.7

the-maldridge:


srcpkgs/cbp2make/template
```diff
@@ -0,0 +1,35 @@
...
+hostmakedepends="doxygen p7zip"
+short_desc="Makefile generation tool for Code::Blocks IDE"
+maintainer="Michael Aldridge <aldridge.mac@gmail.com>"
+license="GPL-3"
+homepage="http://sourceforge.net/projects/cbp2make/"
+distfiles="${SOURCEFORGE_SITE}/$pkgname/cbp2make-stl-rev$_version-all.tar.7z"
+checksum=1b211abb8de00dc3048fccad6ebd076ab03dcb9f672cdff379de33a1346ed129
+
+case "$XBPS_TARGET_MACHINE" in
+	*-musl) broken=Build fails on muslc;;
```
xtraeme: Variables with whitespaces must be quoted...

srcpkgs/cbp2make/template
```diff
@@ -0,0 +1,35 @@
...
+
+pre_build() {
+	cp cbp2make.cbp.mak.unix Makefile
+}
+
+do_build() {
+	sed -i	"s/CXX = g++/CXX = $CXX/" Makefile
+	sed -i	"s/CC = g++/CC = $CC/" Makefile
+	sed -i	"s/LD = g++/LD = $CXX/" Makefile
+	make
```
xtraeme: Rather than changing the `Makefile`, but don't you use `make CC=$CC CXX=$CXX LD=$CXX` instead?

the-maldridge:
PR Updated so all checks pass again.

srcpkgs/cbp2make/template
```diff
@@ -0,0 +1,35 @@
+pkgname=cbp2make
+version=1.4.7
+revision=1
+_version=147 #Same as above, but needed for the download link
```
dominikh: Instead of using an extra variable, you can use `${version//./}` in the URL.

srcpkgs/cbp2make/template
dominikh: @the-maldridge can you address this? Thanks.

the-maldridge: Havind done the line as suggested, this no longer builds, it seems that the values are not being honored by make.  Any suggestions?

PR #3018, opened 2015-11-20
New package: uthash-1.9.9

diogoleal:


srcpkgs/uthash/template
```diff
@@ -0,0 +1,23 @@
+# Template file for 'uthash'
+pkgname=uthash
+version=1.9.9
+revision=1
+short_desc="A hash table for C structures"
+maintainer="Diogo Leal <diogo@diogoleal.com>"
+license="CUSTOM"
```
xtraeme: CUSTOM is not a valid license... in fact this software uses BSD license, iirc.

diogoleal: Fixed

Duncaen:
Sorry, a little bit late but why do you install `license.html`, in the root dir is a `LICENSE` file in plain text?

pullmoll:
It's in the tarball as well. Would you fix it? I have a number of commits in the queue.

Duncaen:
switched the file in 8e5745d

PR #3019, opened 2015-11-20
elfutils: update to 0.164

pullmoll:
Now also *-musl compiles.

The amount of patches for musl libc is tremendous. Most are in the spirit of the previous patch set.

In contrast to the previous version now musl-fts and musl-obstack are used (and integrated into configure.ac).

The most critical change is probably the replacement of on_exit() with atexit() and relying on the global error_message_count to tell if the exit is due to a failure.

Is there a way to read the exit code in an atexit() function?


xtraeme:
By reading the POSIX manual pages, looks like atexit() is the portable interface, and on_exit() is the unportable interface available in glibc and others.

Looks like they should be equivalent.

I think your way of using a global variable is the only way to get the exit code on atexit().

pullmoll:
Yup, I also think so. We can try and see how it behaves on a musl system in the error case.
OTOH I don't know why removing a temporary file in the non-error case could hurt *shrug*

somasis:
I was working on porting elfutils for my distribution last night, are any of these patches upstream?

pullmoll:
No, I devised them from the previous patch set and don't know if @xtraeme had tried to get that upstreamed. I don't know, if Fedora is in any way interested in musl libc support.

For glibc there are currently no patches required.

PR #3020, opened 2015-11-20
New package: freedroidRPG-0.16

pullmoll:


pullmoll:
There's some issue with the in game font for my German environment not displaying the diacritic characters. I'm not sure if this is due to `LANG=de_DE.UTF-8` or if the translations of texts are encoded using the wrong character encoding.

PR #3033, opened 2015-11-22
xerces-c: fix distfile url, add License, add xerces-c-doc package.

yopito:


pullmoll:
The Apache-2.0 license is already installed in `/usr/share/licenses/Apache-2.0`, so there's no need for `vlicense ...` AFAIR.

Instead of using `dos2unix` in the host, you could just use `sed -i doc/*.xml -e's;\015;;'` to remove the extra CRs in doc/*.xml.

yopito:
Thanks for your comments, I will take care of them ASAP.

PR #3034, opened 2015-11-22
New package: gsmartcontrol 0.8.7

yopito:


yopito:
travis-ci fails on armv*-musl : probably related and fixed to recent commit "gtkmm2: rebuild for arm/musl symbol versioning"
(https://github.com/voidlinux/void-packages/commit/b1f9f2c38564784709fd4d634a2bb1601b5e870e)

I will push a fake update of this pull request to make travis-ci try again.

pullmoll:
There's no need to install licenses for libraries this package links against. I don't think it uses a shipped zlib, pcre, or boost. The gpl2 and gpl3 license files are redundant, because they already exist in `/usr/share/licenses`

PR #3039, opened 2015-11-23
gnome: batch update packages

pullmoll:
The `grilo-0.2.14` package has a revbump back from *.so.10 to *.so.1 !?
Seems a little strange, but works after removing the other grilo entries from `common/shlibs`.
The dependencies are either revbumped or updated.

Packages tested to build ok on x86_64 and x86_64-musl.

What shall we do with `gdm`? It is now at `3.18.2` and could probably also be updated?


xtraeme:
gdm must be kept as is because 3.18 does not have support for consolekit and requires systemd.

Gottox:
grilo-plugins-0.2.16_1: broken, unresolvable shlib `libgrilo-0.2.so.10'

Gottox:
Fixed by d8aa7cf0fa7b39c9436edc04aeb00192992780b9. Please don't break shlibs.

PR #3040, opened 2015-11-23
ncmpcpp: updated to 0.7

rolf1304:


srcpkgs/ncmpcpp/template
```diff
@@ -8,13 +8,13 @@ configure_args="
  --enable-visualizer --with-curl --with-taglib --with-fftw"
 hostmakedepends="pkg-config"
 makedepends="boost-devel ncurses-devel readline-devel libmpdclient-devel
- libcurl-devel taglib-devel fftw-devel"
+ libcurl-devel taglib-devel fftw-devel icu-devel"
```
xtraeme: `icu-devel` -> `icu55-devel`

PR #3047, opened 2015-11-25
New package: tinywm-1.3

diogoleal:


pullmoll:
It may be cool to have a 50 line C code window manager, but does it have any practical value?

diogoleal:
It can use it.

I did some tests before and it worked.

PR #3056, opened 2015-11-26
strongswan: updated to 5.3.5.

Vaelatern:
Also ensured subdirectories of `/etc/ipsec.d/` wouldn't just be removed after the tarball extracts them.

### DIFF
--- a/srcpkgs/strongswan/template
+++ b/srcpkgs/strongswan/template
@@ -1,7 +1,7 @@
 # Template file for 'strongswan'
 pkgname=strongswan
-version=5.3.4
-revision=2
+version=5.3.5
+revision=1
 build_style=gnu-configure
 configure_args="--disable-static"
 depends="iproute2 sqlite"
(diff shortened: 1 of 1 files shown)


srcpkgs/strongswan/template
```diff
@@ -12,8 +12,16 @@ maintainer="Toyam <Vaelatern@gmail.com>"
 license="GPL-2"
 homepage="http://www.strongswan.org/"
 distfiles="http://download.strongswan.org/${pkgname}-${version}.tar.bz2"
-checksum=938ad1f7b612e039f1d32333f4865160be70f9fb3c207a31127d0168116459aa
+checksum=2c84b663da652b1ff180a1a73c24a3d7b9fc4b9b8ba6bd07f94a1e33092e6350
 
 post_install() {
+	touch $DESTDIR/etc/ipsec.d/reqs/.empty
```
xtraeme: Those dirs don't exist, so that `touch(1)` fails.

Vaelatern: Except that line works on my machine... It didn't work without `$DESTDIR` there, but the dirs exist. If I don't touch them, the remove-empty-directory hook removes them in `post_install`.

xtraeme:
By default `vmkdir` creates dirs with mode 0755, so that specifying 0755 is redundant.

PR #3064, opened 2015-11-27
Added musl suport to chromium package

amon-ra:
Added patches and tested x86_64-musl


xtraeme:
Did you test glibc build? note that both glibc and musl must build correctly.

amon-ra:
I have not tested other archs, ti takes about 8h to build. But i have
revised the patches and does not modify other archs.


2015-11-28 2:51 GMT+01:00 Juan RP <notifications@github.com>:

PR #3071, opened 2015-11-28
lvm2: update to 2.02.136.

thypon:


thypon:
The current lvm2 distfile was removed.

PR #3073, opened 2015-11-28
New package: fbreader-0.99.4

diogoleal:


pullmoll:
In this case it looks like there should *not* be a -devel subpackage. This is no library but an application. There are no header files, there is no pkconfig file... and there is no fbreader-dev(el) package in any other distribution AFAICT.

pullmoll:
The common/shlibs additions are unnecessary as well...

srcpkgs/fbreader/template
```diff
@@ -0,0 +1,22 @@
+# Template file for 'fbreader'
+pkgname=fbreader
+version=0.99.4
+revision=1
+build_style=gnu-makefile
+hostmakedepends="pkg-config"
+makedepends="fribidi-devel bzip2-devel expat-devel libcurl-devel libunibreak-devel sqlite-devel qt-devel zlib-devel"
+depends="qt"
```
xtraeme: This line is bogus

srcpkgs/fbreader/template
```diff
@@ -0,0 +1,22 @@
...
+license="GPL-3"
+homepage="http://www.fbreader.org/"
+distfiles="http://www.fbreader.org/files/desktop/${pkgname}-sources-${version}.tgz"
+checksum=3d7c31d5ea314589d2a963290ad16f4d3d631a41e802b8b39f8be0c9f71eb8e9
+
+do_build() {
+	sed -e "s:-llinebreak:-lunibreak:" -i makefiles/config.mk zlibrary/text/Makefile
+	export TARGET_ARCH=desktop
+	export UI_TYPE=qt4
+	export TARGET_STATUS=release
```
xtraeme: Those exports are bogus too

Issue #3082, opened 2015-11-29
Package Request: Links


ghost:
http://links.twibright.com/download.php

Vaelatern:
Working on enabling SSL

xtraeme:
Already there

PR #3084, opened 2015-11-30
New package: links-2.12.

Vaelatern:
Request #3082

### DIFF
--- a/srcpkgs/links/template
+++ b/srcpkgs/links/template
@@ -1,0 +1,18 @@
+# Template file for 'links'
+pkgname=links
+version=2.12
+revision=1
+build_style=gnu-configure
+configure_args="--with-ssl --enable-graphics"
+makedepends="libpng-devel libjpeg-turbo tiff-devel bzip2-devel zlib-devel gpm-devel libressl-devel"
+short_desc="Graphics and text mode web browser, released under GPL"
+maintainer="Toyam Cox <Vaelatern@gmail.com>"
+license="GPL-2"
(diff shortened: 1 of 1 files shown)


srcpkgs/links/template
```diff
@@ -0,0 +1,18 @@
+# Template file for 'links'
+pkgname=links
+version=2.12
+revision=1
+build_style=gnu-configure
+configure_args="--with-ssl --enable-graphics"
+makedepends="libpng-devel libjpeg-turbo tiff-devel bzip2-devel zlib-devel gpm-devel libressl-devel"
+depends="libpng>1.2.18 libjpeg-turbo tiff"
```
xtraeme: This depends line is bogus, remove it.

srcpkgs/links/template
```diff
@@ -0,0 +1,18 @@
...
+short_desc="Graphics and text mode web browser, released under GPL"
+maintainer="Toyam Cox <Vaelatern@gmail.com>"
+license="GPL-2"
+homepage="http://links.twibright.com/"
+distfiles="http://links.twibright.com/download/links-${version}.tar.gz"
+checksum=2e3da6193de9d6ee71bc4beab17169ecc58d7cbf2e20192a3f518ed9f7f22e33
+
+post_install() {
+	mv doc/ /usr/share/doc/links
```
xtraeme: This is completely wrong... you are moving a directory to masterdir and not the pkg destdir. Use `mv foo ${DESTDIR}/...`

PR #3088, opened 2015-11-30
New package: io-2015.11.11

diogoleal:


xtraeme:
The development pkg is useless without headers; and so are the changes to `common/shlibs`.

srcpkgs/io/template
```diff
@@ -0,0 +1,28 @@
+# Template file for 'io'
+pkgname=io
+version=2015.11.11
+revision=1
+nocross=yes
+only_for_archs="x86_64 armv6hf armv7hf"
```
xtraeme: `armv[67]hf` is not a valid architecture... use `armv[67]l` instead.

srcpkgs/io/template
```diff
@@ -0,0 +1,28 @@
...
+short_desc="A dynamic prototype-based programming language"
+maintainer="Diogo Leal <diogo@diogoleal.com>"
+license="BSD"
+homepage="http://iolanguage.org"
+distfiles="https://github.com/stevedekorte/io/archive/${version}.tar.gz"
+checksum=00d7be0b69ad04891dd5f6c77604049229b08164d0c3f5877bfab130475403d3
+
+pre_build() {
+	cmake . -DCMAKE_INSTALL_PREFIX=/usr
+}
```
xtraeme: What's the reason to not use `build_style=cmake` ?

diogoleal: I include this cmake because of some tests. already returned

diogoleal:
What do you think of each addons have a package for him?
For example, the directory ```/usr/lib/io/addons/Socket```

It would be the package: ```io-addons-socket``` (example)

For the binary does not need it for basic operation.

Gottox:
https://travis-ci.org/voidlinux/void-packages/jobs/96937548

Vaelatern:
Closing due to no activity. The pkg still points to the latest release, but this doesn't build at all on x86_64.

PR #3091, opened 2015-11-30
dhcpcd: redirect stderr to stdout

abuibrahim:


xtraeme:
why?

abuibrahim:
Need all logs redirected to stdout to be able to add an accompanying logger like bellow.

cat log/run
#!/bin/sh
exec svlogd -tt /var/log/dhcpcd

Duncaen:
dhcpcd already loggs to syslog, which is the "standard". dhcpcd has also has a logfile option as flag or config variable to log to a file instead of syslog.

abuibrahim:
Thanks for feedback Duncaen.

I prefer each daemon having a separate logger (svlogd) instead of all logs going to a single bottleneck (syslog). True, I can add "OPS='--logfile blah' option in conf file but it doesn't give me log rotation and other goodies as svlogd.

I understand syslog is "standard" but was wondering if I can take care of my use case without breaking the "standard" behaviour.

The thing is, with -B  dhcpcd also logs to stdout with an occasional message like "starting wpa_supplicant" going to stderr. Now I see the reason why "1>&2" was there in the first place. 

I can achieve what I want by adding 'exec 2>&1' to my /etc/sv/dhcpcd/conf.

Please ignore this PR.

Regards,
Ruslan

PR #3093, opened 2015-11-30
Use default Monospace font for st.

rogerbraun:
As st configures its font during compile time, running it without the right font can be pretty unusable. If someone wants to run it with a different font they'll have to recompile it anyway.

An example without terminus-font installed:
st_ugly


Duncaen:
Not sure about this, should we better switch to the default monospace font?

rogerbraun:
@duncaen, which font is that?

Duncaen:
That would be just `Monospace` as an alias to a font, fontconfig chooses.
I have to test if that works.

Duncaen:
In a chroot without DejaVu fonts and Terminus installed i get terminus.
```
[duncan@/home/duncan/repos/github.com/Duncaen/void-packages/masterdir /]$ fc-match Monospace
ter-112n.pcf.gz: "Terminus" "Regular"
```

rogerbraun:
I'll compile it with Monospace and see how it looks.

rogerbraun:
This is how it looks, top is DejaVu as default Monospace font, bottom is Terminus.

st_variants

rogerbraun:
Also, the default font size is really small with 12 pixels... I'd suggest bumping it to 14, but that really is another thing.

PR #3099, opened 2015-12-02
New package: shellcheck-0.4.1

diogoleal:


pullmoll:
Setting `LDFLAGS` and `CPPFLAGS` empty is not required (verified locally).
It seems ghc can't cross compile, so `nocross=yes` should be set.

PR #3101, opened 2015-12-02
New package: snapwm-2.0.2

diogoleal:


pullmoll:
Remove `do_buIld() { ... }` and add `CFLAGS="-std=c99"` to the template body.

PR #3105, opened 2015-12-02
New Package: binwalk-2.0.1

ananteris:


xtraeme:
Already pushed by @Duncaen

PR #3107, opened 2015-12-03
New package: hdf5-1.8.16

pulux:


srcpkgs/hdf5/template
```diff
@@ -0,0 +1,28 @@
+# Template file for 'hdf5'
+pkgname=hdf5
+version=1.8.16
+revision=1
+build_style=gnu-configure
+only_for_archs="i686 x86_64 x86_64-musl"
```
xtraeme: what's the reason for `only_for_archs` ?

srcpkgs/hdf5/template
```diff
@@ -0,0 +1,28 @@
...
+homepage="http://hdfgroup.org"
+distfiles="http://www.hdfgroup.org/ftp/HDF5/current/src/hdf5-${version}.tar.gz"
+checksum=ed17178abd9928a7237f30370189ba767b9e39e0db45917c2ac4665eb9cb4771
+
+post_install() {
+	vlicense COPYING
+}
+
+hdf5-devel_package() {
+	depends="zlib-devel"
```
xtraeme: `depends="${makedepends} ${sourcepkg}-${version}_${revision}"`

srcpkgs/hdf5/template
pulux: on armx I get:
```
configure: error: cannot run test program while cross compiling
See `config.log' for more details
=> ERROR: hdf5-1.8.16_1: failed to run do_configure() at line 7.
```
the log says:
`#define PRINTF_LL_WIDTH "unknown"`

pullmoll: The switch to use in this case is `nocross=yes` then, because the package would probably compile on a native arm* build.

xtraeme:
Please rebase to get this merged, thanks.

PR #3109, opened 2015-12-03
common: merge only_for_archs and noarch=yes into one.

Gottox:
* noarch=yes is replaced with archs=noarch
* only_for_archs= is renamed to archs=
* archs= allows the use of wildcards and negations; first matching rule applies:
  * archs="*-musl" will build the pkg only for musl-libcs
  * archs="~*-musl *" will build the pkg only on non-musl-libc
  * archs="x86_64-musl ~*-musl *" will build for x86_64-musl and any non-musl
    arch.
* archs= defaults to "*"


thypon:
why not using "archs=any"?

Gottox:
For convenience:
* `archs="i686 x86_64"` produces the packages `pkgname-version_revision.i686.xbps` and `pkgname-version_revision.x86_64.xbps`
* `archs=noarch` produces `pkgname-version_revision.noarch.xbps`

Gottox:
Any thoughts on this?

Vaelatern:
How does the transition to `archs` work?

Gottox:
```
grep only_for_archs `grep noarch= */template -l`
# ... Resolve them manually ...
sed -i "s/noarch=.*\|/archs=noarch/;s/only_for_archs=/archs=/" srcpkgs/*/template
```
This should be enough.

Vaelatern:
So this needs changes to every template to build correctly after. Probably a good time to make more case statements to mark certain archs as broken, so as to ease transitions when new arches like `aarch64` are added in the future.

Vaelatern:
I'm in favor of merging this, so long as there is a definite policy going forward of what we mark with `archs` and what we mark as broken.

ebfe:
ping

Gottox:
@xtraeme had doubts on this.

Vaelatern:
At this point our workflows are adequate. This doesn't help in the `broken` case, and only_for_archs is only used in 109 templates.

```
files="$(grep --color=never only_for_archs srcpkgs/*/template | cut -f 1 -d ':' | sed 's:/template::')"
for file in $(echo $files); do [ -h ./$file ] && echo $file; done | wc -l
```

Move to close?

Gottox:
I think we need to add backwards compatibility here too.

Vaelatern:
Also I need to find a way to make this style of archs work in dxpb.
Acceptable backwards compatibility: on `read_pkg`, generate `only_for_archs` based on all archs that xbps-src knows.

maxice8:
ping

Gottox:
Any vetos on merging? Will merge tomorrow :)

Gottox:
It's backwards compatible, so as long as we don't change the templates we're good, are we?

Vaelatern:
We are alright so long as noarch is still noarch, and only_for_archs isn't removed.

So this can be merged, but not used until dxpb supports it?

Vaelatern:
Does not seem to be backwards compatible from what I've read.

I can work on a compatibility layer within xbps-src for printing information.... that might be the easiest course of action due to wildcards being completable within xbps-src.

PR #3122, opened 2015-12-05
valgrind: unbreak musl

Duncaen:


srcpkgs/valgrind/patches/musl-fixes.patch
```diff
@@ -0,0 +1,45 @@
...
++#  define  VG_Z_LIBC_SONAME  libcZdZa                // libc.*
+ 
+ #elif defined(VGO_darwin) && (DARWIN_VERS <= DARWIN_10_6)
+ #  define  VG_Z_LIBC_SONAME  libSystemZdZaZddylib    // libSystem.*.dylib
+@@ -274,7 +274,7 @@
+ /* --- Soname of the pthreads library. --- */
+ 
+ #if defined(VGO_linux)
+-#  define  VG_Z_LIBPTHREAD_SONAME  libpthreadZdsoZd0     // libpthread.so.0
++#  define  VG_Z_LIBPTHREAD_SONAME  libcZdZa              // libc.*
```
xtraeme: This won't work for glibc....

Duncaen: Thanks for the review @xtraeme I added a condition for glibc to still use `libpthread`.
Its now a mix of the patch from alpine[0] and openwrt[1].

[0] http://git.alpinelinux.org/cgit/aports/tree/main/valgrind/musl-fixes.patch
[1] https://github.com/openwrt-mirror/openwrt/blob/master/package/devel/valgrind/patches/200-musl_fix.patch

PR #3125, opened 2015-12-06
New package: go-ipfs-0.3.9

CMB:
Closes issue 3037.  Also kudos to user "7185" on github.


Duncaen:
Merged in 95a68f somehow github didn't notice it.

dominikh:
@CMB Why was this package named "go-ipfs"? Upstream seems to call it ipfs, and the binary is called ipfs, too.

CMB:
@dominikh Because the official project name does seem to be go-ipfs.
Any reason not to stick with that?
Also for what it's worth, Arch is using the same name.

dominikh:
> Because the official project name does seem to be go-ipfs

Can you point me to official documentation saying that its name is go-ipfs? All of the documentation I can find only acknowledges that it is implemented in Go, but not that there are alternate implementations. The version string also makes no mention of it:

```
> ipfs version
ipfs version 0.3.8
```

/cc'ing @jbenet to ask him what he would like distributions to call the package for the Go IPFS implementation. `ipfs` or `go-ipfs`?

CMB:
I'm sure I got the idea that the official name was go-ipfs from reading
https://ipfs.io/docs/ where it mentions the go-ipfs implementation.
Also all of their source and binary archives are packaged as
go-ipfs-x.y.z.archive_extension.
I don't have any kind of attachment to the current package name.  As you
say, the version string and binary name are both ipfs.  This is a
strong reason for naming the package ipfs, so a name change would be
fine, unless @jbenet wants it packaged as go-ipfs by the distributions.

PR #3128, opened 2015-12-08
New package: frotz-2.44

diogoleal:


xtraeme:
- Compilation does not use our `CFLAGS` nor `LDFLAGS`
- Compilation does not use the correct directory for the configuration directory

xtraeme:
```
cc -O2 -DCONFIG_DIR="\"/usr/local/etc\""  -DVERSION="\"2.44\"" -DCOLOR_SUPPORT    -o src/curses/ux_init.o -c src/curses/ux_init.c
```

My two comments still remain.

PR #3134, opened 2015-12-08
New package: minised-1.15

dicktyr:


xtraeme:
`license=custom` is wrong. Its license is just a standard 3-clause-BSD:

http://svn.exactcode.de/minised/trunk/LICENSE

Change license to `license="3-clause-BSD"`

dicktyr:
> Change license to `license="3-clause-BSD"`

ok
done

but I'm confused by the manual which says license should be

	A string matching any license file available in /usr/share/licenses

and I see only "BSD"

PR #3143, opened 2015-12-10
New package: weather-2.0

dicktyr:


srcpkgs/weather/template
```diff
@@ -0,0 +1,38 @@
+# Template file for 'weather'
+pkgname=weather
+version=2.0
+revision=1
+# upstream supports both python2 and python3
```
Vaelatern: It is expected you have a different subpackage for different versions of python. https://github.com/voidlinux/void-packages/blob/master/Manual.md#pkgs_python

srcpkgs/weather/template
```diff
@@ -0,0 +1,38 @@
...
+noarch=yes
+depends="python"
```
Vaelatern: "To allow cross compilation, the python-devel package (for python 2.7) must be added to hostmakedepends and makedepends. If any other python version is also supported, for example python3.4, those must also be added as host and target build dependencies."

srcpkgs/weather/template
```diff
@@ -0,0 +1,38 @@
...
+pycompile_module="weather.py"
```
Vaelatern: It's not byte compiled. @xtraeme doesn't this go against documentation to NOT have this byte-compiled? Or am I reading the manual wrong?

srcpkgs/weather/template
```diff
@@ -0,0 +1,38 @@
...
+maintainer="Richard Taityr <dicktyr@yahoo.co.uk>"
+license="ISC"
+homepage="http://fungi.yuggoth.org/weather/"
+distfiles="http://fungi.yuggoth.org/${pkgname}/src/${pkgname}-${version}.tar.xz"
+checksum=807aff002d31cd81aeaae096981c28900323c405b8f7d36625d9c7e48719347a
+
+do_install() {
+	vbin weather
+	vinstall weather.py 644 usr/lib/python2.7/site-packages
+	vinstall weather.py 644 usr/lib/python3.4/site-packages
```
Vaelatern: Again, different subpackages for different python versions.

srcpkgs/weather/template
dicktyr: it's not clear to me how to separate the two (2.7 from 3.4) using only a subpackage

i.e. if the 3.4 subpackage is installed
the 2.7 will also be installed because it's in the main package
but maybe somebody wants only the 3.4

Vaelatern: There are examples in the srcpkgs

dicktyr: yes I've been looking for one

Vaelatern: Good example is `python-urllib`

dicktyr: thanks
but that doesn't answer my question

it's not a python-module so I assume I need do_install
and that's why it's not clear to me
how the two versions should be separated

srcpkgs/weather/template
stpx: This line is correct.

Python code is byte-compiled, during the package configure phase, after the package has been unpacked via `xbps-install(1)`; or manually via `xbps-reconfigure(1)`.

srcpkgs/weather/template
stpx: @dicktyr I think you can just choose the python version of your preference. If or when someone wants/needs the other version, we can add a subpackage.

dicktyr: ok thanks
I don't mind making it work for both
but I need more guidance to do that

`./xbps-src pkg weather` fails with

    => ERROR: python-2.7.11_1: failed to install 'libressl-devel-2.2.5_1' dependency! (error 19)

stpx: Your repo is outdated: 
```sh
$ git pull --rebase upstream master
```

dicktyr: thanks
I'm learning about git and github as well

srcpkgs/weather/template
Vaelatern: I saw, when I tested an install, a `weather.py` file that doesn't look like a `.pyc` file or byte compiled. Perhaps I saw it wrong.

dicktyr:
I believe I've addressed all comments/concerns
but this request remains unaccepted

it's also coming from my master branch
(because I had not yet learned to isolate
each pull request in a separate branch)
and I've been holding back other pull requests
until it is accepted or rejected

PR #3146, opened 2015-12-11
New package: pjproject-2.4.5

CMB:


srcpkgs/pjproject/template
```diff
@@ -0,0 +1,65 @@
...
+}
+
+pjproject-devel_package() {
+	depends="${makedepends} pjproject>=${version}_${revision}"
+	short_desc+=" - development files"
+	pkg_install() {
+		vmove "usr/include"
+		vmove "usr/lib/*.a"
+		vmove "usr/lib/*.so"
+	}
```
xtraeme: `vmove usr/lib/pkgconfig`

srcpkgs/pjproject/template
```diff
@@ -0,0 +1,65 @@
...
+	sed -i -e "s#../../third_party/libsamplerate/src/samplerate.h#samplerate.h#" pjmedia/src/pjmedia/resample_libsamplerate.c
+}
+
+do_build() {
+	make dep LD=$CC
+	make -j1 LD=$CC
+}
+
+
+pjproject-pjsua_package() {
```
xtraeme: I don't see why this subpkg is necessary, has not extra dependencies and is just a single executable.

Could you please merge it into the main pkg? (use vbin via post_install).

CMB: Merged the executable into the main package and moved pkgconfig to
-devel.  Thank you.

PR #3148, opened 2015-12-11
New package: pnmixer-0.6.1

diogoleal:


srcpkgs/pnmixer/template
```diff
@@ -0,0 +1,18 @@
+# Template file for 'pnmixer'
+pkgname=pnmixer
+version=0.6.1
+revision=1
+build_style=gnu-makefile
```
xtraeme: The `configure` script is generated by GNU autoconf? if this is true you should use `build_style=gnu-configure` instead.

srcpkgs/pnmixer/template
```diff
@@ -0,0 +1,18 @@
...
+hostmakedepends="autoconf pkg-config intltool automake"
+makedepends="glib-devel libX11-devel alsa-lib-devel gtk+3-devel libnotify-devel"
```
xtraeme: - `autoconf` dependency is redundant.
- `glib-devel` and `libX11-devel` dependencies are redundant.

srcpkgs/pnmixer/template
```diff
@@ -0,0 +1,18 @@
...
+short_desc="Volume mixer for the system tray"
+maintainer="Diogo Leal <diogo@diogoleal.com>"
+license="GPL-3"
+homepage="https://github.com/nicklan/pnmixer"
+distfiles="https://github.com/nicklan/pnmixer/archive/v${version}.tar.gz"
+checksum=2c611cd8d7bc5b11fbf2f8610ef20a19b541873cc87bc18a0dc5bc9baeadc08b
+
+pre_build() {
```
xtraeme:
```
pre_configure() {
   NOCONFIGURE=1 ./autogen.sh
}
```

xtraeme:
To fix cross compilation add `glib-devel` to `hostmakedepends`.

PR #3153, opened 2015-12-11
toxcore-git: update to 20151211.

Spencer-H:


srcpkgs/toxcore-git/template
```diff
@@ -1,16 +1,16 @@
 # Template file for 'toxcore-git'
 pkgname="toxcore-git"
-version="20150604"
-revision=1
-_commithash="c259250abb86a01ae7e3fc906380417427f38ca1"
+version="20151211"
+revision=2
```
xtraeme: `revision` must be set to 1 for new packages/updates.

Spencer-H:
Yeah. Sorry I had that changed locally to force an update over the version I was using. I forgot to change it back before pushing.

Duncaen:
You can use the `-f` flag to force build/install with `xbps-src` and `xbps-install`.

Spencer-H:
Ah. I didn't know you could use that with xbps-install, that was the problem I was having. It was seeing it was the same version and using the cached version instead.

PR #3178, opened 2015-12-15
make sure that dnsmasq is installed when building libvirt

EaterOfCode:
if dnsmasq isn't installed while building libvirt, it wil be unable to find dnsmasq at runtime. and when enabling NAT for example you will be hit with errors like:
```
Cannot check dnsmasq binary dnsmasq: No such file or directory
```

since finding a path doesn't really matter on which architecture it is, I've added it to hostmakedepends. if this is wrong please correct me.

Source of solution: https://bugs.archlinux.org/task/35183 

`./xbps-src pkg libvirt` ran successfully and without error

### DIFF
--- a/srcpkgs/libvirt/template
+++ b/srcpkgs/libvirt/template
@@ -1,7 +1,7 @@
 # Template file for 'libvirt'
 pkgname=libvirt
 version=1.2.21
-revision=1
+revision=2
 build_style=gnu-configure
 configure_args="--without-hal --with-storage-lvm --with-qemu-user=libvirt
  --with-qemu-group=libvirt --without-netcf --with-interface --disable-static"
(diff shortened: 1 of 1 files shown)


Gottox:
Please increase the revision

EaterOfCode:
@Gottox as requested bumped, I didn't know I was supposed to do that, thanks!

Gottox:
Not a big deal.

Please use `git rebase -i origin/master; git push -f` merge the two commits and change the commit messages to comply with our [guidelines](https://github.com/voidlinux/void-packages/blob/master/CONTRIBUTING.md#committing-your-changes)

The change is okay, will merge afterwards.

EaterOfCode:
This should do the job, sorry for the trouble and thanks :)

Gottox:
:+1:

PR #3184, opened 2015-12-17
new package: i3-gaps

skrzyp:
Git sucks, so I needed to recreate this brach. Now with correct musl fixes.

### DIFF
--- a/srcpkgs/i3-gaps/template
+++ b/srcpkgs/i3-gaps/template
@@ -1,0 +1,40 @@
+# Template file for 'i3-gaps'
+pkgname=i3-gaps
+version=4.11
+revision=1
+build_style=gnu-makefile
+hostmakedepends="pkg-config perl git"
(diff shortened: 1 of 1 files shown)


srcpkgs/i3-gaps/template
```diff
@@ -0,0 +1,40 @@
+# Template file for 'i3-gaps'
+pkgname=i3-gaps
+version=4.11
+revision=1
+build_style=gnu-makefile
+hostmakedepends="pkg-config perl git"
+makedepends="pcre-devel yajl-devel libxcb-devel libev-devel
+ pango-devel startup-notification-devel libxkbcommon-devel
+ xcb-util-keysyms-devel xcb-util-wm-devel xcb-util-cursor-devel"
+short_desc="Improved tiling window manager"
```
xtraeme: I'd like to see a different description compared to the original i3.

PR #3200, opened 2015-12-20
New package: buku-1.5

diogoleal:


xtraeme:
https://github.com/jarun/Buku/blob/master/buku#L1

dominikh:
@diogoleal A note in general: After having addressed our comments and having pushed a new version, please write a comment on the PR. Otherwise we will not be notified by GitHub.

PR #3202, opened 2015-12-20
libdbus-c++: unbreak musl

pullmoll:
I found a way to replace the non-portable mutex init through PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP  with this code fragment at http://stackoverflow.com/a/7963765


xtraeme:
:+1:

PR #3203, opened 2015-12-20
New package: ebook-tools-0.2.2

diogoleal:


srcpkgs/ebook-tools/ebook-tools-devel
```diff
@@ -0,0 +1 @@
+ebook-tools-devel
```
xtraeme: double symlink

diogoleal: fixed.

PR #3204, opened 2015-12-20
New package: font-fira-ttf-4.202

diogoleal:


srcpkgs/font-fira-ttf/template
```diff
@@ -0,0 +1,25 @@
...
+checksum=d86269657387f144d77ba12011124f30f423f70672e1576dc16f918bb16ddfe4
+
+do_install() {
+	vmkdir usr/share/fonts/TTF
+
+	for files in ttf/*.ttf; do
+		vinstall ${files} 0644 \
+		/usr/share/fonts/TTF/
+	done
+	vlicense LICENSE
```
xtraeme: `GPL` licenses do not need to be installed.

PR #3205, opened 2015-12-21
New package: font-fira-otf-4.202

diogoleal:


srcpkgs/font-fira-otf/template
```diff
@@ -0,0 +1,25 @@
...
+checksum=d86269657387f144d77ba12011124f30f423f70672e1576dc16f918bb16ddfe4
+
+do_install() {
+	vmkdir usr/share/fonts/OTF
+
+	for files in otf/*.otf; do
+		vinstall ${files} 0644 \
+		/usr/share/fonts/OTF/
+	done
+	vlicense LICENSE
```
xtraeme: `GPL` licenses do not need to be installed.

diogoleal: @xtraeme  sorry, license correct for this package is OFL-1.1

xtraeme:
OFL-1.1 license is already provided by the `base-files` pkg.

PR #3210, opened 2015-12-22
new package: tolua

antonio-malcolm:


srcpkgs/tolua/template
```diff
@@ -0,0 +1,24 @@
...
+version=5.2.0
+revision=1
+build_style=cmake
+hostmakedepends="pkg-config cmake"
+makedepends="lua52-devel"
+short_desc="A tool that greatly simplifies the integration of C/C++ code with Lua"
+maintainer="Antonio Malcolm <antonio@antoniomalcolm.com>"
+license="GPL-2"
+homepage="https://github.com/LuaDist/tolua"
+distfiles="${homepage}/archive/${version}.tar.gz>${pkgname}-${version}.tar.gz"
```
dominikh: Renaming the file shouldn't be necessary, should it?

dominikh:
Does the package only install a libtolua.so, no versioned file? @xtraeme, how do we handle that?

antonio-malcolm:
I removed the file rename- it was something I had done for testing, locally, and neglected to remove it, before commit.

As for the unversioned .so file, I could add a post_install() which performs a `mv`, to rename the file, then adds a symlink, if it makes a difference to @xtraeme .

xtraeme:
If this package provides an unversioned soname and cannot be changed to be versioned, the soname must be registered via `shlib_provides="libtolua.so"`.

antonio-malcolm:
Added `shlib_provides="libtolua.so"`.

antonio-malcolm:
Is there anything else which needs to be done, here?
I'd like to get toluapp and conky updated, so conky will have correct/working lua binding.

dominikh:
I'm defering to @xtraeme because I'm not familiar with our lua packaging and versioning scheme.

But you're at least missing an entry in common/shlibs

antonio-malcolm:
What's missing in common/shlibs?

dominikh:
`libtolua.so tolua-5.2.0_1`

antonio-malcolm:
Ah, ok, I see the mappings in there- I'll add it. Thanks.

PR #3211, opened 2015-12-22
new package: unoconv

antonio-malcolm:


xtraeme:
`GPL` licenses do not need to be installed.

antonio-malcolm:
vlicense removed.

xtraeme:
- This is a python script and there's no dependency for it.
- This just contains a python script, hence this needs `noarch=yes`.

antonio-malcolm:
Ok, added the noarch flag and python-devel, to ensure it has libpython v2.7

xtraeme:
Your latest `makedepends` change is wrong... change this to `depends="python"`.

antonio-malcolm:
AH, you're right. Changed it to depends="python". That said, don't we need the development package for shared libs?

xtraeme:
Yes, but this pkg only provides a script and won't be linked to any shared library.

python is only required at runtime.

antonio-malcolm:
Ok, because looking at how it's packaged with other distros points to it 
needing the sahred lib. That said, it's working with libreoffice and 
gloobus-preview without issue.

On 12/23/2015 01:02 AM, Juan RP wrote:
>
> Yes, but this pkg only provides a script and won't be linked to any 
> shared library.
>
> python is only required at runtime.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/voidlinux/void-packages/pull/3211#issuecomment-166837956>.
>

-- 
/"Creative ability is best displayed with the most basic tools"/

PR #3212, opened 2015-12-22
New package: iprange-1.0.1

diogoleal:


srcpkgs/iprange/template
```diff
@@ -0,0 +1,16 @@
+# Template file for 'iprange'
+pkgname=iprange
+version=1.0.1
+revision=1
+build_style=gnu-configure
+hostmakedepends="autoconf automake"
```
xtraeme: `autoconf` is redundant.

diogoleal:
Removed autoconf.

PR #3213, opened 2015-12-22
Add sway (i3-compatible window manager for Wayland)

robotanarchy:
* More info about sway [here](https://github.com/SirCmpwn/sway).
* I've patched sway [upstream](https://github.com/SirCmpwn/sway/pull/393) for musl support.
* Tested with x86_64-musl


Gottox:
See [the contribution guidelines](https://github.com/voidlinux/void-packages/blob/master/CONTRIBUTING.md#getting-your-packages-into-void-by-yourself):
> Please note that we do not accept any packages containing non-release versions such as specific git- or svn-revisions anymore.

PR #3214, opened 2015-12-22
darktable: update to 2.0.0.

lemmi:
testing on x64 now. let it sit for a day or two.


lemmi:
dropping musl for now until i reworked the patch

lemmi:
had no problems with my usual workflow so far and no issues with the new database format.
can be merged.
i will remove the install message after 2-3 patch versions again.

PR #3219, opened 2015-12-23
python-click-log: update to 0.1.3.

okapia:


srcpkgs/python-click-log/template
```diff
@@ -1,6 +1,6 @@
 # Template file for 'python-click-log'
 pkgname=python-click-log
-version=0.1.1
+version=0.1.3
 revision=2
```
xtraeme: `revision` must be set to 1 when a new upstream version is used.

PR #3221, opened 2015-12-23
New package: libosmgpsmap-1.1.0

lemmi:
new dep for darktable


Duncaen:
And you could try to disable introspection for cross builds.

PR #3228, opened 2015-12-24
Update qtox, and toxcore to their respective new versions.

Spencer-H:
Toxcore 20151225
qtox 1.2.2

These are grouped to prevent needing to rebuild utox and qtox once toxcore pushes.

### DIFF
--- a/srcpkgs/qtox/template
+++ b/srcpkgs/qtox/template
@@ -1,7 +1,7 @@
 # Template file for 'qtox'
 pkgname=qtox
-version=1.2
-revision=3
+version=1.2.2
+revision=1
 build_style=qmake
 short_desc="QT-based TOX instant messenger client"
 maintainer="Spencer Hill <spencernh77@gmail.com>"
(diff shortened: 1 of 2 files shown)


dominikh:
1) Please change the commit messages to match the following format: `$pkgname: update to $version`
2) utox used to build fine on ARM, this new version doesn't. Can you investigate?

Spencer-H:
The problem with utox is that they are linking their image files into the executable as .o files so that they can be loaded without needing the icons to be located somewhere else. It seems that the linker on ARM refuses to link the png files masquerading as .o files. I don't really know enough about the internals of the linker to fix that, sorry.

dominikh:
The way these issues usually go is that the makefile uses a mix of host and target architectures. It's either trying to use the wrong linker, or the .o files are being generated for the wrong architecture. It seems like the makefile needs to be fixed.

Spencer-H:
Someone pushed utox on their own, so we're going to have to change the revision after to make it rebuild with the newest toxcore.

PR #3230, opened 2015-12-24
add eclipse .project filename to .gitignore

antonio-malcolm:


Gottox:
Whatever you're doing, it sounds so wrong ;)

antonio-malcolm:
HAhaha... :oD
Yeah, I'm using eclipse (or, rather, Aptana, with lots of customization). 
I prefer IntelliJ, but they've been futzing with the licensing in ways that rub me wrong.

chneukirchen:
Add it to you own `$HOME/.config/git/ignore` then.

antonio-malcolm:
Why is there an issue adding it to the repo's .gitignore?
Is there some assumption that I would be the only person ever to use an IDE?
Also, does this break something? Is .gitignore so complicated, that adding the line renders it unmaintainable?

antonio-malcolm:
In any case, if folks are simply uncomfortable updating .gitignore, I'll close this.
It's just food for thought, as more folks start contributing, with different editors and environments, that it might be a good idea to reinforce it.

dominikh:
@antonio-malcolm The point of a repo .gitignore is to ignore build artifacts and similar that are specific to the project and are the same for everyone involved. If you have some additional, personal files that need ignoring, such as junk your editor produces, then you can either use `$HOME/.config/git/ignore` or `git/info/exclude`.

(Yes. the existing .gitignore has some editor-specific files in it. Prior mistakes do not justify making more mistakes.)

antonio-malcolm:
So, ahhh, just no one bothered to remove the remaining editor-specific lines?
I've closed the pull-request, so I s'pose that's one less, then!

PR #3234, opened 2015-12-26
New package: mrxvt-0.5.4

dkociela:


Duncaen:
You don't have to close and reopen pull requests, you can force push to your branch and travis checks it again.
And removing travis environments is not a solution.

dkociela:
Thanks, I'm a newbie struggling with this. I appreciate your input. My package won't build with musl and I'm trying to find a way around. It's a worthwhile package

pullmoll:
If a build fails for travis, you can try to fix it by adding patches to solve the issue(s).
Or you can simply ignore the fact and leave it to others to try to fix the problem later on.

In this case the macro `LASTLOG_FILENAME` is not defined for musl libc. It isn't defined in glibc's `/usr/include` either, so I guess it's defined in some package header file under a condition which isn't true for musl libc.

The line `#define _PATH_LASTLOG "/var/log/lastlog"`, which probably is what's needed here, is defined in `/usr/include/paths.h` for both C libraries.

dkociela:
Thank you, very informative. I decided to "--disable-lastlog" in the
configure_args. I'm learning, and responses like yours are extremely
helpful to me.

On Sat, Dec 26, 2015 at 1:28 AM, Jürgen Buchmüller <notifications@github.com

PR #3243, opened 2015-12-28
[DO NOT MERGE] [PoC] Workaround for Git packages fetched from HEAD.

skrzyp:
While some upstream developers don't care about releases or are lazy in releasing, some users might find useful to use latest commit of their software to get newest features, but only on some software packages, where that's important.

So, here's a proof-of-concept for some impatient people like me who want to bleed from edgy sources.
In this example, we use `i3-gaps`, because its Git branch contains some fixes for ahead current release which are important for me, like not crashing itself when you use titlebars in gaps mode. But of course, I think you can use this for some "lazy" software.

It's not very important and serious as for now, more like proposal, because I want to hear some feedback about this.

### DIFF
--- a/srcpkgs/i3-gaps/template
+++ b/srcpkgs/i3-gaps/template
@@ -1,16 +1,22 @@
 # Template file for 'i3-gaps'
 pkgname=i3-gaps
-version=4.11
-revision=2
+if [ "build_option_git" ]; then
+	version=git1
+else
+	version=4.11
+fi
+revision=1
 build_style=gnu-makefile
 hostmakedepends="pkg-config perl git"
 makedepends="pcre-devel yajl-devel libxcb-devel libev-devel
  pango-devel startup-notification-devel libxkbcommon-devel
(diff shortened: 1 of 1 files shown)


xtraeme:
Nope, I won't accept this. This just adds unnecessary and redundant changes to all templates.

skrzyp:
> redundant changes 

So, here's the another way: Extend `xbps-src` to respect a `git_source` variable, which automagically makes `git` option available in specific templates and grabs source from Git repo on demand (for development purposes, lazy upstream and impatient users).

So, @xtraeme , don't be so grumpy and consider some feedback and ideas, instead of reacting like this:
grumpy-xtraeme

Duncaen:
Its a bad idea, currently you could checkout a old version of a template and it still compile it, without realeses/commits/references you can't do this.
And if your git version introduces new dependencies or changes to the buildsystem, you endup with adding more and more conditions.

Gottox:
Versions must not be defined in conditional branches otherwise the version comparision on the buildserver breaks.

Furthermore, as we said before, we won't pull non-release packages into void-packages anymore. 

closed

PR #3246, opened 2015-12-28
New package: libexplain-1.4

okapia:
I suspect this is going to fail for the cross-compiled architectures. If so, any help would be appreciated.


Duncaen:
Why the subpackage with the same name?

okapia:
I've fixed the subpackage - the result of copying another template without fully understanding it. Any idea where the HZ macro that fails for musl should be coming from? Thanks

Vaelatern:
Symbol does not appear to be in muslc.

okapia:
It isn't even a symbol. It is no more than just this:
/usr/include/asm-generic/param.h:#define HZ 100

Actually, that file comes from kernel-libc-headers and not from musl but glibc-devel will include it via any of sys/param.h, rpc/types.h, resolv.h or arpa/nameser.h. I've made a patch to include asm-generic/param.h.

Only problem is that the Travis builds have now failed for other reasons. Last time it was only the musl one that failed so it is a bit puzzling.

okapia:
The new musl build failure is because it tries to do sizeof(*fp) where fp is of type FILE*. musl hides the implementation of FILE completely so you can't even get it's size. Has this come up elsewhere? Any ideas how to resolve it?

chneukirchen:
The solution is to remove this explain_is_efault_pointer, it will work neither in glibc nor musl.

chneukirchen:
The benefit of this library is highly suspect to me, is it a dep for something?

okapia:
It is a dep of tardy, which is a tool for removing things like usernames from .tar files.
But yes, I'd agree about it being suspect. I can see it being useful during development to get better explanations for errors but I think it is wrong for final software. Like exception handling, it probably encourages a lazy approach to error handling that results in users getting a pile of technical internal crap instead of a simple clear error.

chneukirchen:
I thought bsdtar could do this already, but not when converting archive to archive...

okapia:
Thanks, I didn't know that. bsdtar is good enough for my purposes and as it is proving tricky to get this library to build anyway, I'm closing the pull request.

PR #3248, opened 2015-12-28
New package: pal-0.4.3

dkociela:


srcpkgs/pal/template
```diff
@@ -0,0 +1,20 @@
+pkgname="pal"
+version="0.4.3"
+revision=1
+build_wrksrc=src
+only_for_archs="x86_64 i686"
```
xtraeme: Stop using this to make travis report a success, this is not the solution.

srcpkgs/pal/template
```diff
@@ -0,0 +1,20 @@
+pkgname="pal"
```
xtraeme: The first line should be a comment.

srcpkgs/pal/template
```diff
@@ -0,0 +1,20 @@
...
+version="0.4.3"
+revision=1
+build_wrksrc=src
+only_for_archs="x86_64 i686"
+hostmakedepends="pkg-config"
+makedepends="glib-devel gettext-devel readline-devel"
```
xtraeme: - `gettext-devel` should be in `hostmakedepends` instead.

srcpkgs/pal/template
```diff
@@ -0,0 +1,20 @@
...
+homepage="http://palcal.sourceforge.net/"
+distfiles="${SOURCEFORGE_SITE}/project/palcal/$pkgname/$version/$pkgname-$version.tgz"
+checksum="ce470cb7be76522ff58cd0325ad7817b6cc2132a5e0cea55de8c9eb63b54551c"
+
+do_build() {
+	make
+}
+do_install() {
+	make DESTDIR=${DESTDIR} install
+}
```
xtraeme: Use `build_style=gnu-makefile` instead.

srcpkgs/pal/template
dkociela: OK, thanks. Just curious though, what is only_for_archs for?

srcpkgs/pal/template
dkociela: Thanks for the clue. I had to move a couple of things around between
hostmakedepends
and makedepends but finally got it.

srcpkgs/pal/template
dkociela: Thanks, I did try that but got build failures that I couldn't work around.

dkociela:
I finally had to set "noarch" to get success. I wonder if I should have set "nostrip" instead.

pullmoll:
Setting `noarch=yes` for a package which compiles binaries is certainly wrong.
If you can't manage to get cross compiles working, you should just set `nocross=yes`.
If compiling for musl libc fails, just leave it as is. Perhaps someone else knows how to fix it and does it later on. Or perhaps it just won't work with musl libc.

dkociela:
Everything goes well until the post-install phase (armv6, v7), running the strip/debug hook. Strip exits with "/usr/bin/pal: file format not recognized".
Setting "nostrip" in the template, everything succeeds. 
The package builds and installs fine on my x86-64 sytem, and the binary "/usr/bin/pal" shows as "ELF 64-bit LSB executable, x86-64"
I'm still trying to figure out why "strip' gives that error (only on the arm arch)

dkociela:
Thanks, your comment clarified something for me.

srcpkgs/pal/template
```diff
@@ -0,0 +1,28 @@
...
+build_wrksrc=src
+hostmakedepends="pkg-config gettext-devel ncurses-devel readline-devel glib-devel"
+makedepends="glib-devel"
+short_desc="Command line calendar that can keep track of events"
+maintainer="Dan Kociela <dkociela@gmail.com>"
+license="GPL-2"
+homepage="http://palcal.sourceforge.net/"
+distfiles="${SOURCEFORGE_SITE}/project/palcal/$pkgname/$version/$pkgname-$version.tgz"
+checksum="ce470cb7be76522ff58cd0325ad7817b6cc2132a5e0cea55de8c9eb63b54551c"
+do_build() {
```
xtraeme: You are just duplicating what the `gnu-makefile` build_style already does, please remove `do_{build,install}` and set `build_style=gnu-makefile`

srcpkgs/pal/template
```diff
@@ -0,0 +1,28 @@
+# Template file for 'pal'
+
```
Vaelatern: No newline here

srcpkgs/pal/template
Vaelatern: This was taken care of

Vaelatern:
Is this pull request abandoned?

dkociela:
I couldn't get it to cross compile for arm without setting `nostrip=yes` . Also the build fails when using 
`build_style=gnu-makefile`.
I guess I've given up. Should I close it?

Vaelatern:
Don't close it, I'll take more of a stab at it. You are right about it being hard to build with the `build_style`.

Vaelatern:
At present, this package requires headers like `glib.h` which are not even present on Void, and even has a smattering of specific types like `gint` and `gchar` and calls functions like `g_free()`. I don't see this as getting in the repos without impressive work.

I recommend using `wyrd` instead.

PR #3253, opened 2015-12-29
New package: retroshare-0.6.0.RC2

abelboldu:
Skipped voip plugin for now because the build is broken


pullmoll:
The leading slashes in the target paths for vinstall and vcopy are wrong. You should remove them, i.e.
```
…
vinstall ${FILESDIR}/${pkgname}.desktop 644 usr/share/applications/
…
```

abelboldu:
Fixed

PR #3258, opened 2015-12-30
New package: docker-compose-1.5.2

### LINKED ISSUE #3257: Package Request: docker-compose
https://github.com/docker/compose/releases/tag/1.5.2

radare:
Fix #3257


radare:
Fixed

radare:
I have tried to run docker-compose and realized it was depending on many python modules. as long as packaging packages from other package managers is a boring and imho wrong way to go.. it will be good to have a script or some automated way to convert packages from cpan, pip, npm, ... into xbps packages without having to do all this repetitive work by hand.

The problem im facing now is that docker-compose depends on a specific (old) version of `requests`. By installing the OSX pkg, this one comes with `requests-2.8`, and we have 2.9 in void. I'm not sure if this is really a dependency check that can be just removed by patching the code of docker-compose

radare:
@Duncaen what should I do in this case?

Duncaen:
I just grabbed `requests` in the sourcecode and they only use it in one function in `script/versions.py` `get_github_releases`, crazy to add the dependency just for this one get request.
In the changelog[0] of `requests` are mostly bug fixes and just minor improvements, i think there is no api breakage or it would be very easy because its just used once.

I don't know of a script that helps you to package python modules. I share your views on packaging python modules, but there is no better solution at the moment.

Please split up each new dependency in its own `New Package: $pkgname` commit.

[0] https://github.com/kennethreitz/requests/blob/master/HISTORY.rst

radare:
omg, splitting in several pullreqs is gonna be boring, it is really mandatory?

I have edited the ` /usr/lib/python2.7/site-packages/docker_compose-1.5.2-py2.7.egg-info/requires.txt` that forces that stupid restriction. But then it was still failing, so i tried to import that module by hand and realized that python-requests package was wrongly installed, so i fixed that i realized that it is now depending on another python package. so i did it and im gonna push it in a separate PR.

radare:
What should i do with that docker-compose requires.txt? a post-install patch?

radare:
https://github.com/voidlinux/void-packages/pull/3278

Duncaen:
You dont need to split them into PR requests, just multiple commits in this PR is enough I think.
The `requierements.txt` is normally not needed because xbps installs the dependencies.

radare:
Can you merge this https://github.com/voidlinux/void-packages/pull/3278 i'll cleanup this commit history now so having one commit per package in this PR

asergi:
@radare just add this snippet to the `docker-compose` template:
```
pre_build() {
       sed -i '/requests/s|, < 2.8||' setup.py
}
```

radare:
The push -f wasnt really good for github. i created a new PR here https://github.com/voidlinux/void-packages/pull/3281

PR #3259, opened 2015-12-30
New package: inspircd-2.0.20

crater2150:
Accidentally commited to the wrong branch
https://github.com/voidlinux/void-packages/pull/3254


srcpkgs/inspircd/template
```diff
@@ -0,0 +1,53 @@
...
+configure_args+=" --data-dir=/var/lib/inspircd"
+configure_args+=" --module-dir=${_libpath}/modules"
+configure_args+=" --binary-dir=${_binpath}"
+configure_args+=" --uid inspircd"
+configure_args+=" --disable-interactive"
+configure_args+=" --enable-gnutls"
+configure_args+=" --enable-epoll"
+
+make_dirs="/var/log/inspircd 0755 inspircd inspircd"
+make_dirs+=" /var/run/inspircd 0755 inspircd inspircd"
```
xtraeme: - The `/run` dir should be created by the runit service instead.

srcpkgs/inspircd/template
```diff
@@ -0,0 +1,53 @@
...
+	: ${make_cmd:=make}
+	: ${make_install_target:=install}
+
+	local make_install_args="INSTUID=\"${pkgname}\""
+		make_install_args+=" BINPATH=\"${DESTDIR}/${_binpath}\""
+		make_install_args+=" BASE=\"${DESTDIR}/${_libpath}/inspircd.launcher\""
+		make_install_args+=" MODPATH=\"${DESTDIR}/${_libpath}/modules/\""
+		make_install_args+=" CONPATH=\"${DESTDIR}//etc/${pkgname}\""
+
+	${make_cmd} DESTDIR=${DESTDIR} ${make_install_args} ${make_install_target}
```
xtraeme: Those vars are really unnecessary, just set the cmds that this actually needs, i.e `make DESTDIR=...`

srcpkgs/inspircd/template
```diff
@@ -0,0 +1,53 @@
+# Template file for 'hub'
+pkgname=inspircd
+version=2.0.20
+revision=1
+build_style=configure
+makedepends="gnutls-devel pcre-devel"
+hostmakedepends="gcc-c++ perl pcre-devel pkg-config"
```
xtraeme: - `gcc-c++` is unnecessary, it's always part of the bootstrap pkgs.

srcpkgs/inspircd/files/inspircd/run
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+if [ ! -d /var/run/inspircd ]; then
+	mkdir -m0755 -p /var/inspircd
```
xtraeme: `/var/inspircd` vs `/var/run/inspircd` ?

srcpkgs/inspircd/files/inspircd/run
```diff
@@ -0,0 +1,7 @@
...
+fi
+exec chpst -u inspircd /usr/bin/inspircd \
```
xtraeme: Please do not hardcode the path to `inspircd`.

srcpkgs/inspircd/template
```diff
@@ -0,0 +1,46 @@
+# Template file for 'inspircd'
+pkgname=inspircd
+version=2.0.20
+revision=1
+build_style=configure
+makedepends="gnutls-devel pcre-devel"
+hostmakedepends="perl pcre-devel pkg-config"
```
Vaelatern: `pcre-devel` should go in `makedepends`.

Vaelatern: I believe this is why your cross compliation is currently failing.

crater2150: It already was in makedepends, but the additional entry in hostmakedepends is not needed, I removed it. But it did not change the cross-compile issue, from what I found googling the error message, it could be some kind of toolchain configuration issue with arm hardfloat: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=62099

Duncaen: `Compiler program:		g++` seems like it doesn't use the right compiler,  it should use `arm-linux-gnueabihf-g++`.

srcpkgs/inspircd/template
```diff
@@ -0,0 +1,46 @@
...
+_configure_extras="--enable-extras=m_ssl_gnutls.cpp,m_regex_pcre.cpp"
+configure_args="--prefix=/usr"
+configure_args+=" --config-dir=/etc/inspircd"
+configure_args+=" --log-dir=/var/log/inspircd"
+configure_args+=" --data-dir=/var/lib/inspircd"
+configure_args+=" --module-dir=${_libpath}/modules"
+configure_args+=" --binary-dir=${_binpath}"
+configure_args+=" --uid inspircd"
+configure_args+=" --disable-interactive"
+configure_args+=" --enable-gnutls"
```
Vaelatern: Just letting you know, inspircd is fully compatible with libressl, so you can, and probably should, enable openssl.

srcpkgs/inspircd/template
Vaelatern: How would you change this discovery? I've got a package of Inspircd for OpenBSD but OpenBSD doesn't cross compile.

srcpkgs/inspircd/template
xtraeme: Right, we prefer LibreSSL.

crater2150:
Changed to libressl.
I also switched from the optional pcre-regex-module to the posix-regex-module, as the cross compile fails in m_pcre.so, hope this works around it. The posix one also does not have additional dependencies.

PR #3263, opened 2015-12-31
dnscrypt-proxy: fix some problems

ybkbolio:


srcpkgs/dnscrypt-proxy/template
```diff
@@ -1,12 +1,14 @@
...
 pkgname=dnscrypt-proxy
 version="1.6.0"
-revision=1
+revision=2
 build_style=gnu-configure
 makedepends="libsodium-devel"
 short_desc="DNS proxy that encrypts queries"
 maintainer="kebolio <keboler@cock.li>"
 license="ISC"
+conf_files="/etc/sv/dnscrypt-proxy/run"
```
xtraeme: Nope, `run` files should not be configuration file. Use `service/conf` for this purpose, see other pkgs as example.

PR #3265, opened 2015-12-31
New package: vis-0.1

diogoleal:


pullmoll:
`do_build()` is redundant if not counter-productive (no Void specific flags passed on)

PR #3275, opened 2016-01-02
New package: munge-0.5.11

pulux:


srcpkgs/munge/template
```diff
@@ -0,0 +1,43 @@
+# Template file for 'munge'
+pkgname=munge
+version=0.5.11
+revision=1
+nocross=yes
+build_style=gnu-configure
+hostmakedepends="pkg-config"
+makedepends="bzip2-devel libgcrypt-devel zlib-devel"
+depends="shadow munge-libs>=${version}_${revision}"
```
xtraeme: This `munge-libs` dependency should have been added automagically via `common/shlibs`.

PR #3278, opened 2016-01-02
Add python-docopt package

radare:


radare:
Ready to review + merge. This is needed for the upcoming docker-compose tool.

Duncaen:
You need to add a subpackage for the python3.4 version, see `python-Jinja2` as example.

radare:
Ouch. It bothers me to package python stuff for several reasons, which maybe we can find better solutions than doing it like now we do.

First: packaging packages from other package systems is simply wrong. we will end up repackaging everything or we will miss packages and the user will end up looking at the original repo instead of the distro one. which imho makes sense.

I understand, that having distro packages for those modules makes sense to provide a consistent environment, but we are just dupping the work of all package systems.

xbps should be able to install system-wide node, python, perl, ruby, ... modules without having to type a single line.

All the pkginfo can be extracted from their original pkgmanagers or the website of every project, so all this stuff can be done by a script.

Also, why `python3.4` prefix? what happens with python 3.5 or future 3.6? we will have to rename and re-edit ALL the packages again? imho the only meaningful prefix should be `python3` because its just different from python2.

The same script should be able to detect newer versions of each module and create new packages automatically.

Can I push all those `python3.4-XXX_package()` in a separate commit or do i have to create a new PR with the commit-per-pkg thing again?

radare:
Fixed

radare:
@Duncaen ok to merge now?

Duncaen:
Looks good, can you squash and rebase again, to only have the one `New package: python-docopt-0.6.2` commit?

radare:
done

Duncaen:
@radare `python3.4-docopt` should depend on `python3.4`

radare:
Should be fine now :P

radare:
anyway, any comments on my thoughts about automating this dummy work?

xtraeme:
I agree that duplicating python/perl/XXX packages for xbps is redundant and completely stupid.

radare:
But which is the solution? Add some pip/npm/.. Hooks to automatically register those pkgs into xbps

radare:
soo.. is this going to be merged at anytime?

Duncaen:
Maybe you can copy your input/ideas into another issue and we can try to find a solution there.

radare:
Sure!

PR #3286, opened 2016-01-03
New package: perl-Module-Implementation-0.09

jprjr:

PR #3291, opened 2016-01-03
New package: perl-B-Hooks-EndOfScope

jprjr:
Depends on #3286 and #3290


xtraeme:
If you move all packages that depend on each other into the same pull request travis will build all them just fine.

jprjr:
Can do! I'll get these closed out and combined into one pull request

jprjr:
Replacing with #3319

PR #3295, opened 2016-01-03
dovecot-plugin-pigeonhole: build for current dovecot version

crater2150:
Current version of dovecot in the repo is not ABI compatible to the one the pigeonhole plugin is built against. Setting _dovecot_ver fixes this

### DIFF
--- a/srcpkgs/dovecot-plugin-pigeonhole/template
+++ b/srcpkgs/dovecot-plugin-pigeonhole/template
@@ -1,15 +1,15 @@
 # Template file for 'dovecot-plugin-pigeonhole'
 pkgname=dovecot-plugin-pigeonhole
 # make sure this matches the current dovecot
-_dovecot_ver=2.2.19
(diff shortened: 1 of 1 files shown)


Gottox:
+1 for fixing the style :+1:

Gottox:
Whoops. One minor mistake: revision should be resetted to 1 if you update the version. Keep that in mind for future update.

crater2150:
The version of the package itself did not change, only the version of dovecot, for which this is a plugin. That version is relevant for the build process, but not for the package manager.
Should the revision be reset anyway?

dominikh:
No. However, what's the point of the _dovecot_ver variable in the first place? It's not being used anywhere in the template as far as I can tell.

crater2150:
I'm not sure. I only noticed, that the plugin gave API errors, looked into the template, saw the version number and adapted it to the dovecot version in the repo.

PR #3304, opened 2016-01-05
New package: docker-machine-bin-0.5.5

radare:


thypon:
why bin?

radare:
Because i was unable to build it from source, and didnt wanted to spend more time on that. I can share my template attempt for the source version of it, but not now im afk

radare:
Is this going to be merged at any time? otherwise maybe we should be thinking in having a community driven repository to fix that slow master merge foo

thypon:
We discussed about community/unstable/restricted repositories many times on IRC #xbps channel. Right now isn't difficult at alla creating/using a custom repository from xbps-src. Actually i'd like to have alla the voidlinux work inside the void-packages repository to mantain high standard quality with centralized revisions.

I got got your points, btw. I have a docker container to create custom repository from given branch/repo on the fly. Let me know if you need it/you are interested.

radare:
Then it should be used in other 250 packages that are in the repo:
```
$ grep -C 1 -re do_extract * | grep :$ |wc -l
250
```

radare:
vs
```
$ git grep skip_extraction . |wc -l
10
```

radare:
i was grepping for no_extract=YES or something like this, also, this skip_extraction thing doesnt seems to work for me, ive tried setting the filename as argument, but it does not works

radare:
ping

thypon:
:+1: let's wait for @xtraeme : i'm not sure about the -bin politics.

xtraeme:
I'd prefer to not add more `-bin` packages if possible. If you use `build_style=go` does not build?

radare:
Nope, i was unable to make this pkg build at all

xtraeme:
@radare @dominikh is looking into this package, let's see if he's able to build it with `build_style=go`.

radare:
Cool! Thanks

PR #3308, opened 2016-01-05
New package: milkytracker-0.90.86

bra1nwave:


xtraeme:
I believe that this should also be linked to `libjack`.

bra1nwave:
Now it should be working with `libjack`. Thanks for pointing that out.

PR #3318, opened 2016-01-06
New package: salt-2015.8.3

Vaelatern:
The inclusion of `python-devel` is only because the Manual indicates it is necessary for cross compilation.
I chose not to split into client and server packages because the difference is 1 executable and 1 service folder. Also, neither Gentoo nor Arch split them into different packages.
This is a tool similar to ansible, puppet, and chef. I'm working on including explicit void support (mainly identifying a void system and using xbps) in the Salt code.

### DIFF
--- a/srcpkgs/salt/template
+++ b/srcpkgs/salt/template
@@ -1,0 +1,26 @@
+# Template file for 'salt'
+pkgname=salt
+version=2015.8.3
+revision=1
+build_style=python-module
+noarch=yes
+pycompile_module="salt"
+conf_files="/etc/salt/*"
+hostmakedepends="python python-devel"
+makedepends="python-devel"
+depends="python python-yaml python-MarkupSafe python-Jinja2 python-requests python-pyzmq python-crypto python-M2Crypto python-tornado python-msgpack dmidecode pciutils"
+short_desc="Remote execution system, and configuration manager"
+maintainer="Toyam Cox <Vaelatern@gmail.com>"
+license="Apache-2.0"
+homepage="http://saltstack.org/"
+distfiles="$PYPI_SITE/s/salt/salt-${version}.tar.gz"
+checksum=2e9a262789b018f3443513105c0c6ae98934c9bc105a04cf9e5c073ef706218a
+
+post_install() {
+	vmkdir /etc/salt 0750
+	vcopy conf/* /etc/salt/
+	vsv salt-api
+	vsv salt-master
+	vsv salt-minion
+	vsv salt-syndic
+}
--- a/srcpkgs/salt/files/salt-api/run
+++ b/srcpkgs/salt/files/salt-api/run
@@ -1,0 +1,2 @@
+#!/bin/sh
+exec salt-api
--- a/srcpkgs/salt/files/salt-master/run
+++ b/srcpkgs/salt/files/salt-master/run
@@ -1,0 +1,2 @@
+#!/bin/sh
+exec salt-master
--- a/srcpkgs/salt/files/salt-minion/run
+++ b/srcpkgs/salt/files/salt-minion/run
@@ -1,0 +1,2 @@
+#!/bin/sh
+exec salt-minion
--- a/srcpkgs/salt/files/salt-syndic/run
+++ b/srcpkgs/salt/files/salt-syndic/run
@@ -1,0 +1,2 @@
+#!/bin/sh
+exec salt-syndic


yopito:
Hello @Vaelatern.

I disagree to have only one  package for Salt. A far better solution will be to splitt it among the (python site-package libs (let's say package `salt-core`), `salt-master` and `salt-minion` at least, there are other packages too like `salt-ssh`or `salt-proxy` (see my WIP below).

There is some duplicate efforts on SaltStack packaging ! I'm currently discussing with @niamtokik about it (owner of PR #2250).

I also have some WIP on it, including support of XBPS package manager and `sv` as  service management, see https://github.com/yopito/void-packages/tree/saltstack/srcpkgs/salt-core (branch "saltstack").

So would agree to split your packaging into several subpackages, possibly the same way I do ? It could be then a good starting point for all of us ;)

Vaelatern:
I agree with @xtraeme when he said "I don't get the point of splitting into subpackages if those pkgs depend on the same set of dependencies."

The package from that PR created a `-devel` package where salt is never used to create more packages after that.

While you can split into minions, masters, and syndics, the differences are not impressive enough to warrant it. The difference for syndic, on my machine, is just 167 bytes. For master, 734. These are tiny programs, and the core of salt is necessary for basically every operation. It would be a waste of a person's time looking at packages and my time to add subpackages.

I saw your WIP and decided it would be faster, cleaner, and more likely to get in this week, if I just made a simple package a-la Gentoo, Arch, FreeBSD, OpenBSD, OSX Brew, Macports, pip, Solaris, and Windows.

That said, there are other efforts which should not be duplicated. How is your work on the `saltstack/salt` repo, specifically xbps actions and the like?

I am in a bit of a hurry to get this done, since I want to bring up my servers ideally this week, probably next. While your work is impressive, I just want to get it done.

yopito:
@Vaelatern : thanks for your inputs.

Here my points:

PR #2250 is out of business now, I was discussing only on my own packaging.

IMHO, splitting among sub-packages is important from a functional (and security ?) point of view.
If only one package: you could start a salt-master on a giving minion by mistake for instance.
if splitted packages : People don't have to take care about `salt-core` package. They just have to care about `salt-minion` most of the time or `salt-master`.
BTW, the minion needs `dmidecode` and `pciutils`

My Support of XBPS and sv: all are WIP:
  * package management is basically fine (list, remove, add packages)
  * no repository support yet (only `pkg.list_repos()` is fine)
  * sv: basically fine: list, start, stop. missing adding or removing a service from startup.

So: I can quickly provides a splitted packaging for salt, at least for salt-core, salt-minion and salt-master. Then add my WIP stuff once ready in subsequent updates of these packages.

Vaelatern:
WIP is great. Working and PRed is better. If you want to send me your work on `saltstack/salt`, I'll finish it up and PR it.

What does the minion use those for? I'll add them to my dependencies list. Thank you for catching that.

I disagree with your points for splitted packages. If a user mixes up and enables `salt-master` instead of `salt-minion` the solution is to re-link files. If the user downloads the wrong package because the maintainer judged a matter of a few hundred bytes was enough to split them up, that's more work to fix. Finally, security does not benefit from split packages. The salt master is not exploitable unless it's running. The salt-minion doesn't expect or help run the salt-master itself. If someone gets to the point where they have a root shell, they can just download the other package, tinker with code, whatever they like. If the salt-minion is exploitable, the salt code is at risk, but that wouldn't change the outcome if you had installed all of master and minion and syndic since you were running a syndic setup.

yopito:
OK then, let's go ahead with a single package. I will rework my WIP to match your packaging ASAP.
If this PR is merged quickly, I will open another one for these updates.
minion uses `dmidecode` and `lspci`from `pciutils` in `grains/core.py` for various hardware information (BIOS, GPUs).

Vaelatern:
I've incorporated that in this PR.

Don't waste your time here, especially if you already have things in 'saltstack/salt' you have not yet finished.

srcpkgs/salt/template
```diff
@@ -0,0 +1,25 @@
...
+version=2015.8.3
+revision=1
+build_style=python-module
+conf_files="/etc/salt/*"
+hostmakedepends="python python-devel"
+makedepends="python-devel"
+depends="python<3 python-yaml python-MarkupSafe python-Jinja2 python-requests  zeromq-devel python-pyzmq python-crypto python-M2Crypto python-tornado python-msgpack dmidecode pciutils"
+short_desc="Remote execution system, and configuration manager"
+maintainer="Toyam Cox <Vaelatern@gmail.com>"
+license="Apache"
```
xtraeme: Apache-2.0

srcpkgs/salt/template
```diff
@@ -0,0 +1,25 @@
+# Template file for 'salt'
+pkgname=salt
+version=2015.8.3
+revision=1
+build_style=python-module
+conf_files="/etc/salt/*"
+hostmakedepends="python python-devel"
+makedepends="python-devel"
+depends="python<3 python-yaml python-MarkupSafe python-Jinja2 python-requests  zeromq-devel python-pyzmq python-crypto python-M2Crypto python-tornado python-msgpack dmidecode pciutils"
```
xtraeme: - `python` is just enough because there's no `python>=3`
- Why does it depend on `zeromq-devel` ?

srcpkgs/salt/template
```diff
@@ -0,0 +1,25 @@
...
+checksum=2e9a262789b018f3443513105c0c6ae98934c9bc105a04cf9e5c073ef706218a
+
+post_install() {
+	vmkdir /etc/salt 0750
+	vcopy conf/* /etc/salt/
+	vsv salt-api
+	vsv salt-master
+	vsv salt-minion
+	vsv salt-syndic
+	vlicense LICENSE
```
xtraeme: There's no need to install the license, it's part of base-files already.

srcpkgs/salt/template
yopito: I agree: `zeromq-devel` is useless, only `python-pyzmq` is needed

srcpkgs/salt/template
Vaelatern: That's a nice change :- )

srcpkgs/salt/template
Vaelatern: Old change I forgot to take out.

xtraeme:
Last change: this pkg is `noarch`.

xtraeme:
And this also needs `pycompile_module="salt"`

Vaelatern:
Changes made. Thank you for your insight.

PR #3324, opened 2016-01-07
dvdauthor: add build_options for default video format (NTSC or PAL)

dkociela:
dvdauthor requires that a default video format be set, NTSC for the US, Canada, and most of South America,  or PAL for most of Europe and parts of Asia. This can be done at build time or set in the user's environment.


xtraeme:
`revision` must be bumped. I think `PAL` would be a good default.

dkociela:
OK, thanks. Done.

PR #3329, opened 2016-01-07
rover: update to 0.4.11

diogoleal:


Vaelatern:
Make some tiny change and push again. The only thing that failed was Travis, and not even because any tests failed.

PR #3339, opened 2016-01-08
font-iosevka: update to 1.6.2

diogoleal:


Duncaen:
Any reason why you change the `distfiles` to a zip archive?

diogoleal:
Because in version 1.6, it needs to compile in tar.gz file, and the zip file already compiled all.

PR #3341, opened 2016-01-08
libvirt: use libtirpc

ohanar:
portablexdr is flaky and often doesn't work, so use libtirpc instead.


ohanar:
libtirpc is usable on musl with #3340

PR #3347, opened 2016-01-10
minidlna: use -S instead of -d, redirect stderr to stdout

pbui:
1. -S runs minidlnad in the foreground without excessing logging (-d)
2. Redirecting stderr to stdout allows users to log output via svlogd


Vaelatern:
Only error is xlint: `srcpkgs/minidlna/template:21: license 'BSD', but no use of vlicense` and the license is BSD and GPL-2.

PR #3351, opened 2016-01-10
New package: lua-lualdap-1.2.0

jprjr:


jprjr:
Fixed+rebased - added pkg-config to makehostdepends and removed depends (since they're picked up by xbps-src)

srcpkgs/lualdap/template
```diff
@@ -0,0 +1,17 @@
+# Template file for 'lualdap'
+pkgname=lualdap
+version=1.2.0
+revision=1
+build_style=gnu-makefile
+hostmakedepends="pkg-config"
+makedepends="lua51-devel libldap-devel"
```
xtraeme: You've called it `lualdap` but this depends on `lua-5.1`, any reason to not use our current version? (5.3).

jprjr:
@xtraeme good idea! I went ahead and made the package generate subpackages for Lua 5.1 and 5.2, along with the default 5.3.

I also renamed the package to `lua-lualdap` - I figure that fits in better with how packages for modules are named, like `perl-DateTime`

Vaelatern:
There are three PRs including this one. They either all can be closed or merged, imho. @chneukirchen, can you take a look at this one, #3353 and #3354 ?

srcpkgs/lua-lualdap/files/COPYING
```diff
@@ -0,0 +1,20 @@
...
+Permission is hereby granted, free of charge, to any person obtaining a
+copy of this software and associated documentation files (the "Software"),
+to deal in the Software without restriction, including without limitation
+the rights to use, copy, modify, merge, publish, distribute, sublicense,
+and/or sell copies of the Software, and to permit persons to whom the
+Software is furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
```
Vaelatern: Can't you get this license from elsewhere in the PR?

Vaelatern:
Serious question about testing. How can I independently test that these packages work?

Vaelatern:
#yolo

PR #3356, opened 2016-01-11
New Package: spampd-2.42 + dependency

crater2150:


srcpkgs/spampd/files/spampd/run
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+[ -r conf ] && . ./conf
+
+exec /usr/bin/spampd --port=${PORT:=10025} \
```
xtraeme: Please change `exec /usr/bin/...` to `exec spampd ...`

srcpkgs/spampd/template
```diff
@@ -0,0 +1,27 @@
...
+	pod2man --section=8 --center="Spam Proxy Daemon" spampd.pl > spampd.8
+}
+
+do_install() {
+	vbin spampd.pl spampd
+	vman spampd.8
+}
+
+post_install() {
+	vsv spampd
```
xtraeme: There's no reason to use `post_install` just to install the service, move this to `do_install`.

Gottox:
ToDo: 
* [ ] fix crossbuild of perl-NetAddr-IP

Vaelatern:
Bump @Gottox, will crossbuild be fixed or will we fix it after it's in the tree?

PR #3358, opened 2016-01-11
mypaint: update to 1.2.0-beta.4

Spencer-H:
The reason for the revision is due to being unable to include the -beta.4 part in the version variable, so I had to trigger an update with the revision.


xtraeme:
- `revision` must be always set to `1`for new upstream versions
- what's the reason to update it to the beta version?

Spencer-H:
Bug fixes and increased number of translations.

Gottox:
We won't ship unstable releases as long as there's no urgent reason to do so. Imho we should wait for a stable release here.

xtraeme:
What @Gottox said. Just wait until the next stable version is released.

Thanks

PR #3364, opened 2016-01-13
font-iosevka: update to 1.6.3

diogoleal:


srcpkgs/font-iosevka/template
```diff
@@ -1,9 +1,8 @@
 # Template file for 'font-iosevka'
 pkgname=font-iosevka
-version=1.6.2
+version=1.6.3
 revision=1
 wrksrc=iosevka
-#wrksrc=Iosevka-${version}
```
Gottox: remove that line.

diogoleal: @Gottox it was removed in this PR.

PR #3379, opened 2016-01-15
openimageio: update to 1.6.9.

lemmi:


lemmi:
i don't know why travis fails. x64 definitely builds locally. i686 fails at the same point, not sure why.

PR #3382, opened 2016-01-16
New package: relaysrv-0.12.1

lemmi:


lemmi:
If the whole dependency vendoring wasn't under construction right now I suggested to package godep and maybe even have a hook for it in the go build style.
If someone pokes me, I'll package godep and update this template, too.

srcpkgs/relaysrv/files/relaysrv/run
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+[ -r config ] && . ./config
```
xtraeme: should be `conf` not `config`

srcpkgs/relaysrv/files/relaysrv/config
```diff
@@ -0,0 +1 @@
+RELAY_ARGS='-pools=""'
```
xtraeme: The `service/conf` should not be provided by the package.

PR #3394, opened 2016-01-18
deluge: Added service files

Vaelatern:


srcpkgs/deluge/files/deluge-web/run
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+exec chpst -u delugeweb:delugeweb /usr/bin/deluge-web -d 2>&1
```
xtraeme: Please do not hardcode the path to the cmd, i.e `exec chpst ... deluge-web ...`

PR #3401, opened 2016-01-19
New package: iverilog-10.0

ebfe:


srcpkgs/iverilog/template
```diff
@@ -0,0 +1,23 @@
+# Template file for 'iverilog'
+pkgname=iverilog
+version=10.0
+revision=2
```
xtraeme: `revision` must be set to 1 for new packages and upstream updates.

srcpkgs/iverilog/template
```diff
@@ -0,0 +1,23 @@
...
+wrksrc="${pkgname}-${version/./_}"
+build_style=gnu-configure
+hostmakedepends="autoconf automake bison flex gperf"
```
xtraeme: redundant dependencies, use `hostmakedepends="automake flex gperf"`

srcpkgs/iverilog/template
```diff
@@ -0,0 +1,23 @@
...
+makedepends="readline-devel"
+depends="readline"
```
xtraeme: This is unnecessary, xbps-src adds dependencies automagically if ELF files are dynamic linked.

srcpkgs/iverilog/template
```diff
@@ -0,0 +1,23 @@
...
+homepage="http://iverilog.icarus.com/"
+distfiles="https://github.com/steveicarus/iverilog/archive/v${version/./_}.tar.gz"
+checksum=27a100a63e57081368d9c2db303b296806949b173dd244e7d55d91b120d669df
+
+only_for_archs="i686 i686-musl x86_64 x86_64-musl"
+nopie=yes
+
+pre_configure() {
+	aclocal
+	autoconf
```
xtraeme: `autoreconf -fi` instead.

srcpkgs/iverilog/template
```diff
@@ -0,0 +1,23 @@
...
+depends="readline"
+short_desc="Verilog simulation and synthesis tool"
+maintainer="Michael Gehring <mg@ebfe.org>"
+license="GPL-2, LGPL-2.1"
+homepage="http://iverilog.icarus.com/"
+distfiles="https://github.com/steveicarus/iverilog/archive/v${version/./_}.tar.gz"
+checksum=27a100a63e57081368d9c2db303b296806949b173dd244e7d55d91b120d669df
+
+only_for_archs="i686 i686-musl x86_64 x86_64-musl"
+nopie=yes
```
xtraeme: What's the reason to disable PIE?

srcpkgs/iverilog/template
ebfe: On Mon, Jan 18, 2016 at 09:23:02PM -0800, Juan RP wrote:
> > +pre_configure() {
> > +	aclocal
> > +	autoconf
> 
> `autoreconf -fi` instead.

That's what I tried first, but it fails with the helpful message
"/usr/bin/autoheader failed with exit status: 1"

Switched to using upstream supplied `autoconf.sh`.

srcpkgs/iverilog/template
ebfe: On Mon, Jan 18, 2016 at 09:23:14PM -0800, Juan RP wrote:
> > +nopie=yes
> 
> What's the reason to disable PIE?

CFLAGS end up in iverilog-vpi. Unless `nopie` is set they contain
`--spec=/void-packages/.../hardened-cc1` which doesn't exists in the real system.

xtraeme: Seems reasonable, thanks for the info.

PR #3414, opened 2016-01-20
Add python-tkinter as dependency for python-matplotlib

WKuipers:
The matplotlibrc shipped with matplotlib defines TkAgg as the default backend, but this backend is not set as a dependency. The python-tkinter package supplies this backend, so I added it as a dependency.


asergi:
I set `agg` as the default backend again to avoid any further confusion.

`TkAgg` is one of several optional backends, none of which is required. The system-wide `matplotlibrc` is just a template: it's being created at compile time based upon what is present (iirc Qt5Agg by default). I chose TkAgg as good default because is probably the most commonly used.

So, if you want to use the TkAgg backend or any other at your choice, you need to install the required dependencies on your own (as already explained in the post-install message).

TL;DR matplotlib requires user intervention. It's up to the user to choose which backend to use. For more info, see the message at post-install time.

PR #3419, opened 2016-01-21
New package: pitivi-0.95

Spencer-H:


Spencer-H:
This builds properly and seems to function, but further testing reveals it breaks because gst-python builds with python 2. This can be used if someone can get that to build with 3, but it's beyond me at the moment.

PR #3423, opened 2016-01-22
New package: openshot-2.0.5

Spencer-H:
And its many many dependencies.


Spencer-H:
I submitted a bug report to openshot about the arm and musl builds. After attempting some patches and fixes myself it seemed as if the error wouldn't be a simple on to fix and I decided to leave it up to the upstream devs.

(The arm one may be simple to fix, but it requires messing around with the cmake files, which is something I don't have experience with. This cmake doesn't include <ft2build.h> properly if anyone wants to look into it)

srcpkgs/libopenshot-audio/template
```diff
@@ -0,0 +1,24 @@
...
+only_for_archs="x86_64 i686"
+hostmakedepends="cmake doxygen"
+makedepends="xorg-server-devel libXrandr-devel libXinerama-devel libXcursor-devel alsa-lib-devel"
+short_desc="OpenShot audio library"
+maintainer="Spencer Hill <spencernh77@gmail.com>"
+license="GPL-3"
+homepage="www.openshot.org"
+distfiles="https://launchpad.net/libopenshot/0.0/0.0.9/+download/${pkgname}-${version}.tar.gz"
+checksum=29707bf7603fc08b2a152abff2599405f3499a2d674919b1a1749638cd91f42e
+wrksrc=.
```
xtraeme: use `create_wrksrc=yes` instead.

Spencer-H:
I believe this commit can be merged.

srcpkgs/libopenshot-audio/template
```diff
@@ -0,0 +1,24 @@
+# Template file for 'libopenshot-audio'
+pkgname=libopenshot-audio
+version=0.0.6
+revision=1
+build_style=cmake
+only_for_archs="x86_64 i686"
```
xtraeme: Any reason to disable musl?

srcpkgs/unittest-cpp/template
```diff
@@ -0,0 +1,31 @@
...
+
+post_install() {
+	vlicense LICENSE
+}
+
+unittest-cpp-devel_package() {
+	short_desc+=" - development files"
+	pkg_install() {
+		vmove usr/include
+	}
```
xtraeme: A development pkg that only contains headers is not useful, please remove this subpkg.

srcpkgs/unittest-cpp/template
```diff
@@ -0,0 +1,31 @@
...
+license="MIT"
+homepage="https://github.com/unittest-cpp/unittest-cpp"
+distfiles="https://github.com/unittest-cpp/unittest-cpp/archive/v1.5.0.tar.gz"
+checksum=09426598ee0976f9c908f8ea7282bda7e9f088305a1fe27d674b7c65b8f53379
+
+pre_build() {
+	case $XBPS_TARGET_MACHINE in
+		arm* )  echo "Removing built in test from arm build. Won't run on host x86_64 machine"
+				sed -i '59,61d' CMakeLists.txt ;;
+		* ) ;;
```
xtraeme: The default `case` is always empty, you can drop `*) ;;`.

srcpkgs/openshot/template
```diff
@@ -0,0 +1,17 @@
+# Template file for 'openshot'
+pkgname=openshot
+version=2.0.5
+revision=1
+build_style=python-module
+python_versions="3.4"
+only_for_archs="i686 x86_64"
```
xtraeme: You should change `only_for_archs` to also include `i686-musl` and `x86_64-musl`.

Spencer-H:
The reason that musl is disabled on both openshot and libopenshot-audio is because they both make use of glibc's unravel. I put in an issue requesting that they switch to a library that wasn't glibc dependent, but porting it to libunwind didn't seem very trivial as changing the functions to call the ones from libunwind resulted in a lot of type errors. If someone could help make a patch to change it to use that library I'd love to, but it was beyond my meager experience with C++ to fix.

As for the case and development package, I will switch fix those.

PR #3425, opened 2016-01-22
qtox: update to 1.2.4

Spencer-H:
Also split off the toxcore-git bootstrap daemon and create a runit service for it.

### DIFF
--- a/srcpkgs/toxcore-git/files/toxbootstrapd/run
+++ b/srcpkgs/toxcore-git/files/toxbootstrapd/run
@@ -1,0 +1,3 @@
+#!/bin/sh
+cd /etc/toxbootstrapd
+exec chpst -u toxbootstrapd DHT_bootstrap
--- a/srcpkgs/qtox/template
+++ b/srcpkgs/qtox/template
@@ -1,6 +1,6 @@
 # Template file for 'qtox'
 pkgname=qtox
-version=1.2.2
+version=1.2.4
 revision=1
 build_style=qmake
 short_desc="QT-based TOX instant messenger client"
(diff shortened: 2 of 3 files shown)


srcpkgs/toxcore-git/files/toxbootstrapd/run
```diff
@@ -0,0 +1,2 @@
+su - toxbootstrapd
+exec DHT_bootstrap
```
xtraeme: use `chpst -u <user> ...` instead of `su(1)`

srcpkgs/toxcore-git/files/toxbootstrapd/run
```diff
@@ -0,0 +1,2 @@
+chpst -u toxbootstrapd
```
xtraeme: Clearly you did not test this, because `chpst(1)` expects a command to run.

Use `exec chpst -u <user> <cmd>` instead.

srcpkgs/toxcore-git/files/toxbootstrapd/run
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+exec chpst -u toxbootstrapd /usr/bin/tox-bootstrap-daemon
```
xtraeme: what's the reason to separate this into `tox-bootstrap-daemon` ? the `run` script is enough.

Spencer-H:
At some point someone in the IRC suggested that the daemon be moved to a separate package. If you would prefer it simply be part of the main toxcore package I can change that though.

PR #3434, opened 2016-01-22
openvpn: update to 2.3.10.

ebfe:
Patch taken from upsteam git master for libressl compat.


TheDoctorsLife:
The short description is to long should be less than 72 characters.

ebfe:
> The short description is to long should be less than 72 characters.

Yes, i know the xlint fails, but the `short_desc` is like that since Feb
2013, so I guess that is on purpose?

TheDoctorsLife:
Maybe, shorting it should help kick Travis to redo the build, (since one of them timed out). Plus, fixing errors is a nice thing to do, even if it was left.

ebfe:
Updated.

TheDoctorsLife:
Cool, now Travis is happy

PR #3458, opened 2016-01-25
New package: tagainijisho-1.0.3

salihgerdan:
This is my first time using git, or doing anything for any linux distribution; if I did anything wrong, please help me correct it.

### DIFF
--- a/srcpkgs/tagainijisho/template
+++ b/srcpkgs/tagainijisho/template
@@ -1,0 +1,14 @@
+# Template file for 'tagainijisho'
+
+pkgname="tagainijisho"
+version="1.0.3"
+revision=1
+build_style=cmake
+hostmakedepends="cmake"
+makedepends="qt-devel"
+short_desc="A free software Japanese dictionary and learning assistant"
+maintainer="Salih Gerdan <msalih45@gmail.com>"
+license="GPL-3"
+homepage="http://www.tagaini.net"
+distfiles="https://github.com/Gnurou/tagainijisho/releases/download/${version}/tagainijisho-${version}.tar.gz"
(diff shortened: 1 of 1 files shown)


srcpkgs/tagainijisho/template
```diff
@@ -0,0 +1,15 @@
...
+build_style=cmake
+hostmakedepends="cmake"
+makedepends="qt-devel"
+short_desc="A free software Japanese dictionary and learning assistant"
+maintainer="Salih Gerdan <msalih45@gmail.com>"
+license="GPL-3"
+homepage="http://www.tagaini.net"
+distfiles="https://github.com/Gnurou/tagainijisho/releases/download/${version}/tagainijisho-${version}.tar.gz"
+checksum="6fe3c579ec1cea6c41393ac9a82dd12d0e240032d2bb7837a90e470f480aaf4e"
+depends="qt>=4.5"
```
xtraeme: This depends line is unnecessary, you can remove this.

salihgerdan:
I made the recommended change. Though I don't  understand why it is unnecessary. Could you explain?

pullmoll:
The `xbps-src` scripts detect dependencies of your package on other packages by analysing the shared libraries which your package links against and looking them up in the `common/shlibs` file.
In the build log you see these three lines:
```
   SONAME: libQtGui.so.4 <-> qt>=4.5.3_1
   SONAME: libQtNetwork.so.4 <-> qt>=4.5.3_1
   SONAME: libQtCore.so.4 <-> qt>=4.5.3_1
```
which means the dependency is detected and registered. Only in cases where there is no linking against a required dependency you would have to explicitly specify it.

salihgerdan:
Thanks.

PR #3469, opened 2016-01-26
qtpass: update to 1.1.0

annejan:


xtraeme:
Can you please correct the commit message? it should be in the form of `qtpass: update to 1.1.0.`

PR #3470, opened 2016-01-26
xbps-src: generate python pkgs for multiple versions automatically

asergi:
The patch has been successfully tested on several python modules.

Any suggestion or help is appreciated.


common/xbps-src/shutils/pkgtarget.sh
```diff
@@ -99,6 +100,19 @@ remove_pkg() {
             msg_normal "$f: removing 32bit files from destdir...\n"
             rm -rf ${_destdir}/${f}-32bit-${version}
         fi
+        for pyver in $python_versions; do
```
Vaelatern: Wouldn't a better way to deal with this be to optionally add the `python${pyver}-${pymodn}` to the for loop this is in, to do the full processing on it, perhaps to the `${subpackages}` variable that they currently get processed in?

Manual.md
```diff
@@ -1055,6 +1055,11 @@ by blanks, i.e `pycompile_module="foo blah"`.
 recursively by the target python version. This differs from `pycompile_module` in that any
 path may be specified, i.e `pycompile_dirs="usr/share/foo"`.
 
+- `py2depends`: this variable expects python modules that are only
+needed by the Python 2 version of the package. These are often
+backports of Python 3.x standard library modules for use with Python 2,
+e.g.: `py2depends="python-enum34 python-ipaddress"`.
+
 > NOTE: it's expected that additional subpkgs must be generated to allow packaging for multiple
```
Vaelatern: This PR would remove this `NOTE`

Manual.md
```diff
@@ -1055,6 +1055,11 @@ by blanks, i.e `pycompile_module="foo blah"`.
 recursively by the target python version. This differs from `pycompile_module` in that any
 path may be specified, i.e `pycompile_dirs="usr/share/foo"`.
 
+- `py2depends`: this variable expects python modules that are only
```
Vaelatern: `this variable expects dependencies that are only`

common/hooks/do-pkg/00-gen-pkg.sh
```diff
@@ -129,6 +130,60 @@ hook() {
 		_replaces+=" ${_pkgn}-32bit${_pkgv}"
 	done
 
+	# Generate pythonX.Y-foo pkg.
+	if [ -z "$python_versions" -o "$python_versions" = "${python_versions#2.7}" ]; then
+		return
+	fi
+	for pyver in $python_versions; do
+		if [ "$pyver" != "2.7" ]; then
```
Vaelatern: Should also loop over `depends`, checking for dependencies and automatically changing them to the python `pyver` version of that dependency, if such a thing exists. Alternately a new variable for python 3 specific depends, such as used in the `python-PyQt4` template.

Gottox:
I'm not sure if this is a good idea. This is a lot of added complexity that will work for python only. So the process of building a python template differs from ruby, lua, ...

Until now we had a very simple rule:

> One package, one directory (or link) in void-packages/srcpkgs
> (except for dbg and 32bit which are in its own repository)

This PR breaks this rule, makes it harder to understand how a package is built and adds special treatment for one script language only.

Vaelatern:
Except python 3 subpackages are not an unknown value. It's a lot of retyping of stuff that is relatively basic to pull off a python 3 subpackage.

asergi:
Obsolete. We switched to python3 packages, instead.

PR #3480, opened 2016-01-28
Fix issue #2994: default umask setting is crazy

ayghor:
This lets pam_umask do its job of setting teh default umask, which,
unlike shellscripts, respect teh usergroups policy.

Teh setting is made on '/etc/pam.d/default-umask-hack', instead of teh
default config file, '/etc/login.defs', to work around a `useradd`
issue when creating new user directories.

Our `useradd` programs respect '/etc/login.defs', but not pam_umask or
any of its options. This way, we can use teh first to control it, and
teh second to set teh default behaviour.

In form of Q&A:

Q: Why not on `/etc/login.defs`?

A: It affects `useradd`, which does not respect pam_umask, when
creating new user directories. We want them to be created with umask
077, not 022.

Q: Why not on `/etc/profile`?

A: Because not every shell reads this file, not every process is a
shell and it does not respect teh usergroups umask policy.

Q: What is usergroups umask policy?

A: From pam_umask(8):

> If the user is not root and the username is the same as primary
> group name, the umask group bits are set to be the same as owner
> bits (examples: 022 -> 002, 077 -> 007).

It is intended for use with setgid directories. For example: you may
want to have a directory shared with a group, and that new files there
are read-writable to everyone (umask 002).

People expect this when users are created their own usergroup.

Q: What about `useradd` newly created user directories permissions?

A: They are created with umask 077, as defined in '/etc/login.defs'.
As `useradd` does not respect `pam_umask` at all.

Q: What is teh default umask now?

A: 022 (002 if you are not root and your primary group and username
are equal).

Q: Bugs? Thanks?

A: rogi@skylittlesystem.org


chneukirchen:
Sounds like a good solution but the patch contains a lot of fuzz not part of the change.

dominikh:
`s/teh/the`

ayghor:
By fuzz you mean teh comments? I think it is a lot of text too, but teh
hack is not so obvious. It is really frustrating when you try to
reconfigure things and nothing happens, because it is being overriden in
another place for obscure reasons.

Aww and let my spelling be, plz. =o3

Vaelatern:
Closing because it contains a lot of fuzz, a significant number of misspellings of "the" in materials that would be distributed with the operating system, has ambiguous benefit, and has had no activity in more than two months.

PR #3483, opened 2016-01-28
watchman: updated to 4.4.0

Vaelatern:


Vaelatern:
`sys_siglist` seems to not be used in muslc.

xtraeme:
:+1:

PR #3484, opened 2016-01-28
rxvt-unicode: updated to 9.22

Vaelatern:


Vaelatern:
I can't think of a better way to say `rxvt clone supporting Xft fonts and Unicode` and xlint fails on just that line.

7185:
How about `An rxvt clone supporting Xft fonts and Unicode`?

chneukirchen:
First off, you don't need to fix xlint things just when you want to bump a revision.

Second, I propose "Terminal emulator supporting Xft fonts and Unicode", so people would actually find it...

xtraeme:
+1

Vaelatern:
Description changed to @chneukirchen's suggestion.

PR #3490, opened 2016-01-29
New package: python-marisa-trie-0.7.2

bra1nwave:


srcpkgs/python-marisa-trie/template
```diff
@@ -0,0 +1,30 @@
...
+distfiles="https://github.com/kmike/marisa-trie/archive/${version}.tar.gz"
+checksum=5a5ca6a032fe924097f87439130f013dfb2682e597f4d5414bb87cfe49bc715e
+
+post_install() {
+	vlicense LICENSE
+}
+
+python3.4-marisa-trie_package() {
+	pycompile_version="3.4"
+	pycompile_module="marisa-trie"
```
xtraeme: both `pycompile_module` are wrong, there's not a module named this way.

bra1nwave:
OK, I removed them. Thanks

Issue #3495, opened 2016-01-31
Package request: nvidia (outdated)


nem-arch:
if the latest and stable version are equal to the short-/longlived versions, nvidia is outdated [1].
nvidia-352.63_1 is outdated to shortlived 358.16
nvidia-stable-340.96_1 is outdated to longlived 352.79

[1] https://www.nvidia.com/object/unix.html

xtraeme:
Please do not submit problem reports to just report a package is outdated.

If you need another new version please submit a Pull Request with a proper commit.

THanks

nem-arch:
So a user finds an outdated package and you expect him to submit an updated package himself? I think you are confusing 'user' and 'maintainer' here. If that is really what you expect your void users to do, then void is not the one I was looking for.

The nvidia package was outdated since november by the way!

Duncaen:
@nem-arch we already have a script that creates a list of outdated packages, see http://repo.voidlinux.eu/void-updates/void-updates.txt

I don't see whats the problem with asking for a PR, in most cases you just have to change the version number and update the checksums.

nem-arch:
There's no problem with asking for a PR, but:
A maintainer (or at least someone with commit rights) didn't know any better and pointed me to create an "outdated" issue.
Why was the package outdated for two months if it's just two numbers changing and an automated script should have indicated it?
Why should a user care updating it in the first place if it's the maintainers job? I'm not a maintainer.

xtraeme:
Let's clarify some things:

1- I'm aware the packages were outdated, HOWEVER if you look at the commit logs you'll see that I have created everything related to void: xbps, xbps-src, void-mklive, etc.
2- I waste enough HOURS DAILY for 8 years already, even I'm not getting paid for any of this.
3- My time is precious and I make enough changes daily, can't update EVERYTHING daily.
4- I've got a life and I got to maintain TWO daughters.

So just be patient until someone updates the affected packages, or update them yourself and submit a pull request, or just GTFO.

HTH

nem-arch:
1) I'm aware you are the author of the tools. Is that related to the topic?
2) I'm aware the work is unpaid, not only for you, but for all maintainers.
3) Everybody's time is precious. I'm aware that not everything is updated daily.
4) Everybody has a life to maintain. Is that related to the topic?

I was patient. Two months. Maybe hand off some packages, so other people from the devteam feel responsible before a user needs to intervene. And maybe change this quote from the voidlinux frontpage:
"
Rolling release
[...]

PR #3496, opened 2016-01-31
nvidia-stable: updated to 352.79

Vaelatern:
#3495


Vaelatern:
xlint fails because vman doesn't understand .gz files can be man pages but xlint expects it to.

chneukirchen:
Void does not use compressed manpages, they should be unpacked first.

Vaelatern:
They were not uncompressed in the package before. I shall add decompression.

Vaelatern:
Decompression added, all checks passing.

PR #3497, opened 2016-01-31
[WIP] nvidia: updated to 358.16

Vaelatern:
#3495 
Also fixes xlint warnings.
Still WIP due to missing file on version upgrade.


srcpkgs/nvidia/template
```diff
@@ -159,10 +168,6 @@ do_install() {
 
 	# dkms pkg
 	vmkdir usr/src/nvidia-${version}
-	# add nvidia-uvm dkms submodule too.
```
Vaelatern: Reports of failing to add dkms so this section needs review.

srcpkgs/nvidia/template
```diff
@@ -159,10 +168,6 @@ do_install() {
...
-	if [ "$XBPS_TARGET_MACHINE" = "x86_64" ]; then
-		cat kernel/uvm/dkms.conf.fragment >> kernel/dkms.conf
```
Vaelatern: `kernel/uvm/` no longer exists, `kernel/dkms.conf` still does, and `kernel/nvidia-uvm` doesn't have any conf files in it to append to `kernel/dkms.conf`

Vaelatern:
https://git.io/vg8BO

PR #3498, opened 2016-01-31
New package: slurm-wlm 15.08.7

Vaelatern:


Vaelatern:
Musl compilation is broken -- can't find type `ino_t`.

srcpkgs/slurm-wlm/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'slurm-wlm'
+pkgname=slurm-wlm
+version=15.08.7
+revision=1
+wrksrc=slurm-$version
+build_style=gnu-configure
+hostmakedepends="perl cgit python"
+makedepends="munge-devel lua-devel hdf5-devel pam-devel libressl-devel gtk+
+	libnuma-devel ncurses-devel readline-devel rrdtool-devel mysql"
```
xtraeme: Adding `gtk+` and `mysql` as `makedepends` is useless.

srcpkgs/slurm-wlm/template
```diff
@@ -0,0 +1,19 @@
...
+depends="munge libnuma lua readline ncurses rrdtool libressl pam hdf5 mysql gtk+"
```
xtraeme: This depends line is completely unnecessary.

srcpkgs/slurm-wlm/template
```diff
@@ -0,0 +1,18 @@
+# Template file for 'slurm-wlm'
+pkgname=slurm-wlm
+version=15.08.7
+revision=1
+wrksrc=slurm-$version
+build_style=gnu-configure
+hostmakedepends="perl cgit python pkg-config"
```
xtraeme: Please disable static libs, as you won't need them (no -devel pkg). `configure_args="--disable-static"`.

Vaelatern:
Fixed. In the future it may be useful to re-enable static libs, but currently I can't find anybody who actually builds against slurm-wlm.

Vaelatern:
All issues resolved.

PR #3115, opened 2015-12-04
New package: ndpi-1.6

diogoleal:

PR #3503, opened 2016-02-01
New package: ndpi-1.6

Vaelatern:
Since #3115 seems dead, this is that work with a change in maintainer.


srcpkgs/ndpi/template
```diff
@@ -0,0 +1,35 @@
+# Template file for 'ndpi'
+pkgname=ndpi
+version=1.6
+revision=1
+wrksrc=nDPI-${version}
+build_style=gnu-configure
+configure_args="--prefix=/usr"
```
xtraeme: This line is redundant, `--prefix=/usr` is set by default with `build_style=gnu-configure`.

srcpkgs/ndpi/template
```diff
@@ -0,0 +1,35 @@
...
+hostmakedepends="libtool autoconf automake pkg-config"
```
xtraeme: `autoconf` is redundant, it's a dependency of `automake`

Vaelatern:
Rebased and pushed. All issues addressed.

PR #3504, opened 2016-02-01
openjdk: update to 8u82.

Vaelatern:


Vaelatern:
https://travis-ci.org/voidlinux/void-packages/jobs/106173880
Re-pushed to fix the revision.

PR #3517, opened 2016-02-02
Add hid_rmi module to linux4.3

renatoaguiar:
This module is required to use touchpad on some laptops, like Levono S21e.


xtraeme:
Enabled the module for 4.1, 4.3 and 4.4.

Thanks!

PR #3519, opened 2016-02-02
broadcom-wl-dkms: Patch for x86 on linux >= 4.3

renatoaguiar:


xtraeme:
Please bump revision.

PR #3521, opened 2016-02-02
New package: global

renatoaguiar:


srcpkgs/global/template
```diff
@@ -0,0 +1,14 @@
+# Template file for 'global'
+pkgname=global
+version=6.5.1
+revision=1
+build_style=gnu-configure
+configure_args="--with-exuberant-ctags=/usr/bin/ctags"
```
xtraeme: Please disable static libs, because they are not useful on this pkg, i.e `--disable-static`.

PR #3524, opened 2016-02-02
adobe-flash-plugin: fixed URL

rocx:
Despite the `adobe-flash-plugin` being marked as restrictive and [requiring manual building](https://wiki.voidlinux.eu/Flash), the package would always fail to build as the URL was pointing to the wrong location.  This patch points XBPS to the right place and the package can be installed successfully once built.


xtraeme:
Please fix the commit log messages as explained in CONTRIBUTING.md.

xtraeme:
Well, you've updated this pkg to a new version and your commit log still says "fixed URL/checksum".

PR #3525, opened 2016-02-02
Java: manage profiles scripts by alternatives

Gottox:
This PR moves the profiles file to a different location and adds an alternative to /etc/profile.d/java.sh.

This way we ensure, that the right JAVA_HOME is set when the java alternative is activated on a package.


Gottox:
Not ready yet, need to fix

PR #2612, opened 2015-09-26
New package: seafile-client-qt

yopito:
https://www.seafile.com/ is an "Open Source Cloud Storage for your teams and organizations", including the ability to create your own private storage.

include packaging for libsearpc, ccnet, seafile-libclient.

PR #3529, opened 2016-02-02
New package: libsearpc-3.0.7

yopito:
Following PR #2612: split Seafile package(s) in 2 distincts PR to avoid troubles with Travis CI. This is PR 1/2


yopito:
just rebased from master.

PR #3531, opened 2016-02-03
Virt missing dependency

Vaelatern:
Throws internal error without this.


xtraeme:
`revision` is already set to 3, looks like your repo is outdated.

Vaelatern:
Fixed

PR #3536, opened 2016-02-04
rakudo-star: updated to 2016.01

xaduha:


chneukirchen:
This appears not to work anymore because the sceond step of make install expects rakudo to be installed already outside of $DESTDIR.

Perhaps we should make a separate package for rakudo and use that as a dep for rakudo-star.

chneukirchen:
Doesn't really help, since rakudo-star plainly seems to ignore DESTDIR when installing modules now.

chneukirchen:
This perhaps worked at some point: https://github.com/rakudo/star/commit/226eb20032baac7c4bda074e0d16b741ecbf568a

I guess for now, we only can reasonably package rakudo without modules, I have a template for it ready and the other templates cleaned up.

xaduha:
I'll think of something, if nothing else upstream patches can be made.

chneukirchen:
My stash so far: http://sprunge.us/dQgh

PR #3540, opened 2016-02-05
pm-utils: Add in kernel suspend to both

renatoaguiar:


chneukirchen:
Interesting. May be worth adding to zzz(8).

PR #3546, opened 2016-02-05
New package: libpurple-steam-1.6.1

jprjr:


jprjr:
Looks like I had a problem with CFLAGS on non-cross builds. Fixed+pushed.

PR #3554, opened 2016-02-06
New package/subpackages: opensmtpd-extras-5.7.1

jprjr:


xtraeme:
- Please do not include this huge license file into the repo, use sed/awk/whatever to install it.
- `depends="$pkgname"` is wrong in all subpkgs, use `depends=${sourcepkg}` instead.

jprjr:
I'm actually not sure how to handle the licensing on this package. The source tarball doesn't have a license file, I figured the license from opensmtpd would cover it. Is there a way to somehow reference another package's license?

jprjr:
I'm going to go ahead and close this for now - I should have had the licensing straightened out before opening this PR. Sorry about that!

PR #3565, opened 2016-02-09
New package: openvswitch

renatoaguiar:


Duncaen:
Normally we don't have conditional patches, make them portable to get them upstreamed.

srcpkgs/openvswitch/files/ovs-vswitchd/run
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+/usr/sbin/modprobe openvswitch
```
xtraeme: Please do not hardcode the paths, i.e `modprobe` not `/usr/sbin/modprobe`.

srcpkgs/openvswitch/template
```diff
@@ -0,0 +1,40 @@
+# Template file for 'openvswitch'
+pkgname=openvswitch
+version=2.4.0
+revision=1
+build_style=gnu-configure
+configure_args="--prefix=/usr --sysconfdir=/etc --localstatedir=/var --with-rundir=/run/openvswitch --enable-shared"
```
xtraeme: `--prefix=/usr --sysconfdir=/etc --localstatedir=/var` is set by default with `build_style=gnu-configure`, remove them.

srcpkgs/openvswitch/template
```diff
@@ -0,0 +1,40 @@
...
+hostmakedepends="python perl"
+makedepends="libressl-devel ${hostmakedepends}"
+depends="python"
+short_desc="Production quality, multilayer virtual switch"
+maintainer="Renato Aguiar <contact@renatoaguiar.org>"
+license="Apache-2.0"
+homepage="http://openvswitch.org/"
+distfiles="http://openvswitch.org/releases/${pkgname}-${version}.tar.gz"
+checksum=424c5a8dea1c5b153e8c1803bb041d0495e81379fb6f386dd58ee903a7681b9a
+only_for_archs="i686 x86_64"
```
xtraeme: This will skip x86/musl.

srcpkgs/openvswitch/template
```diff
@@ -0,0 +1,40 @@
...
+
+pre_configure() {
+	case "$XBPS_TARGET_MACHINE" in
+	*-musl)
+		patch -p1 < ${FILESDIR}/musl-if_packet.patch
+		patch -p1 < ${FILESDIR}/0001-ovs-thread-Set-stacksize-to-1M.patch
```
xtraeme: Those patches should be moved to `$PATCHESDIR` and be applied by default.

srcpkgs/openvswitch/files/ovs-vswitchd/run
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+modprobe -q openvswitch || exit 1
+sv check ovsdb-server >/dev/null || exit 1
+exec /usr/sbin/ovs-vswitchd --pidfile=/run/openvswitch/ovs-vswitchd.pid
```
xtraeme: `exec ovs-vswitchd ...`

I really doubt `--pidfile` would be useful because runit services must run in the foreground.

srcpkgs/openvswitch/files/ovsdb-server/run
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+if [ ! -f /etc/openvswitch/conf.db ]; then
+    /usr/bin/install -d /etc/openvswitch
```
xtraeme: Drop hardcoded paths from here too, ditto for `--pidfile`.

xtraeme:
Bash completions:

```/home/travis/build/voidlinux/void-packages/masterdir/destdir//openvswitch-2.4.0/etc/bash_completion.d/ovs-appctl-bashcomp.bash

/home/travis/build/voidlinux/void-packages/masterdir/destdir//openvswitch-2.4.0/etc/bash_completion.d/ovs-vsctl-bashcomp.bash
```

should be moved to `/usr/share/bash-completion/completions`.

srcpkgs/openvswitch/template
```diff
@@ -0,0 +1,44 @@
...
+	ln -s ${_completiondir}/ovs-appctl-bashcomp.bash \
+	   ${DESTDIR}/usr/share/bash-completion/completions/ovs-appctl
+	ln -s ${_completiondir}/ovs-appctl-bashcomp.bash \
+	   ${DESTDIR}/usr/share/bash-completion/completions/ovs-ofctl
+	ln -s ${_completiondir}/ovs-appctl-bashcomp.bash \
+	   ${DESTDIR}/usr/share/bash-completion/completions/ovs-dpctl
+	ln -s ${_completiondir}/ovs-appctl-bashcomp.bash \
+	   ${DESTDIR}/usr/share/bash-completion/completions/ovsdb-tool
+	ln -s ${_completiondir}/ovs-vsctl-bashcomp.bash \
+	   ${DESTDIR}/usr/share/bash-completion/completions/ovs-vsctl
```
xtraeme: Please do not make symlinks, move the files to the correct directory.

asergi:
- Add `vmove usr/lib/*.so` to -devel
- The package contains python files too, so add `pycompile_dirs="usr/share/openvswitch/python"`

renatoaguiar:
The libraries cannot be moved since some binaries installed by openvswitch depend on them.

Vaelatern:
Pardon if this was asked, but why `enable-shared`? Seems a lot of hassle for no gain, unless you really need the libraries shared.

Vaelatern:
Also, @asergi meant that since you were adding things to `common/shlibs` they should be available in a `-devel` subpackage. You shouldn't be using `--enable-shared`, and definitely not adding them to `common/shlibs` unless you can think of a legitimate case for other packages to have access to them.

srcpkgs/openvswitch/template
```diff
@@ -0,0 +1,42 @@
...
+	vsv ovs-vswitchd
+}
+
+openvswitch-devel_package() {
+	depends="${sourcepkg}>=${version}_${revision}"
+	short_desc+=" - development files"
+	pkg_install() {
+		vmove usr/include
+		vmove usr/lib/pkgconfig
+		vmove "usr/lib/*.a"
```
Vaelatern: Quotes not needed.

PR #3567, opened 2016-02-09
gpgme: Use gnupg2 instead of gnupg

renatoaguiar:


xtraeme:
I really doubt this will have any effect, its configure script does not check for any `gpg` command. Did you test this works as expected?

PR #3568, opened 2016-02-09
New package: rkt-1.0.0

thypon:


srcpkgs/rkt/template
```diff
@@ -0,0 +1,57 @@
+# Template file for 'rkt'
+pkgname=rkt
+version=1.0.0
+revision=1
+build_style=gnu-configure
+hostmakedepends="automake wget cpio squashfs-tools bc gnupg git go"
+makedepends="acl-devel zlib-devel"
+short_desc="App Container runtime for Linux"
+maintainer="Andrea Brancaleoni <miwaxe@gmail.com>"
+license="Apache"
```
xtraeme: Apache-2.0

chneukirchen:
It would be nice to also provide the `stage1-kvm.aci` (could be taken from upstream release).

chneukirchen:
`--with-stage1-flavors=fly,kvm` seems to build as well.

Further points:
- `actool` is not installed
- it may be less confusing to not pass `--with-stage1-default-location=/usr/lib/rkt/stage1.aci` and keep the stage named `stage1-fly.aci`... should be found then.

PR #3569, opened 2016-02-09
freeciv: update to 2.5.3.

bra1nwave:
If no one minds, I'll maintain this package.

### DIFF
--- a/srcpkgs/freeciv-qt5
+++ b/srcpkgs/freeciv-qt5
@@ -1,0 +1,1 @@
+freeciv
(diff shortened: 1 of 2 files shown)


xtraeme:
This package was cross compiled before, and due to qt5-devel being in `makedepends` this cannot be cross compiled anymore.

Please change this to only add qt5-devel on native builds, i.e:

```
    makedepends+=" qt5-devel"
    subpackages+=" freeciv-qt5"
fi
```

PR #3580, opened 2016-02-10
afl: fix default helper path

ebfe:


Vaelatern:
Irrelevant. The build-style provides that already.

ebfe:
> Irrelevant. The build-style provides that already.
No, it doesn't.

Vaelatern:
Ah. Probably an oversight.
https://github.com/voidlinux/void-packages/blob/master/common/build-style/gnu-makefile.sh#L7-L18 Should probably be modified with this patch to put things in `/usr`

chneukirchen:
The build-styles generally don't pass this in `do_build`... if we change it, we need to do it everywhere.

ebfe:
Rebased. If #3582 gets merged this can be changed to a simple revbump.

Issue #3579, opened 2016-02-10
SONAME: libarpack.so.2 <-> UNKNOWN PKG PLEASE FIX!


ayghor:
While packaging octave, I got this:

```
=> octave-4.0.0_2: running pre-pkg hook: 04-generate-runtime-deps ...
   SONAME: liboctgui.so.1 <-> octave (ignored)
   SONAME: liboctinterp.so.3 <-> octave (ignored)
   SONAME: libstdc++.so.6 <-> libstdc++>=4.4.0_1
   SONAME: libc.so.6 <-> glibc>=2.8_1
   SONAME: liboctave.so.3 <-> octave (ignored)
   SONAME: libQtNetwork.so.4 <-> qt>=4.5.3_1
   SONAME: libQtOpenGL.so.4 <-> qt>=4.5.3_1
   SONAME: libQtGui.so.4 <-> qt>=4.5.3_1
   SONAME: libQtCore.so.4 <-> qt>=4.5.3_1
   SONAME: libfontconfig.so.1 <-> fontconfig>=2.6.0_1
   SONAME: libfreetype.so.6 <-> freetype>=2.3.9_1
   SONAME: libGL.so.1 <-> libGL>=7.11_1
   SONAME: libGLU.so.1 <-> glu>=9.0.0_1
   SONAME: libutil.so.1 <-> glibc>=2.8_1
   SONAME: libm.so.6 <-> glibc>=2.8_1
   SONAME: libpthread.so.0 <-> glibc>=2.8_1
   SONAME: libgcc_s.so.1 <-> libgcc>=4.4.0_1
   SONAME: libqhull.so.7 <-> libqhull>=2015.2_1
   SONAME: libfltk_gl.so.1.3 <-> fltk>=1.3.0_1
   SONAME: libfltk.so.1.3 <-> fltk>=1.3.0_1
   SONAME: libXfixes.so.3 <-> libXfixes>=4.0.3_1
   SONAME: libXext.so.6 <-> libXext>=1.0.5_1
   SONAME: libXft.so.2 <-> libXft>=2.1.13_1
   SONAME: libXinerama.so.1 <-> libXinerama>=1.0.3_1
   SONAME: libdl.so.2 <-> glibc>=2.8_1
   SONAME: libX11.so.6 <-> libX11>=1.2_1
   SONAME: libarpack.so.2 <-> UNKNOWN PKG PLEASE FIX!
[...]

PR #3581, opened 2016-02-10
arpack-ng: Add shlib and -devel package

Vaelatern:
#3579

### DIFF
--- a/srcpkgs/arpack-ng/template
+++ b/srcpkgs/arpack-ng/template
@@ -1,22 +1,33 @@
 # Template file for 'arpack-ng'
 pkgname=arpack-ng
 version=3.3.0
-revision=1
+revision=2
 build_style=gnu-configure
 hostmakedepends="automake libtool"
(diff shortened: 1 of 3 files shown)


srcpkgs/arpack-ng/template
```diff
@@ -1,22 +1,32 @@
...
 	./bootstrap
 }
 
-post_install(){
+post_install() {
 	vlicense COPYING
 }
+
+arpack-ng-devel_package() {
+	short_desc+=" - development files"
```
xtraeme: `depends="${sourcepkg}>=${version}_${revision}"`

Vaelatern: Added

srcpkgs/arpack-ng/template
```diff
@@ -1,22 +1,33 @@
...
+	depends="${sourcepkg}>=${version}_${revision}"
```
Vaelatern: Added here.

PR #3582, opened 2016-02-10
build-style/gnu-makefile.sh - add PREFIX=/usr

Vaelatern:
It is a common setting as seen at
https://gist.github.com/anonymous/0a7139a16844d4e0d8d7


Duncaen:
Did you look at the grep result, its not every time `/usr`, some Makefiles use `PREFIX` differently.

Vaelatern:
Looked at the grep result, didn't count. Turns out just 6 going on 7 use it.

PR #3584, opened 2016-02-11
New package: armagetronad-0.2.8.3.3

Vaelatern:
Different templates only because the configure script can not configure both dedicated server roles and a game client at the same time.

### DIFF
--- a/srcpkgs/armagetronad/template
+++ b/srcpkgs/armagetronad/template
@@ -1,0 +1,62 @@
+# Template file for 'armagetronad'
+pkgname=armagetronad
+version=0.2.8.3.3
+revision=1
+build_style=gnu-configure
+conf_files="/etc/games/${pkgname}/*.cfg
+ /etc/games/${pkgname}/*.srv"
+configure_args="--disable-useradd"
+hostmakedepends="m4"
+makedepends="libX11-devel libxml2-devel libICE-devel libpng-devel glu-devel
+ libjpeg-turbo-devel SDL_image-devel SDL_mixer-devel"
+short_desc="3D Lightcycle game from the movie Tron with advanced capabilities"
+maintainer="Toyam Cox <Vaelatern@gmail.com>"
+license="GPL-2"
+homepage="http://armagetronad.org"
+distfiles="${SOURCEFORGE_SITE}/$pkgname/$pkgname-$version.src.tar.bz2"
+checksum=ddf2d781b134392e9667f17e0319ae2117c20e6eb7738e3cacd701f3608ea5e8
+
+do_configure() {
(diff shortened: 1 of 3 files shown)


Vaelatern:
Need to add `config_files`

srcpkgs/armagetronad/template
```diff
@@ -0,0 +1,17 @@
+# Template file for 'armagetronad'
+pkgname=armagetronad
+version=0.2.8.3.3
+revision=1
+build_style=gnu-configure
+conf_files="/etc/games/${pkgname}/*.cfg
+		/etc/games/${pkgname}/*.srv"
```
xtraeme: both package handle the same configuration files or are they different?

srcpkgs/armagetronad/template
```diff
@@ -0,0 +1,17 @@
...
+configure_args="--enable-master --enable-main --enable-music"
+hostmakedepends="m4"
+makedepends="libX11-devel libxml2-devel libICE-devel SDL-devel libpng-devel
+		libjpeg-turbo-devel SDL_image-devel SDL_mixer-devel libGL glu-devel"
```
xtraeme: `libGL` is unnecessary, as well as `SDL-devel`.

srcpkgs/armagetronad-dedicated/template
```diff
@@ -0,0 +1,19 @@
...
+version=0.2.8.3.3
+revision=1
+_name=armagetronad
+wrksrc="${_name}-${version}"
+build_style=gnu-configure
+conf_files="/etc/games/${pkgname}/*.cfg
+		/etc/games/${pkgname}/*.srv"
+configure_args="--enable-dedicated"
+hostmakedepends="m4"
+makedepends="libX11-devel libxml2-devel libICE-devel SDL-devel libpng-devel
```
xtraeme: ditto

srcpkgs/armagetronad/template
Vaelatern: Both packages handle the same configuration, but they install it to `/etc/games/${pgkname}`

Vaelatern: I'm adding a patch to put the configuration files in the same place.

Vaelatern: @xtraeme, same config, but similar enough files in both packages that they'd collide otherwise.

Vaelatern:
Now both packages build out of the same template and source.

srcpkgs/armagetronad/template
Vaelatern: Should I change the `conf_files` variable in the subpackage?

PR #3591, opened 2016-02-12
New package: python-bitarray-0.8.1

Vaelatern:
I need these for a different piece of software that I have not yet successfully packaged, but I'm PRing them separately because they may be generally useful.


asergi:
`xbps-query -Rs python-parsing`
`xbps-query -Rs python-inotify`
`xbps-query -Rs python-curl`

Vaelatern:
Searching for pyparsing and pyintofy and pycurl, as the pycompile_modules are, did not turn up results. Thanks for the tip!

asergi:
Merged in 7449114 + fixed.

PR #3606, opened 2016-02-13
Openshot: update to 2.0.6

Spencer-H:


Spencer-H:
This failing appears to be related to #3602

Spencer-H:
Marked so that python is only compiled if it's not a cross compile. This should work now.

PR #3611, opened 2016-02-14
python3.4-cairo: set nocross until upstream fixes waf

the-maldridge:
This PR deals with #3602 until such time as upstream fixes the laughably outdated version of waf they are shipping.


xtraeme:
There's no need to revbump it, just add `nocross`.

the-maldridge:
un-revbumped

PR #3635, opened 2016-02-16
New package: light

renatoaguiar:


xtraeme:
- Missing manpage
- The executable needs the setuid bit set

PR #3641, opened 2016-02-17
[WIP] Openstack

Vaelatern:
Still lacking:
- service files for each openstack component
- No runtime dependencies listed. Some of them are not even packaged yet, like `python-eventlet`

I've just been sitting on this for too long, and hopefully someone with a need for this (I no longer have one) will start working this over.

The commits below should all be correct anyway. They are not worth the work reviewing unless you are trying to finish including OpenStack in Void. The only errors thrown are of log length: 24 packages generate 4 MB of logs very quickly and Travis stops.


Gottox:
reopen when there's some update

PR #3653, opened 2016-02-19
bitcoin: update to 0.11.2.

jannisch:
Notes on changes:
* `--with-libressl` flag: required in order to configure
* qt5 gui toolkit: now default in bitcoin
* `--with-gui=qt4` ignored anyway when building with qt5 dependencies so replaced it with `--with-gui`

### DIFF
--- a/srcpkgs/bitcoin/template
+++ b/srcpkgs/bitcoin/template
@@ -1,7 +1,7 @@
 # Template file for 'bitcoin'
 pkgname=bitcoin
-version=0.10.2
-revision=4
+version=0.11.2
+revision=1
 create_wrksrc=yes
 hostmakedepends="pkg-config yasm"
 makedepends="db-devel protobuf-devel libressl-devel boost-devel miniupnpc-devel"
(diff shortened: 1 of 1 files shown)


jannisch:
I just submitted an updated version which builds with glibc on all arches, and it also builds with musl on my void computer using `./xbps-src -a x86_64-musl pkg bitcoin`. However all Travis musl builds throw the following error and then abort: `test/data/alertTests.raw.h:162:9: error: expected primary-expression before ',' token`
Does anyone know this? And might [this](http://git.alpinelinux.org/cgit/aports/tree/testing/bitcoin/ssize_t.patch?id=1698e598c66b4dd81b50440f4c30a4d14b92d056) patch from Alpine fix it? Thanks!

jannisch:
I think we should care about the musl builds later, please review and merge.

Duncaen:
Please squash your commits into one commit.

jannisch:
@Duncaen Please apologize, I'm a git newcomer.

Duncaen:
No problem, its good now.
Only thing I would revert to the original is the indention in the case statement.
https://github.com/jannisch/void-packages/commit/b17cc9913bf18eaf49fe67fd149a80ffee60722a#diff-b8d315e8a0f2690d7d445b4f2332c683R16

jannisch:
Thanks, it did not paste correctly for some reason.

PR #3654, opened 2016-02-19
mariadb: update to 10.1.12.

stpx:


Vaelatern:
From the failed configure outputs, it seems like `libevent-devel` (for `event.h`), `cracklib-devel` (for `crack.h`), and `pam-devel` (for `security/pam_appl.h`) are wanted by the configure script for armv*.

xtraeme:
unstable version?

stpx:
The `10.1` branch is the current stable release.

From https://mariadb.com/kb/en/mariadb/what-is-mariadb-101/:

> MariaDB 10.1 is the current stable version of MariaDB. It will be supported until October 2020.

PR #3656, opened 2016-02-20
glibc: update to 2.23

pullmoll:


Gottox:
I'm feeling lucky today.

PR #3660, opened 2016-02-20
dovecot-plugin-pigeonhole: add ldap support as a plugin

jprjr:


jprjr:
There was a cross-compiling error, it turns out the build process runs and compiles some C programs. Adding `libldap-devel` to `makehostdepends` fixed it. Rebased + pushed.

PR #3663, opened 2016-02-21
vim: update to 7.4.1376.

ebfe:
`hostmakedepends+="glib-devel"` since the gvim build now uses `glib-compile-resources`.

### DIFF
--- a/srcpkgs/vim/template
+++ b/srcpkgs/vim/template
@@ -1,8 +1,8 @@
 # Template file for 'vim'
 pkgname=vim
-version=7.4.1046
+version=7.4.1376
 revision=1
-hostmakedepends="pkg-config"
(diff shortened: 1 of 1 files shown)


Varkolac:
Shouldn't the "glib-devel" dependency get added to gvim instead of vim?

I guess most people don't use gvim anyway.

ebfe:
> Shouldn't the "glib-devel" dependency get added to gvim instead of vim?
gvim is a subpackage so there is no separate hostmakedepends.

> I guess most people don't use gvim anyway.
gvim doesn't care and still wants to use `glib-compile-resources` to build ;).

chneukirchen:
FTR, glib-complile-resources has to be disabled on llvm-musl, but it seems to build without it as well.

PR #3665, opened 2016-02-21
spotify: update to 1.0.23.

Vaelatern:
With trimming the update check.

Successfully installed and in use on amd64.

The current spotify package is broken because the files to be fetched don't exist for the old version.


Vaelatern:
Merged,  44fa878

PR #3683, opened 2016-02-23
seafile-client-qt: update to 5.0.5 (and dependencies)

yopito:


Duncaen:
A commit for each template file would be better.

yopito:
@Duncaen : no problem, here they are.

PR #3687, opened 2016-02-24
octave: allow default options on all archs.

ayghor:


ayghor:
It did kinda pass... no ARM version though.

srcpkgs/octave/template
```diff
@@ -14,6 +14,62 @@ checksum=cf6be2fac6796fda68971e249c96d026ce4a1b0ed074ce49a5791365a958c474
...
+desc_option_curl="Provides 'urlread' and 'urlwrite' functions and teh 'ftp' class."
+desc_option_fftw3="Improves performance on discrete Fourier transforms."
+desc_option_gui="Graphical User Interface."
+desc_option_graphicsmagick="Provides 'imread' and 'imwrite' functions."
+desc_option_hdf5="Support for HDF data files."
+desc_option_imagemagick="Provides 'imread' and 'imwrite' functions."
+desc_option_qhull="Provides 'convhull{,n}', 'delaunay{,n}' and 'voronoi{,n}' functions."
+desc_option_openmp="Enable support for OpenMP SMP multi-threading"
+desc_option_zlib="Support for compressed data."
+
```
xtraeme: Please fix all occurrences of `teh` with `the`.

srcpkgs/octave/template
```diff
@@ -1,9 +1,9 @@
 # Template file for 'octave'
 pkgname=octave
 version=4.0.0
-revision=1
+revision=2
 build_style=gnu-configure
-hostmakedepends="perl gcc-fortran"
+hostmakedepends="perl gcc-fortran pkg-config gnuplot-wx"
```
xtraeme: The `gnuplot` package should be enough because it's now using alternatives to create `gnuplot`.

Vaelatern:
Please have 1 commit (squash your commits) and make sure that *all occurances* of "teh" are replaced with "the".

Vaelatern:
Better, but your commit message still contains an instance of "teh". Please fix it.

PR #3693, opened 2016-02-24
nautilus-dropbox: renamed to dropbox and description updated

Vaelatern:
The current name of "nautilus-dropbox" makes it unclear the package includes the whole of Dropbox.


Vaelatern:
Currently broken on arm. Not sure why.

Vaelatern:
Fixed, thanks @ebfe

PR #3695, opened 2016-02-25
New package: libreoffice-5.1.0.3

pullmoll:
It builds with musl libc. Installed language is just en-US, because
building many or all takes way too long. For i18n support there are
separate packages using the precompiled RPMs from the website.
(I'll create that PR after this one is merged)

Cross compiling does not work yet, but should be possible after some
more tweaking and patching.

Considered WIP because I may be missing some `--with-system-*` options.


pullmoll:
For reference and your eagle eyes, here's the `./configure --help` output:
http://sprunge.us/DIXU

pullmoll:
May I ask for any thoughts or comments on this?

PR #3702, opened 2016-02-26
New package: jeti-filemanager-1.1.1

mrshampoo:
Lets try again :)


mrshampoo:
i think this is dependent of ncurses >= 6.0, i did not know how to make dependent's based on specifik version numbers :/
maybe this is a reason why it fails? (if the make errors is pointing to main.c,   BUTTON5_PRESSED. then this is the case) or maybe i have missed something more?

Vaelatern:
Install things to `$(DESTDIR)/dir/to/use` instead of to `/dir/to/use`

PR #3703, opened 2016-02-27
New package: jeti-filemanager-1.1.1

mrshampoo:
ok, now there is $(DESTDIR) instead of $(prefix). makefile changed.
what if changing the makefile wasn't an option?


mrshampoo:
wonderful! this time it is even worse! now i686 doesn't work ether... getting tired of this

PR #3710, opened 2016-02-27
New package: lollypop-0.9.87

### LINKED ISSUE #3705: Package Request: Lollypop
https://gnumdk.github.io/lollypop-web/

pullmoll:
Closes #3705 

_Note_: we don't have `python3.4-pylast` and I don't know if using `python3.4-pip`
and running `pip3 install pylast` would be sufficient.

### DIFF
--- a/srcpkgs/lollypop/template
+++ b/srcpkgs/lollypop/template
@@ -1,0 +1,17 @@
+# Template file for 'lollypop'
+pkgname=lollypop
+version=0.9.87
+revision=1
+build_style=gnu-configure
+noarch=yes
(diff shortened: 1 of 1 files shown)


stp242:
Wow that was fast! Thank you so much :smiley:

pylast is an optional dependency and not something I need personally. AFAIK, it will work just fine by installing it with pip.

pullmoll:
I created `python-pylast` and `python3.4-pylast` packages now.
I'm think this package should perhaps be `noarch=yes`. I'm going to try locally and fix that.

PR #3712, opened 2016-02-28
rkt: update to 1.1.0.

thypon:
can't build: libressl bug.


ebfe:
CMS disabled.  https://www.mail-archive.com/tech@openbsd.org/msg19675.html

ebfe:
A patch like this

```diff
--- a/scripts/sign-file.c
+++ b/scripts/sign-file.c
@@ -39,7 +39,7 @@
  * signing with anything other than SHA1 - so we're stuck with that if such is
  * the case.
  */
-#if OPENSSL_VERSION_NUMBER < 0x10000000L
+#if (OPENSSL_VERSION_NUMBER < 0x10000000L || LIBRESSL_VERSION_NUMBER)
 #define USE_PKCS7
 #endif
 #ifndef USE_PKCS7
```

might work if we can sneak it into the build process.

thypon:
I don't get why this bug isn't present in our linux kernels?

ebfe:
> I don't get why this bug isn't present in our linux kernels?
Because CONFIG_MODULE_SIG is not set in the config.

PR #3730, opened 2016-03-01
[WIP]chromium: update the widevine patch to fix chromium-widevine

TheDoctorsLife:


xtraeme:
This patch does not change anything, does it?

TheDoctorsLife:
At the Moment? I don't really think so. I'm trying to figure out how to make widevine work again. Its still broken for void.

xtraeme:
Useless patch; re-open with a proper patch next time.

PR #3741, opened 2016-03-02
New package: mininet

renatoaguiar:


srcpkgs/mininet/template
```diff
@@ -0,0 +1,26 @@
...
+pycompile_module="mininet"
+hostmakedepends="help2man python-setuptools"
+depends="bash inetutils-telnet iperf iputils libcgroup-utils net-tools openvswitch python socat"
+short_desc="Emulator for rapid prototyping of Software Defined Networks"
+maintainer="Renato Aguiar <contact@renatoaguiar.org>"
+license="custom"
+homepage="http://mininet.org/"
+distfiles="https://github.com/mininet/${pkgname}/archive/${version}.tar.gz"
+checksum=e4c20a6ab4ddfe4d5cad06ba25742507967df67a845dbe2e90d627af7cd21705
+
```
Vaelatern: `noarch=yes`

renatoaguiar: It also includes a binary (mnexec) which is arch dependent. This binary is compiled on post_build and installed on post_install.

srcpkgs/mininet/template
```diff
@@ -0,0 +1,26 @@
...
+pkgname=mininet
+version=2.2.1
+revision=1
+build_style=python-module
+pycompile_module="mininet"
+hostmakedepends="help2man python-setuptools"
+depends="bash inetutils-telnet iperf iputils libcgroup-utils net-tools openvswitch python socat"
+short_desc="Emulator for rapid prototyping of Software Defined Networks"
+maintainer="Renato Aguiar <contact@renatoaguiar.org>"
+license="custom"
```
Vaelatern: BSD

Vaelatern:
Disregard my comment about BSD. Was confused.

srcpkgs/mininet/template
```diff
@@ -0,0 +1,26 @@
...
+post_build() {
+	make mnexec
+	make man
+}
+
+post_install() {
+	vbin mnexec
+	vman mn.1
+	vman mnexec.1
+	vlicense LICENSE
```
Vaelatern: Sorry, was wrong, isn't a BSD license, was confused.

renatoaguiar:
In the website it says BSD, but the text in the LICENSE file  is not exactly the same. I'm going to change it back to custom.

"Mininet is actively developed and supported, and is released under a permissive BSD Open Source license." (http://mininet.org)

Vaelatern:
I do suspect the `cc` in the Makefile should be patched to `$(CC)` but since the build is succeeding I'm not sure it matters.

srcpkgs/mininet/template
```diff
@@ -0,0 +1,30 @@
+# Template file for 'mininet'
+pkgname=mininet
+version=2.2.1
+revision=1
+build_style=python-module
+pycompile_module="mininet"
+hostmakedepends="help2man python-setuptools"
+depends="bash inetutils-telnet iperf iputils libcgroup-utils net-tools openvswitch python socat"
```
Vaelatern: `python` is an implied dependency in this case.

Vaelatern:
Last comment, the commit should be `New package: mininet-2.2.1`

PR #3743, opened 2016-03-02
ebtables: Fix #3742

### LINKED ISSUE #3742: [ebtables] segmentation fault
ebtables-2.0.10.4_4 seq faults on void linux (glibc) x86_64

downgrading to ebtables-2.0.10.4_3.x86_64.xbps fixes the issue.

ebfe:


srcpkgs/ebtables/template
```diff
@@ -24,6 +24,16 @@ pre_build() {
 	rm include/linux/if_ether.h
 }
 
+do_build() {
+	LDFLAGS=${LDFLAGS/--as-needed/--no-as-needed}
```
Gottox: use `pre_build()` and remove everything below.

ebfe: > use `pre_build()` and remove everything below.
done

srcpkgs/ebtables/template
```diff
@@ -22,6 +22,7 @@ make_install_args="
 
 pre_build() {
 	rm include/linux/if_ether.h
+	LDFLAGS=${LDFLAGS/--as-needed/--no-as-needed}
```
xtraeme: Use `LDFLAGS="-Wl,--no-as-needed"` in the main block instead.

xtraeme:
:+1:

ebfe:
0b97e9adc489a9658e589e0d42b5538f8f74a636

PR #3744, opened 2016-03-02
New package: burp2-2.0.34

yopito:
Along with burp-{client,server} packages (1.4.40): 
* burp-client 2.x does not work with burp-server 1.x, 
* automatic upgrade of XBPS package would not be avoided from 1.4.40 to 2.0.34 (see issue https://github.com/voidlinux/xbps/issues/153)

So create packages with a new  prefix (burp**2**-... instead of burp-..)

### DIFF
--- a/srcpkgs/burp2-client
+++ b/srcpkgs/burp2-client
@@ -1,0 +1,1 @@
+burp2-server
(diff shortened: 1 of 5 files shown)


yopito:
Hello, anyone to merge this PR ? Thx

PR #3745, opened 2016-03-02
New package: ganeti-2.15.2

Vaelatern:


Vaelatern:
Remaining issue after getting build working: symlinks in `/usr/bin/` are broken.

Vaelatern:
No, it just turns out they are massively complicated.

The whole program cross links things like mad. Do I need to fix that (probably including some code) or can we leave it as is? Stupid, but the way they do it.

PR #3748, opened 2016-03-03
rxvt-unicode: build with unicode3 option

m-wynn:
Using the --enable-unicode3 option allows urxvt to display high-range
unicode characters, including those used by the popular "powerline"
theme.


Vaelatern:
Probably cleaner to change `build_options_default="perl startup_notification"` to `build_options_default="perl startup_notification unicode3"`

xtraeme:
This won't be accepted. There's a build option that you'll have to set to have this enabled by default.

PR #3751, opened 2016-03-03
rxvt-unicode: build with unicode3 by default

m-wynn:
Using the --enable-unicode3 option allows urxvt to display high-range
unicode characters, including those used by the popular "powerline"
theme.

This time, setting the build option to be enabled by default.

### DIFF
--- a/srcpkgs/rxvt-unicode/template
+++ b/srcpkgs/rxvt-unicode/template
@@ -1,7 +1,7 @@
 # Template build file for 'rxvt-unicode'.
 pkgname=rxvt-unicode
 version=9.22
-revision=1
+revision=2
 build_style=gnu-configure
 configure_args="
  --with-terminfo=/usr/share/terminfo --enable-256-color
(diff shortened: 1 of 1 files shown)


chneukirchen:
NACK, this is not worth 25% more memory usage IMNSHO.

dominikh:
What's that 25% in absolute numbers? IMO they might well be worth it to actually support real world usage of unicode.

dominikh:
Or, in other words: Will more people care about the extra memory usage, or about the lacking support of a unicode range? Because that ultimately should determine who has to go through the hassle of building their own package.

m-wynn:
We are already building with "--enable-combining", so characters beyond plane 0 are being stored.  This change would simple enable them to be shown.  Some googling suggests that urxvt will use 21 instead of 16 bits to represent unicode characters.  Other sites say 4 bytes instead of 2.

the-maldridge:
A quick test with a borrowed powerline config showed that the change used only 4% more memory in my configuration.  I have scrollback set to ~10k lines and use urxvtd.  Given the statement about --enable-combining it seems like this should either be merged, or --enable-combining should be removed.

Either way memory is very cheap.

dominikh:
I'm in favour of merging this.

xtraeme:
:+1:

chneukirchen:
FTR, my terminals went up from 10MB resident memory to 17MB. YMMV.

PR #3753, opened 2016-03-04
MultiMC: update to 0.4.11.

Spencer-H:
This uses git because the default source package doesn't include libnbt++, which is needed to compile. I'm instead using a commit hash that point's to the release commit.


pullmoll:
But you don't use `_commithash` anywhere?

FWIW Instead of using git you can download `https://github.com/MultiMC/MultiMC5/archive/a3cd3d5ff18215a475ed6b5f231f3f02ccd668f7.tar.gz`
and perphaps rename it like `>MultiMC5-0.4.11.tar.gz`.

Spencer-H:
Yep. Sorry I was really tired and had forgotten which changes I had made. I believe there may have been some sort of conflict with that package not being able to update the git sub modules, but I'll check once I get home.

Spencer-H:
Oh yeah. The issue was that if we download through archives, it strips all of the git data out of the package, which is needed to get the right version of libnbt++

srcpkgs/MultiMC/files/MultiMC.sh
```diff
@@ -1,2 +1,2 @@
-#/bin/sh
+#!/bin/bash
```
xtraeme: why?

Spencer-H: Because there was a lot of moving files around between the version github
has and the version I ended up pushing. Part of that moving involved
deleting that file and then rewriting it. Sorry.

xtraeme: This change is bogus... the script does not need bash.

Spencer-H: Bit of a bad force of habit there. I'll change it when I'm back at my
computer.

Spencer-H:
And this is now... Not terrible. I believe.

srcpkgs/MultiMC/template
```diff
@@ -1,17 +1,24 @@
...
-wrksrc=${pkgname}5-$version
+wrksrc=${pkgname}5
+
+do_fetch() {
+	git clone "https://github.com/${pkgname}/${pkgname}5.git"
+	cd ${wrksrc}
+	git checkout stable
+	git submodule init
+	git submodule update
+}
```
Gottox: Please use http download instead.

Spencer-H:
I don't see anything else wrong. So I believe that this is ready to merge.

Gottox:
Please do not use `git clone`. Instead point to the real release tarball: https://github.com/MultiMC/MultiMC5/archive/0.4.11.tar.gz

To explain why:

We're currently working on reproducible build support. For this to work we need to be sure that xbps-src fetches the same source regardless of when it is build.

For http download we're using the sha256sum to verify this. When a git branch or tag is used we can't be sure as there's no check for the integrity. An attacker with commit access could push any source code he wants without a way we can notice that when we later rebuild this package.

If the source code can't be fetched with http you may use git (which is not the case on github) but you must always address a specific revision using the githash.

Spencer-H:
I understand the concept of an attacker and that's why we need the sum. I'll work on a way to get it to build from http, the issue is that MultiMC is split into multiple git projects and its components are not included in the release Tarball.

I'll work on collecting the components through http though.

PR #3754, opened 2016-03-04
lightdm: update to 1.17.5.

ebfe:


xtraeme:
Is not this an unstable version?

ebfe:
> Is not this an unstable version?

It's non-obvious, but it looks like you are correct.

PR #3759, opened 2016-03-05
libgit2: update to 0.24.0 (soname bump).

ebfe:
libgit2-glib currently broken by incompatible API changes.


Vaelatern:
Did we figure `libgit2-glib` should be marked broken until upstream releases a new version that addresses the incompatibilities?

ebfe:
marked libgit2-glib as broken

ebfe:
libgit2-glib updated to 0.24

PR #3769, opened 2016-03-06
build-style/cmake: Add cmake as hostmakedepends

Vaelatern:


xtraeme:
I would prefer to not add dependencies automagically based on `build_style`.

Vaelatern:
Except that in this case, the build style leads to epople necessarily adding cmake to `hostmakedepends` in every case it is used: https://gist.github.com/anonymous/4497e282473acc8b098f

dominikh:
> I would prefer to not add dependencies automagically based on build_style.

Really? We're already doing that for the `go` build style, and that's largely why we even made it possible to add dependencies in build styles. (The `go` build style adds a build dependency on `go`, and on `go-cross-linux` when cross compiling.)

xtraeme:
@dominikh you are right, I completely forget about this.

In this case this PR would be ok :-)

Issue #3765, opened 2016-03-06
timestamp-macros.h modified CFLAGS/CXXFLAGS leaking into packages


pullmoll:
When trying to build the `alsa-tools` package there are errors regarding a missing `/builddir/.xbps-fltk/timestamp-macros.h` file. See http://sprunge.us/iigc

FWIW the `/builddir/.xbps-alsa-tools/timestamp-macros.h` file repeats the macro definitions like this:
```
#undef __DATE__
#define __DATE__ "Nov 10 2015"
#undef __TIME__
#define __TIME__ "18:08:26"
#undef __DATETIME__
#define __DATETIME__ "Nov 10 2015 18:08:26"
#undef __DATE__
#define __DATE__ "Nov 10 2015"
#undef __TIME__
#define __TIME__ "18:08:26"
#undef __DATETIME__
#define __DATETIME__ "Nov 10 2015 18:08:26"
```

Similar problems arise trying to build the `ImageMagick` package. There it is `/builddir/.xbps-perl/timestamp-macros.h` which is missing. See http://sprunge.us/CPPU

In case of `ImageMagick` the sequence of `hostdir=...` and `hostdir+=" perl"` in the template is wrong for non-cross builds. Fixing it (moving the `if [ -z "$CROSS_BUILD" ]; then ...` block down) does not fix above problem, though.

Gottox:
>When trying to build the alsa-tools package there are errors regarding a missing /builddir/.xbps-fltk/timestamp-macros.h file. See http://sprunge.us/iigc

I cannot reproduce this issue. Have you tried to xbps-clean before building?

pullmoll:
I even tried with `./xbps-src zap` and `./xbps-src binary-bootstrap`. I'll try with an empty repository, i.e. without my locally built packages.

pullmoll:
Closing because it seems to be an issue which only appears locally.

pullmoll:
Here it is again on the build bots:
http://build.voidlinux.eu/builders/x86_64_builder/builds/21467/steps/shell_3/logs/stdio

Gottox:
Did you have perl in your local binpkgs?

pullmoll:
Yes. I usually do `./xbps-src -N pkg base-system` and then packages I need with `-N pkg <name>`.

Perhaps it has to do with `subpackages+=...` in the template?

Gottox:
can you try this patch?

```diff
[...]

PR #3771, opened 2016-03-07
timestamp-macros: fix creation and add cleanup

pullmoll:
As described in #3765

Afterwards the package `perl` must be rebuilt to clean the leaked flags.
There may be more packages which were built since the timestamp-macros were introduced.


Gottox:
It's an ugly hack, but I think it's reasonable enough till the real issue is found.

ebfe:
Can we have a flag to disable the timestamps machinery from the template ('nofixedtimestamps=yes')? This likely also affects other packages where this workaround doesn't work.

Gottox:
This is a bug in xbps-src, not in the packages. There's no reason to disable it for specific packages.

ebfe:
> This is a bug in xbps-src, not in the packages. There's no reason to disable it for specific packages.
Same could be said for most of the nocross=yes instances...

pullmoll:
@Gottox I disagree. There is no bug in xbps-src. If you add something to the CFLAGS or CXXFLAGS, like the timestamps stuff does, any development package can assume the flags are valid for building other packages depending on them.

A possible solution might be to modify the CFLAGS and CXXFLAGS only after the package to be built was configured. That could perhaps happen in a pre-build hook. In case the package creates shell scripts, configuration files, package config files or similar which in the configure stage, there's a good chance the timestamps additions won't leak.

FWIW I now built about 800 packages without trouble with this workaround.

pullmoll:
As I saw just now in the package `ImageMagic` the timestamp macros flags leak into an XML file:
```
=> WARNING: Can't clean /usr/lib/ImageMagick-6.9.3/config-Q16/configure.xml (mime-type: application/xml) ...
```
So the `07-cleanup-timestamp-macros.sh` should include `"application/xml"` as well.

Gottox:
Adding a flag to disable reproducible builds is not an option in any way.

But we need a fix for that. I'm more and more considering the compiler patches instead of the CFLAGS.

pullmoll:
Seriously, switching to a gcc (cpp) patch is the only way to prevent demolishing a lot of packages!

It should be considered and implemented rather soon, because every updated package may now pick up the leaked, wrong C{,XX}FLAGS and later on cause trouble for everyone.

PR #3777, opened 2016-03-07
python-apsw: update to 3.11.1r1.

bra1nwave:


pullmoll:
You should not remove `version=` because this is a mandatory variable.

To reflect the `r1` in the version you could perhaps use `version=3.11.1.1` and reformat this `$version` for the wrksrc and distfile like `${version%.*}-r${version##*.}`.

pullmoll:
Oops! Disregard the comment, I overlooked the `version=` line in the diff.

bra1nwave:
No problem. Thanks anyway :)

Vaelatern:
Why a release candidate (if I'm reading the version line correctly)

bra1nwave:
Well, it's not a release candidate. The upstream dev names his releases like this. See: https://github.com/rogerbinns/apsw/releases

PR #3785, opened 2016-03-08
synergy: update to 1.7.5

jaagr:
- Update version to 1.7.5
- Fix build with Qt 5.5
- Install services


pullmoll:
If you create a patch which adds `#include <stdint.h>` to `src/ZeroconfService.h` the build for musl libc might succeed.

Issue #3119, opened 2015-12-05
Request for Packages: Assorted network auth and related


the-maldridge:
Two of the sites I manage are considering a switch away from Ubuntu.  Void is currently the first choice do to the simplicity of maintenence and the use of LibreSSL (we value security very highly).  This list is the software that we have found to be missing in Void.  While we plan to package as much as possible on our own, this list is here in the even others wish to help out.  Package names are from Ubuntu but should be clear anyway.

Obviously not all can be packaged for Void, but I defer to the more senior members of the community to point those out.
- [ ] icinga2
- [ ] icingaweb2
- [x] lightdm-webkit-greeter
- [x] krb5-kdc
- [x] libpam-krb5
- [x] ~~shutter~~ (duplicate of #2929 )
- [x] krb5-admin-server
- [x] libpam-ldapd
- [ ] icinga2-ido-mysql
- [ ] nagios-nrpe-server
- [x] ~~nagios-plugins~~ (duplicate of #3858)
- [x] libsasl2-modules-gssapi-mit
- [ ] python-mysqldb
- [x] libpam-mount
- [x] ~~gap~~
- [x] spim
- [x] krb5-kdc-ldap
- [x] ~~nagios-nrpe-plugin~~ (duplicate of #3858)
- [x] google-chrome-stable
- [x] ~~libcrypt-openssl-bignum-perl~~ (Pending review of irssi)
- [x] ~~WPS Office~~ (duplicate of #1711)

chneukirchen:
We have mit-krb5 (not newest version tho).

the-maldridge:
@chneukirchen To my knowledge that contains the client programs, not the server ones.

Vaelatern:
Working on `libpam-krb5` version 4.7.

Vaelatern:
Working on pam-mount.

EaterOfCode:
When monitoring plugins is through, I'll add Icinga (monitoring plugins is a dependecy of icinga)

the-maldridge:
This ticket should have been closed ages ago.  All the core software we needed for our migration has been available for some time.

PR #3786, opened 2016-03-08
New package: wps-office-bin

### LINKED ISSUE #1711: Package Request: Kingsoft WPS Office [ALPHA]
"Established in 1989, KINGSOFT Office is a leading global developer of office software solutions, developing productivity applications for the consumer and enterprise markets. With over 800 million registered users worldwide, its flagship product WPS Office (“best Microsoft Office alternative…” – CNET) is one of the world’s most popular cross-platform office suites supporting iOS and Android, as well as Windows and Linux PCs."
http://wps-community.org/download.html

Packages are available as .deb/.rpm/.tar.xz for both 32- & 64-bit.
The .tar.xz contains already functional binaries and a fonts-installer script. This should (in theory) make a really simple script.

WPS is still in Alpha for linux. Not sure, if that matters, but I guess it's worth mentioning.
AFAIK, this is non-free software (no source-files available).

the-maldridge:
Package for wps-office by KingSoft.  This package is marked restricted as it requires an old, custom version of libpng, and the english translation of the license agreement is not entirely clear for redistribution rights.

This PR resolves #1711 and continues progress on #3119.


srcpkgs/wps-office-bin/template
```diff
@@ -0,0 +1,58 @@
...
+if [ "$XBPS_TARGET_MACHINE" = "x86_64" ]; then
+	_arch=x86_64
+	_distTar="${_suiteName}_${version}~${_releaseIncrement}${_patchLevel}_${_arch}.tar.xz"
+	distfiles="${_disturl}/${_distTar}"
+	checksum=9a43a32db385a604e9959487abc330bb4602724527c9bca229a0ddf98f462e41
+elif [ "$XBPS_TARGET_MACHINE" = "i686" ]; then
+	_arch=x86
+	_distTar="${_suiteName}_${version}~${_releaseIncrement}${_patchLevel}_${_arch}.tar.xz"
+	distfiles="${_disturl}/${_distTar}"
+	checksum=6f71131835d743594e0c89a4b34a4d924d72f48b0c02f7ea6fe7270f273f71e9
```
Vaelatern: With the exception of `$_arch` and `$checksum` these can be set outside the if construct.

srcpkgs/wps-office-bin/template
```diff
@@ -0,0 +1,58 @@
+_suiteName=wps-office
```
Vaelatern: First line should be a comment. Probably `# Template file for 'wps-office-bin'`

Vaelatern: changed

the-maldridge: Yeah, kind of wierd that github didn't pick up on the very obvious diff.

srcpkgs/wps-office-bin/template
```diff
@@ -0,0 +1,58 @@
...
+maintainer="Michael Aldridge <aldridge.mac@gmail.com>"
+homepage="http://wps-community.org"
+license="Kingsoft WPS Community License"
+#Full license is at: http://wps-community.org/license.md (Not downloadable)
+short_desc="Linux office suite with similar appearance to MS Office"
+allow_unknown_shlibs=yes
+nodebug=yes
+hostmakedepends="rsync"
+depends="wps-office-libpng12"
+allow_unknown_shlibs=yes
```
dominikh: You set this variable twice

srcpkgs/wps-office-bin/template
```diff
@@ -0,0 +1,57 @@
+# Template file for 'wps-office-bin' (restricted)
+_suiteName=wps-office
+pkgname=${_suiteName}-bin
+version=10.1.0.5503
+revision=1
+_releaseIncrement=a20
+_patchLevel=p2
```
Vaelatern: Would be cleaner to include all the relevant information in the version, so if the patch level increases then you don't need to also revbump to have the package considered of a new version

srcpkgs/wps-office-bin/template
```diff
@@ -0,0 +1,57 @@
...
+maintainer="Michael Aldridge <aldridge.mac@gmail.com>"
+homepage="http://wps-community.org"
+license="Kingsoft WPS Community License"
+#Full license is at: http://wps-community.org/license.md (Not downloadable)
```
Vaelatern: Maybe (and I am not certain this is correct by any stretch) have `license="http://wps-community.org/license.md"` but urls in that field may be wrong.

the-maldridge: All other packages point to the name of the license.

the-maldridge:
Any reason this package is stalled?  I'm happy to make changes but afaik this package is good to merge.

Vaelatern:
Travis failures give a good indication: `=> ERROR: version contains invalid character: -`

the-maldridge:
If you'll read just above that, you'll see the build actually fails because the travis builders don't build restricted software.

Vaelatern:
`version contains invalid character: -`

the-maldridge:
@Vaelatern once again, if you look above, you'll see xlint passes that on both travis and local builds.  I assume at this point I should file a bug on xlint for inconsistent behavior?  I expect it to fail with that output in all places if it is in fact a build failure...

the-maldridge:
Fixed and ticket opened on xtools (xtools #38), xlint should not pass if xbps is going to fail on a reserved character.

the-maldridge:
bump

srcpkgs/wps-office-bin/template
```diff
@@ -0,0 +1,57 @@
+# Template file for 'wps-office-bin' (restricted)
+_suiteName=wps-office
+pkgname=${_suiteName}-bin
+version=10.1.0.5503${_releaseIncrement}${_patchLevel}
```
xtraeme: you are using two vars here that are set later on... this won't work correctly.

srcpkgs/wps-office-libpng12/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'wps-office-libpng12' (restricted)
```
xtraeme: Adding a separate pkg just for libpng12 does not seem too useful to me. Please merge this into wps-office-bin.

the-maldridge:
I have incorperate the feedback provided and the package now embeds its own version of libpng.

PR #3792, opened 2016-03-08
New Package: google-chrome-49.0.2623.75

dinoocch:
Package for the Google Chrome web browser.  The templating and install is based extensively on the chrome-pepper-flash package.
This PR continues progress on #3119.


xtraeme:
I would prefer to set `restricted=yes` and get rid of `INSTALL`.

Vaelatern:
Should we also do that for things like spotify? Having INSTALL and not having `restricted=yes` means people without the whole repository set up on their computer can use it.

xtraeme:
The problem with those packages that use INSTALL scripts is that they are dangerous:

- the file needs to be downloaded, extracted and unpacked into a location
- files need to be removed via REMOVE
- files are not tracked by xbps

With `restricted` those packages will be fully tracked by xbps and will work as expected once built.

Vaelatern:
In which case, how should a user who is not so technically adept get a package like spotify or google chrome? Do we care? Should we have a set of repositories under `current/restricted`, so to get an individual package you need to add that specific repository, but don't need to use the template?

the-maldridge:
Is this PR still alive?

Vaelatern:
@dinoocch has not yet made the changes to remove the `INSTALL` script.

Vaelatern:
In addition, `nopie=yes` must be set.

Vaelatern:
Superseded.

PR #3794, opened 2016-03-09
New package: lastpass-cli-0.9.0

jaagr:
Package for command line interface to LastPass.com

### DIFF
--- a/srcpkgs/lastpass-cli/template
+++ b/srcpkgs/lastpass-cli/template
@@ -1,0 +1,30 @@
+# Template file for 'lastpass-cli'
+pkgname=lastpass-cli
+version=0.9.0
+revision=1
+makedepends="asciidoc libressl-devel libcurl-devel libxml2-devel"
+depends="libressl curl libxml2"
+short_desc="Command line interface to LastPass.com"
+maintainer="Michael Carlberg <c@rlberg.se>"
+license="GPL-2"
+homepage="https://lastpass.com/"
+distfiles="https://github.com/lastpass/lastpass-cli/archive/v${version}.tar.gz"
+checksum=e7314f0dfeec86add7c19c053ee34bb7a176e462e71727d2b345481d2d136800
+
+do_build() {
+	cd $wrksrc
+	make all doc
+}
+
+do_install() {
+	cd $wrksrc
+	make DESTDIR="$DESTDIR" install install-doc
+
+	vbin lpass
+	vman lpass.1
+
+	vinstall contrib/lpass_bash_completion 644 usr/share/bash-completion/completions lpass
+
+	vmkdir usr/share/examples/lpass
+	vcopy contrib/examples/* usr/share/examples/lpass
+}
--- a/srcpkgs/lastpass-cli/template
+++ b/srcpkgs/lastpass-cli/template
@@ -2,8 +2,11 @@
 pkgname=lastpass-cli
 version=0.9.0
 revision=1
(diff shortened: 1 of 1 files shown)


srcpkgs/lastpass-cli/template
```diff
@@ -0,0 +1,30 @@
...
+makedepends="asciidoc libressl-devel libcurl-devel libxml2-devel"
+depends="libressl curl libxml2"
+short_desc="Command line interface to LastPass.com"
+maintainer="Michael Carlberg <c@rlberg.se>"
+license="GPL-2"
+homepage="https://lastpass.com/"
+distfiles="https://github.com/lastpass/lastpass-cli/archive/v${version}.tar.gz"
+checksum=e7314f0dfeec86add7c19c053ee34bb7a176e462e71727d2b345481d2d136800
+
+do_build() {
```
Vaelatern: `build_style=gnu-makefile`
`make_build_target="all doc"
`make_install_target="install install-doc"`

Vaelatern: This whole function is unecessary.

srcpkgs/lastpass-cli/template
```diff
@@ -0,0 +1,30 @@
...
+	cd $wrksrc
+	make all doc
+}
+
+do_install() {
+	cd $wrksrc
```
Vaelatern: No need for `cd $wrksrc`

srcpkgs/lastpass-cli/template
```diff
@@ -0,0 +1,30 @@
+# Template file for 'lastpass-cli'
+pkgname=lastpass-cli
+version=0.9.0
+revision=1
+makedepends="asciidoc libressl-devel libcurl-devel libxml2-devel"
```
Vaelatern: `asciidoc` should be in `hostmakedepends`

srcpkgs/lastpass-cli/template
```diff
@@ -0,0 +1,30 @@
...
+depends="libressl curl libxml2"
```
Vaelatern: `depends="curl libxml2"`

dominikh: Why are curl and libxml2 in depends? Unless it's actually shelling out to the curl binary, curl isn't needed. And unless it is using one of the utilities in libxml2, that's not needed, either. Library dependencies get detected automatically.

srcpkgs/lastpass-cli/template
ebfe: Not entirely, with the `gnu-makefile` do_build it requires a Makefile patch, so not sure if this is an improvement.

srcpkgs/lastpass-cli/template
```diff
@@ -0,0 +1,32 @@
...
+homepage="https://lastpass.com/"
+distfiles="https://github.com/lastpass/lastpass-cli/archive/v${version}.tar.gz"
+checksum=e7314f0dfeec86add7c19c053ee34bb7a176e462e71727d2b345481d2d136800
+
+pre_build() {
+	sed -i 's/^CFLAGS +=/override CFLAGS +=/g' Makefile
+	sed -i 's/^LDLIBS +=/override LDLIBS +=/g' Makefile
+}
+
+do_install() {
```
Vaelatern: If you change this to `post_install` then you can remove the `make` line, the `vbin` line, and the `vman` line.

PR #3796, opened 2016-03-09
Quassel

Vaelatern:
Make the 3 templates a single template so when the update to `0.12.3` happens (pending this PR merging), it's a one-stop change instead of 3.
I do some strange-to-me things with subpackages so would like more eyes on this.

### DIFF
--- a/srcpkgs/quassel/template
+++ b/srcpkgs/quassel/template
@@ -1,7 +1,7 @@
 # Template file for 'quassel'
 pkgname=quassel
-version=0.12.2
-revision=2
+version=0.12.3
+revision=1
 build_style=cmake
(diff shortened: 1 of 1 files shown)


srcpkgs/quassel/template
```diff
@@ -1,17 +1,68 @@
...
+	conflicts="quasselcore>=0"
+	depends+=" qt-plugin-sqlite quassel-shared>=${version}_${revision}"
+	short_desc+=" - server"
+	pkg_install() {
+		vmove /usr/bin/quasselcore
+		vmove /etc/sv
+	}
+}
+
+quassel_package() {
```
xtraeme: This one is unnecessary and those vars should be set as part of the main package.

PR #3797, opened 2016-03-09
build-style/gnu-makefile: allow ?= in Makefiles

Vaelatern:
``` 
05:23 <ebfe> gnu-makefile doesnt work with makefiles that do CFLAGS?=... CFLAGS+=...
05:24 <ebfe> which is pretty common
...
05:28 <ebfe> in my view, the makefiles do the right thing with CFLAGS?=<can be changed> 
CFLAGS+=<required>. The gnu-makefile build-style is wrong imo.
...
05:35 <Vaelatern> Why does gnu-makefile not work with ?=?
05:36 <ebfe> because it does `make CFLAGS=` which does not allow further modification
05:36 <ebfe> maybe should have been `CFLAGS=  make` but it's probably too late to change that
05:38 <Vaelatern> Only 343 templates
05:38 <Vaelatern> Less because of symlinks
05:39 <Vaelatern> To test it I'd need to get all the builds working on my machine, test it on every template, fix the then-broken ones...
05:39 <Vaelatern> If it was `CFLAGS= make` would that allowe CFLAGS?=... CFLAGS+=... ?
05:39 <ebfe> yes
```

Thoughts?


chneukirchen:
Most Makefiles use `CFLAGS = ...` and then ignore ours.

Vaelatern:
Would a `make CFLAGS+=` work better? @ebfe @chneukirchen

ebfe:
> Would a `make CFLAGS+=` work better?
No, thats basically the same as it is now. Probably best to leave
things as they are, and use explicit do_build, do_install for sane
Makefiles.

PR #3798, opened 2016-03-09
jo: update to 0.7

jpmens:
jo: update to 0.7


dominikh:
Wrong commit message.

jpmens:
I was trying to be helpful.
It would help if you'd explain what's wrong about the commit message (other than that it's empty)

dominikh:
The commit message (which includes the "subject") for package updates should be of the form "<package>: update to <version>", so in this case "jo: update to 0.7" instead of "bump to v0.7".

Sorry if I came across as rude.

dominikh:
I've rewritten the commit message to fit. You only updated the PR subject, not the actual git commit; but there's no point in extra roundtrips for a trivial commit. Thanks for the update.

PR #3799, opened 2016-03-09
firefox: update to 45.0.

jannisch:


Gottox:
Please test your build before opening a PR. :smile:

jannisch:
I thought it's not so uncommon to rely on Travis since there are so many architectures and it takes very long to build but thanks for the advise.

Duncaen:
I would prefer to merge something like firefox only if its tested, there where some updates in the past that made problems and everyone had to downgrade, would be nice if this is avoidable.

jannisch:
I see. Building firefox from the development sources is even more annoying.
Do we need to test this on several platforms? I only have a x86-musl notebook right now, hopefully compiling has finished by the end of this evening :laughing:

jannisch:
This is strange, I built it successfully on my x86-64-musl machine and it also works as expected but the Travis build fails.

srcpkgs/firefox/patches/firefox-install-dir.patch
```diff
@@ -11,7 +11,7 @@
 +includedir := $(includedir)/$(MOZ_APP_NAME)
 +idldir = $(datadir)/idl/$(MOZ_APP_NAME)
 +installdir = $(libdir)/$(MOZ_APP_NAME)
-+sdkdir = $(libdir)/$(MOZ_APP_NAME)-devel
- ifndef TOP_DIST
- TOP_DIST = dist
- endif
++sdkdir = $(libdir)/$(MOZ_APP_NAME)-deve
```
xtraeme: typo `devel` vs `deve`

xtraeme:
Travis couldn't build firefox probably due to not having enough RAM. The compiler process was killed.

If you've built this locally there shouldn't be any issue.

xtraeme:
Also please rebase your branch with master and use a proper commit.

jannisch:
I hope all necessary work is done now. Firefox 45.0 definitely builds locally and runs stable.
Thanks for your support.

PR #3800, opened 2016-03-10
haproxy: fixed run script so the daemon no longer forks to the background

jasonMoggridge:
The service for haproxy was forking to the background on start causing runit to repeatably spawn new instances eventually crashing the system.

Added a command line flag to prevent the daemon from forking into the background. 

Relevant flag documentation: [haproxy manual](http://cbonte.github.io/haproxy-dconv/configuration-1.7.html#3.1-daemon)


Vaelatern:
Service works fine on my machine. Suggest you try running the daemon yourself to see if it actually forks to background, and if so, if that's somehow in your configuration.

Gottox:
From the manpage:
> For debugging, the '-db' option is very useful as it temporarily disables daemon mode and multi-process mode.

This is not what we want. haproxy needs the `-D` flag to fork to backround. Without any flag it has the desired behavior.

jasonMoggridge:
If the daemon variable is set within the configuration it will still fork, the -db will override that. It seems a simple preventative measure to avoid an entire system crashing.

PR #3801, opened 2016-03-10
hex2hcd: remove incorrect conflict variable

antonio-malcolm:
The conflict added, at the previous update (revision 2), with bluez, is incorrect.
It was added with a comment, which mentioned /usr/bin/hexdump, and I think there's some confusion, here. This template installs a small binary /usr/bin/hex2hcd, and does nothing with hexdump. Nor should there exist any conflict with bluez. Probably, the chance of this conflicting with anything is very, very small.

If there is some conflict with bluez, could someone please clarify what, exactly, that conflict is? Because, on the surface, it makes no sense. The last revision made was by efbe (Michael Gehring).


Vaelatern:
No. https://gist.github.com/5353bf33d9bacfbe36fe it contains a binary of the same name and path. It conflicts.

Gottox:
but good point, we should rename that binary. imho.

antonio-malcolm:
Ok. I have the core bluez package installed, and an install of hex2hcd, from my local repo, and aren't experiencing any issues (the bluez package didn't install a hex2hcd binary). I don't think the original author of hex2hcd maintains it, anymore (I opened a pull request, for an update I made, with no response, and it doesn't look like he's updated it, inn some time). I'd be opened to renaming it hextohcd, or something along those lines, but that sounds similar enough to be confusing to someone. 

Literally, the binary converts a .hex file to a Linux-compatible .hcd file. That's it. That's its mission in life. As such, I'm thinking naming conventions are a tad limited.

Vaelatern:
Can you install bluez again and see if that overwrites the binary? It really should.

Vaelatern:
https://gist.github.com/anonymous/7cd66b6f75f25cf6c6d4 both packages install the same file. The correct course of action would be to change the name of one of the binaries.

antonio-malcolm:
Actually, I'm thumbing through the source code ot both, right now...

antonio-malcolm:
Well, they're both written to convert file formats for firmware files. I had thought, for a moment, one may be derived from the other (as the standalone was developed in 2012, and Bluez released their's last year), but they're quite different. I can rename the binary- I think hextohcd is fine, considering the differences. I think it might be best to change the package name, too.

Vaelatern:
Different kinds of firmware files?

antonio-malcolm:
They're both targeted at converting broadcom firmware files.

antonio-malcolm:
(The big difference is in how they're writ, with the older version using more boilerplate code.l

Vaelatern:
Why does bluez distribute the binary? May be worth putting it in a separate subpackage....

Gottox:
@antonio-malcolm are they both option compatible? If yes, we can just remove the file from bluez.

antonio-malcolm:
No. The Bluez binary takes the output arg as an option (hex2hcd <input file> -o <output file>), and only requires the first arg (with which it will output to the same location as the input file, by default).

The older binary takes two args (hex2hcd <input file> <output file>), and requires both.

If you attempt to call the Bluez binary the same way as the older (specifiying the output file without the -o option), it creates an empty file.

It's probably easiest if I take some time, this weekend, to rename the binary and package. It's a tiny program, really, so this shouldn't be difficult. I'd much rather leave Bluez' stuff as-is.

I'm not sure why Bluez distributes their own binary. I imagine they're trying to do folks a favor (though I wonder how many know it's there). I found the older binary, as I needed a means of converting the broadcom firmware, for the controller in my Thinkpad.

antonio-malcolm:
BTW, best I can tell, they both work fine. If we wanted to offer Bluez' hex2hcd, as its own package, instead, that's good, too.

PR #3802, opened 2016-03-10
perl-Glib: fix cross builds

the-maldridge:
For certain definitions of "fixed".

Builds fail with a myriad of type width errors, then finally with an error trying to load a glib which is the wrong ELFCLASS for the target.


xtraeme:
Can you please fix the commit message? this is not a fix :-)

the-maldridge:
I have updated the commit message as requested

Issue #2929, opened 2015-11-10
Package Request: shutter


sunaku:
http://shutter-project.org/
> Shutter is a feature-rich screenshot program for Linux based operating systems such as Ubuntu. You can take a screenshot of a specific area, window, your whole screen, or even of a website – **apply different effects to it, draw on it to highlight points, and then upload** to an image hosting site, all within one window. Shutter is free, open-source, and licensed under GPL v3.

dominikh:
- Written in perl
- Needs perl-Gnome2 (+ some perl-Gnome2-* packages)
- Needs libbonobo, which adds some odd activation server
- Needs libgnomeui

Maybe someone who is bored wants to package this and its dependencies.

sunaku:
Ouch, that's quite the laundry list of unmet dependencies. :cold_sweat:

the-maldridge:
This software relies on absolutely ancient dependencies which cannot be built on a modern system.  Ultimately the final nail in the coffin for this program is Gnome-VFS, the precursor to gvfs.  Certain types and threading primatives were removed from later versions of the Gnome libraries and thus the dependencies for shutter cannot be built without significant modernization of libraries.

I suggest closing this issue.

cr6git:
@the-maldridge still open, why?

PR #3803, opened 2016-03-10
New package: perl-Gnome2-Canvas

the-maldridge:
MUST be applied after e871cbc, not entirely sure how that commit got into this PR.  This PR is part of the dependency chain for: #2929.


the-maldridge:
Original package turns out to have impossible to satisfy dependencies, therefore this package is not needed.

PR #3807, opened 2016-03-11
soulseekqt: update to 2016-1-17

jarbowski:


jarbowski:
yes, it seems that soulseekqt devs have chosen this. you can check the urls on the site and compare the sha256sums, everything is ok

srcpkgs/soulseekqt/template
```diff
@@ -1,22 +1,21 @@
...
 	_pkgver=$_pkgver-64bit
-	_disttarball="${_url}/SoulseekQt-$_pkgver.tgz"
-	_distcksum=a098d9bfdf68b6f944ef83716e3c94e7abc20738cede7af5c9d78aa23a62207e
+	_disttarball="https://dl.dropboxusercontent.com/u/7226803/SoulseekQt-$_pkgver.tgz"
+	_distcksum=8c11af264ffdd045de1093090533fc7b86e8814442a14d366c1df1dd6ed58316
 else
 	_pkgver=$_pkgver-32bit
-	_disttarball="${_url}/SoulseekQt-$_pkgver.tgz"
-	_distcksum=c2d19e8511ebba298bf44243fb0e467717b4e06ecfcbc7237ae1157586f73979
+	_disttarball="https://dl.dropboxusercontent.com/u/7226803/SoulseekQt-$_pkgver.tgz"
```
Vaelatern: This variable can be set outside the if statement

srcpkgs/soulseekqt/template
```diff
@@ -1,22 +1,21 @@
 # Template file for 'soulseekqt'
 pkgname=soulseekqt
-version=20150625
+version=20160117
 revision=1
-_pkgver=2015-6-25
-_url=http://www.soulseekqt.net/SoulseekQT/Linux
+_pkgver=2016-1-17
```
Vaelatern: Would likely be cleaner to have a `version=2016:1:17` and `_pkgver=${version//:/-}` so not duplicated information.

jarbowski:
ok for the distfiles url but I don't understand the purpose of the version thing. sorry, but I have limited skills, as you can see

Vaelatern:
Well basically it's just less to update when there is an update to the software. Especially since the information is essentially duplicated as it currently exists, it's just more confusing to someone trying to bump the verison.

jarbowski:
commit and rebase but I still don't understand the version naming

PR #3808, opened 2016-03-11
New package: gnome-colors-icon-theme-5.5.1

pbui:


srcpkgs/gnome-colors-icon-theme/template
```diff
@@ -0,0 +1,22 @@
...
+homepage="https://github.com/gnome-colors/gnome-colors"
+distfiles="http://gnome-colors.googlecode.com/files/gnome-colors-${version}.tar.gz"
+checksum=1e91d845a753eb38671e7614a786f852ab858a12983111e780da4a47c9debe58
+wrksrc=${pkgname}
+
+do_extract() {
+	mkdir -p "${wrksrc}"
+	tar -xf "${XBPS_SRCDISTDIR}/${pkgname}-${version}/gnome-colors-${version}.tar.gz" \
+	    -C "${wrksrc}"
+}
```
xtraeme: This is really unnecessary... set `create_wrksrc` instead.

pbui: Huh... I tried that earlier and it didn't work... but now it does...  I most have forgotten to clean up between builds.

Anyway, I've made the change.

PR #3815, opened 2016-03-12
New package: vt-setcolors-1.0.0

jaagr:


Vaelatern:
Please use a tagged release. Relevant: EvanPurkhiser/linux-vt-setcolors#5

srcpkgs/vt-setcolors/template
```diff
@@ -0,0 +1,43 @@
+# Template file for 'vt-setcolors'
+pkgname=vt-setcolors
+version=1.0.0
+revision=1
+only_for_archs="i686 i686-musl x86_64 x86_64-musl"
```
Vaelatern: Mind saying why? A comment or a case statement marking arm* broken would be appreciated.

PR #3819, opened 2016-03-12
intellij-idea-community-edition: update to 15.0.4

Grauwolf:


ebfe:
current is 2016.1 btw.

PR #3822, opened 2016-03-13
New package: vte3-ng-0.42.4

jaagr:
VTE build with required patch to build Termite


srcpkgs/vte3-ng/template
```diff
@@ -0,0 +1,55 @@
...
+vte3-ng-devel_package() {
+	replaces="vte3-devel>=0"
+	conflicts="vte3-devel>=0"
+	depends="vte3-ng>=${version}_${revision}"
+	short_desc+=" - development files"
+
+	post_extract() {
+		vteversion=$(sed -nr 's/^VTE = (vte-[0-9.]+)$/\1/p' Makefile)
+		shlib_provides="libvte-${_vte_version}.so.0"
+	}
```
xtraeme: This `post_extract` won't have any effect... only `pkg_install` is processed in subpkgs.

srcpkgs/vte3-ng/template
```diff
@@ -0,0 +1,55 @@
...
+ $(vopt_if gir 'gobject-introspection vala-devel')"
+makedepends="gtk+3-devel ncurses-devel gnutls-devel"
+short_desc="Enhanced Virtual Terminal Emulator widget for use with GTK3"
+maintainer="Michael Carlberg <c@rlberg.se>"
+homepage="https://github.com/thestinger/vte-ng"
+license="LGPL-2.1"
+distfiles="${GNOME_SITE}/vte/${version%.*}/vte-${version}.tar.xz"
+checksum=08c88bf5c0aa1dfa3711e2e83b784fb5ea82025661f30b54e93eebd5f4bce113
+
+provides="vte3-${version}_${revision} vte3-ng-${version}_${revision}"
```
xtraeme: `vte3-ng-...` is redundant.

srcpkgs/vte3-ng/template
jaagr: Would it work if I moved the function to the parent scope, storing the `vteversion` variable and then later in `pkg_install` call `shlib_provides`?

xtraeme: This is not the way to register shlibs, use `common/shlibs` instead.

xtraeme: And the shlib must be stored in the main package, not the development one.

Gottox:
I'm not a huge fan of slightly patched variants of packages we already have. I don't think it's worth the effort just to get yet another terminal into void.

xtraeme:
What @Gottox said

jaagr:
I would agree if there was a good replacement for the terminal emulator.

Vaelatern:
What's special about this terminal emulator as opposed to every one we have in the repos?

Gottox:
I guess this won't fly.

Vintodrimmer:
Good evening,

Have your opinion on the subject changed? I would also like to see termite in the repos, if possible.

the-maldridge:
@Vintodrimmer iirc something came up on IRC recently about this.  The basic idea is that we don't want to be maintaining a bunch of different incompatible versions of the same library.  As soon as this can work without a custom build, we're happy to merge it.

PR #3828, opened 2016-03-13
pinentry: build pinentry-emacs.

CMB:
This requires some extra code on the emacs side to work.
With 24.5, that's a package from elpa.  So I won't be
the least bit disappointed if this PR gets rejected.

### DIFF
--- a/srcpkgs/pinentry-emacs
+++ b/srcpkgs/pinentry-emacs
@@ -1,0 +1,1 @@
+pinentry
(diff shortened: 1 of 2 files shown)


Vaelatern:
What in the file list changes when you make this change? If the change is just a new set of files included with the rest, perhaps this should be a subpackage?

CMB:
Ah, you get the pinentry-emacs executable, and that's it.
This should definitely be a subpackage, and I have updated the pull
request accordingly.  Thanks.

PR #3830, opened 2016-03-14
vivaldi: update to 1.0.403.24

diogoleal:


Vaelatern:
Still a package with a version it claims as: beta 3. How legitimate is this package, and how legitimate is this upgrade?

PR #3832, opened 2016-03-15
New package: snapraid-10.0

evandeaubl:


srcpkgs/snapraid/template
```diff
@@ -0,0 +1,11 @@
+# Template file for 'snapraid'
+pkgname=snapraid
+version=10.0
+revision=1
+build_style=gnu-configure
+short_desc="A RAID-style backup program for disk arrays"
+maintainer="Evan Deaubl <evan@deaubl.name>"
+license="GPL-3"
+homepage="http://snapraid.it"
```
Vaelatern: Bogus homepage.

evandeaubl:
Thanks for the feedback, fixed. Thanks to my browser for glossing over the difference between www/non-www...

srcpkgs/snapraid/template
```diff
@@ -0,0 +1,16 @@
...
+depends="smartmontools"
+short_desc="A RAID-style backup program for disk arrays"
+maintainer="Evan Deaubl <evan@deaubl.name>"
+license="GPL-3"
+homepage="http://www.snapraid.it"
+distfiles="https://github.com/amadvance/${pkgname}/releases/download/v${version}/${pkgname}-${version}.tar.gz"
+checksum=f7dcf19480256fc2c1db9ab976aa12f786e76da6044cc397f0451524e8031ad6
+
+post_install() {
+	vsconf snapraid.conf.example
```
Vaelatern: This is an example, not the actual configuration. Should not be `vsconf`.

evandeaubl: Per the manual, `vsconf` copies the file into `/usr/share/examples`. Judging from other templates, that should be for example configurations, correct?

Vaelatern: Ah, missed the `s`

Vaelatern:
Is there a specific configuration location for the config files? Such as `/etc/snapraid.conf` or some such? If so, it should be in `conf_files` so xbps treats it correctly.

evandeaubl:
Hmm, kind of a thorny issue. SnapRAID looks at /etc/snapraid.conf by default, but it doesn't really own it (meaning the install process doesn't install anything there by default). I decided to put a stub config in /etc/snapraid.conf that simply has a comment to copy in the example config. It allows the conf_files handling to work, and it seemed the safest, because the example config could accidentally work if your system just happened to have the paths as configured in the example (unlikely I know, but...)

Other distros are confused on the issue as well: the Ubuntu package dumps the example config in /etc, the Fedora package dumps nothing in /etc...

Vaelatern:
Proper way to do the echos in your post_install probably is `vconf snapraid.conf.example snapraid.conf` to change the name and do the copying already.

evandeaubl:
Not sure why I would do the vconf when that will copy the example config into /etc/snapraid.conf, which is what I don't want. Does vconf have other side effects?

Vaelatern:
Why don't you want the example config in the configuration file location?

evandeaubl:
I guess I thought it would be a little safer putting in a stub config that is guaranteed to not work, because the example config contains a valid configuration that could accidentally work if you had the correct paths available on the target system. That said, it's highly unlikely for that to be the case, but I guess I prefer certainty to luck. :-)

No distro seems to package SnapRAID the same way, and I didn't see a consistent convention for handling it in Void. If you think that copying the example config in /etc/snapraid.conf would be fine, I don't have any objections.

evandeaubl:
I made the change to put the example config in /etc/snapraid.conf. Just wanted to check in because there was no movement. Are there any other issues holding up this pull request?

Vaelatern:
Given that the example configuration could work, it's just a couple paths to change for some working configurations, it's probably the right move to install the example as a configuration file. I'll review this again later.

PR #3835, opened 2016-03-15
New package: xastir-2.0.6

the-maldridge:
Some more amatuer radio software.  The build is marked broken on all arm and musl systems.  Years of code cruft mean this is not likely to change anytime soon, however conversations with the upstream project suggest they will consider patching thier build in the next release cycle, which if current rate is any example, should be some time next year.

### DIFF
--- a/srcpkgs/xastir/template
+++ b/srcpkgs/xastir/template
@@ -1,0 +1,13 @@
+# Template file for 'xastir'
+pkgname=xastir
+version=2.0.6
+revision=1
+build_style=gnu-configure
+only_for_archs="i686 x86_64"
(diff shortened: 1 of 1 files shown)


the-maldridge:
I have updated my PR after futher conversation with the upstream devs.  There is currently no interest from upstream to target ARM or musl.  As such I have changed the broken statements to instead reflect this package is only_for_archs x86 x86_64.

srcpkgs/xastir/template
```diff
@@ -0,0 +1,13 @@
+# Template file for 'xastir'
+pkgname=xastir
+version=2.0.6
+revision=1
+build_style=gnu-configure
+only_for_archs="x86 x86_64"
```
xtraeme: s/x86/i686/

the-maldridge:
Token fixed.

PR #3848, opened 2016-03-15
[NOMERGE] Bump isl

Vaelatern:
@chneukirchen maybe this will do the job?


Vaelatern:
Unless perhaps we need to push it one commit at a time. Review is necessary and Travis is not able to provide it.

Vaelatern:
Abandoned until gcc 5.

ebfe:
gcc6 already here.

Issue #3853, opened 2016-03-16
WIP: Qt5 5.6.0


pullmoll:
The new version 5.6.0 has changes which require a little more work as usual to be done

+ [x] some (musl) patches became obsolete or need to be adapted
+ [x] `qtquick1` is gone; there will be no `qt5-quick1{,-devel}` package anymore
+ [x] the `qt5-doc` hackery stopped working; remove `qt5-doc` subpackage from `qt5`
+ [x] Create new pkg `qt5-doc` and depend on `qt5-tools-devel` and what else is needed
+ [x] besides `qt5-quickcontrols` there is a new subpkg `qt5-quickcontrols2`
+ [x] new dependency: `qtwebengine` needs `libXScrnSaver-devel` (at least)
+ [x] `qtwebkit` is gone as part of the main archive; remove `qt5-webkit{,-devel}` from `qt5`
+ [x] `qtwebkit` is is still available for `5.6.0` as a separate tarball; build new pkgs `qt5-webkit{,-devel}`
    + See: https://projects.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/qt5-webkit
+ [x] create a new subpkg `qt5-examples` after removing the `-nomake examples` option

pullmoll:
After trying to upgrade from `5.5.1` to `5.6.0` without success (due to now missing subgpkgs) instead create a new package `qt56-5.6.0`.
Use `/usr/lib/qt56` as the prefix and rename anything `qt5` to `qt56` to allow parallel installation and upgrading of dependent packages one after another.

equalsraf:
FYI I was testing the Qt 5.6.0 binaries earlier and ran into this issue https://github.com/QupZilla/qupzilla/issues/1870#issuecomment-189643688

asergi:
@pullmoll about the upgrade problem: if haven't got it wrong, I had a similar issue back when upgrading IPython to 4.0.

As a workaround, just add to the `qt5` template: `replaces="qt5-doc<5.6.0 qt5-foo<5.6.0"`, where `qt5-foo` is the now non-exisisting subpackage.

So, this way there is no need to rename (unless there are other reasons).
Hope it helps.

pullmoll:
@asergi I'm stuck with the try to rename, because I get wrong dependencies to qt5-* packages in the qt56-branch due to duplicates in common/shlibs.

Now that at least qt5-webkit-5.6.0 builds I'm trying again to replace qt5.

[...]

PR #3855, opened 2016-03-17
qt5: update 5.6.0

pullmoll:
See #3853


pullmoll:
As it is now `xbps-install -S -R /data/void-glibc/binpkgs/qt5.6 qt5` does not work.

The missing packages `qt5-doc`, `qt5-webkit{,-devel}`, `qt5-quick1{,-devel}` prevent this.

I'm working on separate `qt5-doc` and `qt5-webkit{,-devel}` packages but `qt5-quick1` is definitely gone. How do I remove it without the need to rebuild every(?) dependency on `qt5`?

pullmoll:
From my point of view the PR is now completed. I could update `qt5` with my locally built packages for `x86_64` and pkgs depending on `qt5` continue to work as expected.

I'd be happy if someone would scrutinize the changes and merge, if everything seems ok.

xtraeme:
:+1:

srcpkgs/qt5-doc/template
```diff
@@ -0,0 +1,47 @@
+# Template file for 'qt56-doc'
```
Vaelatern: May as well be updated to be the same as the pkgname.

pullmoll:
Actually there are some duplicate files left in the qt5 packages. A complete list is here: http://sprunge.us/deKU

The functions `_install_subpkg()` and `_install_devel()` need to `rm` some more files. I'll take a look at what's left that should not be.

srcpkgs/qt5-doc/template
pullmoll: Good catch :)

pullmoll:
Duplicate files are gone, all seems fine now.

Gottox:
**Never ever** remove (sub-)packages if you don't know *exactly* what you are doing. What happened:

1. qtcreator-full depends on qt5-quick1-devel
2. this PR removed qt5-quick1-devel but didn't update the corresponding packages that depend on that package
3. This:

```
# sudo xbps-install -u 
MISSING: qt5-script-5.5.1_4
Transaction aborted due to unresolved dependencies.
```

pullmoll:
D'OH! I didn't think about that. I should have checked for anyone using qt5-quick1.

The qt5-quick1 is definitely gone and non pkg really linked against it - this was what I verified and what wasn't sufficient.

Gottox:
maybe it's time for a xbps-src check-consistency :)

PR #3857, opened 2016-03-17
lemonbar: add xft and xft_monitor build options

jaagr:


Vaelatern:
Only one commit is necessary.

chneukirchen:
and please write an informative subject like "lemonbar: add xft and xft_monitor build options".

chneukirchen:
also, is xtraeme ok with you adopting the package?

jaagr:
@chneukirchen I never asked him, but I reverted the maintainer change

PR #3864, opened 2016-03-18
sudo: update to 1.8.16.

ebfe:
xlint failure is bogus.


Vaelatern:
Alternately those 2 lines can be made one xlint.

PR #3866, opened 2016-03-18
mpv: only enable rubberband if not cross compiling

Vaelatern:
This seems to be a problem with the builders partially; they keep trying to build mpv despite not being able to due to cross issues.

### DIFF
--- a/srcpkgs/mpv/template
+++ b/srcpkgs/mpv/template
@@ -1,7 +1,7 @@
 # Template file for 'mpv'
 pkgname=mpv
 version=0.16.0
-revision=5
+revision=6
 short_desc="Video player based on MPlayer/mplayer2"
 maintainer="Juan RP <xtraeme@voidlinux.eu>"
 license="GPL-3"
@@ -17,7 +17,7 @@ makedepends="
  pulseaudio-devel libbluray-devel libcdio-paranoia-devel libdvdread-devel
  MesaLib-devel harfbuzz-devel libXScrnSaver-devel jack-devel libdvdnav-devel
  wayland-devel libuuid-devel libguess-devel libXrandr-devel samba-devel
- $(vopt_if build_option_rubberband rubberband-devel)"
+ $(vopt_if rubberband rubberband-devel)"
 depends="desktop-file-utils hicolor-icon-theme youtube-dl"
 conf_files="/etc/mpv/*.conf"
 distfiles="https://github.com/mpv-player/${pkgname}/archive/v${version}.tar.gz"


srcpkgs/mpv/template
```diff
@@ -24,6 +24,8 @@ checksum=fc3619de0ede16fbb023ac72589090e8e77fd9d9e03a81adc728105d50ef38ba
 
 if [ -z "$CROSS_BUILD" ]; then
 	configure_args+=" --enable-zsh-comp"
+	makedepends+=" rubberband-devel"
+	# ^ doesn't cross compile and gets stuck on the builders for it.
```
Gottox: I would prefere a build_option that is default disabled on CROSS_BUILD.

srcpkgs/mpv/template
```diff
@@ -16,14 +16,19 @@ makedepends="
...
 conf_files="/etc/mpv/*.conf"
 distfiles="https://github.com/mpv-player/${pkgname}/archive/v${version}.tar.gz"
 checksum=fc3619de0ede16fbb023ac72589090e8e77fd9d9e03a81adc728105d50ef38ba
 
+build_options="rubberband"
+build_options_default=""
+
 if [ -z "$CROSS_BUILD" ]; then
 	configure_args+=" --enable-zsh-comp"
+	build_options_default+=" rubberband"
```
Vaelatern: Is this sort of mechanism appropiate? Alternately could `build_options_default="rubberband"` and `build_options_default=${build_options_default/rubberband/}` but that doesn't clear it up any more.

srcpkgs/mpv/template
```diff
@@ -16,14 +16,18 @@ makedepends="
...
 depends="desktop-file-utils hicolor-icon-theme youtube-dl"
 conf_files="/etc/mpv/*.conf"
 distfiles="https://github.com/mpv-player/${pkgname}/archive/v${version}.tar.gz"
 checksum=fc3619de0ede16fbb023ac72589090e8e77fd9d9e03a81adc728105d50ef38ba
 
+build_options="rubberband"
+
 if [ -z "$CROSS_BUILD" ]; then
 	configure_args+=" --enable-zsh-comp"
+	build_options_default+=" rubberband"
```
Vaelatern: This isn't "Disable if crossing" but rather "Enable if not," which has the meager gain of being easier to write. Okay, @Gottox ?

srcpkgs/mpv/template
```diff
@@ -16,14 +16,18 @@ makedepends="
  libXv-devel libxkbcommon-devel libva-glx-devel libvdpau-devel alsa-lib-devel
  pulseaudio-devel libbluray-devel libcdio-paranoia-devel libdvdread-devel
  MesaLib-devel harfbuzz-devel libXScrnSaver-devel jack-devel libdvdnav-devel
- wayland-devel libuuid-devel libguess-devel libXrandr-devel samba-devel rubberband-devel"
+ wayland-devel libuuid-devel libguess-devel libXrandr-devel samba-devel
+ $(vopt_if build_option_rubberband rubberband-devel)"
```
dominikh: This should be `vopt_if rubberband`

PR #3872, opened 2016-03-20
New package: html-xml-utils-6.9

ebfe:


srcpkgs/html-xml-utils/patches/musl.patch
```diff
@@ -0,0 +1,12 @@
+--- hxnum.c.orig	2016-03-20 10:29:20.905978225 +0100
++++ hxnum.c	2016-03-20 10:30:27.952140032 +0100
+@@ -50,6 +50,9 @@
+   "%d.%d.%d.%d.%d.", "%d.%d.%d.%d.%d.%d."};
+ static int skipping = 0;			/* >0 to suppress output */
+ 
++#if !defined(__GLIBC__)
++extern int optreset; /* musl */
```
chneukirchen: optreset is declared in getopt.h, just include that...

Else +1.

ebfe:
c066a33257eadd0e655427ec083f974dd6df6c7f

PR #3875, opened 2016-03-20
Split p7zip into p7zip and p7zip-unrar

dominikh:
The RAR support in p7zip is based on unrar's code and unfree.


chneukirchen:
Why not vmkdir? Else LGTM.

dominikh:
> Why not vmkdir? 

Stylistic choice, to match the mv on the next line.

PR #3878, opened 2016-03-21
crypto++: update to 563

DestroyFX:


xtraeme:
Please squash those commits into one.

Thanks :+1:

PR #3788, opened 2016-03-08
ffmpeg: update to 3.1.3.

Vaelatern:
Still needs any revdeps added. What's the command to find reverse dependencies that need rebuilding after a shlib bump?

2.8-3.0 is a relatively simple changelog for ffmpeg, from a packaging perspective.

PR #3880, opened 2016-03-21
New package: libvidstab-0.98b

lemmi:


Gottox:
May conflict with https://github.com/voidlinux/void-packages/pull/3788

Vaelatern:
Definitely would. But so does every revbump... when I rebase I do so very carefully to not lose commits.
I doubt ffmpeg would remove support for this when they updated to v3.0, but I have read no literature that supports this.

srcpkgs/libvidstab/template
```diff
@@ -0,0 +1,32 @@
...
+license="GPL-2"
+homepage="http://public.hronopik.de/vid.stab/"
+distfiles="https://github.com/georgmartius/vid.stab/archive/release-${version}.tar.gz"
+checksum=530f0bf7479ec89d9326af3a286a15d7d6a90fcafbb641e3b8bdb8d05637d025
+
+pre_configure() {
+	sed -i -e 's/include (FindSSE)//' CMakeLists.txt
+	case "$XBPS_TARGET_MACHINE" in
+		i686*|x86_64*)
+			configure_args+=" -D SSE2_FOUND:BOOL=true"
```
xtraeme: `-DSSE2_FOUND=1` is enough too

ebfe:
merge conflict

Issue #3879, opened 2016-03-21
Python shebangs


chneukirchen:
Python shebang lines should explicitly mention the Python version used.

On my system, this is violated by 

- [ ] /usr/bin/bno_plot.py:#! /usr/bin/env python
- [ ] /usr/bin/d-feet:#!/usr/bin/python
- [ ] /usr/bin/dstat:#!/usr/bin/env python
- [ ] /usr/bin/gdbus-codegen:#!/usr/bin/python
- [ ] /usr/bin/git-clang-format:#!/usr/bin/env python
- [ ] /usr/bin/gnome-tweak-tool:#!/usr/bin/env python
- [ ] /usr/bin/gsettings-schema-convert:#!/usr/bin/env python
- [ ] /usr/bin/gtester-report:#!/usr/bin/python
- [ ] /usr/bin/gtk-builder-convert:#!/usr/bin/env python
- [ ] /usr/bin/mountstats:#!/usr/bin/env python
- [ ] /usr/bin/mpaste:#!/usr/bin/python
- [ ] /usr/bin/nfsiostat:#!/usr/bin/python
- [ ] /usr/bin/purple-remote:#!/usr/bin/env python
- [ ] /usr/bin/purple-url-handler:#!/usr/bin/env python
- [ ] /usr/bin/pygmentize:#!/usr/bin/python
- [ ] /usr/bin/svgtoipe:#!/usr/bin/env python

Gottox:
We can add a post_install hook to rewrite the shebangs.

ebfe:
usr/bin/avahi-bookmarks:#!/usr/bin/python
usr/bin/avahi-discover:#!/usr/bin/python
usr/bin/chardetect:#!/usr/bin/python
usr/bin/csscapture:#!/usr/bin/python
usr/bin/csscombine:#!/usr/bin/python
usr/bin/cssparse:#!/usr/bin/python
usr/bin/cvs2bzr:#!/usr/bin/python
usr/bin/cvs2git:#!/usr/bin/python
usr/bin/cvs2svn:#!/usr/bin/python
usr/bin/econnman-bin:#!/usr/bin/python
usr/bin/gajim:exec "/usr/bin/python" -OO $APP.py "$@"
usr/bin/gajim-history-manager:exec "/usr/bin/python" -OO $APP.py "$@"
usr/bin/gajim-remote:exec "/usr/bin/python" -OO $APP.py "$@"
usr/bin/grc:#! /usr/bin/python
[...]

PR #3883, opened 2016-03-21
hooks/post-install: add hook to rewrite unversioned python shebangs.

Gottox:
https://github.com/voidlinux/void-packages/issues/3879

### DIFF
--- a/common/hooks/post-install/07-rewrite-python-shebang.sh
+++ b/common/hooks/post-install/07-rewrite-python-shebang.sh
@@ -15,7 +15,7 @@ hook() {
 	: ${pyver:=2.7}
 	shebang="#!/usr/bin/python$pyver"
 	find ${PKGDESTDIR} -type f -print0 | \
-		xargs -0 grep -l -m 1 "^#!.*\([[:space:]]\|/\)python\([[:space:]]\|$\)" -- | while read f; do
+		xargs -0 grep -l -m 1 "^#!.*\([[:space:]]\|/\)python\([[:space:]]\|$\)" -- | while IFS= read -r f; do
 		if [ "$warn" ]; then
 			msg_warn "$pkgname: multiple python versions defined!"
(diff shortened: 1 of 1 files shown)


common/hooks/post-install/07-rewrite-python-shebang.sh
```diff
@@ -0,0 +1,27 @@
...
+	shebang="#!/usr/bin/python$pyver"
+	find ${PKGDESTDIR} -type f -print0 | \
+		xargs -0 grep -l -m 1 "^#!.*/python\([[:space:]]\|$\)" -- | while read f; do
+		if [ "$warn" ]; then
+			msg_warn "$pkgname: multiple python versions defined!"
+			msg_warn "$pkgname: using $pyver for shebang"
+			warn=
+		fi
+		echo "   Unversioned shebang replaced by '$shebang': ${f#$PKGDESTDIR}"
+		sed -i "1s@.*@${shebang}@" -- "$f"
```
Vaelatern: Possibly should make sure to depend on the correct version of python to provide the python in the versioned shebang.

Gottox: Not that easy as it sounds. Python packages don't need to depend on pythonX.X directly, instead they're indirect dependants of python through other python packages.

common/hooks/post-install/07-rewrite-python-shebang.sh
```diff
@@ -0,0 +1,27 @@
...
+			warn=1
+			break;
+		fi
+		pyver=$i
+	done
+
+	: ${pyver:=2.7}
+	shebang="#!/usr/bin/python$pyver"
+	find ${PKGDESTDIR} -type f -print0 | \
+		xargs -0 grep -l -m 1 "^#!.*\([[:space:]]\|/\)python\([[:space:]]\|$\)" -- | while read f; do
```
chneukirchen: never use `read` without `-r` and IFS=

PR #3893, opened 2016-03-23
New package: swift-2.2

radare:


Vaelatern:
Please attempt to patch the build script to allow for compilation on other systems.

srcpkgs/swift/template
```diff
@@ -0,0 +1,75 @@
+# Template file for 'swift'
+pkgname=swift
+version=2.2
+revision=1
+makedepends="libxml2-devel libuuid-devel libbsd-devel icu-devel libedit-devel sqlite-devel"
+only_for_archs="i686 x86_64 x86_64-musl"
```
Vaelatern: Why no arm?

srcpkgs/swift/template
```diff
@@ -0,0 +1,75 @@
...
+_arch="${XBPS_TARGET_MACHINE}"
+hostmakedepends="cmake pkg-config ninja cmake icu swig clang python perl"
```
Vaelatern: You list cmake twice.

srcpkgs/swift/template
```diff
@@ -0,0 +1,75 @@
...
+	vlicense LICENSE.txt
+	cd ../build/Ninja-ReleaseAssert
+	cd swift-linux-${_arch}
+	vmkdir usr/bin
+	vinstall bin/sil-opt 755 /usr/bin
+	vinstall bin/sil-extract 755 /usr/bin
+	vinstall bin/swift 755 /usr/bin
+	vinstall bin/swiftc 755 /usr/bin
+	vinstall bin/swift-demangle 755 /usr/bin
+	vinstall bin/swift-autolink-extract 755 /usr/bin
```
xtraeme: You can `vbin path` instead.

radare:
updated

srcpkgs/swift/template
```diff
@@ -0,0 +1,72 @@
+# Template file for 'swift'
+pkgname=swift
+version=2.2
+revision=1
+makedepends="libxml2-devel libuuid-devel libbsd-devel icu-devel libedit-devel sqlite-devel"
+only_for_archs="i686 x86_64 armv7l aarch64 mips mipsel"
```
xtraeme: `only_for_archs="i686 x86_64"`

at least until we figure out how to cross compile this beast.

jannisch:
I think we should create subpackages for LLDB (the Swift programming language debugger and REPL) and Foundation (core library).

srcpkgs/swift/template
```diff
@@ -0,0 +1,73 @@
+# Template file for 'swift'
+pkgname=swift
+version=2.2
+revision=1
+nocross=yes
+makedepends="libxml2-devel libuuid-devel libbsd-devel icu-devel libedit-devel sqlite-devel"
+only_for_archs="i686 x86_64 armv7l aarch64 mips mipsel"
+hostmakedepends="cmake pkg-config ninja cmake icu swig clang python perl"
```
Vaelatern: cmake still listed twice.

srcpkgs/swift/template
Vaelatern: Not just `nocross=yes`?

Vaelatern:
The build script itself seems limited. I'll work a bit on running cmake directly, since the build script does not have us in mind at all.

radare:
the repl is basically an lldb script

On 25 March 2016 at 02:57, Jannis Christ <notifications@github.com> wrote:

> I think we should create subpackages for LLDB (the Swift programming
> language debugger and REPL) and Foundation (core library).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/voidlinux/void-packages/pull/3893#issuecomment-201106844>
>



-- 
--pancake

srcpkgs/swift/template
radare: it's using ninja too

On 25 March 2016 at 04:21, Toyam Cox <notifications@github.com> wrote:

> In srcpkgs/swift/template
> <https://github.com/voidlinux/void-packages/pull/3893#discussion_r57415888>
> :
>
> > @@ -0,0 +1,73 @@
> > +# Template file for 'swift'
> > +pkgname=swift
> > +version=2.2
> > +revision=1
> > +nocross=yes
> > +makedepends="libxml2-devel libuuid-devel libbsd-devel icu-devel libedit-devel sqlite-devel"
> > +only_for_archs="i686 x86_64 armv7l aarch64 mips mipsel"
> > +hostmakedepends="cmake pkg-config ninja cmake icu swig clang python perl"
>
> cmake still listed twice.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/voidlinux/void-packages/pull/3893/files/4e832e930397e3d316a2e3260f443391977fc055#r57415888>
>



-- 
--pancake

Vaelatern: Why does that mean `cmake` must be listed twice?

radare:
Removed the dupped cmake, squashed commits. it is ready to merge now?

srcpkgs/swift/template
```diff
@@ -0,0 +1,73 @@
...
+		done
+		ln -sf swift-swift-${version}-RELEASE swift
+	)
+	export LDFLAGS="-ldl -lpthread"
+	utils/build-script -R -t -j${XBPS_MAKEJOBS}
+}
+
+do_install() {
+	vlicense LICENSE.txt
+	cd ../build/Ninja-ReleaseAssert/swift-linux-${XBPS_TARGET_ARCH}
```
jannisch: `XBPS_TARGET_ARCH` seems to empty which caused the fail of [the last Travis build](https://travis-ci.org/voidlinux/void-packages/jobs/118805151).
Maybe you could replace the line with `cd ../build/Ninja-ReleaseAssert/swift-linux-*`

radare:
updated

radare:
Not sure if this works at all on Ubuntu or Archlinux, but a simple `import Glibc` fails if trying to be compiled with swiftc because of missing include files

radare:
We should probably split this pkg in `swift-runtime` because compiled swift programs depend on libswiftCore.so which is owned by this pkg and installed in `/usr/lib/swift/linux/x86_64/libswiftCore.so` we should probably also include all the *.so files from `/usr/lib/swift/linux/x86_64/*` in that swift-runtime pkg. what do you think?

jannisch:
Oh yes, Swift definitely expects headers in the wrong directory, for example `/usr/include/stddef.h` instead of `/usr/include/linux/stddef.h`. A very similar problem is fixed in the AUR buildscript [here](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=swift-language#n64).

I agree that there should be different packages for only runtime and development tools. I think that `swift-runtime` isn't a good name however, maybe just keep `swift`.
By the way the libswiftCore.so in my older swift build is under `/usr/lib/swift/linux/libswiftCore.so`, did you change that?
The developement tools (ǹamed `swift-lang` or `swiftc` for example) should also depend on the compiler stuff like llvm. They might also depend on `swift-lldb` which we don't have yet, I know the REPL is already included in swift.

radare:
oh nevermind, dunno why i put x86_64 in that directory. its ok

radare:
can you send a PR for that include path issue?

radare:
This path is specified in swift-llvm/lib/Driver/ToolChains.cpp there’s no need to find all directories to do that sed :P

radare:
sorry swift-clang. im building right now with this patch to see if it works and i’ll push soon

radare:
i have pushed , but cant confirm if it works because i think im installing the same pkg all the time

Vaelatern:
@radare Tip: `./xbps-src -f pkg <name>` will force replace the old package you built if there is one.

radare:
Doesnt seems to fix the issue, there are still many problems with missing basic include files. strace -f -o shows nothing wrong.. so im open to ideas

Vaelatern:
Maybe there should be more dependencies?

radare:
no. see other distro packages, and its all about system includes

jannisch:
Just discovered some broken or strange stuff in my swift installation (not the newest build though):
-  `/usr/lib/swift/clang -> /builddir/build/Ninja-ReleaseAssert/llvm-linux-x86_64/lib/clang/3.8.0`
- `/usr/lib/swift/linux/` includes both all the libswift* files AND an x86_64 subdirectory with the same libswift* files (I hashed them) in addidtion to other files. Might not cause any problems however.
- `/usr/lib/swift/install-tmp/install-tmp/clang -> ../clang/3.8.0`, there is no such file or directory (no link as well)

srcpkgs/swift/template
```diff
@@ -0,0 +1,74 @@
...
+	cd ../build/Ninja-ReleaseAssert/swift-linux-${XBPS_TARGET_MACHINE}
+	vbin bin/sil-opt
+	vbin bin/sil-extract
+	vbin bin/swift
+	vbin bin/swiftc
+	vbin bin/swift-demangle
+	vbin bin/swift-autolink-extract
+	vman docs/tools/swift.1
+	vmkdir usr/lib/swift
+	cp -rf lib/swift/* ${PKGDESTDIR}/usr/lib/swift
```
jannisch: better use [vcopy()](https://github.com/voidlinux/void-packages/blob/master/Manual.md#global-functions) here

srcpkgs/swift/template
```diff
@@ -0,0 +1,74 @@
+# Template file for 'swift'
+pkgname=swift
+version=2.2
+revision=1
+nocross=yes
+makedepends="libxml2-devel libuuid-devel libbsd-devel icu-devel libedit-devel"
+only_for_archs="i686 x86_64 armv7l aarch64 mips mipsel"
```
jannisch: i686 is currently broken, should we really keep it here amongst the other ones?

Vaelatern: An implementation detail. Not important yet.

radare: we can probably stick to x86_64 for now. swift is a moving target and i expect much more changes and enhancements. for example, for swift3, there will be a stable pkg manager and there will be some important syntax changes in the language.

It will be great if official swift gets a `make install` at some point. or at least some more suppotr for other OSs and archs (official support), because right now is possible to use swift on freebsd, arm, android, powerpc, etc...

Vaelatern: Their build script is just, so far as I can see, a wrapper for a cmake call.

jannisch:
I wonder if this is the glibc problem is actually caused by the swift package. Here is the error message by swiftc:
``` <module-includes>:7:10: note: in file included from <module-includes>:7:
#include "/usr/include/locale.h"
         ^
/usr/include/locale.h:28:10: error: 'stddef.h' file not found
#include <stddef.h>
         ^
<unknown>:0: error: could not build Objective-C module 'SwiftGlibc'

File: srcpkgs/swift/template (hunk 8 above)
Comment (Vaelatern) [inline srcpkgs/swift/template]: If you follow it, you can figure out what cmake call it eventually makes. Have not gotten that far yet, so many other things to do, but if you figure what cmake call they end up making, cross building might become possible and maybe even some more archs than just `x86_64`

**Comment by jannisch** [comment]
I wonder if this is the glibc problem is actually caused by the swift package. Here is the error message by swiftc:
```
<module-includes>:7:10: note: in file included from <module-includes>:7:
#include "/usr/include/locale.h"
         ^
/usr/include/locale.h:28:10: error: 'stddef.h' file not found
#include <stddef.h>
         ^
<unknown>:0: error: could not build Objective-C module 'SwiftGlibc'
```
The imported `/usr/include/locale.h` belongs to package glibc-devel. The import statement in **this** file is wrong in fact. There is no such `/usr/include/stddef.h`, only a `/usr/include/linux/stddef.h`

**Comment by Vaelatern** [comment]
So patch it.

**Comment by jannisch** [comment]
I can't really believe this because so many packages depend on glibc-devel but I try...

**Comment by Vaelatern** [comment]
Many packages, meaning
```
srcpkgs/glibc/template
srcpkgs/base-devel/template
srcpkgs/llvm/template
srcpkgs/gcc-multilib/template
srcpkgs/base-chroot/template
srcpkgs/gcc/template
srcpkgs/dkms/template
srcpkgs/cross-vpkg-dummy/template
```

**Comment by radare** [comment]
FreeBSD patchset is here http://git.snooey.net/Mirrors/swift/commit/16de5d1fc498a523dee2bb9f5f751d641270ee14

And this seems related: https://github.com/dcci/swift/blob/8f301b6554563990c9dfdc365473ebbb48fad4d6/stdlib/public/Glibc/module.freebsd.map.in

**Comment by radare** [comment]
Another related patch:     `sed -i 's/\/x86_64-linux-gnu//g' "${pkgdir}/usr/lib/swift/glibc/module.map"`

**Comment by radare** [comment]
There's no such thing as include/stddef.h in any file installed in the swift pkg... maybe my system is broken. can anyone confirm the issue?

**Comment by Vaelatern** [comment]
Your system is just fine.

File: srcpkgs/swift/template (hunk 8 above)
Comment (radare) [inline srcpkgs/swift/template]: Yep already changed that in the lastcommit

**Comment by radare** [comment]
Maybe /linux is also in the includepath ?

**Comment by radare** [comment]
Maybe related to GCC vs CLANG?

**Comment by radare** [comment]
Just filled https://bugs.swift.org/browse/SR-1090 because i'm also missing the manpages

**Comment by radare** [comment]
Ive been playing with the module.map file and its weird that the paths are always absolute. And i can override the -Xcc -I but including linux/ produces more errors related to gnuc_va_* which makes me think that maybe debian is patching something there. I will try to test this on archlinux to see if their pkg is good

**Comment by radare** [comment]
Just installed the pkg for debian from the repository and it is not even working :P

**Comment by jannisch** [comment]
Ok but `glibc-devel` is still important if only gcc, clang and so forth depend on it I guess.

At least I had some success linking the wrong `stddef.h` and `stderr.h` from `glibc-devel` into the right directory because now I'm stuck at a missing `stdarg.h`.
According to [this](http://stackoverflow.com/questions/6712493/where-is-stdarg-h) StackOverflow thread it's very compiler specific so I looked into my packages repo. Swift offers these two:
```
builddir/swift-clang-swift-2.2-RELEASE/lib/Headers/stdarg.h
builddir/build/Ninja-ReleaseAssert/llvm-linux-x86_64/lib/clang/3.8.0/include/stdarg.h
```
and inside the masterdir:
```
masterdir/usr/lib/gcc/x86_64-unknown-linux-gnu/4.9/include/stdarg.h
masterdir/usr/include/c++/4.9/tr1/stdarg.h
```

We probably need a `swift-clang` and a `swift-llvm` package but you guys know more about compilers and linkers in general, what do you think?

At least we can check out the Ubuntu build :smile:

**Comment by radare** [comment]
I think i'm gonna squash all those commits into a single one and for me it's ready to be merged. What do you think?

**Comment by radare** [comment]
Removed commented blocks

**Comment by jannisch** [comment]
Looks great now. A `swift-packagemanager` release exists in fact, it's [0.1.0](https://github.com/apple/swift-package-manager/releases/tag/0.1.0) from end of last year.
We should also add a `swift-lldb` subpackage later on.

**Comment by radare** [comment]
yep. swift-build is not available either, and im not sure if this should be in packagemanager or in core swift..

**Comment by radare** [comment]
can we get this pkg merged? the `swift-build` binary seems to be generated by this swift-package-manager repo. we can just add it as a dependency when installing swift.

**Comment by jannisch** [comment]
I think we should discuss additional swift packages in a separate issue now.
`/usr/bin/swift`, `/usr/bin/swiftc` and `/usr/bin/swift-autolink-extract` are identical files, so the latter two should only be symbolic links, can you fix this?

**Comment by radare** [comment]
done

**Comment by jannisch** [comment]
Did you already test it? I doubt that this symlink will be included in package or only exists in the builddir.
For some reason I now have to compile everything again before crafting the package.

**Comment by radare** [comment]
will test tonight, tested without building the pkg

**Comment by jannisch** [comment]
It doesn't work yet, the symbolic links are not included in the final package.
But I can't tell you how to fix it, it doesn't work in the `post_install()` function either.

File: srcpkgs/swift/template [hunk 9]
```diff
@@ -0,0 +1,62 @@
...
+}
+
+do_install() {
+	vlicense LICENSE.txt
+	cd ../build/Ninja-ReleaseAssert/swift-linux-${XBPS_TARGET_MACHINE}
+	vbin bin/sil-opt
+	vbin bin/sil-extract
+	vbin bin/swift
+	ln -fs bin/swift bin/swiftc
+	ln -fs bin/swift bin/swift-autolink-extract
```
xtraeme: those symlinks are not being copied to pkg DESTDIR.

radare: the pkg is no longer working for me because of the libressl changes i guess that i should rebase the branch from master again. Could you send a PR?

radare:
Please @xtraeme can you take a look on this and suggest how to fix that symlink issue? i cant manage to get the symlinks included in the final pkg :S

srcpkgs/swift/template
```diff
@@ -0,0 +1,62 @@
...
+
+do_install() {
+	vlicense LICENSE.txt
+	cd ../build/Ninja-ReleaseAssert/swift-linux-${XBPS_TARGET_MACHINE}
+	vbin bin/sil-opt
+	vbin bin/sil-extract
+	vbin bin/swift-demangle
+	vbin bin/swift
+	ln -fs swift ${DESTDIR}/usr/bin/swiftc
+	ln -fs swift ${DESTDIR}/usr/bin/swift-autolink-extract
```
Vaelatern: These two symlinks are the issue @radare?

Vaelatern: Maybe you want to `cd bin` before and `cd ..` after making the symlinks.

radare:
Confirmed. symlinks work as expected. i was just testing the same pkg all the time.

srcpkgs/swift/template
radare: that was my first approach, but didnt worked..probably because i was using an old pkg all the time...

this is how its done in other packages..

radare:
ping

radare:
\o/

PR #3900, opened 2016-03-25
[NOMERGE] Libressl

Vaelatern:
Patches to fix xlint are welcome, but not considering xlint blocking. Testers of various packages with the latest libressl are welcome. I've successfully built opensmtp, but have failed with john, not sure if it's related to libressl.
Limiting the Travis builds to 3 at a time so travis isn't blocked from sourceforge, which has caused test travis builds to fail today.

### DIFF
--- a/srcpkgs/znc/template
+++ b/srcpkgs/znc/template
@@ -1,7 +1,7 @@
 # Template file for 'znc'
 pkgname=znc
 version=1.6.3
-revision=1
+revision=2
 build_style=gnu-configure
 configure_args="--enable-python --enable-perl --enable-tcl --enable-cyrus"
 hostmakedepends="pkg-config perl"


Vaelatern:
John, as it turns out, wants `SHA_{Init,Update,Final}` to be present, and I think it wants them to be in OpenSSL.

Vaelatern:
Most pressingly, I commented out some logic in the libressl template. Should we have that logic stay, even though it seems to not be a concern of ours?

Vaelatern:
Confirmed. Jack depends on code removed in the latest libressl. Should it just be marked broken or should we patch the functionality out of jack (or find a different SHA_Init source?)

chneukirchen:
You mean John, right?

Looks like SHA0 got dropped from LibreSSL. If you can disable it in John, just do that, its not a very important hash I think.

srcpkgs/john/template
```diff
@@ -16,6 +17,11 @@ homepage="http://www.openwall.com/john/"
 distfiles="$homepage/j/$pkgname-${version}-jumbo-${_jumbover}.tar.xz"
 checksum=bac93d025995a051f055adbd7ce2f1975676cac6c74a6c7a3ee4cfdd9c160923
 
+post_extract() {
+	# remove SHA0 support, as libressl no longer has it
+	rm src/rawSHA0_fmt_plug.c
+}
+
```
Vaelatern: @chneukirchen This fixes john. I meant to say "john" when I said "jack" in that last comment, yes. I tried using other `sha.h` files but none worked out of the box.

srcpkgs/libressl/template
```diff
@@ -9,18 +9,18 @@ maintainer="Juan RP <xtraeme@voidlinux.eu>"
...
-	configure_args+=" --disable-hardening"
-elif [ "$XBPS_TARGET_MACHINE" = "armv5tel" ]; then
-	configure_args+=" --disable-asm"
-fi
+#if [ "$XBPS_TARGET_MACHINE" = "i686-musl" ]; then
+#	# XXX disable SSP
+#	configure_args+=" --disable-hardening"
+#elif [ "$XBPS_TARGET_MACHINE" = "armv5tel" ]; then
+#	configure_args+=" --disable-asm"
+#fi
```
Vaelatern: This comment out, is it okay? @xtraeme, also considering just complete removal. The logic seems to not be pertinent to Void at this time, and we seem to build just fine on all supported arches.

srcpkgs/libssh/template
```diff
@@ -1,9 +1,9 @@
 # Template file for 'libssh'
 pkgname=libssh
 version=0.7.3
-revision=1
+revision=2
 build_style=cmake
-hostmakedepends="cmake pkg-config"
+hostmakedepends="pkg-config"
```
Vaelatern: I also removed `cmake` from `hostmakedepends` wherever `build_style=cmake` in this set of revbumps.

srcpkgs/nix/template
```diff
@@ -1,20 +1,20 @@
...
-# Extra run-time dependencies.
+# Extra build-time dependencies.
 makedepends="bzip2-devel libressl-devel sqlite-devel gc-devel libcurl-devel"
 # Default configuration file.
 conf_files="/etc/nix/nix.conf"
 # Create required build users/groups.
 system_groups="nixbld"
 system_accounts=
-for f in $(seq 1 9); do
+for f in {1..9}; do
```
Vaelatern: One less function call.

srcpkgs/xchat/template
```diff
@@ -20,9 +20,9 @@ pre_configure() {
 	local perlflags=
 	# Fix perl library path
 	if [ "$CROSS_BUILD" ];then
-		perlflags=`perl -MExtUtils::Embed -e ldopts | sed \
+		perlflags=$(perl -MExtUtils::Embed -e ldopts | sed \
```
Vaelatern: Fixes an xlint issue. Seems to have no negative impact.

Vaelatern:
Currently waiting primarily on @Gottox to confirm the build servers can handle the build without breaking the ecosystem.

Gottox:
DO NOT MERGE. Either we wait till xbps-0.52 with the staging repository or do a progressive update on these packages. I would prefer the latter as it won't break the whole distribution if something blows up.

Vaelatern:
@xtraeme As resolved in IRC, we can safely do this (after more git rebasing). Can I get a sign-off from you for these changes? I've commented on all the parts that are not just revbumps.

srcpkgs/libressl/template
xtraeme: @Vaelatern probably those still are valid. These archs are not used by default in the buildbot builders but there are some users.

xtraeme:
Good to go, thanks :+1:

srcpkgs/libressl/template
Vaelatern: So keep them commented out, or undo the comments and re-push?

PR #3903, opened 2016-03-26
New package: font-tewi-2.0.0

asiekierka:


Vaelatern:
Upstream calls it `tewi-font`, you will save yourself some pain if you just name the whole package `tewi-font`.

asiekierka:
I feel that it's better to adhere to a standard naming, to make it easier to find the package.

srcpkgs/font-tewi/template
```diff
@@ -0,0 +1,26 @@
+# Template file for 'font-tewi'
+pkgname=font-tewi
+version=2.0.0
+revision=1
+wrksrc=tewi-font-${version}
+noarch="yes"
+makedepends="font-util bdftopcf python"
```
Vaelatern: This variable should be: `hostmakedepends`

Vaelatern:
I present you with `google-fonts-ttf` `ohsnap-font` `termsyn-font` `terminus-font` and xorg-fonts`. Just to name the ones that fit on my page. Yes, many are `font-such-and-such`, but they are (for the ones I've checked) named that way upstream.

asiekierka:
I am aware, but the majority of the fonts begin with font- after all. I can change it if it's that big a problem...

Vaelatern:
The fonts that are named that way, as far as I can tell, are either not named  with `font` in the name by upstream, or are named in exactly that way by upstream. Generally I think we go with what upstream sets. `xbps-query -Rs` can find it for us. The big deal is the `hostmakedepends` thing though.

asiekierka:
Okay then, let's try again...

PR #3905, opened 2016-03-26
catalyst: linux-4.4 support

lemmi:
frankly, i just copied patches i could find. works for me on 4.4 and 4.2. use with caution :)


Gottox:
Can s/o verify, that it's working and not breaking older versions?

PR #3916, opened 2016-03-27
Python3.5 / fix blender

Gottox:
Just enough of python3.5 to make blender build.


asergi:
If you're interested, I made a working python3.5 template (since 15/09/2015 to be exact).
I already addressed cross-compiling and all the conflicts in order to allow multple python3.x versions.

As a side note, I made PR #3470 to generate multi-versioned python pkgs on-the-fly, but I didn't get a review and nobody seems to care much.

Gottox:
@asergi Ah, cool. I'll have a look.

PR #3922, opened 2016-03-28
monkey: update to 1.6.8.

Vaelatern:
File list for simpler review: https://gist.github.com/anonymous/75ac2c51ab2797ccaecd

### DIFF
--- a/srcpkgs/monkey/template
+++ b/srcpkgs/monkey/template
@@ -1,45 +1,40 @@
 # Template file for 'monkey'
 pkgname=monkey
-version=1.5.6
-revision=4
+version=1.6.8
+revision=1
 build_style=configure
+hostmakedepends="cmake"
 makedepends="mbedtls-devel"
-configure_args+="--prefix=/usr --plugdir=/usr/share/monkey/plugins "
+configure_args+="--prefix=/usr --sbindir=/usr/bin --libdir=/usr/lib "
 configure_args+="--sysconfdir=/etc/monkey/ --enable-plugins=mbedtls "
-configure_args+="--datadir=/srv/httpd --mandir=/usr/share/man "
+configure_args+="--webroot=/srv/httpd --mandir=/usr/share/man "
 configure_args+="--pidfile=/var/run/monkey.pid --logdir=/var/log/monkey "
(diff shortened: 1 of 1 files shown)


Vaelatern:
Note there is a `-devel` package without any `common/shlibs` entries, or even any versioned shlibs.

Gottox:
> Note there is a -devel package without any common/shlibs entries, or even any versioned shlibs.

That's okay. We currently have no packages that link against monkey.

ebfe:
Why did you remove the pkg-config file?

srcpkgs/monkey/template
```diff
@@ -1,45 +1,40 @@
...
- /etc/monkey/plugins/auth/monkey.users"
+ /etc/monkey/plugins/*/*.conf
+ /etc/monkey/plugins/dirlisting/themes/*/*.theme
+ /etc/monkey/sites/default
+ /etc/monkey/monkey.conf
+ /etc/monkey/plugins.load
+ /etc/monkey/monkey.mime"
 
 system_accounts="httpd"
 httpd_descr="Monkey HTTP server"
```
Gottox: I think we should remove the httpd description from all httpds. Currently the description of the user depends on the first httpd that was installed.

Vaelatern:
@ebfe Please note that it isn't anywhere in the file list. I don't think it exists in this version.

srcpkgs/monkey/template
Vaelatern: Should I add that to this PR?

ebfe:
@Vaelatern there's monkey.pc.in, so maybe an upstream bug that it doesn't get installed?

ebfe:
I think the .so shouldn't be in the `-devel` package. These are plugins to be loaded by the webserver. If it has to be a separate package monkey-plugins would make more sense.

Vaelatern:
Doesn't need to be a separate package at all. I'll poke at the pkg config file.

srcpkgs/monkey/template
Gottox: No. just wanted to mention it somewhere :)

Gottox: we can push these changes this directly

PR #3923, opened 2016-03-28
New package: visitors-0.7

asiekierka:


srcpkgs/visitors/template
```diff
@@ -0,0 +1,23 @@
+# Template file for 'visitors'
+pkgname=visitors
+version=0.7
+revision=1
+wrksrc="${pkgname}_${version}"
+short_desc="Lightweight web log analyzer"
+maintainer="Adrian Siekierka <asiekierka@gmail.com>"
+license="GPL-2"
+homepage="http://www.hping.org/visitors/"
+distfiles="http://www.hping.org/visitors/visitors-0.7.tar.gz"
```
Vaelatern: Use `${version}`

asiekierka: Forgot! Sorry.

srcpkgs/visitors/template
```diff
@@ -0,0 +1,23 @@
+# Template file for 'visitors'
+pkgname=visitors
+version=0.7
+revision=1
+wrksrc="${pkgname}_${version}"
+short_desc="Lightweight web log analyzer"
+maintainer="Adrian Siekierka <asiekierka@gmail.com>"
+license="GPL-2"
+homepage="http://www.hping.org/${pkgname}/"
```
Vaelatern: No `${pkgname}` in this please :)

srcpkgs/visitors/template
```diff
@@ -0,0 +1,20 @@
+# Template file for 'visitors'
+pkgname=visitors
+version=0.7
+revision=1
+wrksrc="${pkgname}_${version}"
+short_desc="Lightweight web log analyzer"
+maintainer="Adrian Siekierka <asiekierka@gmail.com>"
+license="GPL-2"
```
Vaelatern: Not adequate in this case. There is custom copyright if you read the COPYING file you also need to account for.

asiekierka: How is it not adequate? The copyright notice is MIT, while the compiled binary (what's being packaged) is GPLv2 - the complete product as a whole is then also GPLv2.

Vaelatern:
You don't want the `iis2apache.pl` file? What about the .png files?

asiekierka:
@Vaelatern We'd have to rename it to something like visitors-iis2apache to prevent potential conflicts with other software? As for the PNG files, I don't particularly see why? They're useful for the HTML mode, but that always uses relative paths so they're not useful in a packaged variant.

srcpkgs/visitors/template
Vaelatern: Then `license="GPL-2, MIT"` but the MIT needs a vlicense of it's own, so you may want to chop out the first X lines of the file and put that as a license.

Vaelatern:
Okay, no png files. There is no other `iis2apache` in the repos.

Vaelatern:
Nevermind about the `iis2apache` file, turns out if someone really needs IIS stuff, they know enough to find it.
Currently waiting on you setting accurate license information, and vlicensing the BSD license from the C file.

PR #3925, opened 2016-03-28
rakudo-star: updated to 2016.03

xaduha:

PR #3930, opened 2016-03-29
Rakudo

Vaelatern:
Fewer pieces of magic. `panda` and `rakudo-star` are not yet correct or building, can you help @xaduha ?

Makes the mess of changes in #3925 not necessary, and fixes the templates to comply more with the rest of the package repo.

Also features perl 6 on more platforms than before.

Review very welcome.


xaduha:
I value practical things over doing everything perfectly, as long as it works it's fine with me. I'm happy with the way I did it, it builds, it works. I don't want to spend much time on it, but if anyone want to take over maintaining this package - I'll give it up.

Vaelatern:
If you can suggest changes to this version of the panda and rakudo-star templates, that would be great.

Vaelatern:
I mostly disagree that the changes introduced in 733e08f were correct or necessary.

xaduha:
I'd be glad not make these changes, but that was the only solution that worked properly. Do you actually use Rakudo Star or you're just talking as a mantainer of void-packages repo? The hope is that proper changes would be made upstream and things would work as they were before, but until that happens this is what I came up with.

Vaelatern:
What broke upstream that necessitated a wholly different installation method?

xaduha:
I have a vague theory, but I don't really care what broke it upstream. I'm not interested in trying it again until a new version comes out. A previous PR (2016.02) with those changes is already included in void-packages, my 2016.03 PR of rakudo-star is just an increment with 3 dependencies included only once where they actually needed, which is MoarVM. I'm not against improving it later.

Vaelatern:
Every single time you do `makedepends="${hostmakedepends}"` it is wrong by definition. Both hostmakedepends and makedepends are present on the system when the package is being built. One is there for the cross-compiled arch (and that brings in `-devel` packages) and the other is the programs that just need to be run, like `perl` or `git`.

xaduha:
If it's a big deal, then I'll change it, no problem. Anything else?

Vaelatern:
I think the first three sets of changes in this PR are correct, at the very least. Can you test that they work as you expect? I am willing to say panda and rakudo-star don't need the sweeping changes I made, but I believe the first three templates to be as necessary.

xaduha:
I've updated my PR.

xtraeme:
Already pushed

PR #3931, opened 2016-03-29
gpgme: change to gnupg2

Vaelatern:
Closes #3567 and removes extra variable.

### DIFF
--- a/srcpkgs/gpgme/template
+++ b/srcpkgs/gpgme/template
@@ -1,15 +1,14 @@
 # Template file for 'gpgme'
 pkgname=gpgme
 version=1.6.0
-revision=1
+revision=2
 build_style=gnu-configure
 configure_args="--enable-fd-passing --disable-gpgconf-test
  --disable-gpg-test --disable-gpgsm-test
  --with-libgpg-error-prefix=$XBPS_CROSS_BASE/usr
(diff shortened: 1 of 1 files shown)


xtraeme:
The same question I did in #3567 remains.

Vaelatern:
No, it is not necessary at build at all. That is why I remove it here.
So far I have determined either of `gnupg` or `gnupg2` satisfies basic runtime depends.

ebfe:
gpgme prefers gpg2 and checks at runtime which binaries are available. So the `depends=` change should be fine.

xtraeme:
Took a look at the source code (src/posix-util.c):

```
/* Return the full file name of the GPG binary.  This function is used
   if gpgconf was not found and thus it can be assumed that gpg2 is
   not installed.  This function is only called by get_gpgconf_item
   and may not be called concurrently.  */
```

So that even if it depends on gpg2, it will use gpg it gpgconf is in PATH.

ebfe:
On Tue, Mar 29, 2016 at 06:30:06AM -0700, Juan RP wrote:
> So that even if it depends on gpg2, it will use gpg it gpgconf is in PATH.

I think you read that wrong. gpgconf is only in gpg2.

xtraeme:
Hmm you are right. In this case it will prefer gpg2.

Thanks for the clarification.

This PR can be merged :+1:

PR #3938, opened 2016-03-29
fontconfig: update to 2.11.94

renatoaguiar:


ebfe:
I think .9x versions are prelease/beta.

renatoaguiar:
@ebfe You are right. I just checked in the oficial website. Thanks.

PR #3945, opened 2016-03-31
mopidy: update to 2.0.0

allan:


Duncaen:
afaik mopidy 2.0.0 switched to gstreamer1, did you test it?

allan:
It probably worked because `gstreamer1` got pulled in by `gst-plugins-good1` et al.

The manpage one has to generate with Sphinx is real minimal and kind of useless.  To provide documentation inside the package we could move all the `doc/*.rst` files to /usr/share/doc or generate the HTML/man.  Or keep the hint to the online docs in the INSTALL.msg.  Dunno about that.  What do you think?

Vaelatern:
Let's do all the documentation options if it helps people.

Vaelatern:
Any way we can stick with just one version of gstreamer with this?

PR #3953, opened 2016-04-01
intellij-idea-community-edition: update to 2016.1.1

Grauwolf:


ebfe:
Fails to start with openjdk:

    $ idea
    OpenJDK 64-Bit Server VM warning: ignoring option MaxPermSize=350m; support was removed in 8.0
    
    JDK Required: 'tools.jar' seems to be not in IDEA classpath.
    Please ensure JAVA_HOME points to JDK rather than JRE.

Can be fixed by setting JAVA_HOME=/usr/lib/openjdk.

Vaelatern:
Should we add an entry, within this package, in `/etc/profile.d/` that will change that setting?

TheDoctorsLife:
You should, It would really help. In my case I had to play around for an hour or so to get JAVA_HOME set correctly of IDEA.

ebfe:
I don't think /etc/profile.d is the right place for this. Either the current /etc/profile.d/10_openjdk.sh is wrong in setting JAVA_HOME to the jre subdir then that should be fixed (maybe @Gottox knows). Otherwise idea has the wrong expectations and then the idea launch script should be patched (or similar). But changing JAVA_HOME on wether you have this package installed or not is not a good idea.

Vaelatern:
Can we have a conclusion on this, @Gottox @ebfe ? I prefer an `INSTALL.msg` to tell the user they may want to set `JAVA_HOME`

Gottox:
The oracle-j* packages are wrong. Idealy there should be one /etc/profiles.d/10_java.sh and one /etc/profiles.d/09_jdk.sh provided through an alternative.

Gottox:
See https://github.com/voidlinux/void-packages/pull/3525

ebfe:
http://blog.jetbrains.com/blog/2016/05/11/security-update-for-intellij-based-ides-v2016-1-and-older-versions/

Vaelatern:
#4213

PR #3957, opened 2016-04-01
tigervnc: Integrate xorg source build so that Xvnc is built

evandeaubl:
My previous packaging of tigervnc didn't properly build the Xvnc program because it requires the xorg-server source to build. The build instructions closely mirror the documented instructions at tigervnc.org, plus a patch from the Arch Linux PKGBUILD to properly support xorg-server-1.18 source trees.

### DIFF
--- a/srcpkgs/tigervnc/template
+++ b/srcpkgs/tigervnc/template
@@ -1,17 +1,43 @@
 # Template file for 'tigervnc'
 pkgname=tigervnc
 version=1.6.0
-revision=1
+revision=2
+_xorg_version=1.18.2
 build_style=cmake
 short_desc="VNC client forked from TightVNC"
 maintainer="Evan Deaubl <evan@deaubl.name>"
 license="GPL-2"
-hostmakedepends="cmake"
-makedepends="fltk-devel zlib-devel libXtst-devel libjpeg-turbo-devel $(vopt_if gnutls gnutls-devel)"
+hostmakedepends="automake xorg-util-macros font-util pkg-config libtool xtrans"
+makedepends="fltk-devel zlib-devel libXtst-devel libjpeg-turbo-devel pixman-devel
+	randrproto fontsproto videoproto compositeproto scrnsaverproto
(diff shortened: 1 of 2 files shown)


srcpkgs/tigervnc/template
```diff
@@ -1,17 +1,46 @@
...
 homepage="http://www.tigervnc.org"
-distfiles="https://github.com/TigerVNC/tigervnc/archive/v${version}.tar.gz>tigervnc-${version}.tar.gz"
-checksum=98ffe98fcfe883e6c35aec579295b53d73d2ccf62e0f6e53a73ecad993b096ca
+distfiles="https://github.com/TigerVNC/tigervnc/archive/v${version}.tar.gz>tigervnc-${version}.tar.gz
+	${XORG_SITE}/xserver/xorg-server-${_xorg_version}.tar.bz2"
+checksum="98ffe98fcfe883e6c35aec579295b53d73d2ccf62e0f6e53a73ecad993b096ca
+	022142b07f6477d140dcc915902df326408a53ca3a352426a499f142b25d632d"
 
 build_options="gnutls"
 build_options_default="gnutls"
```
Vaelatern: Would it use libressl if we didn't have this option by default?

evandeaubl: No, tigervnc only support GnuTLS. If the configure doesn't find it, then it turns off that support.

Vaelatern: @pullmoll Can you provide feedback as to why this is a build option at all?

pullmoll: Oops! Me? No, I am clueless. I think the original author's PR had it like this and I didn't ask if it makes any sense when I helped him with #2417. I see no good reason why anyone would want to disable gnutls, though. It's available for all supported platforms AFAIK.

evandeaubl: I'm okay with un-optioning gnutls. Blame it on my previous Gentoo heritage, where anything and everything is an option. :-)

PR #3959, opened 2016-04-02
youtube-dl: add python(3.4)-youtube-dl subpackage

### LINKED ISSUE #2722: pafy/youtube-dl: unable to find moduel youtube_dl
Pafy decided to drop their stream code in favor of youtube_dl.  Well, it now requires the youtube_dl module to build so I added it to hostmakedepends.

Pafy can't seem to find it during build, and I'm a bit at a loss of what to do.

ebfe:
Fixes #2722


srcpkgs/youtube-dl/template
```diff
@@ -21,3 +22,24 @@ post_install() {
 	vinstall ${pkgname}.zsh 644 usr/share/zsh/site-functions _${pkgname}
 	vinstall ${pkgname}.fish 644 usr/share/fish/completions ${pkgname}
 }
+
+python-youtube-dl_package() {
```
Vaelatern: Maybe these packages should depend on python

ebfe: > Maybe these packages should depend on python
They did at first, but i checked a few other python- packages and most
don't. I don't think it's necessary if they don't contain runnable
binaries and doesn't matter in practice at all.

Vaelatern:
:+1:

ebfe:
e97df4b8fe4f765362fd9b06d79e3306a0b2e40e

PR #3968, opened 2016-04-03
New package: libtcd-2.2.7

dicktyr:


Vaelatern:
A change to common/shlibs for this is probably in order. The whoe thing is essentially a `-devel` package. File list [here](https://gist.github.com/17377492232782168b78db279fc53c6e).

dicktyr:
sorry, I'm not sure what to change

I'm trying to package xtide which uses libtcd
but libtcd is also used by other separate packages

Vaelatern:
The normal thing to do here is put a set of the files normally installed in the main package into a `-devel` subpackage. Examples are [libbidstab](https://gist.github.com/2fa6bb4c9bb88faa21db6655940a5316), [libfastjson,](https://gist.github.com/4c88f033fecb08658b1cce649eed3a93) and [a larger example, trousers](https://gist.github.com/498b26f81eb6757d8a107888d6805570), with example templates, in order, [libvidstab](https://gist.github.com/87f5e6f1dcca22c5e6398609b48ca9b6), [libfastjson](https://gist.github.com/215e857fb1ef9c7e57ffeae69fb439f4), and [trousers](https://gist.github.com/717973460c03cfbc5d2f640f9d4a3e48)

srcpkgs/libtcd/template
```diff
@@ -0,0 +1,13 @@
+# Template file for 'libtcd'
+pkgname=libtcd
+version=2.2.7
+revision=1
+_release="-r2"
+noarch=yes
```
Vaelatern: Wrong.

Vaelatern:
Please read the material at [the Manual](https://github.com/voidlinux/void-packages/blob/master/Manual.md#deps_runtime) for doing `common/shlibs` entries, as you need in this case.

Vaelatern:
This also needs a `-devel` subpackage.
```
libtcd-devel_package() {
	pkg_install() {
		vmove usr/include
		vmove usr/lib/*.a
		vmove usr/lib/*.so
	}
}
```

PR #3969, opened 2016-04-04
New package: fontmanager-0.7.2

antonio-malcolm:
Shiny new package...

### DIFF
--- a/srcpkgs/fontmanager/template
+++ b/srcpkgs/fontmanager/template
@@ -1,0 +1,20 @@
+# Template file for 'fontmanager'
+pkgname=fontmanager
+version=0.7.2
+revision=1
+wrksrc="master-${version}"
+build_style=gnu-configure
+disable_parallel_build=yes
+# glib-devel needs to be present in hostmakedepends,
+# gucharmap-devel and libgee08-devel need to be present in both hostmakedepends and makedepends,
+# in order to build correctly across all supported architectures
+hostmakedepends="pkg-config gobject-introspection intltool yelp-tools vala
+ file-roller glib-devel gucharmap-devel libgee08-devel"
+makedepends="libgee08-devel gucharmap-devel json-glib-devel sqlite-devel gtk+3-devel cairo-devel
+ libxml2-devel freetype-devel fontconfig-devel"
+short_desc="Simple font management for GTK+ desktop environments"
+maintainer="Antonio Malcolm <antonio@antoniomalcolm.com>"
+license="GPL-3"
+homepage="http://fontmanager.github.io/"
+distfiles="https://github.com/FontManager/master/archive/${version}.tar.gz"
+checksum=557e6f6130ff498d791c7a86f7c2b0f954f65fd680d95b2477582b90b4864295


srcpkgs/FontManager/template
```diff
@@ -0,0 +1,16 @@
...
+build_style=gnu-configure
+hostmakedepends="pkg-config"
+makedepends="gobject-introspection vala-devel glib-devel json-glib-devel
+	libgee08-devel gucharmap-devel sqlite-devel gtk+3-devel cairo-devel
+libxml2-devel freetype-devel fontconfig-devel intltool yelp-tools file-roller"
+short_desc="Simple font management for GTK+ desktop environments"
+maintainer="Antonio Malcolm <antonio@antoniomalcolm.com>"
+license="GPL-3"
+homepage="http://fontmanager.github.io/"
+distfiles="https://github.com/${pkgname}/master/archive/${version}.tar.gz"
```
Vaelatern: This looks like you didn't test it.

srcpkgs/FontManager/template
```diff
@@ -0,0 +1,16 @@
+# Template file for 'FontManager'
+pkgname=FontManager
+version=0.7.2
+revision=1
+wrksrc="master-${version}"
+build_style=gnu-configure
+hostmakedepends="pkg-config"
+makedepends="gobject-introspection vala-devel glib-devel json-glib-devel
+	libgee08-devel gucharmap-devel sqlite-devel gtk+3-devel cairo-devel
+libxml2-devel freetype-devel fontconfig-devel intltool yelp-tools file-roller"
```
Vaelatern: Please keep a consistent indentation with these. Single space is the standard for this part.

antonio-malcolm: Yeah, not sure what happened, there, but that's fixed.
I also like teh clean indents.

antonio-malcolm:
BTW, in response to a comment which now has disappeared, I compiled and installed this, and it works. I wouldn't open a pull request, without testing locally.

antonio-malcolm:
Ok, this was passing, previously, is now failing the i686 test, and the only change made was to the indents. It still builds, installs, and runs locally, with no issues.

antonio-malcolm:
Can someone rerun the tests, to verify? Thanks.

pullmoll:
You can just `git commit --amend` and `git push -f` to see if it was a temporary hiccup of travis-ci.

Vaelatern:
Try setting `disable_parallel_build=yes` because the job ran when I restarted it, and this was the second time the same job has failed.

antonio-malcolm:
Ok, that seems to have fixed it. Thanks!

srcpkgs/FontManager/template
```diff
@@ -0,0 +1,17 @@
+# Template file for 'FontManager'
+pkgname=FontManager
+version=0.7.2
+revision=1
+wrksrc="master-${version}"
+build_style=gnu-configure
+disable_parallel_build=yes
+hostmakedepends="pkg-config"
+makedepends="gobject-introspection vala-devel glib-devel json-glib-devel
```
xtraeme: `gobject-introspection intltool yelp-tools` need to be in `hostmakedepends` instead.

antonio-malcolm:
Well, that broke things, again. Do I need to move vala-devel or file-roller?

Vaelatern:
Try removing `vala-devel` in favour of `vala` in `hostmakedepends`, if that fails put back `vala-devel` but keep `vala` in `hostmakedepends`

antonio-malcolm:
BTW, perhaps a silly question to some of _you_, but could someone better clarify the difference, between 'makedepends' and 'hostmakedepends'? Common sense would dictate that one is installed to the _host machine_, and the other installed and used for the _build, only_, however, the documentation only notes that both are used for the build, and both are installed installed to the _master directory_, which I understand, on the more immediate level, as I've looked over the heirarchy of void-packages, whilst building. The lack of explanation is in the distinction- this can be confusing, as there is _no such distinction, when building and installing a project from source (i.e., without XBPS)_.

dominikh:
hostmakedepends are installed for the host's architecture, as they
usually contain binaries that need to be executable (pkg-config, git,
...). makedepends, on the other hand, are installed for the target
architecture. These are usually libraries and headers.

The distinction only occurs when you're cross compiling, say on an
amd64 system for an ARM system. The tools in hostmakedepends will
execute during building, while makedepends will contain code that gets
linked into the binary for ARM.

On Sat, Apr 09, 2016 at 11:50:22AM -0700, Antonio Malcolm wrote:
> BTW, perhaps a silly question to some of _you_, but could someone better clarify the difference, between 'makedepends' and 'hostmakedepends'? Common sense would dictate that one is installed to the _host machine_, and the other installed and used for the _build, only_, however, the documentation only notes that both are used for the build, and both are installed installed to the _master directory_, which I understand, on the more immediate level, as I've looked over the heirarchy of void-packages, whilst building. The lack of explanation is in the distinction- this can be confusing, as there is _no such distinction, when building and installing a project from source (i.e., without XBPS)_.
> 
> ---
> You are receiving this because you are subscribed to this thread.
> https://github.com/voidlinux/void-packages/pull/3969#issuecomment-207831422

-- 
Dominik Honnef

antonio-malcolm:
Aho, ok, I get host vs target, and that when we compile outside of void-packages, we are compiling only for the architecture of our host machine, and target is assumed to be host- I suppose I lack the understanding of why, in each case of a package, in the case of each dependency, we have the need to make the distinction- it seems almosty arbitrary, and we run into issues, such as these failing tests... 

Then again, I don't run anything on anything other than x86_64, so admittedly, this may just be a lack of cross-architecture experience, on my part, and, of course, some packages require a different target, than host, whilst others may not.

antonio-malcolm:
Ok, on some architectures, there appears to be some crossover, between what's expected for the target architecture, and what's expected for the binary. Noted.

antonio-malcolm:
Is there anything else which needs to be done, here?

srcpkgs/FontManager/template
```diff
@@ -0,0 +1,17 @@
+# Template file for 'FontManager'
+pkgname=FontManager
+version=0.7.2
+revision=1
+wrksrc="master-${version}"
+build_style=gnu-configure
+disable_parallel_build=yes
+hostmakedepends="pkg-config gobject-introspection intltool yelp-tools vala
+ file-roller glib-devel gucharmap-devel libgee08-devel"
```
Vaelatern: Without a good reason, no `-devel` in `hostmakedepends`.

antonio-malcolm: Not sure if it's a good enough reason, but, as I already did, 11 days ago, I just repeated attempts to build, both with and without the `-devel` packages in `hostmakedepends`. It only builds sucessfuilly, when they're present.

Vaelatern:
Make the name lowercase? Makes it easier for people to find. You'll need to define a wrksrc to make this change.

antonio-malcolm:
Done.

srcpkgs/fontmanager/template
```diff
@@ -0,0 +1,17 @@
+# Template file for 'fontmanager'
+pkgname=fontmanager
+version=0.7.2
+revision=1
+wrksrc="master-${version}"
+build_style=gnu-configure
+disable_parallel_build=yes
+hostmakedepends="pkg-config gobject-introspection intltool yelp-tools vala
+ file-roller glib-devel gucharmap-devel libgee08-devel"
```
Vaelatern: A good reason and possible alternative:

> `gucharmap-devel` is needed at configure and compile time. Thus for the target arch. Thus in `makedepends`. Thus not duplicated here.
> Same for `libgee08-devel`.
> `glib-devel` is the only package that provides the ELF binary `glib-compile-resources` which is run to compile the source, thus is run by the host, thus stays in `hostmakedepends`.
> 
> It's not that simple though. For the VALAC steps, `error: Package 'gee-0.8' not found in specified Vala API directories or GObject-Introspection GIR directories` is thrown. What provides that in the correct directories? `libgee08-devel-0.18.0_1 /usr/share/vala/vapi/gee-0.8.vapi`. So that needs to be in `hostmakedepends`. And `error: Package 'Gucharmap-2.90' not found in specified Vala API directories or GObject-Introspection GIR directories`, and guess what...
> ```
> gucharmap-devel-8.0.0_1 /usr/share/gir-1.0/Gucharmap-2.90.gir
> ```
> 
> But that doesn't mean we need to include the whole development package twice for one file on the host and necessary includes for build time. Perhaps someone else knows a reasonable workaround, or if this is a reasonable action to take.
> 
> Whatever we come up with, the existance of `-devel` packages in `hostmakedepends` should be commented on in the template, for people in the future.

antonio-malcolm: I think I've already attempted all of the possible combinations, here, for all three packages. It simply won't pass all of the tests, without the `-devel` packages, in both 'makedepends` and `hostmakedepends`. That said, I'm open to suggestions. I'm always in favor of documenting code, when and where it's needed, tho.

Vaelatern: And I just explained why the `-devel` packages are needed in both places (except `glib-devel` which is only in hostmakedepends).

antonio-malcolm: Yeah, I'm just leaving the door open, if there are any further thoughts about it. OItherwise, I'll document the code this evening, once I'm home from work.

Vaelatern:
Just one commit please

antonio-malcolm:
Done.

Vaelatern:
This commit should be named as directed by the end of the [Manual (link)](https://github.com/voidlinux/void-packages/blob/master/Manual.md#contributing-via-git).

antonio-malcolm:
Done.

PR #3970, opened 2016-04-04
Cron alternatives

Vaelatern:
@xtraeme but note that all but `fcron` can't be installed at the same time at present.

### DIFF
--- a/srcpkgs/dcron/files/dcron/run
+++ b/srcpkgs/dcron/files/dcron/run
@@ -1,2 +1,2 @@
 #!/bin/sh
-exec crond -f 2>&1
+exec dcrond -f 2>&1
--- a/srcpkgs/dcron/template
+++ b/srcpkgs/dcron/template
@@ -1,16 +1,23 @@
 # Template file for 'dcron'
 pkgname=dcron
 version=4.5
-revision=29
+revision=30
 conf_files="/var/spool/cron/root"
-replaces="cron-daemon>=0"
-provides="cron-daemon-1_1"
+provides="cron-daemon-0_1"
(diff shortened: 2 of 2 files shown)


ebfe:
This might be a good opportunity to fix the file clashes between all the crons.

    cronie and dcron conflict for
    /usr/bin/crond
    /usr/bin/crontab
    /usr/share/man/man1/crontab.1
    /usr/share/man/man8/crond.8
   
    cronie and scron conflict for
    /etc/crontab
    /usr/bin/crond
    
    dcron and scron conflict for
    /usr/bin/crond

Alternatives for /etc/sv doesn't make a lot of sense if the packages overwrite their binaries.

Vaelatern:
I agree fully. I'll make the changes in separate commits to easily drop if xtraeme doesn't like them, to be squashed before comitting.

Vaelatern:
Not sure what to do about `/etc/crontab`

chneukirchen:
Leave `/etc/crontab` for the users to create.

Vaelatern:
Saying remove it in both packages that have it. Okay.

Vaelatern:
Action taken. Alternatives group should be fully implemented. Should only be one commit per template, unless it's important to show these changes.

Vaelatern:
@xtraeme @chneukirchen ready for review.

xtraeme:
The `replaces` var needs to be removed.

Vaelatern:
Commits added.

Vaelatern:
@xtraeme Ready again for review.

xtraeme:
@Vaelatern looks good, thanks :+1:

PR #3971, opened 2016-04-04
burp2: update to 2.0.36

yopito:


Vaelatern:
Your added comment to the top is not very informative.

yopito:
oops, forgot it. I've squashed it.

PR #3974, opened 2016-04-04
jreen-qt5, tomahawk(-qt5): fix build

ebfe:
cc @Duncaen @pullmoll


ebfe:
24cd0bfc753efd726cda0cef77fdf57797322ad3

PR #3976, opened 2016-04-04
SMC: fix build

ebfe:
cc @pullmoll


ebfe:
musl cross builds fail. For non-obvious reasons the cegui07 package doesn't ship /usr/lib/pkgconfig/CEGUI-OPENGL.pc on arm*-musl but does so on arm glibc.

ebfe:
Fixed with a rebuild of cegui07. e1ca81abbde7b1da01ee49e10f026b85b042de9b

PR #3979, opened 2016-04-05
Added lttng-ust and liburcu

t0b1nux:
Here is some work to add support for lttng-ust, which is required to build Microsoft .NET coreclr. I also included liburcu because it is itself required by lttng-ust.


t0b1nux:
Looks like it doesn't work well with musl.

t0b1nux:
Maybe I should rename the commits ? And separate the two packages ? Like 2 commits:

srcpkgs/lttng-ust/template
```diff
@@ -0,0 +1,23 @@
...
+build_style=gnu-configure
+short_desc="LGPLv2.1 userspace RCU (read-copy-update) library"
+maintainer="Simon Thoby <simonthoby@live.fr>"
+license="LGPL-2.1"
+homepage="http://lttng.org"
+distfiles="${homepage}/files/${pkgname}/${pkgname}-${version}.tar.bz2"
+checksum="b03f62fd0b2832fad20c10c337f3a0ee1c8f0ff3c9f9c52f28c72fc2ec1fba2e"
+makedepends="liburcu-devel"
+depends="liburcu"
+only_for_archs="i686 x86_64 armv6l armv7l aarch64"
```
Vaelatern: Correct way to do this would be a blacklist of musl.

`
	*-musl) broken="link_to_build"
esac
`

But it should be patched instead of just blacklisted unless that is really impractical.

t0b1nux:
@Vaelatern That's true, thanks for the reply. Nevertheless, if you look at the travis log, you will see a lot of errors trigered by the musl library itself, like this one:
```
/usr/armv7l-linux-musleabihf/usr/include/sched.h:101:1: error: expected ';', ',' or ')' before '*' token
 __CPU_op_func_S(XOR, ^)
 ^
```
Is it normal or is the cross-compilation going wrong ? 
Anyway, the error sems to concern a pretty high numbers of files and maintaining patches of this size doesn't look like a easy nor good idea.

t0b1nux:
@Vaelatern I have integrated your template modification in the meantime. However, as expressed before, I am unsure making a patch is the good solution.

srcpkgs/lttng-ust/template
```diff
@@ -0,0 +1,25 @@
...
+revision=1
+build_style=gnu-configure
+short_desc="LGPLv2.1 userspace RCU (read-copy-update) library"
+maintainer="Simon Thoby <simonthoby@live.fr>"
+license="LGPL-2.1"
+homepage="http://lttng.org"
+distfiles="${homepage}/files/${pkgname}/${pkgname}-${version}.tar.bz2"
+checksum="b03f62fd0b2832fad20c10c337f3a0ee1c8f0ff3c9f9c52f28c72fc2ec1fba2e"
+makedepends="liburcu-devel"
+depends="liburcu"
```
xtraeme: This `depends` line is unnecessary.

srcpkgs/lttng-ust/template
```diff
@@ -0,0 +1,25 @@
+pkgname="lttng-ust"
+version="2.7.2"
+revision=1
+build_style=gnu-configure
+short_desc="LGPLv2.1 userspace RCU (read-copy-update) library"
```
xtraeme: Same description than the other pkg?

srcpkgs/lttng-ust/template
```diff
@@ -0,0 +1,25 @@
+pkgname="lttng-ust"
```
xtraeme: The first line must be a comment like:

`# template file for 'lttng-ust'`

srcpkgs/lttng-ust/template
```diff
@@ -0,0 +1,25 @@
...
+esac
+
+lttng-ust-devel_package() {
+	depends="${sourcepkg}-${version}_${revision}"
+	short_desc+=" - development files"
+	pkg_install() {
+		vmove /usr/share/doc
+		vmove /usr/include
+		vmove /usr/lib/pkgconfig
+	}
```
xtraeme: The development pkg must contain the soname symlinks, aka:

`vmove usr/lib/*.so`

t0b1nux:
Thanks, Is it better now ?

srcpkgs/lttng-ust/template
```diff
@@ -0,0 +1,27 @@
...
+build_style=gnu-configure
+short_desc="Low-overhead tracing capabilities for userspace"
+maintainer="Simon Thoby <simonthoby@live.fr>"
+license="LGPL-2.1"
+homepage="http://lttng.org"
+distfiles="${homepage}/files/${pkgname}/${pkgname}-${version}.tar.bz2"
+checksum="b03f62fd0b2832fad20c10c337f3a0ee1c8f0ff3c9f9c52f28c72fc2ec1fba2e"
+makedepends="liburcu-devel"
+case "$XBPS_TARGET_MACHINE" in
+	*-musl) broken="link_to_build"
```
Vaelatern: By `link_to_build` I mean put an actual link to the failing travis build in there :)

srcpkgs/lttng-ust/template
```diff
@@ -0,0 +1,27 @@
...
+esac
+
+lttng-ust-devel_package() {
+	depends="${sourcepkg}-${version}_${revision}"
```
Vaelatern: `depends="${sourcepkg}>=${version}_${revision}"` is normal.

srcpkgs/lttng-ust/template
```diff
@@ -0,0 +1,27 @@
...
+	short_desc+=" - development files"
+	pkg_install() {
+		vmove /usr/share/doc
+		vmove /usr/share/man
```
Vaelatern: Don't think the manpages should only be in the development subpackages, same for the rest of the documentation.

t0b1nux:
@Vaelatern I was unsure about it, I thought maybe it tell the script to not compile the package because the compilation fail at link time...
Thanks nevertheless and sorry for all thoses errors, this is my first package submission

Vaelatern:

And yes, the `.a` file should go in the `-devel` package.

t0b1nux:
Shall I close this and start two new PR, based on upstream ?

Vaelatern:
No, you should instead force push your commits.

Vaelatern:
You can rebase and squash commits, that would work quite well actually. Then `git push -f`

Vaelatern:
Also, only 1 PR, 2 commits.

t0b1nux:
Oh, I didn't think I would be allowed to do things like rebase and force push...

Vaelatern:
With the version, please. according to how the manual says it.

t0b1nux:
All right, sorry

srcpkgs/lttng-ust/template
Vaelatern: github didn't notice, but this has been changed.

Vaelatern:
Please try to put the relevant common/shlib lines in the relevant package's commit?

t0b1nux:
Oh, sorry, You're right, I had completely forgottent it.

common/shlibs
```diff
@@ -2419,3 +2419,13 @@ liburcu-mb.so.4 liburcu-0.9.1_1
 liburcu-qsbr.so.4 liburcu-0.9.1_1
 liburcu-signal.so.4 liburcu-0.9.1_1
 liburcu.so.4 liburcu-0.9.1_1
+liblttng-ust-tracepoint.so.0
```
Duncaen: the pkgname is missing here

t0b1nux: Yeah, it was a copy/paste error

t0b1nux:
I guess anybody is interested in it currently, and I'm more using Rust than .Net currently. SO I don't really plan to maintain this. That's why I close this PR. Feel free to reopen a new PR from your own fork (of void-packages, ofc) if you want it to be merged.

jasonwilliams200OK:
@mjeanson, is there anyone at lttng using void linux? I am using Void Linux musl (Void Linux also comes with glibc). I guess musl variant would require the same patch you did for Alpine/aports.

mjeanson:
Yes, you'll need the same patches as those in Alpine for everything to build and work properly on Musl.

Vaelatern:
Nope. Those patches were upstreamed.

Vaelatern:
Even Alpine doesn't have them anymore.

jasonwilliams200OK:
@mjeanson, the lttng-ust v2.8.1 package has landed in the xbps repos (as you yourself upstreamed all the musl patches for 2.8.1 :smile:) But you are right, for lttng-tools, however, we would require your patches.

@Vaelatern, Note that for dotnet, we just need user-space lib (lttng-ust). However, it would be nice to have lttng-tools for kernel event tracing etc. available for the platform as Alpine now does with @mjeanson's benign patches: http://git.alpinelinux.org/cgit/aports/tree/testing/lttng-tools. :)

PR #3980, opened 2016-04-05
[WIP] fvwm: update to 2.6.6.

Vaelatern:
Newly hosted on git instead of by project, so update not tracked. Recent update to a previously dead project.

Does not build, needs some autoconf (I think) magic to create the configure. Any tips, @chneukirchen (as you have maintainership).


Vaelatern:
Already been done by maintainer.

PR #3981, opened 2016-04-05
New package: neomutt-20160404

Vaelatern:


Vaelatern:
Do I want to add something like
```
provides="mutt-${version}_${revision}"
replaces="mutt>=0"
```
?

Duncaen:
Or add alternatives to `neomutt`, `mutt` and `mutt-kz`.

Vaelatern:
Given the file lists, I am not sure alternatives are the best option. There are 9
manpages in `mutt-kz` and 7 in `mutt`, there are muttrc config files in both, but
Mutt has an extra file in there. There are several similarly named binaries as well.

If it's probably the better way to go, I can start writing alternatives.

asergi:
I really doubt someone wants to install/use all mutt flavors at the same time.
`provides` and `replaces` are enough, imo. Other than that, LGTM.

logenkain:
On Fri, Apr 08, 2016 at 06:07:29AM -0700, Alessio Sergi wrote:
> I really doubt someone wants to install/use all mutt flavors at the same time.
> `provides` and `replaces` are enough, imo. Other than that, LGTM.
> 
> ---
> You are receiving this because you are subscribed to this thread.
> https://github.com/voidlinux/void-packages/pull/3981#issuecomment-207424404

Don't underestimate how much I love mutt and want to play with its flavors, but I agree.  I doubt anyone actually wants to use multiple flavors at the same time. I suppose for that we can always use a VM.

chneukirchen:
Could be relevant on a multi-user system...

Vaelatern:
So what should the flavors be renamed? Where things need renaming that is.

Vaelatern:
Okay, going to go with provides and replaces. Should I bother with replaces or just provides?

Vaelatern:
This PR has been updated to have all the mutts using alternative groups. Should sysconfdir be changed, however? Might be cleaner to keep sysconfdir at `/etc` and pollute `/etc` with `Muttrc-$pkgname`

srcpkgs/mutt/template
```diff
@@ -1,32 +1,66 @@
...
-	rm ${DESTDIR}/etc/Muttrc.dist
-	rm ${DESTDIR}/etc/mime.types{,.dist}
-	rm ${DESTDIR}/usr/bin/{flea,muttbug}
-	rm ${DESTDIR}/usr/share/man/man1/{flea,muttbug}.1
+	# provided by mime-types
+	rm ${DESTDIR}/etc/${pkgname}/mime.types{,.dist}
+
+	rm ${DESTDIR}/etc/${pkgname}/Muttrc.dist
+
 	vinstall contrib/gpg.rc 644 etc Muttrc.gpg.dist
```
Vaelatern: Don't know if this should be `vconf` at all. I'm thinking `vsconf` is all that is necessary.

Vaelatern:
Was merged.

PR #3984, opened 2016-04-05
glib: bring back quark_init_on_demand.patch

ebfe:
Add back minimally adjusted quark_init_on_demand.patch. This patch got removed in the last update, but without this ardour (and maybe others?) crash at startup on x86_64-musl
(g_assert (quark_seq_id == 0)). 

cc @xtraeme


xtraeme:
Hmm I thought they applied a fix somehow... I supposed wrongly. Thanks :+1:

PR #3989, opened 2016-04-06
New package: libaccounts-qt-1.14, libaccounts-glib-1.21

subsonik:
libaccounts-qt is a dependency of kaccounts-integration, part of KDE Plasma 5. libaccounts-glib is a dependency of libaccounts-qt.


srcpkgs/libaccounts-qt/template
```diff
@@ -0,0 +1,32 @@
...
+}
+
+libaccounts-qt-devel_package() {
+	short_desc+=" - development files"
+	depends="${sourcepkg}>=${version}_${revision}"
+	pkg_install() {
+		vmove usr/include
+		vmove usr/lib/*.so
+		vmove usr/lib/pkgconfig
+		vmove usr/share/doc/accounts-qt
```
Vaelatern: Unless the docmentation is only necessary for the development package, the documentation probably should go in the main package.

subsonik: The documentation consists of a brief introduction and then just class/API references. I presume this should be a part of the -devel package? Sorry if not, I'm new to this!

Vaelatern: In this case, that makes sense.

subsonik:
Updated, thanks for your feedback.

asergi:
Please rename the `libaccounts-qt` package to `libaccounts-qt5`.

subsonik:
I've renamed the package but the doc folder is still called `libaccounts-qt`, which is hard-coded into the Makefile, I can knock up a patch to fix this if needed.

common/shlibs
```diff
@@ -2412,3 +2412,5 @@ libcmark.so.0 cmark-0.24.1_1
 libspotify.so.12 libspotify-12.1.51_1
 libwiredtiger-2.8.0.so wiredtiger-2.8.0_1
 libvidstab.so.0.9 libvidstab-0.98b_1
+libaccounts-glib.so.0 libaccounts-glib-1.21_1
+libaccounts-qt5.so.1 libaccounts-qt-1.14_1
```
asergi: `libaccounts-qt5.so.1 libaccounts-qt5-1.14_1`

srcpkgs/libaccounts-qt5/template
```diff
@@ -0,0 +1,29 @@
+# Template file for 'libaccounts-qt5'
+
+pkgname="libaccounts-qt5"
+_pkgname="libaccounts-qt"
```
asergi: You can drop `_pkgname` var.

srcpkgs/libaccounts-qt5/template
```diff
@@ -0,0 +1,29 @@
...
+build_style=qmake
+short_desc="Client library to the accounts database, with a Qt5-based API"
+maintainer="John Rowley <enterthevoid@codesector.co>"
+license="LGPL-2.1"
+homepage="https://gitlab.com/accounts-sso/libaccounts-qt"
+distfiles="${homepage}/repository/archive.tar.gz?ref=VERSION_${version}>${pkgname}-${version}.tar.gz"
+checksum="53d2e295b26295166db5c16333c91e32a599b36c2eee296b803ce39d6a99a8d9"
+hostmakedepends="pkg-config doxygen"
+makedepends="libaccounts-glib-devel qt5-devel"
+wrksrc="${_pkgname}-VERSION_${version}-a34ca4b6d250529c900b0382559553b6e5885918"
```
asergi: wrksrc="${pkgname/5/}-VERSION_${version}-a34ca4b6d250529c900b0382559553b6e5885918"

asergi:
> I've renamed the package but the doc folder is still called libaccounts-qt, which is hard-coded into
> the Makefile, I can knock up a patch to fix this if needed.

Not an issue for now.

Vaelatern:
@subsonik Please rebase on top of master and push again.

Vaelatern:

PR #4002, opened 2016-04-08
New package: arc-theme

renatoaguiar:


srcpkgs/arc-theme/template
```diff
@@ -0,0 +1,15 @@
+# Template file for 'arc-theme'
+pkgname=arc-theme
+version=20160331
+revision=1
+build_style=gnu-configure
+configure_script="./autogen.sh"
+hostmakedepends="automake pkg-config"
+makedepends="gtk+3-devel"
```
Vaelatern: Why did you include this?

xtraeme:
I refuse to push any THEME available on inet just because it can be packaged.

This theme can be downloaded by the user and unpacked into its home directory. 

There's no need to make an xbps package.

Thanks!

PR #4003, opened 2016-04-08
New package: vertex-theme

renatoaguiar:


srcpkgs/vertex-theme/template
```diff
@@ -0,0 +1,15 @@
...
+revision=1
+build_style=gnu-configure
+configure_script="./autogen.sh"
+hostmakedepends="automake pkg-config"
+makedepends="gtk+3-devel"
+depends="gnome-themes-standard gtk-engine-murrine"
+short_desc="Vertex is a theme for GTK 3, GTK 2, Gnome-Shell and Cinnamon"
+maintainer="Renato Aguiar <contact@renatoaguiar.org>"
+license="GPL-3"
+homepage="https://github.com/horst3180/${pkgname}"
```
Vaelatern: Please no `${pkgname}` in the `homepage` field.

srcpkgs/vertex-theme/template
```diff
@@ -0,0 +1,15 @@
+# Template file for 'vertex-theme'
+pkgname=vertex-theme
+version=20160329
+revision=1
+build_style=gnu-configure
+configure_script="./autogen.sh"
+hostmakedepends="automake pkg-config"
+makedepends="gtk+3-devel"
```
Vaelatern: Why did you include this?

xtraeme:
I refuse to push any THEME available on inet just because it can be packaged.

This theme can be downloaded by the user and unpacked into its home directory. 

There's no need to make an xbps package.

Thanks!

PR #4005, opened 2016-04-08
New package: snort-2.9.8.2, daq-2.0.6, libdnet-1.11

subsonik:
Snort and its two dependencies. Will need to add pulledpork later as a new package.


subsonik:
Need to integrate patches to fix musl build, I'll use the ones from Alpine Linux.

Vaelatern:

subsonik:
Thanks for your help, @Vaelatern. I'm still getting acquainted with this whole process, it doesn't help that I'm also new to using GitHub as a committer. It's a bit overwhelming but I'm getting there. :)

Vaelatern:
`nocross=yes`

srcpkgs/daq/template
```diff
@@ -0,0 +1,33 @@
...
+distfiles="${homepage}/downloads/snort/${pkgname}-${version}.tar.gz"
+checksum=b40e1d1273e08aaeaa86e69d4f28d535b7e53bdb3898adf539266b63137be7cb
+disable_parallel_build=yes
+nocross="./configure fails when cross-compiling; cannot run test program"
+
+pre_configure() {
+	case "$XBPS_TARGET_MACHINE" in
+		*-musl) patch -Np1 -i ${FILESDIR}/fix-musl-build.patch;;
+	esac
+}
```
xtraeme: This is nonsense... move the patches to the "patches" directory and remove this function completely.

srcpkgs/daq/template
```diff
@@ -0,0 +1,33 @@
...
+hostmakedepends="bison flex"
+makedepends="libpcap-devel"
+short_desc="Snort's abstraction layer for libpcap"
+maintainer="John Rowley <enterthevoid@codesector.co>"
+license="GPL-2"
+homepage="https://www.snort.org"
+distfiles="${homepage}/downloads/snort/${pkgname}-${version}.tar.gz"
+checksum=b40e1d1273e08aaeaa86e69d4f28d535b7e53bdb3898adf539266b63137be7cb
+disable_parallel_build=yes
+nocross="./configure fails when cross-compiling; cannot run test program"
```
xtraeme: Please stop abusing "nocross", if the pkg fails to cross compile just leave it as is.

srcpkgs/daq/template
```diff
@@ -0,0 +1,33 @@
+# Template file for 'daq'
+pkgname=daq
+version=2.0.6
+revision=1
+build_style=gnu-configure
+hostmakedepends="bison flex"
```
xtraeme: flex depends on bison, bison is redundant.

srcpkgs/libdnet/template
```diff
@@ -0,0 +1,27 @@
+# Template file for 'libdnet'
+pkgname=libdnet
+version=1.11
+revision=1
+build_style=gnu-configure
+hostmakedepends="pkg-config"
+short_desc="A portable interface to several low-level networking routines"
+maintainer="John Rowley <enterthevoid@codesector.co>"
+license="custom"
```
xtraeme: "custom" is not a proper license, please set it specifically.

srcpkgs/snort/template
```diff
@@ -0,0 +1,65 @@
...
+hostmakedepends="flex"
+makedepends="pcre-devel libpcap-devel libdnet-devel daq-devel libtirpc-devel"
+short_desc="A lightweight network intrusion detection system"
+maintainer="John Rowley <enterthevoid@codesector.co>"
+license="GPL-2"
+homepage="https://www.snort.org"
+distfiles="${homepage}/downloads/snort/${pkgname}-${version}.tar.gz"
+checksum=4075012d350dfa47a0200b7a920323f15cb7c370790f2a47367c03aba4009333
+conf_files="/etc/snort/*"
+system_groups="snort:29"
```
xtraeme: `system_groups` is redundant here, because `system_accounts` creates a system user/group by default.

radare:
oh right. lgtm

Vaelatern:
Please rebase on master to remove the conflicts and trigger a travis rebuild.

ebfe:
Needs another rebase unfortunately.

PR #4009, opened 2016-04-09
New package: libsoxr-0.1.2

subsonik:


srcpkgs/libsoxr/template
```diff
@@ -0,0 +1,30 @@
+# Template file for 'libsoxr'
+pkgname=libsoxr
+version=0.1.2
+revision=1
+wrksrc="${pkgname/lib/}-${version}-Source"
+build_style=cmake
+configure_args="
+	-DWITH_AVFFT=ON
```
Duncaen: we usually indent in variables with one space only.

38github:
Thank you guys for looking into this!

PR #3941, opened 2016-03-30
icu, libvpx: update. (soname bump)

ebfe:

PR #4010, opened 2016-04-09
swift: rebuild for icu-libs-57.1_1

jannisch:
Swift is currently broken because a system upgrade can't resolve `libicui18n.so.56` and `libicuuc.so.56` which are from the old icu-libs package. This should fix it.


jannisch:
Oh, didn't see [#3941](https://github.com/voidlinux/void-packages/pull/3941). I guess you should simply close my PR.

ebfe:
> Oh, didn't see #3941. I guess you should simply close my PR. 

No you were right. That commit slipped through the cracks somehow. it's rebuilding now.

PR #4012, opened 2016-04-10
New package: Vundle.vim-0.10.2

konimex:


Vaelatern:
This really does not require an OS package to install at all. Nothing can ever possibly require this to be installed, and the user can just create this any which way they like for themselves.

konimex:
So, should I just.. Y'know, close it?

PR #4013, opened 2016-04-10
antimicro: update broken distfiles

7185:
See Ryochan7/antimicro-packaging#1


Duncaen:
In case you didn't know, void is mirroring the distfiles too.
http://sources.voidlinux.eu/sources/antimicro-2.21/

PR #4017, opened 2016-04-11
gloobus-preview: Update Video Codec Support

antonio-malcolm:
Making the video codec support complete (via the gstreamer plugins), so videos preview properly.

### DIFF
--- a/srcpkgs/gloobus-preview/template
+++ b/srcpkgs/gloobus-preview/template
@@ -1,15 +1,14 @@
 # Template file for 'gloobus-preview'
 pkgname=gloobus-preview
 version=2015.12.21
-revision=1
+revision=2
 build_style=gnu-configure
-configure_args="--prefix=/usr"
-hostmakedepends="automake autoconf libtool pkg-config python3 gettext-devel xz"
+hostmakedepends="automake libtool pkg-config python3 gettext-devel xz"
 makedepends="gettext-devel boost-devel gtk+-devel gtk+3-devel gtksourceview-devel
 	cairo-devel gstreamer1-devel gst-plugins-base1-devel gst-plugins-bad1-devel poppler-glib-devel
 	libspectre-devel djvulibre-devel libgxps-devel freetype-devel glib-devel libX11-devel
 	libgxps-devel libarchive-devel"
-depends="python3 python-dbus gst-plugins-good1 gst-plugins-ugly1 ImageMagick p7zip tar unoconv unrar unzip"
+depends="python3 python-dbus gst-libav gst-plugins-good1 gst-plugins-bad1 gst-plugins-ugly1 ImageMagick p7zip tar unoconv unrar unzip"
 short_desc="A file previewer"
 maintainer="Antonio Malcolm <antonio@antoniomalcolm.com>"
 license="GPL-3"


antonio-malcolm:
Anything else I need to do, here?

xtraeme:
revision must be bumped

antonio-malcolm:
Oops! Yeah, that's now done, thanks.

srcpkgs/gloobus-preview/template
```diff
@@ -1,15 +1,15 @@
 # Template file for 'gloobus-preview'
 pkgname=gloobus-preview
 version=2015.12.21
-revision=1
+revision=2
 build_style=gnu-configure
 configure_args="--prefix=/usr"
```
Vaelatern: This line should really not be necessary, the `configure_args`

srcpkgs/gloobus-preview/template
```diff
@@ -1,15 +1,15 @@
...
 hostmakedepends="automake autoconf libtool pkg-config python3 gettext-devel xz"
```
Vaelatern: Remove `autoconf` from this line

srcpkgs/gloobus-preview/template
antonio-malcolm: Except that it is- I've built this from source, several times, both with and without xbps. Packages installed by a package manager should be installed into /usr, not /usr/local. This will default to /usr/local, without the arg..

srcpkgs/gloobus-preview/template
antonio-malcolm: This will not build without autoconf.

ebfe: automake depends on autoconf.

srcpkgs/gloobus-preview/template
Vaelatern: Please see `common/environment/configure/gnu-configure-args.sh`, especially the part where export configure_args="--prefix=/usr` .
https://gist.github.com/5af614218c096983cd1948f96d824b7a

antonio-malcolm: I'll run another local test, but when last I did, it installed to `/usr/local`

srcpkgs/gloobus-preview/template
antonio-malcolm: When building this locally, using `./xbps-src`, without autoconf, I get the following error:
`=> ERROR: wayland-1.10.0_1: failed to run pre_configure() at line 17.`

antonio-malcolm: And, here's the full stack of errors:
```
Can't load '/usr/lib/perl5/core_perl/auto/POSIX/POSIX.so' for module POSIX: /usr/lib/libm.so.6: version `GLIBC_2.23' not found (required by /usr/lib/perl5/core_perl/auto/POSIX/POSIX.so) at /usr/share/perl5/core_perl/XSLoader.pm line 70.
 at /usr/lib/perl5/core_perl/POSIX.pm line 36.
Compilation failed in require at /usr/share/autoconf/Autom4te/FileUtils.pm line 266.
BEGIN failed--compilation aborted at /usr/share/autoconf/Autom4te/FileUtils.pm line 266.
Compilation failed in require at /usr/bin/autoreconf line 42.
BEGIN failed--compilation aborted at /usr/bin/autoreconf line 42.
```

Vaelatern: ..... Why is it wayland that is failing when this package is not wayland?

antonio-malcolm: According to the CLI ouput, it's building wayland as a dependency of GTK+3 (gtk3+devel), which is a runtime dependency of gloobus-preview.

A better question is, why is this building differently, and subsequently breaking, locally, then on Travis.

Vaelatern: Rebase your branch on top of master.

PR #4024, opened 2016-04-11
New package: dotgit-1.2.0

oliver-cfc:


Vaelatern:
The software has no problem with just being installed in a bin folder by a user. No compilation necessary. Not sure we want this included.

oliver-cfc:
Fair enough.  I thought it might be nice to get some automated updates.

Vaelatern:
Someone still has to package those updates.

oliver-cfc:
I meant updates to the script (and bash-completion if there are ever new features)... Wouldn't the buildbot take care of everything once the version number was bumped?

Vaelatern:
No, that is not automatic, since changes can happen between version numbers that require modifying the template significantly.

PR #4025, opened 2016-04-12
Manual.md: Add package requirements

Vaelatern:
This is to help ensure effort is not wasted trying to package software, like jekyll, which doesn't need to or have a good argument for being included in the repository.

@xtraeme, for your review.


Vaelatern:
Numbered lists work as I have done here: see https://github.com/Vaelatern/void-packages/blob/requirements/Manual.md#quality_requirements

PR #4027, opened 2016-04-12
Mpv: add motion estimation via vapoursynth

lemmi:
this will need a little bit extra work, as some dependencies don't compile for anything but x86{,_64}.


srcpkgs/vapoursynth/template
```diff
@@ -0,0 +1,17 @@
+# Template file for 'vapoursynth'
+pkgname=vapoursynth
+version=R31
+revision=1
+build_style=gnu-configure
+hostmakedepends="base-devel yasm python3.4-devel python-Cython python3.4-Cython"
```
Vaelatern: `-devel` things ought to go in `makedepends` and on no accounts should you depend on `base-devel`

srcpkgs/vapoursynth-mvtools/template
```diff
@@ -0,0 +1,18 @@
+# Template file for 'vapoutsynth-mvtools'
+pkgname=vapoursynth-mvtools
+version=12
+revision=1
+build_style=gnu-configure
+hostmakedepends="base-devel yasm"
+makedepends="fftw-devel vapoursynth zimg-devel"
+depends=""
```
pullmoll: Remove this line.

srcpkgs/zimg/template
```diff
@@ -0,0 +1,32 @@
+# Template file for 'zimg'
+pkgname=zimg
+version=2.0.4
+revision=1
+wrksrc=zimg-release-${version}
+build_style=gnu-configure
+hostmakedepends="base-devel"
```
pullmoll: This should most probably be `hostmakedepends="automake libtool pkg-config"`.
As mentioned elsewhere, `base-devel` does definitely not belong here.

xtraeme:
- vapoursynth is missing a development pkg

PR #4028, opened 2016-04-13
go-ipfs: update to 0.4.0.

CMB:
The fs-repo-migrations package isn't really a dependency, but
included here because people will need it when upgrading.
Also added two hostmakedepends.

### DIFF
--- a/srcpkgs/fs-repo-migrations/template
+++ b/srcpkgs/fs-repo-migrations/template
@@ -1,0 +1,18 @@
+# Template file for 'fs-repo-migrations'
+pkgname=fs-repo-migrations
+version=1.0.0
+revision=1
+build_style=go
+go_import_path="github.com/ipfs/fs-repo-migrations"
+short_desc="Tool to upgrade the IPFS filesystem repository"
+maintainer="Christopher Brannon <chris@the-brannons.com>"
+license="MIT"
+homepage="https://github.com/ipfs/fs-repo-migrations"
+distfiles="https://${go_import_path}/archive/v${version}.tar.gz"
+checksum=b8cd8ea22f59740be3eb7a5c0c6da769a0b26101d65237da2971ee0670e2e44d
+
+post_install() {
+	vlicense ipfs-1-to-2/go-datastore/LICENSE
+	vdoc README.md
+	vdoc run.md
+}
--- a/srcpkgs/go-ipfs/template
+++ b/srcpkgs/go-ipfs/template
@@ -1,7 +1,9 @@
 # Template file for 'go-ipfs'
 pkgname=go-ipfs
-version=0.3.11
-revision=3
+version=0.4.1
+revision=1
+create_wrksrc=yes
+build_wrksrc=${pkgname}-${version}
 build_style=go
 go_import_path="github.com/ipfs/${pkgname}"
 go_package="${go_import_path}/cmd/ipfs"
@@ -9,16 +11,24 @@ short_desc="Global versioned P2P merkle DAG file system"
 maintainer="Christopher Brannon <chris@the-brannons.com>"
 license="MIT"
 homepage="https://ipfs.io"
-distfiles="https://${go_import_path}/archive/v${version}.tar.gz"
-checksum=1806fea22ccee0c8005fffdd6391531c8df44f23571b3e9539f6fa3a0d4efb01
-broken=yes
+distfiles="https://${go_import_path}/archive/v${version}.tar.gz
+ http://the-brannons.com/ipfs-gx-deps-20160427.tar.xz"
+checksum="b6bcab4d9da11339523cd178165c24f7784abda9dd820a8f3490c78da67be786
+ 3c6d99f4847eba3d3956d1b2f2eb07e7e0fe108d6711f5353bbf3aaf059e4dde"
 
 do_build() {
-	GO15VENDOREXPERIMENT=0 go get -x ${go_package}
+	local path="${GOPATH}/src/${go_import_path}"
+	mkdir -p "$(dirname ${path})"
+	ln -fs $PWD "${path}"
+	ln -s "${wrksrc}/ipfs-gx-deps-20160427/src/gx" "${GOPATH}/src"
+	cd cmd/ipfs
+	go build
 }
 
-post_install() {
+do_install() {
+	vbin cmd/ipfs/ipfs
 	vinstall misc/completion/ipfs-completion.bash 644 usr/share/bash-completion/completions ipfs
 	vlicense LICENSE
 	vdoc README.md
+	vdoc CHANGELOG.md
 }
--- a/srcpkgs/go-ipfs/INSTALL.msg
+++ b/srcpkgs/go-ipfs/INSTALL.msg
@@ -1,0 +1,6 @@
+Please note that go-ipfs 0.4.0 introduces a change that breaks compatibility
+with earlier versions at the networking layer.  Thus, it will be unable to
+communicate with peers running versions of go-ipfs older than 0.4.0.
+If you are upgrading from version 0.3.x, you should install
+the fs-repo-migrations package and run the fs-repo-migrations tool
+to upgrade your IPFS filesystem repository.


srcpkgs/fs-repo-migrations/files/LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2014 Juan Batiz-Benet
```
Vaelatern: Why was this file not included with the tarball?

srcpkgs/go-ipfs/template
```diff
@@ -1,23 +1,31 @@
 # Template file for 'go-ipfs'
 pkgname=go-ipfs
-version=0.3.11
-revision=2
+version=0.4.0
+revision=1
 build_style=go
 go_import_path="github.com/ipfs/${pkgname}"
 go_package="${go_import_path}/cmd/ipfs"
+hostmakedepends="git gx gx-go"
```
Vaelatern: `hostmakedepends="git gx-go"`

srcpkgs/go-ipfs/template
```diff
@@ -1,23 +1,31 @@
...
 
 do_build() {
-	GO15VENDOREXPERIMENT=0 go get -x ${go_package}
+	local path="${GOPATH}/src/${go_import_path}"
+	mkdir -p "$(dirname ${path})"
+	ln -fs $PWD "${path}"
+	gx --verbose install --global
+	cd cmd/ipfs
+	go build
 }
```
Vaelatern: Is there a reason you don't use the recommended install method? Calling an arbitrary package manager and using it to produce an arbitrary build is not necessary, and definitely not ideal.

srcpkgs/gx-go/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'gx-go'
+pkgname=gx-go
+version=1.0.0
+revision=1
+build_style=go
+go_import_path="github.com/whyrusleeping/gx-go"
+depends="gx"
+hostmakedepends="git git-perl"
```
Vaelatern: `hostmakedepends="git-perl"`

srcpkgs/gx/template
```diff
@@ -0,0 +1,18 @@
+# Template file for 'gx'
+pkgname=gx
+version=0.4.0
+revision=1
+build_style=go
+go_import_path="github.com/whyrusleeping/gx"
+hostmakedepends="git git-perl"
```
Vaelatern: Same.

srcpkgs/fs-repo-migrations/files/LICENSE
CMB: Toyam Cox <notifications@github.com> writes:

> In srcpkgs/fs-repo-migrations/files/LICENSE:
>
>> @@ -0,0 +1,21 @@
>> +The MIT License (MIT)
>> +
>> +Copyright (c) 2014 Juan Batiz-Benet
>
> Why was this file not included with the tarball?

I had to find it in their git repo.  I don't know why they didn't
include it.

srcpkgs/go-ipfs/template
CMB: Toyam Cox <notifications@github.com> writes:

> Is there a reason you don't use the recommended install method?
> Calling an arbitrary package manager and using it to produce an
> arbitrary build is not necessary, and definitely not ideal.

Well I really don't have any choice on the package manager, since that's
what upstream is using for their dependencies now.
Here's what the ArchLinux guys are doing, more or less:

	local path="${GOPATH}/src/${go_import_path}"
	mkdir -p "$(dirname ${path})"
	ln -fs $PWD "${path}"
	make toolkit_upgrade
	export PATH="$PATH:$GOPATH/bin"
	make build

And that may be even worse, since it just fetches gx with go get and
then uses it.

To be honest, I don't know of any better options here,.

srcpkgs/go-ipfs/template
CMB: Toyam Cox <notifications@github.com> writes:

> In srcpkgs/go-ipfs/template:
>
>>  build_style=go
>>  go_import_path="github.com/ipfs/${pkgname}"
>>  go_package="${go_import_path}/cmd/ipfs"
>> +hostmakedepends="git gx gx-go"
>
> hostmakedepends="git gx-go"

Even simpler, git was unnecessary to begin with.
Fixed this and the other deps you mentioned in other comments.
Thanks for reviewing.

srcpkgs/go-ipfs/template
Vaelatern: Can we package the parts gx installs?

srcpkgs/fs-repo-migrations/files/LICENSE
Vaelatern: Where in the repo is it?

srcpkgs/fs-repo-migrations/template
```diff
@@ -0,0 +1,18 @@
...
+go_import_path="github.com/ipfs/fs-repo-migrations"
+short_desc="Tool to upgrade the IPFS filesystem repository"
+maintainer="Christopher Brannon <chris@the-brannons.com>"
+license="MIT"
+homepage="https://github.com/ipfs/fs-repo-migrations"
+distfiles="https://${go_import_path}/archive/v${version}.tar.gz"
+checksum=b8cd8ea22f59740be3eb7a5c0c6da769a0b26101d65237da2971ee0670e2e44d
+
+post_install() {
+	vlicense ${FILESDIR}/LICENSE
```
Vaelatern: Remove the special license file. Just `vlicense LICENSE`

CMB: Toyam Cox <notifications@github.com> writes:

> In srcpkgs/fs-repo-migrations/template:
> Remove the special license file. Just vlicense LICENSE

It's not in the top-level of the 1.0.0 archive,
but they added it to git:
https://github.com/ipfs/fs-repo-migrations/blob/master/LICENSE

Turns out that there's another copy buried in their archive as well:
ipfs-1-to-2/go-datastore/LICENSE

I suppose I could use that one if you like.

srcpkgs/go-ipfs/template
CMB: Toyam Cox <notifications@github.com> writes:

> Can we package the parts gx installs?

In theory yes.
But it depends on something like 45 separate packages, all of which have
hashes in their import path.
So basically that'd be some 45 Void packages just for IPFS.
Here's what my $GOPATH looks like after fetching all this stuff locally:
http://sprunge.us/EfTU

If we're going for reproducibility, those import paths use IPFS hashes,
so the build should be reproducible, right?

srcpkgs/fs-repo-migrations/template
Vaelatern: Yes please.

CMB: Toyam Cox <notifications@github.com> writes:

>> +	vlicense ${FILESDIR}/LICENSE
>
> Yes please.

Ok, $FILESDIR/LICENSE is gone.

CMB:
This keeps failing on travis for arm* with the message:

/void-packages/common/build-style/go.sh: line 13: go: command not found

It fails while building gx-go.
So I wonder if there's not some sort of parallelism issue going on?
Should I try splitting this up into multiple PRs or something?

Vaelatern:
Don't split into multiple PRs.

abelboldu:
Would be nice to add runit service files also, this one works for me:
```
# cat /etc/sv/ipfs/run 
#!/bin/sh
[ -r conf ] && . ./conf
exec 2>&1
exec ipfs daemon
```

abelboldu:
Would be nice to add runit files for the daemon too:
```
18:00:57 [root@gulik ~]# cat /etc/sv/ipfs/log/run 
#!/bin/sh
install -d -m0750 -o root -g wheel /var/log/ipfs
echo "Logging to /var/log/ipfs/current"
exec chpst -u root:daemon svlogd -t /var/log/ipfs
18:01:03 [root@gulik ~]# cat /etc/sv/ipfs/run 
#!/bin/sh
[ -r conf ] && . ./conf
exec 2>&1
exec ipfs daemon

```

Vaelatern:
@xtraeme can you have a look at this, specifically the build failures?
First build is successful: https://travis-ci.org/voidlinux/void-packages/jobs/123658377#L1254-L1281
Then: https://travis-ci.org/voidlinux/void-packages/jobs/123658377#L1470-L1481

```
14:11 <Vaelatern> dominikh: PR 4028 https://github.com/voidlinux/void-packages/pull/4028 is having some weird issues. go-ipfs on cross builds doesn't pick up that there are any dependencies, so it fails to build entirely.
14:26 <Vaelatern> Does seem to be a build system failure, these errors in crossing.
14:36 <dominikh> meh, I'm taking a look
14:40 <dominikh> it's building gx and gx-go twice, which in and of itself looks wrong. also I think it's a bug I've seen before, where it installs build deps, builds one package, removes teh build deps, then tries to build the next package, without actually installing the deps
14:40 <dominikh> see line 1470
14:40 <dominikh> this was fixed once before; maybe it's a regression. ping xtraeme
```

CMB:
Yes, I like the idea of a system-wide ipfs, supervised by runit.
Here are a few extra considerations I should make if I do that.

The IPFS daemon doesn't need root permissions,
so we also need a system account.

The default data path for go-ipfs is ~/.ipfs.  That's not so friendly
for a system-wide installation.  It can be changed via the IPFS_PATH
environment variable.  So I'd make it something like /var/lib/ipfs.

Before the daemon is started for the first time, we also have to run
ipfs init.  I could check the state of the IPFS repository in
the runit script and then run ipfs init if it is empty.

There's also ipfs mount, which lets you make a readonly mount of IPFS
and IPNS using fuse.  The daemon needs to be running first for this to
work.  By default, the mounts are only accessible to the user running
ipfs.  Getting ipfs mount to work right in the system-wide case probably
isn't worth the trouble, though.

I thought about some of this when I made the first go-ipfs package.
Looking at the defaults and docs, it seemed like it wasn't quite meant
to be a system service, so I never added runit files.  But it's doable.

7185:
Any progress on this?
Most of the packages are now out of date.

CMB:
No.
And there's a further problem.  None of the released versions of gx will
build, because they try to pull in dependencies with go get, and the
repo layout of one of those has changed.
I can fix that with a one-line patch.  But now I'm wondering if I
shouldn't find some way to bundle all of gx's dependencies.
I have hosting space for the bundle.

In fact, if the bundle idea sounds good, I could just dispense with
packaging gx altogether.  Instead, I'd build it locally, use gx
to grab the deps for go-ipfs, and then bundle them up in a source
tarball.

It would work, if there were no objections.

Vaelatern:
I like that approach more than gx doing whatever it wants. Consistency is good.

CMB:
Yes, it seems to be the best approach so far.
And the latest has no build issues on travis; I'm happy with it.

CMB:
So is this ready to merge or what?

Issue #4031, opened 2016-04-14
pd: audio broken


hjek:
The version of **pd** currently available in **voidlinux** is only compiled with **jack** and **oss** support (no **alsa** or **portaudio**). **oss** appears not to work, and if **jack** is selected, it segfaults. So the **pd** package is currently broken (apart from the gui).

ebfe:
For alsa support, could you give #4033 a try? No idea how to actually use this.

hjek:
Cool. I assume that means the **pd** package will be compiled with **alsa** support once someone with write-access to this repo has approved the pull request?

Duncaen:
#4033 is already merged, did that fix your issues @hjek?

cr6git:
I wonder what this program is supposed to do?
````
$ pd
sh: 1: wish: not found
watchdog: signaling pd...
watchdog: signaling pd...
watchdog: signaling pd...
watchdog: signaling pd...
watchdog: signaling pd...
watchdog: signaling pd...
watchdog: signaling pd...
watchdog: signaling pd...
watchdog: signaling pd...
^CPd: signal 2
````
Very interesting...

PR #4033, opened 2016-04-14
pd: enable alsa output

ebfe:
#4031

### DIFF
--- a/srcpkgs/pd/template
+++ b/srcpkgs/pd/template
@@ -1,13 +1,13 @@
 # Template file for 'pd'
 pkgname=pd
 version=0.46.7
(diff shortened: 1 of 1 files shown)


xtraeme:
You can also enable pulseaudio, it will be linked to libpulseaudio and will behave correctly in all cases.

ebfe:
There doesn't seem to be a flag for pulseaudio. Are you thinking of portaudio?

ebfe:
Enabling portaudio is tricky since it wants to use it's own version of it and that conflicts with our portaudio package. There seems to be no option to link to the systems portaudio.

PR #4062, opened 2016-04-20
Opencv

Vaelatern:


Vaelatern:
nomacs and -qt5 should be bumped. Will take care if it later if nobody else does it today.

Vaelatern:
Man that's complicated,

Vaelatern:
With the `nomacs` version bump, we'd need to lose `nomacs` and replace it with the `nomacs-qt5` variant, since upstream doesn't support `libopencv-3.1` and `qt4` at the same time.

asergi:
`nomacs` bumped. All issues should be solved now.

asergi:
Apply the following patch to `opencv`: https://patch-diff.githubusercontent.com/raw/Itseez/opencv/pull/5861.patch. That should fix your issue with `gmic`.

asergi:
`gmic` package fixed.

Vaelatern:
Now `nomacs`, `gmic`, both build fine. It's now `actiona` that's at fault. I'll have a poke later.

Vaelatern:
There are no patches in the actiona tree to fix this

Issue #3355, opened 2016-01-11
Proposal: New metadata field "changelog"


chneukirchen:
I propose to add an optional string to each package which contains a
URL pointing to the changelog of a program by upstream, ideally for
the particular release if available.

I think this is useful
1) for users, since we don't package changelogs in the package
2) for maintainers, since they don't need four clicks on every version change to see if anything needs works

Now start bikeshedding about the name.  I like `changelog`.

chneukirchen:
It's a bit annoying that adding this really requires a XBPS patch...

tac1ica:
As the average Joe user I'd consider this feature as the next best thing right after a switch in xbps-query to print the changelog :)

maxice8:
xbps-create got -c/--changelog , any plans to continue this ?

maxice8:
changelog is now a thing!

ebfe:
> changelog is now a thing!

No it isn't.

PR #4068, opened 2016-04-22
[RFC] xbps-src: handle changelog metadata field

ebfe:
#3355, depends on voidlinux/xbps#169. Not sure if I updated everything that's necessary, also needs some kind of version check to determine if xbps-create accepts `--changelog`.

### DIFF
--- a/common/environment/setup/sourcepkg.sh
+++ b/common/environment/setup/sourcepkg.sh
@@ -2,7 +2,7 @@
 # the package template (excluding subpackages).
 
 ## VARIABLES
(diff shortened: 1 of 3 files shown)


xtraeme:
I'd wait until 0.52 is released and then increasing the min xbps version required in xbps-src should dtrt.

xtraeme:
Looks ok btw :+1:

Vaelatern:
Can you also add a note in the Manual about this variable, so when this PR is merged the documentation is updated as well?

Vaelatern:
This OK?

PR #4075, opened 2016-04-24
New package: netcat-openbsd-1.105

tml:
Because traditional netcat doesn't know how to talk to Unix sockets.

Originally pulled from http://anonscm.debian.org/cgit/collab-maint/netcat-openbsd.git and massaged to build on Void Linux.


tml:
Already exists anyway - not sure how I overlooked it. :sigh:

PR #4077, opened 2016-04-25
New package: netdata-1.1.0

the-maldridge:
Automatic, at a glance, system monitoring.


srcpkgs/netdata/INSTALL
```diff
@@ -0,0 +1,8 @@
+case ${ACTION} in
+    post)
+        chown -R netdata:netdata /srv/netdata/
+	
+	chown root /usr/libexec/netdata/plugins.d/apps.plugin
```
Vaelatern: Default. Did you mean to change the group with this? Otherwise line not necessary.

the-maldridge: In my original testing it seemed it needed to actually be fully owned by the netdata user/group.  I will conduct furthur testing and reply if it is in fact necessary.

chneukirchen: Don't use absolute paths in hooks! PWD is the root of the target dir.

Vaelatern: Couldn't this entire hook be replaced with `make_dirs`? Is there a reason we don't encourage use of `make_dirs`?

the-maldridge:
I have incorperated the suggestions from above.

srcpkgs/netdata/INSTALL
```diff
@@ -0,0 +1,8 @@
+case ${ACTION} in
+    post)
+	chown netdata:netdata -R srv/netdata/
+
+	chown root usr/libexec/netdata/plugins.d/apps.plugin
```
xtraeme: - chown root:netdata
- chmod 4750

why? because otherwise any user might execute this with root privileges. This way only users in netdata group will.

the-maldridge: good point, I will update my local copy and re-push

Vaelatern:
`<xtraeme> maldridge: please change netdata to usr/share/$pkgname/webroot`

the-maldridge:
Updated per webroot change

PR #4087, opened 2016-04-27
New package: bvi-1.4.0

oliver-cfc:


srcpkgs/bvi/template
```diff
@@ -0,0 +1,12 @@
+# Template file for 'bvi'
+pkgname="bvi"
+version="1.4.0"
+revision="1"
+build_style="gnu-configure"
+makedepends="ncurses-devel"
+short_desc="A display-oriented editor for binary files"
+maintainer="oliver <ohemming@gmail.com>"
+license="GPL-1"
```
Vaelatern: Seriously?

oliver-cfc: Just thought I'd add pkgs not in the repo that I use fairly regularly.  If
it's not appropriate (and if not please let me know why so I won't make the
mistake again) then feel free to just close the request.

dominikh: According to the COPYING file bundled with bvi, it's licensed under version 2 of the GPL, not version 1.

Vaelatern: Yes. I don't know of any projects that still use GPLv1.

Vaelatern:
Since I was here anyway, eb2babb is the merged code, with the license fixed.

oliver-cfc:
Thanks all

PR #4089, opened 2016-04-27
New Package: libsass 3.3.6

arete:


srcpkgs/libsass/template
```diff
@@ -0,0 +1,32 @@
+# Template file for 'libsass'
+pkgname=libsass
+version=3.3.6
+revision=1
+wrksrc="$pkgname-$version"
```
Duncaen: thats the default

arete:
ok , i have removed the line 5.
regards

Duncaen:
`usr/lib/libsass.a` should be moved to the `-devel` package too.
And I would remove the `(library)` from `short_desc`.

Vaelatern:
Do the move with `vmove usr/lib/*.a`

arete:
i have update pull request

srcpkgs/libsass/template
```diff
@@ -0,0 +1,32 @@
...
+	autoreconf  -i
+}
+
+
+post_install() {
+	vlicense LICENSE
+}
+
+libsass-devel_package() {
+	depends="libsass>=${version}_${revision}"
```
ebfe: Could use `depends="${sourcepkg}>=${version}_${revision}"`

arete:
ok i have modify the commit wiut yours suggestion

PR #4090, opened 2016-04-28
vivaldi: update to 1.1.453.47

diogoleal:


srcpkgs/vivaldi/template
```diff
@@ -9,18 +9,16 @@ maintainer="Diogo Leal <diogo@diogoleal.com>"
 license="Proprietary"
 homepage="https://vivaldi.com"
 repository="nonfree"
-restricted=yes
```
Vaelatern: What changed?

diogoleal: I removed because with it, this package not is saw in repository
This package is stable and I think equivalent to Opera Browser.

ebfe: It's not in restricted due to instability but due to the license.

Duncaen: I think we should remove `opera` or mark it as restricted too.
From operas EULA:
> User shall not and shall not allow any third party to:
> (d) sublicense, lease, rent, loan, or distribute the Software or Services to any third party;

xtraeme: Indeed. opera needs to set `restricted` as well.

Thanks for catching this! :+1:

Vaelatern: Thus, @diogoleal, please set restricted again.

diogoleal: returned restricted=yes. =)

PR #4091, opened 2016-04-28
libsoup: update to 2.54.1. (rebuild packages built against libsoup-2.54.0)

ebfe:
2.54.0 contained an unintended ABI break (http://ftp.acc.umu.se/pub/GNOME/sources/libsoup/2.54/libsoup-2.54.1.news), so rebuild the packages that were built against it.


ebfe:
bump webkit2gtk instead of rebuild.

ebfe:
travis cancelled since it times out.

ebfe:
Rebuild webkit2gtk (cc @Gottox)

PR #4093, opened 2016-04-28
calibre: added hostdependencies: tslib, tslib-devel; dependency pytho…

apichlkostner:
- added dependency python-PyQt5-svg which is needed to convert books to mobi format
- added dependencies tslib-devel and tslib because build was not possible without ("/usr/bin/ld: cannot find -lts" for target /builddir/calibre-2.55.0/src/calibre/plugins/libheadless.so")
- added calibre.desktop to have an icon in the start menu

I think at least calilbre.desktop should be usefull for everyone.


ebfe:
revision needs to be incremented.

srcpkgs/calibre/template
```diff
@@ -1,20 +1,20 @@
...
-revision=2
+revision=3
 build_style=python-module
 hostmakedepends="python pkg-config python-dateutil python-lxml python-Pillow
 python-PyQt5-webkit python-apsw python-cssutils"
 makedepends="python-devel qt5-qmake python-PyQt5-devel glib-devel
 fontconfig-devel libmagick-devel libressl-devel icu-devel sqlite-devel
 libchmlib-devel libpodofo-devel qt5-devel libusb-devel libmtp-devel
-libinput-devel libxkbcommon-devel"
+libinput-devel libxkbcommon-devel tslib-devel tslib"
```
xtraeme: `tslib-devel` is enough, remove `tslib`.

Issue #4100, opened 2016-04-29
Instant Segmentation Fault in redis-cli on Void Musl x86_64


bitshark:
Problem appears to occur due to some sort of bug or misconfiguration in the system-wide jemalloc libraries, which the void redis package is linked against.

```bash
[user@dolphin redis-3.0.7]$ redis-cli
Could not connect to Redis at 127.0.0.1:6379: Connection refused
not connected> quitSegmentation fault
```

```bash
[user@dolphin redis-3.0.7]$ gdb redis-cli
GNU gdb (GDB) 7.11
Reading symbols from redis-cli...(no debugging symbols found)...done.
(gdb) run
Starting program: /usr/bin/redis-cli 
Could not connect to Redis at 127.0.0.1:6379: Connection refused
not connected> q
                Program received signal SIGSEGV, Segmentation fault.
                                                                    0x00007ffff7b20856 in free () from /lib/libjemalloc.so.2
(gdb) 
#0  0x00007ffff7b20856 in free () from /lib/libjemalloc.so.2
#1  0x000055555556a7c2 in linenoise ()
#2  0x000055555555d359 in main ()
```

and this
```
[user@dolphin redis-3.0.7]$ strace redis-cli 
[user@dolphin redis-3.0.7]$ strace redis-cli 
execve("/usr/bin/redis-cli", ["redis-cli"], [/* 27 vars */]) = 0
arch_prctl(ARCH_SET_FS, 0x7f56457b2988) = 0
set_tid_address(0x7f56457b29c0)         = 3693
open("/etc/ld-musl-x86_64.path", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
open("/lib/libjemalloc.so.2", O_RDONLY|O_CLOEXEC) = 3
fcntl(3, F_SETFD, FD_CLOEXEC)           = 0
fstat(3, {st_mode=S_IFREG|0755, st_size=260208, ...}) = 0
read(3, "\177ELF\2\1\1\0\0\0\0\0\0\0\0\0\3\0>\0\1\0\0\0\3403\0\0\0\0\0\0"..., 960) = 960
mmap(NULL, 2363392, PROT_READ|PROT_EXEC, MAP_PRIVATE, 3, 0) = 0x7f56452ce000
mmap(0x7f564550a000, 20480, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_FIXED, 3, 0x3c000) = 0x7f564550a000
[...]

PR #4101, opened 2016-04-29
Bugfix for redis-cli segfaulting on Void Musl x86_64

bitshark:
-Redis-cli binary as-distributed is causing instant-segfault on Void musl x86_64  see #4100

-This commit configures redis to use it's own internal version of jemalloc which is distributed with
the redis source, rather than the system-wide jemalloc.  This solves the segmentation fault bug
on void musl x86_64

Before:
[user@dolphin redis-3.0.7]$ redis-cli
Could not connect to Redis at 127.0.0.1:6379: Connection refused
not connected> quitSegmentation fault

After:
[user@dolphin redis-3.0.7]$ ./src/redis-cli
Could not connect to Redis at 127.0.0.1:6379: Connection refused
not connected> quit


Gottox:
I'll include this workaround. Nevertheless can you open a seperate issue to fix the real issue?

ebfe:
Not convinced this is a jemalloc problem. Looks more like linenoise is at fault.

Gottox:
This PR breaks cross compiling. Please fix.

bitshark:
I can take a look at the cross compile, but ARM is a bit out of my area of expertise.  If you can offer a suggestion on either (a) how to fix the underling issue causing the segfault (I think it's probably in jemalloc .. what makes you suggest it's linenoise?)  ,  or (b) how I can resolve the build with cross compiling (ie. general tips you could offer), I'd appreciate it, and I'll see what  I can do.  Thanks.

xtraeme:
This PR is unapplicable right now, the issue was fixed.

PR #4123, opened 2016-05-02
openimageio: update to 1.6.13.

lemmi:


Vaelatern:
Chances are really good (have not built to test) that `openimageio-1.6.13` will need to revbump the dependencies on the current SONAME, `libOpenImageIO.so.1.6.12` and `libOpenImageIO_Util.so.1.6.12` and change the `common/shlibs` line.

Vaelatern:
https://gist.github.com/f7a47fe0584305659e752f5583f86029 are the packages that depend on one of those sonames

Vaelatern:
No, no revbumps necessary.

lemmi:
Yeah, this is only a patch-level update. But to confirm it, blender still runs with the updates version.

PR #4130, opened 2016-05-04
New package: lilyterm-0.9.9.4

kurumushi:


srcpkgs/lilyterm/template
```diff
@@ -0,0 +1,16 @@
+# Template file for 'lilyterm'
+
+pkgname="lilyterm"
+version="0.9.9.4"
+revision=1
+build_style="configure"
```
xtraeme: build-style=gnu-configure

srcpkgs/lilyterm/template
```diff
@@ -0,0 +1,16 @@
...
+configure_args="--prefix=/usr"
```
xtraeme: this line is unnecessary with `gnu-configure`

srcpkgs/lilyterm/template
```diff
@@ -0,0 +1,16 @@
...
+short_desc="A fast and lightweight terminal emulator based off of libvte"
+maintainer="Kris Scott <kris@syntosis.net>"
+license="GPL-3"
+homepage="http://lilyterm.luna.com.tw"
+distfiles="http://lilyterm.luna.com.tw/file/lilyterm-${version}.tar.gz"
+hostmakedepends="pkg-config glib-devel gtk+-devel vte-devel"
```
xtraeme: remove `gtk+-devel` and `vte-devel`

srcpkgs/lilyterm/template
```diff
@@ -0,0 +1,16 @@
...
+makedepends="pkg-config glib-devel gtk+-devel vte-devel"
```
xtraeme: remove `pkg-config` and `glib-devel`

srcpkgs/lilyterm/template
```diff
@@ -0,0 +1,16 @@
...
+only_for_archs="i686 x86_64 armv6l armv7l"
```
xtraeme: Remove this line completely

kurumushi:
All changed. hostmakedpends and makedepends still confuses me.

kurumushi:
Actually I made the changes you said, but now the build fails on arm. Going to tweak it around a bit...

Vaelatern:
Failing is fine. Failing means we can try to fix it.
@xtraeme What do you think of on musl, `broken="canonicalize_file_name not in musl"`?
@kurumushi Could you try adding gtk+-devel to hostmakedepends? The configure check might need to see it.

ebfe:
canonicalize_file_name(name) -> realpath(name, NULL);

kurumushi:
So now I added everyting back into hostdepends and makedepends. We are compiling everywhere but on musl, should I be trying to make a patch for musl? It seems the failure is due to libc fuctions that are not in musl.

Also, as @xtraeme suggested, I am using gnu_configure instead of configure, but doing so creates many warnings.

Vaelatern:
Yes, please create a patch, as per the replacement @ebfe noted, that `canonicalize_file_name(name)` could be replaced with `realpath(name, NULL);`

srcpkgs/lilyterm/template
Vaelatern: Please remove `glib-devel` and `vte-devel` until it's shown that they are actually necessary here.

kurumushi: Removed glib-devel, but vte was required to build on arm.

kurumushi:
Ok, looks like we're building everywhere now.

srcpkgs/lilyterm/template
```diff
@@ -0,0 +1,22 @@
...
+makedepends="pkg-config vte-devel gtk+-devel"
+checksum="b72626431d4f5645e0e0f18449319fbb69bef4eab4345d5e37cde0ae712a5d74"
+
+post_extract() {
+	case "$XBPS_TARGET_MACHINE" in
+		*-musl) patch -p0 < ${FILESDIR}/lilyterm-musl.patch
+	esac
+}
+
+# vim: set noexpandtab:
```
Vaelatern: Remove this line.

srcpkgs/lilyterm/template
```diff
@@ -0,0 +1,22 @@
...
+license="GPL-3"
+homepage="http://lilyterm.luna.com.tw"
+distfiles="http://lilyterm.luna.com.tw/file/lilyterm-${version}.tar.gz"
+hostmakedepends="pkg-config gtk+-devel vte-devel"
+makedepends="pkg-config vte-devel gtk+-devel"
+checksum="b72626431d4f5645e0e0f18449319fbb69bef4eab4345d5e37cde0ae712a5d74"
+
+post_extract() {
+	case "$XBPS_TARGET_MACHINE" in
+		*-musl) patch -p0 < ${FILESDIR}/lilyterm-musl.patch
```
Vaelatern: Please move the patch to `srcpkgs/lilyterm/patches/lilyterm-musl.patch` and remove this function. Patches are applied automatically.

srcpkgs/lilyterm/files/lilyterm-musl.patch
```diff
@@ -0,0 +1,14 @@
+--- ./src/profile.c.orig	2016-05-07 00:22:09.702161948 +0900
++++ ./src/profile.c	2016-05-07 00:24:21.566170424 +0900
```
Vaelatern: Instead of `./src/`, `src/` should work

srcpkgs/lilyterm/template
kurumushi: Is there a way to name it such that it only get automatically applied for musl? According to the man page realpath is broken by design, and I'm not sure if it's the best choice unless required.

Vaelatern: Apply it unconditionally.
If you kept reading that section of the man page, it concludes:
```
       The resolved_path == NULL feature, not standardized in
       POSIX.1-2001, but standardized in POSIX.1-2008, allows this design
       problem to be avoided.
```
Since the function prototype is `char *realpath(const char *path, char *resolved_path);` it stands to reason that the function, as called, is not broken by design.

srcpkgs/lilyterm/template
```diff
@@ -0,0 +1,20 @@
...
+version="0.9.9.4"
+revision=1
+build_style="gnu-configure"
+short_desc="A fast and lightweight terminal emulator based off of libvte"
+maintainer="Kris Scott <kris@syntosis.net>"
+license="GPL-3"
+homepage="http://lilyterm.luna.com.tw"
+distfiles="http://lilyterm.luna.com.tw/file/lilyterm-${version}.tar.gz"
+hostmakedepends="pkg-config gtk+-devel vte-devel"
+makedepends="pkg-config vte-devel gtk+-devel"
```
Vaelatern: `pkg-config` in `makedepends` makes no sense.

kurumushi:
Ok, all the changes made. Anyone see anything else that needs fixed?

Vaelatern:
Now that we know the `-devel` packages are detected, just in the wrong directories, could you take a look and see what to pass `./configure` so it detects the libraries in the right places?

kurumushi:
Ok, found the culprit. The configure script checks for pkg-config using `whereis "pkg-config" | tr -s ' ' '\n' | grep "bin/""pkg-config""$$" | head -n 1`. This returns the host's pkg-config instead of the wrapper when cross-compiling.

I made a patch that replaces these lines in the configure script and makefiles with `which pkg-config` and it finds the correct file. Is there a more preferred way to handle this?

srcpkgs/lilyterm/template
```diff
@@ -0,0 +1,15 @@
+# Template file for 'lilyterm'
+
+pkgname="lilyterm"
+version="0.9.9.4"
+revision=1
+build_style="gnu-configure"
+configure_args="--enable-debug=Y --enable-verbosity=Y"
```
Vaelatern: Is debug mode for production software, or just during development?

Vaelatern:
Your patches to fix the `pkg-config` issues looks right.

srcpkgs/lilyterm/template
kurumushi: Oops, sorry, those were just for me. Removed them.

Vaelatern:
Your template needs the line: `conf_files="/etc/lilyterm.conf"`

PR #4131, opened 2016-05-04
fix permissions in duplicity template for man-page

netzverweigerer:
man-page file was created with 0600 previously, this should fix that.


Gottox:
please increment the revision

srcpkgs/duplicity/template
```diff
@@ -17,4 +17,6 @@ checksum=594c6d0e723e56f8a7114d57811c613622d535cafdef4a3643a4d4c89c1904f8
 post_install() {
 	# drop doc
 	rm -rf ${DESTDIR}/usr/share/doc
+	# set correct permissions on man-page file
+	chmod 0644 /usr/share/man/man1/duplicity.1
```
xtraeme: this won't work at all... use `${DESTDIR}/path` instead.

netzverweigerer:
thanks!

PR #4132, opened 2016-05-04
New package: california-0.4.0

elroncio:


elroncio:
Why there is no gobject-introspection for arm?

pullmoll:
`gobject-introspection` cannot be cross compiled.
Look at e.g. `srcpkgs/bamf/template` how to handle this issue. In short: add a `build_options="gir"` and only `if [ -z "$CROSS_BUILD" ]` set `build_options_default="gir"`. Then see how `$(vopt_if gir ...)` is used for the optional dependency in `hostmakedepends`.

srcpkgs/california/template
```diff
@@ -0,0 +1,17 @@
+pkgname="california"
+version="0.4.0"
+revision=1
```
Vaelatern: First line should be a comment. See most other templates.

elroncio:
Ok, done both. Still it don't compile for arm. What should I do to fix that?

srcpkgs/california/template
Vaelatern: github didn't pick up that this happened.

Vaelatern:
Well, according to the log, `gobject-introspection` can't be found. https://travis-ci.org/voidlinux/void-packages/jobs/127885300

The solution is, if it is actually required as it seems, to remove the `gir` build option and assorted logic, and just have `gobject-introspection` in `hostmakedepends`.

Vaelatern:
Please also add a comment explaining that you put `gtk+-devel` in `hostmakedepends` because the configure check doesn't properly search for it.

elroncio:
I forgot to add `gtk.devel` to `hostmakedepends`, sorry.

If I simply add `gobject-introspection` to `hostmakedepends` the build still fails for arm.

Vaelatern:
Disregard about `gtk+-devel`, got disoriented.

elroncio:
Do you think it's not necessary?

What about `gobject-introspection`?

srcpkgs/california/template
```diff
@@ -0,0 +1,14 @@
...
+build_style=gnu-configure
+short_desc="A new calendar application for GNOME 3"
+maintainer="Luca - <elroncio@gmx.it>"
+license="LGPL-2.1"
+patch_args="-p1"
+homepage="https://wiki.gnome.org/Apps/California"
+hostmakedepends="pkg-config intltool itstool vala xdg-utils yelp-tools gobject-introspection"
+makedepends="libgee08-devel evolution-data-server-devel gtk+3-devel"
+distfiles="https://download.gnome.org/sources/${pkgname}/${version%.*}/${pkgname}-${version}.tar.xz"
```
xtraeme: use `$GNOME_SITE` instead.

Vaelatern:
California seems to rely on gobject-introspection. Easy fix is to actually put gobject-introspection in makedepends to avoid setting nocross.

pullmoll:
I entirely missed this question, sorry. @Vaelatern no, quite a few packages can't be cross compiled because of a dependency, so this would be nothing new. If california requires gobject-introspection so be it.

PR #4134, opened 2016-05-05
swift: split into compiler and shared libraries

jannisch:


Vaelatern:
Are either of the subpackages useable separately?

jannisch:
Looks like they published a new 2.2.1 release without changing the version number, the old one was already downloaded on my computer.

`libswift` is sufficent to run precompiled swift applications, so it saves over 400MB if you don't want to install the compiler and clang. It is however required by the compiler.

`swift-devel` should include all headers inside _/usr/lib/swift_ (not required for runtime). They are not directly required by the compiler (e.g. a hello world program works without them), but could be incuded in the `swift` package. Do you think this is too strict? I'm not certain if there isn't any usecase where you can use them seperately.

jannisch:
WIll submit a fix soon after rebuilding the whole stuff.

Vaelatern:
I understand the use of `libswift`, but do not understand the requirement for a `-devel` subpackage, which has a very specific meaning within Void as the components necessary for other programs to compile against that software.

jannisch:
Ok, I first though this is also the case here.
Should be fixed now.

srcpkgs/swift/template
```diff
@@ -51,11 +51,18 @@ do_install() {
...
+	vman docs/tools/swift.1
+}
+
+libswift_package() {
+	short_desc+=" - shared libraries"
+	pkg_install() {
+		cd ../build/Ninja-ReleaseAssert/swift-linux-${XBPS_TARGET_MACHINE}
+		vmkdir usr/lib/swift
+		vcopy lib/swift/linux usr/lib/swift
+		vcopy lib/swift/glibc usr/lib/swift
```
xtraeme: You are copying the same files that are available in the main package (swift).

The subpackage should move the files instead.

Use `vmove usr/lib/swift/...`.

jannisch: How can I specify the targetdir with vmove?
I don't think I copy the same files because I copy different folders in the subpackage and also checked it with `xbps-query -f`.

xtraeme: Oh, I see. You are right. Thanks :-)

Issue #4127, opened 2016-05-03
cdrecord -scanbus failure


nkcyham:
Hi,

I am trying to set up a script that uses cdrecord to burn cd's. Since cdrecord no longer supports opening drive by dev number, I removed the -dev=/dev/sr0 from the command line in the script. However cdrecord cannot find the device "automagically", as the man page says:

[1747][root@5b4az:/home/neoklis]> cdrecord -scanbus
Cdrecord-ProDVD-ProBD-Clone 3.01 (x86_64-unknown-linux-gnu) Copyright (C) 1995-2015 Joerg Schilling
cdrecord: No such file or directory. Cannot open '/dev/pg*'. Cannot open or use SCSI driver.
cdrecord: For possible targets try 'cdrecord -scanbus'.
cdrecord: For possible transport specifiers try 'cdrecord dev=help'.

[1748][root@5b4az:/home/neoklis]> cdrecord dev=help
Cdrecord-ProDVD-ProBD-Clone 3.01 (x86_64-unknown-linux-gnu) Copyright (C) 1995-2015 Joerg Schilling
Supported SCSI transports for this platform:

Transport name:         sg
Transport descr.:       Generic transport independent SCSI
Transp. layer ind.:
Target specifier:       bus,target,lun
Target example:         1,2,0
SCSI Bus scanning:      supported
Open via UNIX device:   not supported

Transport name:         pg
Transport descr.:       SCSI transport for ATAPI over Parallel Port
Transp. layer ind.:
Target specifier:       bus,target,lun
Target example:         1,2,0
SCSI Bus scanning:      supported
Open via UNIX device:   not supported

Transport name:         ATA
Transport descr.:       ATA Packet specific SCSI transport
Transp. layer ind.:     ATAPI:
Target specifier:       bus,target,lun
Target example:         ATAPI:1,2,0
SCSI Bus scanning:      supported
Open via UNIX device:   not supported

Transport name:         ATA
Transport descr.:       ATA Packet specific SCSI transport using sg interface
Transp. layer ind.:     ATA:
Target specifier:       bus,target,lun
Target example:         1,2,0
SCSI Bus scanning:      supported
Open via UNIX device:   not supported

Transport name:         RSCSI
Transport descr.:       Remote SCSI
Transp. layer ind.:     REMOTE:
Target specifier:       rscsi@host:bus,target,lun
Target example:         REMOTE:rscsi@host:1,2,0
SCSI Bus scanning:      supported
[...]

PR #4135, opened 2016-05-05
cdrtools: Ensure kernel module loaded at boot

Vaelatern:
#4127 is it possible that the fix should be in the package itself?


pullmoll:
Perhaps add an `INSTALL` to load the `sg` module? Otherwise a reboot is required.

Vaelatern:
`        rm /usr/lib/modules-load.d/cdrtools.conf` Would this be necessary if the package owns the file?

pullmoll:
Hmm.. no, it would not be necessary - my bad. Posting before the 1st coffee :)

Vaelatern:
Added the INSTALL, not the remove. The `sg` module could legitimately be used by other programs, and is not provided in this package.

xtraeme:
IMHO xbps should never attempt to load a kernel module via package scripts.

modules-load.d is enough. The user can run `modules-load` manually.

pullmoll:
The master's voice :smile: Then perhaps a INSTALL.msg hint what's needed to make `cdrecord -scanbus` work without reboot would be useful?

Vaelatern:
Is the solution to these build errors `nocross`?

Vaelatern:
I'll add an `INSTALL.msg` informing the user they may wish to run `modules-load` to ensure all the necessary kernel modules are loaded.

srcpkgs/cdrtools/INSTALL
```diff
@@ -0,0 +1,4 @@
+case "$ACTION" in
+	post)
+	    [ "$UPDATE" != "yes" ] && modprobe sg;;
+esac
```
xtraeme: Please remove this INSTALL script.

Vaelatern:
INSTALL script replaced with an INSTALL.msg
Should I set nocross on this package?

PR #4136, opened 2016-05-05
pixman: try to fix musl stack size issue

pullmoll:
Use `static __thread` storage for a large buffer which was
allocated in the stack frame. Musl has a rather small per
thread stack frame size.


ebfe:
At some point it might be easier to increase musls default stack size a bit.

pullmoll:
Yeah, but there will always be applications excessing that increased size but just 1 byte :smile:

PR #4148, opened 2016-05-07
zimg: update to 2.1.

lemmi:


lemmi:
I filed an issue upstream.

ebfe:
https://github.com/ebfe/void-packages/commit/170b02932c51bbca5530bdf67de101290de58caf

Gottox:
Any news on this?

Vaelatern:
@lemmi please see the patch @ebfe provided, and possibly mention it to upstream.

lemmi:
there isn't a release yet with the fix. the fix is from after version 2.1, which fails to build unfortunately. if anyone instists on this update (vapoursynth beeing the only package using this afaik), i'll patch it, but i don't really see the point.

Vaelatern:
@lemmi Please include the patch as in @ebfe's branch in this PR. Just put a comment in the template file linking to the issue where you reported to upstream that something was wrong.

srcpkgs/zimg/template
```diff
@@ -1,6 +1,9 @@
 # Template file for 'zimg'
 pkgname=zimg
-version=2.0.4
+# 2.1 needs a patch to cross compile.
+# Please remove un update
```
Vaelatern: typo ;)

Vaelatern:
Seems like the shasum changed on you.

PR #4150, opened 2016-05-07
New package: python-ldap-2.4.25

the-maldridge:


Vaelatern:
No python 3 subpackage?

the-maldridge:
Upstream currently does not support building on 3.x

srcpkgs/python-ldap/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'python-ldap'
+pkgname=python-ldap
+version=2.4.25
+revision=1
+build_style=python-module
+pycompile_module="ldap"
```
Vaelatern: Should also have the `.py` files under `site-packages` in a space separated list.

Vaelatern: Examples include the `z3` package.

the-maldridge:
Updated per request

srcpkgs/python-ldap/template
```diff
@@ -0,0 +1,23 @@
+# Template file for 'python-ldap'
+pkgname=python-ldap
+version=2.4.25
+revision=1
+build_style=python-module
+pycompile_module="lidf.py sasl.py logger.py register.py dds.py
+ tokenizer.py subentry.py models.py modlist.py filter.py syncrepl.py
+ ldapobject.py functions.py deref.py pwdpolicy.py openldap.py
+ libldap.py readentry.py sessiontrack.py ppolicy.py simple.py
+ psearch.py dn.py async.py cidict.py dsml.py ldapurl.py"
```
Vaelatern: Apologies, should have specified .py files to a depth of 1. Thus: `ldap ldif.py ldapurl.py dsml.py`
Still not sure a shared object file should be in site-packages.

PR #4157, opened 2016-05-08
mongodb: build with mmapv1 suport for arm32

Vaelatern:
```
19:49 <tudurom> repo maintainer: on arm mongodb needs to be built with mmapv1 suport. it's currently unusable there
19:49 <tudurom> *maintainers
19:51 <tudurom> see https://github.com/archlinuxarm/PKGBUILDs/pull/1365
19:51 <tudurom> thank you
```

> I have the same problem on my ARMv7 board, which is caused by a change in storage engines. In short, 3.2 introduced wiredTiger, which is now the default storage engine and replaces mmapv1. However, as we run a 32-bit build, wiredTiger is not available. Therefore, we have to stick to mmapv1.

https://archlinuxarm.org/forum/viewtopic.php?f=15&t=10236


Duncaen:
And i686?

Vaelatern:
I left it out because upstream by default leaves it on for `i686` as per: https://github.com/mongodb/mongo/commit/ff8061cd1d97e99f141b02a7fe36e83f302b909c

That said it could have changed. I'll explicitly enable it.

Vaelatern:
Well, since that commit is tagged `3.3.5` and we only are on `3.2.6` it could well change in the future.

PR #4166, opened 2016-05-10
New package: pam-u2f-1.0.4

m-wynn:


srcpkgs/pam-u2f/template
```diff
@@ -0,0 +1,21 @@
+# Template file for 'pam-u2f'
+pkgname=pam-u2f
+version=1.0.4
+revision=1
+wrksrc="${pkgname/-/_}-${version}"
+build_style=gnu-configure
+configure_args="--with-pam-dir=/usr/lib/security"
+depends="libu2f-host libu2f-server"
```
Vaelatern: Bogus

srcpkgs/pam-u2f/template
```diff
@@ -0,0 +1,21 @@
...
+hostmakedepends="autoconf automake libtool pkg-config"
```
Vaelatern: `s/autoconf //`

srcpkgs/pam-u2f/template
```diff
@@ -0,0 +1,21 @@
...
+makedepends="libu2f-host-devel libu2f-server-devel libressl-devel pam-devel"
```
Vaelatern: Fair point. Remove `libressl-devel`

srcpkgs/pam-u2f/template
Vaelatern: No `depends` needed, xbps sees the link against those packages and will depend on them appropiately.

PR #4175, opened 2016-05-11
pam-u2f: fix homepage and add version check

m-wynn:


ebfe:
`pattern` goes into the update file

m-wynn:
Got it.  Still new at this.

PR #4176, opened 2016-05-11
qtcreator: updated to 4.0.0

t0b1nux:
Released today (https://blog.qt.io/blog/2016/05/11/qt-creator-4-0-0-released/).
Nothing special to say...


pullmoll:
I'll take care of this. There are some more qt5 subpackages since 5.6.0 which I wanted to add to qtcreator-full with the next version.

PR #4184, opened 2016-05-12
etc/repos-remote.conf: add correct location for repo3 and repo4

Duncaen:


xtraeme:
:+1:

Duncaen:
Added in b1ab524.

PR #4190, opened 2016-05-14
New package: influxdb-0.13.0

Grauwolf:


srcpkgs/influxdb/template
```diff
@@ -0,0 +1,49 @@
...
+checksum=661dc5719efa2db0ff5e5a4676c31852dd259a952dd8fc837766cb291e06d31b
+
+system_accounts="_$pkgname"
+influxdb_homedir="/var/lib/$pkgname"
+influxdb_shell="/bin/false"
+
+conf_files="/etc/influxdb/influxdb.conf"
+make_dirs="
+	/etc/influxdb 0755 root root
+	$influxdb_homedir 0755 $system_accounts $system_accounts"
```
Vaelatern: Convention is, for extending these variables past a single line, to use only a single space instead of a tab for indentation.

srcpkgs/influxdb/template
```diff
@@ -0,0 +1,49 @@
...
+	go install ./...
+}
+
+pre_install() {
+	rm -f $GOPATH/bin/gdm
+	rm -f $GOPATH/bin/stress_test_server
+	rm -f $GOPATH/bin/test_client
+	rm -f $GOPATH/bin/urlgen
+
+	vsv $pkgname
```
Vaelatern: The rest of these lines, from this to the end of the function, seem to be things that belong in a `post_install`

srcpkgs/influxdb/template
```diff
@@ -0,0 +1,49 @@
...
+distfiles="https://${go_import_path}/archive/v${version}.tar.gz"
+checksum=661dc5719efa2db0ff5e5a4676c31852dd259a952dd8fc837766cb291e06d31b
+
+system_accounts="_$pkgname"
+influxdb_homedir="/var/lib/$pkgname"
+influxdb_shell="/bin/false"
+
+conf_files="/etc/influxdb/influxdb.conf"
+make_dirs="
+	/etc/influxdb 0755 root root
```
xtraeme: This is not necessary because the dir will be created due to `conf_files`.

Grauwolf:
Thanks! Adapted the template according to your comments.

srcpkgs/influxdb/template
```diff
@@ -0,0 +1,49 @@
...
+	rm -f $GOPATH/bin/gdm
+	rm -f $GOPATH/bin/stress_test_server
+	rm -f $GOPATH/bin/test_client
+	rm -f $GOPATH/bin/urlgen
+}
+
+post_install() {
+	vsv $pkgname
+
+	vinstall etc/config.sample.toml 644 etc/influxdb influxdb.conf
```
Vaelatern: Not `vconf etc/config.sample.toml`?

Vaelatern: Ah, okay.

Issue #4192, opened 2016-05-15
Blueman appears to not work


Vaelatern:
https://gist.github.com/364c9635d175bba83701a266af1e7182
Error output above. dbus and bluetoothd are enabled and running fine.

When I try to run what appears to be the failing command, I get:

```
┌-(1:35) aviator@minerva ~
└» /usr/libexec/bluetooth/bluetoothd -n
bluetoothd[5060]: Bluetooth daemon 5.37
D-Bus setup failed: Connection ":1.11" is not allowed to own the service "org.bluez" due to security policies in the configuration file
bluetoothd[5060]: Unable to get on D-Bus
┌-(1:35) aviator@minerva ~
└» sudo !!
sudo /usr/libexec/bluetooth/bluetoothd -n
Password: 
bluetoothd[5083]: Bluetooth daemon 5.37
D-Bus setup failed: Name already in use
bluetoothd[5083]: Unable to get on D-Bus
```

Vaelatern:
It works if I start `blueman-applet` as root. This should not be necessary.

xtraeme:
The issue you are seeing is due to permissions in polkit... unless you add explicit rules to allow access you'll need a ConsoleKit2 session.

Vaelatern:
That's as easy as enabling a consolekit service?

xa0:
I'm having the same problem. Enabling consolekit service doesn't seem to help.

Vaelatern:
@xa0 Please update bluez and add yourself to the `bluetooth` group. Consolekit not required.

xa0:
Thank you Vaelatern. That worked :)

PR #4193, opened 2016-05-15
blueman: add consolekit, install message

Vaelatern:
#4192 ?


Vaelatern:
This is wrong. blueman doesn't need consolekit, bluez does.

PR #4194, opened 2016-05-15
zfs: remove zfs-import service

dominikh:
The service has several flaws:

1) The sequences 'start' and 'start, stop, start' produce different
system states. The first invocation of start will import and mount all
file systems. Stopping the service will unmount them, but not export
them. Running start again will not do anything, leaving the user with
unmounted file systems.

2) It imports all storage pools, not just those that are noted in the
cache. It's usually ill-advised to import all available pools.

3) It is different from all the other file systems and storage layers
that Void supports, which are all handled in stage 1.

This change depends on voidlinux/void-runit#46 for the replacement
solution.


Vaelatern:
:+1:

Vaelatern:
Is it just me or did this PR get confused?

dominikh:
Well, I fucked things up. Let that be a lesson, never create PRs from master.

dominikh:
#4207

PR #4196, opened 2016-05-15
New package: dub-0.9.24_1

jaagr:


srcpkgs/dub/template
```diff
@@ -0,0 +1,25 @@
+# Template file for 'dub'
+pkgname=dub
+version=0.9.24
+revision=1
+hostmakedepends="dmd pkg-config libcurl-devel"
```
Vaelatern: What does `libcurl-devel` provide that it is a `hostmakedepends`? `/usr/bin/curl-config`?

srcpkgs/dub/template
```diff
@@ -0,0 +1,25 @@
...
+depends="curl"
+short_desc="Package and build management system for D"
+maintainer="Michael Carlberg <c@rlberg.se>"
+license="MIT"
+homepage="https://github.com/dlang/dub"
+distfiles="${homepage}/archive/v${version}.tar.gz"
+checksum=88fe9ff507d47cb74af685ad234158426219b7fdd7609de016fc6f5199def866
+
+# Same restrictions as "dmd" apply
+only_for_archs="i686 x86_64"
```
Vaelatern: Remove these two lines.

srcpkgs/dub/template
Vaelatern: Even if the host must run `/usr/bin/curl-config` you must put `libcurl-devel` in `makedepends`. The host does not appear to run `curl-config` so it should be removed from `hostmakedepends`

jaagr: The `libcurl-devel` package is required for the `pkg-config` metadata file.

srcpkgs/dub/template
jaagr: Removing them breaks the build.

Gottox:
Before merging this armv[67] should be fixed.

srcpkgs/dub/template
```diff
@@ -0,0 +1,25 @@
+# Template file for 'dub'
+pkgname=dub
+version=0.9.24
+revision=1
+only_for_archs="i686 x86_64"
```
Vaelatern: As per @Gottox's comment, please remove this line so as to help diagnose arm builds.

jaagr: Still don't see why though? Since a required dependency (`dmd`) will only build on `i686` and `x86_64`

Vaelatern: So what?

Vaelatern: Let xbps take care of that. A piece of software does not inherently take on a snapshot of the limitations of the software it depends on. By leaving the facts to xbps to sort out, when `dmd` or some other piece of software suddenly becomes available (like `qt` on `arm*`), we will not need to manually fix templates.

jaagr: Won't put time into patching something that "might" work in the future. I'll leave the PR if anyone else would like to take a look - otherwise you can close this.

the-maldridge:
Closed as "Abandoned by Author".

PR #4200, opened 2016-05-16
xbps-src: error message if proot is missing

o1ux:


pullmoll:
`$XBPS_CHROOT_CMD` is used in xbps-src, not explicitly or exclusively `proot`. This test is wrong
and will make xbps-src fail for everyone who uses e.g. `uchroot` or `uunshare`. See `etc/defaults.conf` in your tree.

PR #4203, opened 2016-05-17
blender: Add support for COLLADA format

Spencer-H:
Opencollada is already in the repo's so compiling blender with Collada support is reasonable.


Gottox:
please increment the revision.

Spencer-H:
Fixed the Revision.

PR #4211, opened 2016-05-19
openldap: Enable more modules

the-maldridge:
Currently a lot of functionality for openldap just isn't being built.


the-maldridge:
Do not merge until @dinoocch approves these changes as sufficient to enable modular support for core modules.

dinoocch:
Looks good to me

PR #4213, opened 2016-05-20
intellij-idea-community-edition: update to 2016.1.2b.

asiekierka:


srcpkgs/intellij-idea-community-edition/template
```diff
@@ -1,14 +1,15 @@
...
 revision=1
 depends="virtual?java-environment giflib libXtst"
 short_desc="Java integrated development environment"
 maintainer="Adrian Siekierka <asiekierka@gmail.com>"
 license="Apache-2.0"
 homepage="http://www.jetbrains.org/"
 distfiles="http://download.jetbrains.com/idea/ideaIC-${version}.tar.gz"
-checksum=d8d889c42dfde1f41030e9d3f5850d364109df73a6a35081edc72d7d95f22fbf
+checksum=9e3e0f4538707da37237106697fff4e0b5ed651085f456671764e1fbc9651bbe
+nopie=yes
```
Vaelatern: Why?

asiekierka: Using the binary builds of IntelliJ IDEA because Java takes ages to compile for large projects, and the binary builds' executables do not do PIE.

Technically, one could build just the natives from source, I suppose.

Vaelatern: But they do provide source? Just asking here. Also still not completely certain as to why you added `nopie=yes`.

asiekierka: They do, the license is Apache-2.0 and all source is provided.

I added it because some of the executables included with the binary distribution are not compiled with PIE, so xbps-src would error upon trying to create a package.

Vaelatern: Can you look in to actually compiling it all ourselves?

asiekierka: I could, but it might be a good idea to let it slide this one time due to the fact the update is fairly urgent:

https://blog.jetbrains.com/idea/2016/05/important-security-update-for-intellij-idea-2016-1-and-older-versions/

PR #4219, opened 2016-05-22
New package: seq2gif-0.10.4

abenson:


Vaelatern:

And please don't open more than one PR. You can change your commits locally and then forcibly push to github and the PR will be updated, and there will be no need to open new PRs. This only serves to spam the developer's chat room.

PR #4231, opened 2016-05-25
dmd: update to 2.071.0

jaagr:


srcpkgs/dmd/template
```diff
@@ -18,15 +18,15 @@ fi
 
 only_for_archs="i686 x86_64"
 conf_files="/etc/dmd.conf"
-
-hostmakedepends="unzip"
+nopie=yes
```
Vaelatern: Can't you make it honour our CFLAGS and LDFLAGS instead?

Vaelatern: D doesn't PIE, it seems.

PR #4232, opened 2016-05-25
New package: lemonbuddy-0.1.2_1

jaagr:


srcpkgs/lemonbuddy/template
```diff
@@ -0,0 +1,29 @@
+# Template file for 'lemonbuddy'
+pkgname=lemonbuddy
+version=0.1.2
+revision=1
+wrksrc="${pkgname}-${version}"
+#only_for_archs="i686 x86_64"
+build_style=cmake
+hostmakedepends="cmake clang git-perl pkg-config"
```
Vaelatern: Don't need `cmake` listed.

PR #4237, opened 2016-05-27
new package: google-chrome-bin-51.0.2704.63

the-maldridge:
google-chrome as a managed package.  After conversation with @dinoocch this PR replaces #3792 .

### DIFF
--- a/srcpkgs/google-chrome-bin/template
+++ b/srcpkgs/google-chrome-bin/template
@@ -1,0 +1,34 @@
+# Template file for 'google-chrome'
+_chromeVersion=51.0.2704.63
+_chromeRevision=1
+_channel=stable
+
+pkgname=google-chrome-bin
+version=${_chromeVersion}.${_chromeRevision}
+revision=1
+maintainer="Michael Aldridge <aldridge.mac@gmail.com>"
+homepage="http://www.google.com/chrome/"
+license="chrome"
+short_desc="An attempt at creating a safer, faster, and more stable browser"
+only_for_archs="x86_64"
+repository=nonfree
+restricted=yes
+nostrip=yes
+
+_baseUrl="http://dl.google.com/linux/chrome/deb/pool/main/g/google-chrome-stable"
+_filename="google-chrome-${_channel}_${_chromeVersion}-${_chromeRevision}_amd64.deb"
+_chromeUrl="${_baseUrl}/${_filename}"
+
+distfiles="${_chromeUrl}"
+checksum=1cffc224a943cf8d5dc8d0252f96ec70997cd4c986fcfea70a1b4d255cef56ce
+
+
+do_extract() {
+	mkdir -p ${DESTDIR}
+	ar x ${XBPS_SRCDISTDIR}/google-chrome-bin-${version}/${_filename}
+}
+
+
+do_install() {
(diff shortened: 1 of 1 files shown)


srcpkgs/google-chrome/template
```diff
@@ -0,0 +1,35 @@
+# Template file for 'google-chrome'
+pkgname=google-chrome
```
Vaelatern: `pkgname`, `version`, and `revision` must appear together, in order.

srcpkgs/google-chrome/template
```diff
@@ -0,0 +1,35 @@
...
+
+
+do_extract() {
+	mkdir -p ${DESTDIR}
+	ar x ${XBPS_SRCDISTDIR}/google-chrome-${version}/${_filename}
+	tar xf data.tar.xz -C ${DESTDIR}
+}
+
+do_install() {
+	:
```
Vaelatern: You don't want to do the untarring in this function?

srcpkgs/google-chrome/template
```diff
@@ -0,0 +1,35 @@
+# Template file for 'google-chrome'
+pkgname=google-chrome
+
+_chromeVersion=51.0.2704.63
+_chromeRevision=1
+_channel=stable
+
+version=${_chromeVersion}
```
Vaelatern: Is there any case where you can not simply use `version` instead of `_chromeVersion` in this template?

srcpkgs/google-chrome/template
```diff
@@ -0,0 +1,35 @@
...
+homepage="http://www.google.com/chrome/"
+license="chrome"
+short_desc="An attempt at creating a safer, faster, and more stable browser"
+only_for_archs="x86_64"
+repository=nonfree
+restricted=yes
+nostrip=yes
+
+_baseUrl="http://dl.google.com/linux/chrome/deb/pool/main/g/google-chrome-stable"
+_filename=google-chrome-${_channel}_${_chromeVersion}-${_chromeRevision}_amd64.deb
```
Vaelatern: Please quote this value

Vaelatern:
`nopie=yes`

srcpkgs/google-chrome/template
the-maldridge: No, as there is not installing that actually gets done, only extraction.

srcpkgs/google-chrome/template
the-maldridge: moved

srcpkgs/google-chrome/template
the-maldridge: version did not contain the correct information, see the updated files

srcpkgs/google-chrome/template
the-maldridge: quoted

srcpkgs/google-chrome/template
Vaelatern: Extracting into destdir, and not extracting (directly) of the distfiles.

Vaelatern: Unless I am mistaken, the tar step is the only that puts files in destdir

Gottox:
@Vaelatern feel free to merge if you think it's ready :smile:

PR #4241, opened 2016-05-28
git-annex: update to 6.20160527

evandeaubl:


evandeaubl:
Ugh, I figured out why files disappeared from the tarball. Looks like tarballs from Hackage aren't including all the files any more. Will resubmit pull request once I have the template working with the tarball from github.

Vaelatern:
You can just force push to update a PR to keep all work in the same place, and possibly rename the PR `[WIP] <title>` while fixing.

PR #4242, opened 2016-05-28
git-annex: update to 6.20160527

evandeaubl:


srcpkgs/git-annex/template
```diff
@@ -13,8 +13,8 @@ short_desc="Git addon for managing large files"
 maintainer="Evan Deaubl <evan@deaubl.name>"
 license="GPL-3"
 homepage="http://git-annex.branchable.com"
-distfiles="http://hackage.haskell.org/package/${pkgname}-${version}/${pkgname}-${version}.tar.gz"
-checksum=1eac609eeedbf01cf088461577b478a3aa99f7ecefa668214308e3b5509c1506
+distfiles="http://github.com/joeyh/${pkgname}/archive/${version}.tar.gz>${pkgname}-${version}.tar.gz"
```
Vaelatern: Why the change in distfiles?

Vaelatern:
In fact you pushed to the same remote branch. No need to close PRs during this! It just serves to be confusing.

srcpkgs/git-annex/template
evandeaubl: The version in Hackage changed starting with this version to only contain the files strictly needed for the Cabal build, so it's missing things like the bash completion scripts. The Github release tarball contains all of the support files.

evandeaubl:
I was thinking that it might take a while to resolve, so I didn't want an open pull request hanging around that wasn't immediately addressible. Sorry for the confusion!

Vaelatern:
Suggestion is to change the name to start `[WIP]` while issues are being worked out.

PR #4250, opened 2016-05-29
New package: OpenDoas-0.1

Duncaen:
@chneukirchen

### DIFF
--- a/srcpkgs/opendoas/files/doas.pam
+++ b/srcpkgs/opendoas/files/doas.pam
@@ -1,0 +1,6 @@
+#%PAM-1.0
+#auth            sufficient      pam_timestamp.so timestamp_timeout=300 verbose
+#session         optional        pam_timestamp.so timestamp_timeout=300
+auth            include         system-auth
+account         include         system-auth
+session         include         system-auth
--- a/srcpkgs/opendoas/INSTALL
+++ b/srcpkgs/opendoas/INSTALL
@@ -1,0 +1,8 @@
+#
+# This script sets up correct perms for usr/bin/doas.
+#
+case "${ACTION}" in
+post)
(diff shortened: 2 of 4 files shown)


Vaelatern:
Can we have a lowercase name? Since the uppercase names are so confusing to find...

srcpkgs/OpenDoas/INSTALL
```diff
@@ -0,0 +1,8 @@
+#
+# This script sets up correct perms for usr/bin/doas.
+#
+case "${ACTION}" in
+post)
+	chmod 4111 usr/bin/doas
```
Vaelatern: Can't be done in `post_install()`?

Duncaen: afaik tar would drop the suid bit, sudo etc. use the `INSTALL` script too.

Duncaen:
The standard is using the upstream name, there is doas in small in the short_description, so it matches lowercase too.

srcpkgs/OpenDoas/INSTALL
Vaelatern: Please see `pkgname=iputils` line 29

Vaelatern: I know this example because I broke it in the libressl version bump.

srcpkgs/opendoas/files/doas.pam
```diff
@@ -0,0 +1,4 @@
+#%PAM-1.0
```
chneukirchen: Perhaps add an commented example for pam_timestamp here?

Duncaen: Yes why not, added it.

Gottox:
So excited! Could this replace sudo in base?

PR #4254, opened 2016-05-30
leiningen: update to 2.6.1

PythonJedi:
bin/lein autoinstalls and autoupdates leiningen, which is not desired behavior.
bin/lein-pkg is designed to be used by a package and does not install or update
by itself.

Also updating to upstream's latest release, 2.6.1


srcpkgs/leiningen/template
```diff
@@ -1,20 +1,24 @@
...
+skip_extraction="leiningen-${version}-standalone.zip"
+checksum="9e6a6c7b6e032921a37d5be8fcdc2c0288d3698f759e22bdd29de4eb327ebeab
+ d70078fba85d5f405d042a6d7bad3a1e5b4aafae565c2d581feb999e95ae6394"
 
 do_install() {
-	vbin bin/lein
+	vbin bin/lein-pkg lein
 	vman doc/lein.1
 
+	vinstall ${XBPS_SRCDISTDIR}/${pkgname}-${version}/leiningen-${version}-standalone.zip 644 usr/share/java leiningen-${verion}-standalone.jar
```
ebfe: `usr/share/java` -> `usr/share/java/leinignen`
`${verion}` -> `${version}`

PythonJedi: Fixing the misspell, but the installed script expects the jar 'leiningen-${version}-standalone.jar' to be in /usr/share/java. Is there a convention that each package makes its own folder in /usr/share/java?

PR #4256, opened 2016-05-31
freetds: update to 1.00

diogoleal:


Gottox:
Could you add a fix for the musl builds? looks like some includes are missing. Should be easy to fix.

PR #4262, opened 2016-06-01
New package: ejabberd-16.04

Vaelatern:


Vaelatern:
Problems:
[ ] It doesn't seem to cross compile and travis has a weird error for x86_64-musl.
[ ] Build process pulls things from git (all version tagged) but seems such a large part of the build, and it is version, so I don't think it's worth fixing unless someone really cares.

PR #4276, opened 2016-06-03
spotify: update to 1.0.31

teh-jazzman:


ebfe:
This release randomly segfaults for me.

teh-jazzman:
Does ist work for you at all? Do you hear sound?
I can reproduce segfaults when using alsa for sound output. There seems to be something wrong with their alsa driver. It's running stable on pulseaudio for me so far though.

Since this is closed source there is not much I can do about it. Maybe we should skip this version and hope for a fix in the next one?

ebfe:
I don't run pulseaudio. The furthest I got was pressing the play button...

I think we should skip this release. The problem is the 0.29 .deb is
no longer available, which means currently installing spotify fails. We might want to
actually downgrade to 0.28 since the .deb for that is still available...

teh-jazzman:
Agreed. I'll see how the next version performs, it should be around in 1-2 weeks.

PR #4277, opened 2016-06-03
mit-krb5: enable ldap database support

the-maldridge:
Since kerberos is often used in conjunction with ldap, it is often preferable to use ldap as the backing database for the master KDC.  This also permits the operation of multi-master KDC environments.


common/shlibs
```diff
@@ -531,6 +531,7 @@ libkdb5.so.8 mit-krb5-libs-1.13.1_1
 libkrb5.so.3 mit-krb5-libs-1.8_1
 libkrad.so.0 mit-krb5-libs-1.12.1_1
 libkrb5support.so.0 mit-krb5-libs-1.8_1
+libkdb_ldap.so.1 mit-krb5-libs-1.8_1
```
Vaelatern: `mit-krb5-libs-1.14.2_2`

the-maldridge:
Are there other changes needed for this PR?

PR #4285, opened 2016-06-05
New package: owncloud-9.0.2

Gottox:
I've got enough of of updating owncloud by hand. rfc ;)


srcpkgs/owncloud/INSTALL
```diff
@@ -0,0 +1,10 @@
+case "$ACTION" in
+post)
+	mkdir -p var/db/owncloud/data var/db/owncloud/apps var/log/owncloud
+	chown -R _owncloud var/db/owncloud etc/owncloud
```
Vaelatern: Not more with `make_dirs`?

Gottox: true, fixed.

Vaelatern:
I like the notion of xbps handling things for webapps. I don't know how much of this needs to be in the tree; I'd prefer patching the default `php.ini` if one is provided.
It's great having the package manager do things for you instead of yourself.

Vaelatern:
@voidlinux/pkg-committers Anybody want to weigh in on if things like webapps should be merged and handled by xbps?

Gottox:
I'll replace this pkg with nextcloud anyway in the next few days.

PR #4288, opened 2016-06-05
burp2-server: update to 2.0.40

yopito:


Vaelatern:
https://github.com/grke/burp/commit/7e49ed02bc33ce978121c835a9be73fd06ae71d3 is the origin of the patch

Vaelatern:
https://github.com/grke/burp/commit/b56782eac4c432c5937c42bcc5c56d6a3220c001 This seems also to be a reasonable patch, but at that point we are including what is essentially master in a release.

yopito:
* grke/burp@7e49ed0 : yes, my patch (`burp-2.0.40-status-monitor.patch
`) contains the right info (long line):  
    ```
git diff f5d731461eeaff5030df281f0add7fde49dc0330..7e49ed02bc33ce978121c835a9be73fd06ae71d3
    ```  
    I will use short hash instead ;)

* grke/burp@b56782e : thanks for pointing this. also grke/burp@f5d7314

I hesitate to retrieve all patchs since release then ...

Vaelatern:
https://github.com/grke/burp/commit/f5d7314 isn't a bugfix. Unless it's a rather severe usability or security bugfix, it definitely should not be a patch we keep track of. Wait for the next release for those.

srcpkgs/burp2-server/template
```diff
@@ -55,6 +56,9 @@ post_install() {
 	vmkdir "${_docdir}/client"
 	vcopy "configs/client/cron.example" "${_docdir}/client"
 	vcopy "configs/client/zfs_script" "${_docdir}/client"
+
+	# remove client config from -server package
+	rm "${DESTDIR}/etc/burp/clientconfdir/testclient"
```
Vaelatern: Wrong. The correct thing to do is `vmove` the relevant files in the `-client_package`

yopito: I don't agree, but I probably misexplained myself. file `burp/clientconfdir/testclient`define a client's configuration for the server (named "testclient"). looks dangerous to have such a client already defined in the server. This is file is not related with the -client package.  Please notice that the -doc package contains this file as a configuration's example.

Vaelatern: If upstream puts it there, and we don't really have very good reasons for removing it, leave it there.

yopito: upstream focus on automated test, not on packaging . It's really annoying to have a backup server with a client already defined "by default".

yopito: upstream focus on automated test, not on packaging . It's really annoying to have a backup server with a client already defined "by default".
Also: why a package upgrade should add a backup client definition on the server ?

Vaelatern: Fair. Remove the file.

PR #4295, opened 2016-06-07
chromium*: update to 51.0.2704.84.

ebfe:


ebfe:
travis fails on musl but works ok locally.

PR #4300, opened 2016-06-07
New package: go-ethereum-1.4.6

abelboldu:


srcpkgs/go-ethereum/template
```diff
@@ -0,0 +1,24 @@
...
+build_style=go
+go_import_path="github.com/ethereum/${pkgname}"
+short_desc="The Go command line for Ethereum"
+maintainer="Abel Vdo <vdo@greyfaze.net>"
+license="LGPL-3"
+homepage="https://ethereum.github.io/go-ethereum/"
+depends="leveldb"
+distfiles="https://github.com/ethereum/${pkgname}/archive/v${version}.tar.gz"
+checksum=3babd8b2279eea726a57c2e0de55e66c2ddc360a42a8e84bc12dc2924e452a57
+hostmakedepends="gcc go gmp"
```
Vaelatern: Please remove at least `gcc` and `go`.

srcpkgs/go-ethereum/template
```diff
@@ -0,0 +1,24 @@
...
+
+do_build() {
+	make
+}
```
Vaelatern: Why do you have a custom `do_build()`?

abelboldu:
@Vaelatern  Make is required to build the geth binary

dominikh:
I don't see anything in the makefile that we can't do ourselves. And by overwriting do_build like that, you are most likely breaking cross-compilation.

Vaelatern:
@dominikh Any input on what needs to be done here?

dominikh:
Stop relying on the makefile, stop using `build_style=gnu-makefile`, start using the `go` build style, start depending on leveldb in makedepends as it's required for building.

abelboldu:
@dominikh  I'm lost, I don't see the way to use `go` buildstyle as the Makefile has many targets. Any help?

Vaelatern:
Closed due to inactivity.

PR #4302, opened 2016-06-08
seafile-client-qt: update to 5.1.2

yopito:
(and its deps)

### DIFF
--- a/srcpkgs/seafile-client-qt/template
+++ b/srcpkgs/seafile-client-qt/template
@@ -1,11 +1,11 @@
 # Template file for 'seafile-client-qt'
 pkgname=seafile-client-qt
-version=5.1.1
+version=5.1.2
 revision=1
 _sourcename="seafile-client"
 wrksrc="${_sourcename}-${version}"
 build_style=cmake
-configure_args="-D CMAKE_BUILD_TYPE=Release -D USE_QT5=ON ."
+configure_args="-DUSE_QT5=ON"
 
 # nocross since qt unavailable for arm* architecture
 nocross=yes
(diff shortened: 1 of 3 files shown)


Vaelatern:
Mark `seafile-client-qt` broken for musl?
```
case "${XBPS_TARGET_MACHINE}" in
	*-musl) broken="requires fts.h";;
esac

**Comment by ebfe** [comment]
[musl-fts](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/musl-fts/template)

**Comment by Vaelatern** [comment]
Huh, it seems `xlocate` doesn't pick that up.

**Comment by yopito** [comment]
commit updated: Travis CI builds are fine on -musl flavor now.
Many thanks for your accurate checks and the tip about fts support with musl libc. 

Had to (dirty) patch cmake's file to **append** "`-lfts`" on linking stage ... If you have a better workaround ...

File: srcpkgs/seafile-client-qt/template [hunk 1]
```diff
@@ -1,11 +1,11 @@
 # Template file for 'seafile-client-qt'
 pkgname=seafile-client-qt
-version=5.1.1
+version=5.1.2
 revision=1
 _sourcename="seafile-client"
 wrksrc="${_sourcename}-${version}"
 build_style=cmake
-configure_args="-D CMAKE_BUILD_TYPE=Release -D USE_QT5=ON ."
```
Vaelatern: It no longer requires the `CMAKE_BUILD_TYPE=Release`?

ebfe: provided by the build-style.

PR #4304, opened 2016-06-08
catalyst: patched for linux4.6 & update to 15.302

ivierlin:
Obviously should be an upstream patch, but AMD doesn't care about Linux anyway.
So, here is some ducktape.

Still needs xorg-server-1.7 and xf86-input-evdev-2.10.0 to work. Maybe put them back in the current repo, so users don't have to build it themselves? The archives don't go back that far.


ivierlin:
Working fine on x86_64.
Untested on x86.

Gottox:
#yolo

PR #4307, opened 2016-06-09
New package: ldb-1.1.26 & tevent-0.9.28

the-maldridge:
Part of the sssd dependency chain


common/shlibs
```diff
@@ -2457,3 +2457,9 @@ libxlsreader.so.1 libxls-1.4.0_1
 libR.so R-3.3.0_2
 libmarkdown.so.2 discount-2.2.0_3
 libkeystone.so.0 keystone-0.9_1
+libtalloc.so.2 tevent-0.9.21_1
```
Vaelatern: Where do you get this from?

ebfe:
Why does this ship it's own copy of libtevent.so and talloc.so and tdb.so in /usr/lib/ldb? Can't use the system ones?

common/shlibs
```diff
@@ -2457,3 +2457,6 @@ libxlsreader.so.1 libxls-1.4.0_1
 libR.so R-3.3.0_2
 libmarkdown.so.2 discount-2.2.0_3
 libkeystone.so.0 keystone-0.9_1
+libtalloc.so.2 ldb-1.1.26_1
+libtevent.so.0 ldb-1.1.26_1
+libtdb.so.1 ldb-1.1.26_1
```
ebfe: i don't think these are correct. Most certainly other packages should depend on our talloc tevent or tdb package not on this one.

the-maldridge: "our" talloc is in fact an embeddable copy of the library and does not provide the full interface to the library.  Without explaining the questionable design decisions of the Samba project in a github ticket, we may be at an impasss here given that the versions of libtalloc, libtevent and libtdb (as well as ldb from another PR) are incomplete and cannot be linked against with sssd (my end goal with this PR).

Do you have a suggestion wherin we can have functional versions of libtevent, and ldb without conflicting with the incomplete embeddable versions provided by other Samba family packages?

ebfe: I haven't looked into it in any detail, but even if these .sos are required i don't think they should be listen in common/shlibs. This only creates confusion with multiple incompatible packages providing e.g. 'talloc.so' and since they ship in `/usr/lib/ldb` they wont be on the default ld seach path anyways.

As for how to get "complete" versions of these libs, i have no idea.

the-maldridge:
Commits are stacked in this PR to account for ldb needing tevent to build.

Vaelatern:
Then it should be just one PR.

the-maldridge:
Looks like one PR to me...

the-maldridge:
This PR has been subsumed by #4333

PR #4308, opened 2016-06-09
New package: tevent-0.9.28

the-maldridge:
Part of the sssd dependency chain


srcpkgs/tevent/template
```diff
@@ -0,0 +1,27 @@
...
+short_desc="Event system based on the talloc memory management library"
+maintainer="Michael Aldridge <aldridge.mac@gmail.com>"
+license="GPL-3"
+homepage="https://tevent.samba.org"
+distfiles="https://www.samba.org/ftp/tevent/tevent-${version}.tar.gz"
+checksum=04d953379025b1560af5c4ffcce58a3ee84db7aaa09c9f1e3eff5b2945a13529
+
+case "$XBPS_TARGET_MACHINE" in
+	armv[67]l) broken="https://s3.amazonaws.com/archive.travis-ci.org/jobs/136274420/log.txt";;
+	armv[67]l-musl) broken="https://api.travis-ci.org/jobs/136272275/log.txt?deansi=true";;
```
Vaelatern: Both?

the-maldridge: Yes, if you read the provided logs they fail for different reasons.

common/shlibs
```diff
@@ -2457,3 +2457,5 @@ libxlsreader.so.1 libxls-1.4.0_1
 libR.so R-3.3.0_2
 libmarkdown.so.2 discount-2.2.0_3
 libkeystone.so.0 keystone-0.9_1
+libtalloc.so.2 tevent-0.9.21_1
```
Vaelatern: Which is in a subdir of `/usr/lib` instead of in `/usr/lib`. Is it really meant to  be linked against, or present when used?

PR #4315, opened 2016-06-10
New package: corkscrew-2.0

pulux:


srcpkgs/corkscrew/template
```diff
@@ -0,0 +1,11 @@
+# Template file for 'corkscrew'
+pkgname=corkscrew
+version=2.0
+revision=1
+build_style=gnu-configure
+short_desc="Corkscrew is a tool for tunneling SSH through HTTP proxies"
+maintainer="pulux <pulux@pf4sh.de>"
+license="GPL-2"
+homepage="http://www.agroman.net/corkscrew/"
```
Gottox: That's a 404 page

pulux:
Hm, and why Travis and I hasn't a problem with that page?

ayghor:
> Hm, and why Travis and I hasn't a problem with that page?

Idk, but I get 404 too. Also, http://www.agroman.net/ gives "Welcome to nginx!".


-- 
Igor

srcpkgs/corkscrew/template
Vaelatern: Not any more.

PR #4318, opened 2016-06-10
gnutls: update to 3.4.13. (soname bump)

ebfe:
The only thing still broken is dwb (unrelated to the gnutls update). Last update was in 2014 so we might as well remove it? (cc @Gottox)


Gottox:
dwb can be dropped.

Gottox:
Before dropping, we should ask the developer for a new release. The last commits aren't that long ago: https://bitbucket.org/portix/dwb/commits/all

Gottox:
Asked the upstream dev for a new release.

ebfe:
@asergi fixed dwb already. travis failures are due to build order.

Gottox:
So we're ready to merge?

ebfe:
yes, but needs pushing in the right order.

Vaelatern:
Will it not build correctly if just ensuring the deps are in order?

ebfe:
No

Gottox:
@ebfe is it possible to make a progressive update like we do with libressl?

ebfe:
The issue is that the builder/xbps-src doesn't look at dependencies of dependencies when determining the build order. if a depends on b which depends on c. and both a and c depends on the bumbed libray, it still might try to build a before c which results in an "unresolved shlib" error when trying to install the c dependency since it always uses the newest version of the library. This is not specific to gnutls in any way, it's like that with every soname bump. The workaround is to push the rebuild commits in the correct order. Having a separate package for every soname might work around this too, but the real fix is to have the builder track dependencies more than one level deep when determining the build order.

Gottox:
Yea, but besides the build issue a progressive build would help to keep the repo consistent while rebuilding.

ebfe:
So gnutls is safe to use with 2 different versions in the same address space?

Gottox:
If A depends on B and both depend on gnutls, both must depend on the same version. But of A does not depend on B it's safe to use both.

ebfe:
That's the theory. And a recipe for headache with e.g. openssl.

ebfe:
If someone wants to implement/test this, feel free to adopt.

ebfe:
Rebased and bumped to 3.4.14 as packages start to depend on gnutls 3.4.x (filezilla).

ebfe:
#4806

ebfe:
de6d71b2df052ec6159f5887ab1a3d50937a7bd4

PR #4327, opened 2016-06-13
libcaca: update source and add x11 driver.

alxf:
Download source from official github repository tags and add libX11-devel as build depend for the x11 driver.

### DIFF
--- a/srcpkgs/libcaca/template
+++ b/srcpkgs/libcaca/template
@@ -1,16 +1,16 @@
 # Template file for 'libcaca'
 pkgname=libcaca
 version=0.99.beta19
-revision=6
+revision=7
 build_style=gnu-configure
 hostmakedepends="libtool automake pkg-config"
 short_desc="Graphics library that outputs text instead of pixels"
 maintainer="Juan RP <xtraeme@voidlinux.eu>"
-makedepends="ncurses-devel"
+makedepends="libX11-devel ncurses-devel"
 homepage="http://caca.zoy.org/wiki/libcaca"
 license="WTFPL"
-distfiles="http://caca.zoy.org/files/libcaca/libcaca-${version}.tar.gz"
-checksum=128b467c4ed03264c187405172a4e83049342cc8cc2f655f53a2d0ee9d3772f4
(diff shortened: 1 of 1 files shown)


Gottox:
Is the X11 driver useful for anyone? - I'm not sure if it just introduces an unused dependency. What do you guys think?

Vaelatern:
@alxf For what reason do you think `libX11-devel` provides a useful component for this package?

Gottox:
It's useful if you want performant caca rendering on X as most terminals are too slow. But I see no reason why we should have caca if there's graphical output anyway :)

ebfe:
Make it build option, default to off?

alxf:
Hi,

It was for performance reason, the rendering under the X11 driver is better. But indeed, it's not a real issue, if you mind to add this dep.

srcpkgs/libcaca/template
```diff
@@ -6,12 +6,16 @@ build_style=gnu-configure
...
-makedepends="libX11-devel ncurses-devel"
+makedepends="ncurses-devel $(vopt_if x11 libX11-devel)"
 homepage="http://caca.zoy.org/wiki/libcaca"
 license="WTFPL"
 distfiles="https://github.com/cacalabs/libcaca/archive/v${version}.tar.gz"
 checksum=7ed29a00cc7f017424d8b2994f001f137ed5bc4441987b711d78c6734fdf3493
 
+# package build options
+build_options="x11"
+desc_option_x11="Enable x11 driver"
```
Vaelatern: `Enable X11 driver`

Duncaen: https://github.com/voidlinux/void-packages/blob/master/common/options.description
`desc_option_x11="Enable support for X11"`

Gottox: This line can be removed, as @Duncaen stated, there's a global descrition defined.

alxf:
@Gottox 
That why i wanted to update the original source url, i'm not sure how you prefer to handle this diff, i can see with the project owner to make a new tag, and at this point only update source url with the real new release ?

Vaelatern:
Since we don't have epochs yet I don't think we have a clean way to deal with a new release without a new number. Upstream making a new tag would be best.

PR #4330, opened 2016-06-14
gcc: update to 5.4.0 (WIP)

xtraeme:


Gottox:
\o/
nice to have you back :)

chneukirchen:
I think we need `--with-default-libstdcxx-abi=gcc4-compatible` here, else we have to rebuild all C++ code?

pullmoll:
Perhaps we can have a `gcc5` package besides `gcc` so that everyone can test and see if their packages also compile with `gcc5` (using the compatibility switch)?

chneukirchen:
Also we need to consider https://llvm.org/bugs/show_bug.cgi?id=23529 and figure out if that patch is in a released LLVM then.

pullmoll:
Reading that thread and hardly understanding all the implications, I think the decision to make is whether or not to do what other distros seem to have done in using the default _GLIBCXX_USE_CXX11_ABI.

If I got that right it implies rebuilding every piece of c++ code in the repository or rather filling up a parallel gcc5 repository and thinking about how users can transition their systems from gcc4 to gcc5 in one go?

chneukirchen:
I think most distros chose for the rebuild.

Arch first migrated to 5.2, then rebuild. So you don't have this fall out and that fall out at once. May be a good idea.

chneukirchen:
FTR, LLVM 3.9 now has support for this ABI tag stuff. I'm preparing an update.

pullmoll:
The LFS site already describes bootstrapping with gcc-6.2. Should we perhaps skip the 5.x branch altogether and start with 6.2?

chneukirchen:
I'd have disagreed with 6.1, but 6.2 sounds viable.

pullmoll:
FWIW I managed to bootstrap gcc-6.2.0 for x86_64 and [here is how](http://misc.voidlinux.de/gcc-6.2.0-bootstrap.txt).
Instead of removing the `srcpkgs/gcc/patches/libcpp-source_date_epoch.patch` I guess I should fix it for gcc-6.2.0, though.

Gottox:
obsolete #4825

PR #4332, opened 2016-06-14
git: update to 2.9.0.

Vaelatern:


Vaelatern:
https://github.com/git/git/blob/v2.9.0/Documentation/RelNotes/2.9.0.txt

Gottox:
I can't see any changes that may break stuff from the repo. Second opinion?

xtraeme:
:+1:

PR #4333, opened 2016-06-14
New package: sssd-1.13.4

the-maldridge:


the-maldridge:
For reasons not entirely clear, travis fails this from platforms that have broken deps.  I'm not entirely sure why this happens but it looks right when built outside of travis.

srcpkgs/sssd/template
```diff
@@ -0,0 +1,27 @@
+# Template file for 'sssd'
+pkgname=sssd
+version=1.13.4
+revision=1
+build_style=gnu-configure
+configure_args="--with-selinux=no --with-semanage=no --disable-cifs-idmap-plugin --without-samba --disable-config-lib --without-python3-bindings"
+hostmakedepends="pkg-config nscd bind xmlcatmgr docbook-xsl"
+makedepends="pam-devel popt-devel talloc-devel tdb-devel tevent-devel ldb-devel ding-libs-devel libldap-devel pcre-devel mit-krb5-devel c-ares-devel glib-devel libxslt-devel python-devel nss-devel libnfsidmap-devel"
```
Vaelatern: Please break up these variables to be about 80 columns at most? The correct method is something like:
```
makedepends="pam-devel popt-devel talloc-devel tdb-devel tevent-devel ldb-devel
 ding-libs-devel libldap-devel mit-krb5-devel c-ares-devel glib-devel
 libxslt-devel python-devel nss-devel libnfsidmap-devel"
```
Where the space at the beginning of subsequent lines is there by convention. Also, `pcre-devel` is redundant here, `glib-devel` pulls it in.

the-maldridge: Oh yeah, that should probably happen.

the-maldridge:
Is this PR waiting on anything from me?  Its been established that its a possible bug in xbps-src that causes it to fail in Travis last I checked.

Vaelatern:
Merging soon with git-merge-pr. Fixing indentation errors. Fixing unresolved shlib errors. Removing redundant backslashes at end of lines.

Travis errors seem unrelated to actual issues.

PR #4339, opened 2016-06-16
New package: mp3gain-1.5.2

lemmi:


srcpkgs/mp3gain/template
```diff
@@ -0,0 +1,20 @@
+# Template file for 'mp3gain'
+pkgname=mp3gain
+version=1.5.2
+revision=1
+_rev=_r2
+wrksrc=${pkgname}-${version}${_rev}
```
Vaelatern: Best to actually have the revision information in the version variable.
```
_distver=1.5.2
_rev=r2
version=${_distver}_${_rev}
```
or something like that.

lemmi:
yeah, either way it's messy. you can't have _ in the version string, so need to work around that and the zip doesn't contain a directory.
also i'm a little bit opposed to adding more ".", just to replace them later on. makes it kind of opaque.

Vaelatern:
Doesn't need a `.`. Just needs a way that, when a new tarball is downloaded, for a version change to happen instead of a revision bump.

lemmi:
Let's give this a try then :)

srcpkgs/mp3gain/template
```diff
@@ -0,0 +1,21 @@
+# Template file for 'mp3gain'
+pkgname=mp3gain
+_distver=1.5.2
+_rev=r2
+version=${_distver}.${_rev}
+revision=1
+wrksrc=${pkgname}-${version}
```
Vaelatern: Line no longer needed

PR #4356, opened 2016-06-21
apache and apr-util cross

Gottox:
fixes apr-util and apache cross compilation.


srcpkgs/apache/template
```diff
@@ -45,6 +45,27 @@ pre_configure() {
 		-i docs/conf/httpd.conf.in
 
 	cat ${FILESDIR}/xbps.layout >> config.layout
+
+	if [ "$CROSS_BUILD" ]; then
+		configure_args+=" --with-apr=${XBPS_WRAPPERDIR}/apr-1-config
+			 --with-pcre=${XBPS_WRAPPER_DIR}/pcre-config"
```
xtraeme: this should be unnecessary, XBPS_WRAPPERDIR appears in PATH before /usr/bin.

srcpkgs/apr-util/template
```diff
@@ -19,7 +18,17 @@ license="Apache-2.0"
...
 checksum=a6cf327189ca0df2fb9d5633d7326c460fe2b61684745fd7963e79a6dd0dc82e
 
-nocross="http://build.voidlinux.eu/builders/armv7l_builder/builds/7623/steps/shell_4/logs/stdio"
+pre_configure() {
+	if [ "$CROSS_BUILD" ]; then
+		configure_args+=" --with-apr=${XBPS_WRAPPERDIR}/apr-1-config"
+	else
+		configure_args+=" --with-apr=/usr/bin/apr-1-config"
+	fi
+}
```
xtraeme: This is again unnecessary, apr-1-config in XBPS_WRAPPERDIR should be found via PATH automagically.

Gottox: without it:
```
configure: error: APR could not be located. Please use the --with-apr option.
```
The build system is broken.

srcpkgs/apache/template
Gottox: doesn't work. the buildsystem is broken.

asergi:
@pullmoll added cross support to `apr-util` and `apache` with 3dcf63b209d6c33f40eec481c41e1220d773d574 and aad6395993278b7227418ac1f76146c04de127af.

PR #4363, opened 2016-06-24
New package: trojita-0.7

jannisch:
This package also provides `libtrojita_plugins.so`, should I include it in `common/shlibs`? I currently don't see any use of it.

### DIFF
--- a/srcpkgs/trojita/template
+++ b/srcpkgs/trojita/template
@@ -1,0 +1,19 @@
+# Template file for 'trojita'
+pkgname=trojita
+version=0.7
+revision=1
+build_style=cmake
+makedepends="qt5-devel qt5-plugin-mysql qt5-plugin-pgsql qt5-plugin-sqlite
+ qt5-plugin-tds qt5-plugin-odbc qt5-webkit-devel qt5-tools-devel qt5-svg-devel
+ zlib-devel"
+depends="qt5-svg qt5-plugin-sqlite"
+short_desc="A fast Qt IMAP/SMPTP client"
+maintainer="Jannis Christ <hello@jannis.ovh>"
+license="GPL-2,GPL-3"
+homepage="http://trojita.flaska.net/"
+distfiles="${SOURCEFORGE_SITE}/$pkgname/$pkgname-$version.tar.xz"
+checksum=e08da881721cafa00ecd772eaef6b99fd37d068cb703eeb532de0ed1ea0136d9
+
+post_install() {
+	vlicense LICENSE
+}


pullmoll:
As long as there is no `libtrojita-devel` package which would rely on finding a `*.so.<number>` file in `common/shlibs` to set up its shlib dependencies, there is no need to add shared libraries like plugins to `common/shlibs`.

jannisch:
Thanks for the explanation.

srcpkgs/trojita/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'trojita'
+pkgname=trojita
+version=0.7
+revision=1
+build_style=cmake
+hostmakedepends="cmake"
```
Vaelatern: Line not necessary.

srcpkgs/trojita/template
```diff
@@ -0,0 +1,19 @@
...
+makedepends="qt5-devel qt5-plugin-mysql qt5-plugin-pgsql qt5-plugin-sqlite
+ qt5-plugin-tds qt5-plugin-odbc qt5-webkit-devel qt5-tools-devel qt5-svg-devel
+ zlib-devel"
+short_desc="A fast Qt IMAP/SMPTP client"
+maintainer="Jannis Christ <hello@jannis.ovh>"
+license="GPL-2/GPL-3"
```
Vaelatern: Separate with `, ` or just list `GPL-2`.

Duncaen:
And maybe add a `INSTALL.msg` that one of the database plugins is needed?

jannisch:
Do you mean sqlite for example? Trojita has got `qt5-webkit` as a dependency and this package already depends on `sqlite`.

Duncaen:
No, the `qt-plugin` ones like `qt5-plugin-sqlite`.

jannisch:
Oh sorry, I see. Do you wish that I add `qt5-plugin-sqlite` as a dependency in order to provide a default? I think experienced people could still force the installation of a different plugin but I don't want to break a philosophy or something similar.

jannisch:
I just added `qt5-svg` in the meantime which is also required (for icons) but not automatically added as dependency.

pullmoll:
IMO installing `qt5-plugin-sqlite` as a default or fallback doesn't hurt. It is small and `sqlite` itself will already be installed on many systems.

jannisch:
Yes, it requires just 17KB and as I mentioned earlier, `sqlite` is already an indirect dependency.

PR #4365, opened 2016-06-24
nvidia: update to 367.27.

Duncaen:
@xtraeme is this `common/shlibs` change correct, should i update the updater libs too?


Duncaen:
Hm I think this update broke steam for me.

aeosynth:
any updates? can i help?

PR #4379, opened 2016-06-27
New package: hxtools

Nexolight:
This is my first package ever for any linux distro and I hope I got the manual right - This one works for me as intendet


ayghor:
You should probably rebase your work on top of master instead of merging: dba8e8c.
Squash teh commits into a single "New package: hxtools" with `git rebase -i`.
Revision should be 1.

Vaelatern:
https://sourceware.org/bugzilla/show_bug.cgi?id=14553 this is why it doesn't work on musl. Will look in to ways to fix it.

The rule within Void is that for every template altered, there should be one commit. Those commits should be named as directed by the end of the [Manual (link)](https://github.com/voidlinux/void-packages/blob/master/Manual.md#contributing-via-git). Until a template is merged into the distrobution, the revision should be 1.

Nexolight:
I excluded the musl builds. we can include them again when the bug is somehow fixed.

srcpkgs/hxtools/template
```diff
@@ -0,0 +1,13 @@
...
+build_style=gnu-configure
+short_desc="Hxtools collection. See http://inai.de/projects/hxtools/"
+maintainer="nexolight <snow.dream.ch@gmail.com>" #package
+license="WTFPL-2.0+, LGPL-2.1/LGPL-3.0, GPL-2.0+"
+homepage="http://inai.de/projects/hxtools"
+distfiles="http://jftp.inai.de/hxtools/hxtools-20150304.tar.xz"
+checksum="55265b16191831c4e332b0235b057e3f1e952298abf34734a8fa2ee9bb7be382"
+hostmakedepends="pkg-config"
+makedepends="libHX-devel"
+only_for_archs="x86_64 aarch64 i686 armv5te armv5tel armv6l armv6hl armv7l armv7hl"
```
Vaelatern: Don't do this. See `srcpkgs/yoshimi/template` for a better example of how to exclude musl.

Nexolight:
Indeed it's in the documentation as well - sry didn't saw that. I will change it.

srcpkgs/hxtools/template
```diff
@@ -0,0 +1,15 @@
+pkgname=hxtools
+version=20150304
+revision=1
+build_style=gnu-configure
+short_desc="Hxtools collection. See http://inai.de/projects/hxtools/"
```
Vaelatern: Please have a better description. `xbps-query -RS hxtools` will show you the homepage if people are curious. This is to give them a description of the package.

Nexolight: Well there's no description even not from the developer "a colllection of tools over the years" but well...  I needed this package because  of  ofl  which is  used by pam_mount.

I will change it. But apart from that is everything right?

srcpkgs/hxtools/template
```diff
@@ -0,0 +1,15 @@
...
+maintainer="nexolight <snow.dream.ch@gmail.com>" #package
```
Vaelatern: # package
?

srcpkgs/hxtools/template
```diff
@@ -0,0 +1,15 @@
+pkgname=hxtools
```
Vaelatern: First line must be a comment. See most template files for an example.

srcpkgs/hxtools/template
Vaelatern: @the-maldridge Can you confirm that pam_mount needs ofl?

@Nexolight Seems right.

the-maldridge:
I am successfully using pam_mount without the tools in this package on a test network.  That being said pam_mount is very configurable and can shell out to other tools to perform advanced actions.  Perhaps there is a use case here that is not entirely obvious.

Nexolight:
I can confirm it as soon as I'm in my office... Theres a comment on the logout tag in the config of pam_mount which sais requires ofl from hxtools in order to work. It doesn't really work even with it or it's blocked by ongoing processes but anyway the ofl tool itself does work.

It's a package it doesn't need any usecase I'm pretry sure the tools are useful for at least someone. In my case it was the unmount on logout functionality in pam_mount.

srcpkgs/hxtools/template
Nexolight: I'm not the dev

srcpkgs/hxtools/template
Nexolight: didn't knew it's a requirement since the first comment was almost the same in each package and basically said template for package. where the file is named template in a srcpkg folder with packages in it.

Vaelatern:
Please use the variable `${version}` wherever you use the version in the template. Obvious place: The distfile.

srcpkgs/hxtools/template
```diff
@@ -0,0 +1,16 @@
+#Template file for 'hxtools'
+pkgname=hxtools
+version=20150304
+revision=1
+build_style=gnu-configure
+short_desc="Collection of small tools over the years by j.eng (inai.de)"
+maintainer="nexolight <snow.dream.ch@gmail.com>"
+license="WTFPL-2.0+, LGPL-2.1/LGPL-3.0, GPL-2.0+"
+homepage="http://inai.de/projects/hxtools"
+distfiles="http://jftp.inai.de/hxtools/hxtools-20150304.tar.xz"
```
Vaelatern: Basically `${version}` here

Vaelatern:
Merged with a patch to fix musl and appropiate variable substitution.

the-maldridge:
It seems I was incorrect, this is needed for additional functionality within pam_mount that was not in use at the time of my previous comment.  I will consider opening another issue to modify the depends on pam_mount to include hxtools as a lack of the programs included by hxtools can cause unexpected failures which have no descriptive emails.

Vaelatern:
5d76dca

Issue #3909, opened 2016-03-26
Third party package managers should install their packages to /usr/local


Gottox:
Currently tools like `npm -g`, `pip`, or `gem` polute the `/usr`-namespace which should be exclusively used by void. We should change those package managers to install packages to /usr/local and make sure, that python, node, ruby, etc still search for modules in `/usr` and `/usr/local` (the latter be the prefered one)

* [ ] gem
* [ ] npm
* [ ] pip
* maybe more?

biopsin:
dpkg if it's sane

Vaelatern:
Probably:
[ ] ocaml-findlib
[ ] opam

chneukirchen:
opam installs to ~ by default.

Note that ruby uses rubygems internally to find some libs, so this would need to look into both paths (not sure that's supported).

ebfe:
cargo too installs to ~ unless told otherwise.

chneukirchen:
So does gem and pip (and i think npm) when not run as root.

Perhaps expecting users to be smart enough not to run these as root is enough?

Vaelatern:
Some of these package managers have issues sticking to the home directory.

Duncaen:
For python27 the following works
`echo "/usr/local/lib/python2.7/site-packages/" > /usr/lib/python2.7/site-packages/usrlocal.pth`
`echo "[global]\nprefix = /usr/local" > /etc/pip.conf`

Vaelatern:
Is `/etc/pip.conf` needed?

PR #4381, opened 2016-06-27
initial collection of Ruby gems - basic Rails framework

bmjoan:


chneukirchen:
How do you intent to keep the different version requirements consistent?

I think a manual installation of rails using gem is easier...

ayghor:
+1 for manual gem installation.

Anyway, why package ruby gems if we already have them via `gem install`? IMHO this is duplication of already present functionality.

bmjoan:
@chneukirchen What do you mean by *different versions requirements* and how is that different to other sw like C libraries ? 

Versions in templates comply to specific requirements of a main/popular gem, rails meta package in this case. It is a starting point make such software available by standard packaging system and many popular Linux distributions like Debian, Ubuntu, Fedora and Arch Linux (in AUR) do offer rubygems in repo packages.

Btw. system-wide `gem install` is a bad practice, as it would clash with xbps. It may overwrite existing files, installed files are not tracked, operations like `xbps-query -o` won't work etc.

bmjoan:
@ayghor No, this is no duplication. `gem install` is ortogonal to `xbps-install`. See post above. Why Void even offers python PyPi packages ? They may also be installed with `pip install`.

ayghor:
> Debian, Ubuntu, Fedora and Arch Linux (in AUR) do offer rubygems in repo packages.

I can't say for Arch, but Debian and Ubuntu packages for ruby and its gems are really outdated and never ever served me in any way.

http://packages.ubuntu.com/search?keywords=ruby
https://packages.debian.org/search?keywords=ruby

Matter of fact: it is a real pain to keep Rails and its deps up to date. And any update on teh system packages will require teh user to `bundle install` locally their Gemfile locked versions.

> Btw. system-wide gem install is a bad practice, as it would clash with xbps.

We have `/usr/local` for that, no?

> Why Void even offers python PyPi packages ? They may also be installed with pip install.

Idk. I would too appreciate enlightenment on this.

Vaelatern:
>> Why Void even offers python PyPi packages ? They may also be installed with pip install.
> Idk. I would too appreciate enlightenment on this.

We only include PyPi packages where they meet certain conditions, which approximately (this list is just off the top of my head) are:

1. Must be compiled to be used
2. Required component for another program, either at compile time or at run time.
3. Is a tool independent of python modules, such as python-ipython.

Vaelatern:
And about `/usr/local`, see https://github.com/voidlinux/void-packages/issues/3909

bmjoan:
@ayghor 
> I can't say for Arch, but Debian and Ubuntu packages for ruby and its gems are really outdated and never ever served me in any way.
>
> http://packages.ubuntu.com/search?keywords=ruby
> https://packages.debian.org/search?keywords=ruby

What specific *outdated* gems had you actually on mind ? Rails gems are pretty up-to-date in fact, as you can see for [Ubuntu 16.04 LTS](http://packages.ubuntu.com/xenial/ruby-rails) and [Debian Testing](https://packages.debian.org/stretch/ruby-rails).

> Matter of fact: it is a real pain to keep Rails and its deps up to date. And any update on teh system packages will require teh user to bundle install locally their Gemfile locked versions.

What are you talking about ? If you have some specific project requiring locked down versions of some gems, then keep it updated in your bundler environment.
But if we speak about default recent latest stable versions for repository packaging purposes, it is quite simple. Single dedicated environment without version limits and after a `bundle update` you will have most recent versions at hand. Just update numbers in templates accordingly, run `xgensum -i` and you are done. May be automated with cron job or CI build system. Actually a bit simpler then classic C libraries as `gem` handles dependencies (including new ones) for you automatically.

bmjoan:
@Vaelatern 
> We only include PyPi packages where they meet certain conditions, which approximately (this list is just off the top of my head) are any of:
>
>   1. Must be compiled to be used
>   2. Required component for another program, either at compile time or at run time.
>   3. Is a tool independent of python modules, such as python-ipython.

I'm afraid many packages actually won't meet above criteria, like

`python-Babel, python-BeautifulSoup4, python-CherryPy, python-Django, python-Flask, python-Jinja2, python-Mako, python-Markown, python-MarkupSafe, python-Sphinx, python-Werkzeug, python-twitter` and possibly others.

I can't see any real difference here, only the target language. Some modules are for Python and some for Ruby.

ayghor:
Perhaps those are dependencies to some other that meets teh criteria?

ayghor:
Anyway, if it is teh case that those packages were accepted without good reason, it should not be an excuse for others. Instead, they should be reviewed and removed if appropriate.

bmjoan:
@ayghor I'm sorry but still miss what are your claims based upon. Also called your previous responses into a question, but didn't get a single straight answer with arguments.

ayghor:
Whoops I missed your post! In fact, I was wrong about Ubuntu!

What I'm saying that I just see maintaining Rails on void as a waste of time. It is already packaged into `gem` and people don't need `rails` unless they are developing an app.

Vaelatern:
@ayghor People need rails if they are developing an app. People only need swift if they are developing an app, yet we do package it.

ayghor:
> @ayghor People need rails if they are developing an app. People only need swift if they are developing an app, yet we do package it.

Yea we could say teh same for pretty much any developer tool, of which many should obviously be. My bad, sry.

But we put Rails in, why not other popular gems? And if we include other popular gems, how popular should they be? At least a nice guideline is necessary, not "it was done in another package" or "it was done in another distro", as their maintainers may have taken poor decisions.

Also, Rails is not only used to develop an app, it is part of teh stuff, and developers specify in their own `Gemfile` which version they want, which gets locked into `Gemfile.lock` for a good reason. If you include Rails into void, which version(s) will be? Latest? Who are you supporting with this? If you update teh package, you break me =o(.

There are programs which suffer from version problems, but are manageable. Void has `python` and `python3.4`. If old `python` gets updated, to a _compatible_ version, stuff won't (shouldn't) break. However, even if you only update Rails to _compatible_ versions, apps often will, because of their strict specs. Ruby suffers from this...

You may say teh developer should `bundle update` their apps, but this is often not possible. Or that they can `bundle install` their local copy --- but then what is teh point?

Don't get mad at me plz, I'm just sharing my experience that packaged Rails is complicated and rarely useful. You may package some Rails version into void, I just wonder if it is worth.

<3

ayghor:
Sorry for being so `-v`. =o(

thypon:
I completely lost this PR, sorry: cherrypy and a lot of the python packages you told are calibre dependencies, an end user program.

I'd also like not to include devel dependencies when not needed by end user programs (not devel tools), but your work is fantastic.

PR #4382, opened 2016-06-27
Automate packaging of Ruby gems for XBPS !

bmjoan:


Gottox:
Personally I find this very useful. But unfortunately the general direction of void is to let tools like gem, pip, npm, etc do it's job and not get in their way.

@xtraeme, what do you say?

ayghor:
Perhaps teh discussion around #4381 is meaningful to this subject?

srcpkgs/gem2xbps/template
```diff
@@ -0,0 +1,20 @@
+# Template file for 'gem2xbps'
+pkgname=gem2xbps
+version=0.1
+revision=1
+short_desc="Tool to generate XBPS template files for Ruby gems packaging"
+maintainer="JoanBM <no.public@email.sorry>"
```
DirectorX: github providing email addresses masking, your masked email address is `bmjoan@users.noreply.github.com`. it's very important to put in an email address so it will be possible to contact you.

the-maldridge: Our stated package policy requires publicly reachable email addresses, the noreplay.github.com addresses are not acceptable.

PR #4383, opened 2016-06-27
nginx: update to 1.10.1.

Vaelatern:
still not happy on arm.


Gottox:
#yolo

Duncaen:
In `${FILESDIR}` is a pregenerated header file for `arm*` which has to be synced.
Fixed it in 0bfa499 and 04f7ebe.

Vaelatern:
Where does it get synchronized from?

Duncaen:
`./xbps-src configure nginx` and then `vimdiff srcpkgs/nginx/files/ngx_auto_config.h.armv6l masterdir/builddir/nginx-$version/objs/ngx_auto_config.h`.

Gottox:
cool, can you document this in the template?

Vaelatern:
Just did

PR #4384, opened 2016-06-28
rclone: update to 1.30

diogoleal:


srcpkgs/rclone/template
```diff
@@ -1,6 +1,6 @@
 # Template file for 'rclone'
 pkgname=rclone
-version=1.29
+version=1.30
 revision=2
```
Vaelatern: `=1`

diogoleal: Sorry. Fixed!

PR #4386, opened 2016-06-28
glusterfs: update to 3.5.9

pulux:
GlusterFS 3.8 is released, the 3.5 version will become unmaintained upstream


pulux:
upstream is on 3.8.0

PR #4393, opened 2016-06-28
musl: fix failure to obtain EOWNERDEAD status for process-shared robust mutexes

### LINKED ISSUE #4392: Need a musl patch which fixes pthread_exit code to match kernel behavior
Please apply a patch to musl-libc, which fixes the pthread_error code issue.

musl-libc commit: http://git.musl-libc.org/cgit/musl/commit/?id=384d103d94dba0472a587861f67d7ed6e8955f86
(refers to bug in this doc: https://www.kernel.org/doc/Documentation/robust-futex-ABI.txt)

Alpine port: https://github.com/alpinelinux/aports/commit/8b4bc947a74e8ca71490014a622ae0d8e4e72bcc

Original issue: dotnet/coreclr#5872

Gottox:
Added upstream patch:
>Linux's documentation (robust-futex-ABI.txt) claims that, when a
>process dies with a futex on the robust list, bit 30 (0x40000000) is
>set to indicate the status. however, what actually happens is that
>bits 0-30 are replaced with the value 0x40000000, i.e. bits 0-29
>(containing the old owner tid) are cleared at the same time bit 30 is
>set.
>
>our userspace-side code for robust mutexes was written based on that
>documentation, assuming that kernel would never produce a futex value
>of 0x40000000, since the low (owner) bits would always be non-zero.
>commit d338b506e39b1e2c68366b12be90704c635602ce introduced this
>assumption explicitly while fixing another bug in how non-recoverable
>status for robust mutexes was tracked. presumably the tests conducted
>at that time only checked non-process-shared robust mutexes, which are
>handled in pthread_exit (which implemented the documented kernel
>protocol, not the actual one) rather than by the kernel.
>
>change pthread_exit robust list processing to match the kernel
>behavior, clearing bits 0-29 while setting bit 30, and use the value
>0x7fffffff instead of 0x40000000 to encode non-recoverable status. the
>choice of value here is arbitrary; any value with at least one of bits
>0-29 set should work just as well,

fixes #4392

### DIFF
--- a/srcpkgs/musl/template
+++ b/srcpkgs/musl/template
@@ -1,7 +1,7 @@
 # Template file for 'musl'.
 pkgname=musl
 version=1.1.14
-revision=1
+revision=2
 build_style=gnu-configure
 configure_args="--prefix=/usr --disable-gcc-wrapper"
 conflicts="glibc>=0"
--- a/srcpkgs/musl-bootstrap/template
+++ b/srcpkgs/musl-bootstrap/template
@@ -1,7 +1,7 @@
 # Template file for 'musl-bootstrap'.
 pkgname=musl-bootstrap
 version=1.1.14
-revision=3
+revision=4
 lib32disabled=yes
 wrksrc="musl-${version}"
 build_style=configure
--- a/srcpkgs/musl/patches/fix-failure-to-obtain-EOWNERDEAD-status-for-process-shared-robust-mutexes.patch
+++ b/srcpkgs/musl/patches/fix-failure-to-obtain-EOWNERDEAD-status-for-process-shared-robust-mutexes.patch
@@ -1,0 +1,76 @@
+From 384d103d94dba0472a587861f67d7ed6e8955f86 Mon Sep 17 00:00:00 2001
+From: Rich Felker <dalias@aerifal.cx>
+Date: Mon, 27 Jun 2016 15:18:13 -0400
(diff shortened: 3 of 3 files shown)


srcpkgs/musl/template
```diff
@@ -1,7 +1,7 @@
 # Template file for 'musl'.
 pkgname=musl
 version=1.1.14
-revision=1
+revision=2
 build_style=gnu-configure
 configure_args="--prefix=/usr --disable-gcc-wrapper"
 conflicts="glibc>=0"
```
jasonwilliams200OK: I think we need `vlicense LICENSE` before line 45 to make the linter happy. :smile:

jasonwilliams200OK:
Thanks @Gottox! :tada:

srcpkgs/musl/template
Gottox: the linter is wrong here. cc @chneukirchen

chneukirchen: ... the linter is not wrong. You are not installing a license?

Gottox: see line 45

Gottox:
We'll fix linting later.

srcpkgs/musl/template
chneukirchen: The linter complains about musl-bootstrap, not musl.

PR #4394, opened 2016-06-28
glusterfs: update to 3.8.0

pulux:


srcpkgs/glusterfs/template
```diff
@@ -17,11 +17,9 @@ maintainer="Juan RP <xtraeme@voidlinux.eu>"
 license="GPL-2, LGPL-3"
 homepage="http://www.gluster.org/"
 distfiles="http://download.gluster.org/pub/gluster/glusterfs/${version%.*}/${version}/${pkgname}-${version}.tar.gz"
-checksum=ef58b467a60fd6bd0685d4167a45f0860895414285b3908167b264aa63d276c3
+checksum=8a2d89a37bf7c99bc153826dbeaf2113a7e45c77bc6978a2850c9033ddd77511
+only_for_archs="i686 x86_64"
```
Vaelatern: Reasoning?

pulux: It's the current upstream Version.

"""
A reminder for all users of the 3.5 stable series; when GlusterFS 3.8 is released, the 3.5 version will become unmaintained. We do our best to maintain three versions of Gluster, with the 3.8 release it will be 3.8, 3.7 and 3.6. Users still running version 3.5 are highly encouraged to start planning their upgrade process. If there are no critical problems reported against the 3.5 version, and no patches get sent, 3.5.9 might well be the last 3.5 release.
"""

http://blog.gluster.org/2016/04/glusterfs-3-5-9-is-available-would-it-be-the-last-3-5-release/

Vaelatern:
I meant for the arch limitations. Why not try to build for all our platforms?

srcpkgs/glusterfs/template
Vaelatern: I meant for the arch limitations. Why be so limiting?

pulux: I'm not able to fix the musl issue, configure failed with: `rpcgen not found` and tho old version is't also not on mus, may you can help to fix it.

Vaelatern: Then you may mark it broken on musl. `only_for_archs` is often wrong.

srcpkgs/glusterfs/template
```diff
@@ -17,11 +17,9 @@ maintainer="Juan RP <xtraeme@voidlinux.eu>"
...
 
-pre_configure() {
-	./autogen.sh
-}
```
Vaelatern: Why did you remove this configuration script?

pulux: upstream says:
```
0. If you have cloned from git, run ./autogen.sh.

1. Run ./configure.
```
https://github.com/gluster/glusterfs/blob/release-3.8/INSTALL

PR #4407, opened 2016-07-01
New package: tag-1.2.0

thypon:


srcpkgs/tag/template
```diff
@@ -0,0 +1,18 @@
+# Template file for 'tag'
+pkgname=tag
+version=1.2.0
+revision=1
+build_style=go
+hostmakedepends="git-perl mercurial"
```
Vaelatern: `git-submodule` no longer needs perl, are you still sure you want `git-perl`?

PR #4408, opened 2016-07-02
New package: jpegoptim-1.4.3

archaeme:


srcpkgs/jpegoptim/template
```diff
@@ -0,0 +1,15 @@
+# Template file for 'jpegoptim'
+pkgname=jpegoptim
+version=1.4.3
+revision=1
+wrksrc=${pkgname}-RELEASE.${version}
+build_style=gnu-configure
+hostmakedepends="libjpeg-turbo-devel"
+makedepends="${hostmakedepends}"
```
Duncaen: `makedepends="libjpeg-turbo-devel"` is enough, you can remove `hostmakedepends` and `depends.`

`xbps-src` detects runtime dependencies itself, see https://travis-ci.org/voidlinux/void-packages/jobs/141785251#L585

PR #4409, opened 2016-07-02
New package: lily-0.16

diogoleal:


srcpkgs/lily/template
```diff
@@ -0,0 +1,16 @@
+# Template file for 'lily'
+pkgname=lily
+version=0.16
+revision=1
+build_style=cmake
+hostmakedepends="cmake"
+short_desc="An interpreted language with a focus on expressiveness and type safety"
+maintainer="Diogo Leal <diogo@diogoleal.com>"
+license="CUSTOM"
```
Duncaen: `MIT`

diogoleal: Changed

srcpkgs/lily/template
```diff
@@ -0,0 +1,16 @@
+# Template file for 'lily'
+pkgname=lily
+version=0.16
+revision=1
+build_style=cmake
+hostmakedepends="cmake"
```
ebfe: could be removed, the build_style already provides the host dependency on cmake now.

diogoleal: Removed.

PR #4419, opened 2016-07-03
New package: skk-jisyo

rail44:
SKK is system of Japanese input method.
This package provides dictonaries to various input method using SKK.


ebfe:
#4420 was merged, #4421 was closed, what's the state on this?

rail44:
Sorry but I cannot found what make #4421 failuare on ARM platform.

PR #4420, opened 2016-07-03
New package: libskk

rail44:
SKK is system of Japanese input method.
This package is used by input method such as ibus-skk or fcitx-skk.

PR #4421, opened 2016-07-03
New package: ibus-skk

rail44:
depends on: #4419 #4420 

SKK is system of Japanese input method.
This package is frontend of SKK with ibus.

### DIFF
--- a/srcpkgs/okular/template
+++ b/srcpkgs/okular/template
@@ -1,7 +1,7 @@
 # Template file for 'okular'
 pkgname=okular
 version=4.14.3
-revision=2
+revision=3
 build_style=cmake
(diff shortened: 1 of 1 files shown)


Gottox:
Thanks for this PR!

As you see, our CI has failed with your package. Please join the depending PRs so we can actually see if the commits build.

[See *Starting a pull request* in CONTRIBUTING.md](https://github.com/Gottox/void-packages/blob/master/CONTRIBUTING.md#starting-a-pull-request):
> Most pull request should only contain a single package and its dependencies which are not part of void-packages yet.

In general PR should be as minimal as possible. But we cannot accept PRs that can't pass any platform on our CI. Exceptions are long running build processes that are tested by hand.

PR #4422, opened 2016-07-03
New package: tinycron-0.2

diogoleal:


srcpkgs/tinycron/template
```diff
@@ -0,0 +1,18 @@
+# Template file for 'tinycron'
+pkgname=tinycron
+version=0.2
+revision=1
+only_for_archs="i686 x86_64"
```
Vaelatern: Can you confirm why these are the case? Since we are building it, I'd prefer `broken` statements with links to failed travis builds, and a case statement as in the manual or other templates.

srcpkgs/tinycron/template
```diff
@@ -0,0 +1,18 @@
...
+build_style=go
+go_import_path="github.com/bcicen/tinycron"
+hostmakedepends="git-perl"
```
ebfe: If this is for `git submodule` you can drop it. It's provided by the `git` package now.

diogoleal: Ok, I changed for ```hostmakedepends="git"```

srcpkgs/tinycron/template
diogoleal: I removed this variable.

Gottox:
@ebfe @Vaelatern: is this PR ready?

PR #4426, opened 2016-07-04
uthash: update to 2.0.0

diogoleal:


srcpkgs/uthash/template
```diff
@@ -1,23 +1,27 @@
...
 do_install() {
 	vmkdir usr/include
+	vcopy src/ringbuf.h usr/include
+	vcopy src/utmm.h usr/include
+	vcopy src/utvector.h usr/include
+	vcopy src/libut.h usr/include
 	vcopy src/utarray.h usr/include
 	vcopy src/uthash.h usr/include
 	vcopy src/utlist.h usr/include
-	vcopy src/utstring.h usr/include
```
Vaelatern: Can you refactor this part into two loops over the relevant file names, one with `src` and one with `doc`?

diogoleal: Done. I made loops of files.

srcpkgs/uthash/template
```diff
@@ -1,23 +1,25 @@
...
-	vcopy src/utlist.h usr/include
-	vcopy src/utstring.h usr/include
-	vdoc doc/userguide.txt
-	vdoc doc/utarray.txt
-	vdoc doc/utlist.txt
-	vdoc doc/utstring.txt
-	vlicense LICENSE
+
+	for files in src; do
+		vcopy src/$file usr/include
```
Duncaen: `files != file` and `src/*`

srcpkgs/uthash/template
```diff
@@ -1,23 +1,25 @@
...
+	done
+
+	cd doc/
+	for docs in *.txt; do
+		vdoc $docs
+	done
```
Vaelatern: So make these loops, if they are actually supposed to be every file, just two lines:
```
vcopy src/* usr/include
vdoc doc/*.txt
```
and this will be ready for merging.

diogoleal:
When I use vdoc doc/*.txt, ocorreus this error:
```
=> ERROR: uthash-2.0.0_1: vinstall: cannot find 'doc/*.txt'...
=> ERROR: uthash-2.0.0_1: failed to run do_install() at line 140.
```
This is motive is in commit.

```for files in src; do
              vcopy src/$file usr/include
done```

PR #4432, opened 2016-07-05
fish: update to 2.3.1

aeosynth:
2.3.1 is not available from fishshell.com; use the release page instead

### DIFF
--- a/srcpkgs/fish-shell/template
+++ b/srcpkgs/fish-shell/template
@@ -1,6 +1,6 @@
 # Template file for 'fish-shell'
 pkgname=fish-shell
-version=2.3.0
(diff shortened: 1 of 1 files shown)


Vaelatern:
Any reason they no longer use fishshell.com?

aeosynth:
i assume it's just easier to let github keep track of releases. the arch [pkgbuild](https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/fish) uses the github release. i can change the url to the `archive` url, i'm not sure if there's a difference.

Vaelatern:
Looks good to me, since it does seem they use the git page now.

PR #4435, opened 2016-07-05
openssh: enable kerberos5 support

the-maldridge:


Gottox:
Do we want kerberos in base? cc @voidlinux/pkg-committers

the-maldridge:
From conversations with @bougyman there is desire to manage the servers a bit better than "here are some servers".  Adding kerberos into ssh is one of the technologies that would make that substantially easier, though I suppose we could do a custom build with a vopt for void infrastructure if it is deemed too heavy for base.

Seems kind of silly though that void infrastructure wouldn't be using mainline void packages though.

Vaelatern:
It's about 10 megabtytes for mit-krb5 + dependencies, so it's not oppressive. It would not be useful to a live CD, but it would be nice to have it present in a package for a developer who wants to change login methods.

I support it so long as it's a build option.

Gottox:
We could switch from openssh to dropbear in base.

chneukirchen:
Please keep openssh in base.

These issues are mainly build issues, not that base-system is too big.

Vaelatern:
Against dropbear in base.

the-maldridge:
I also am weary of anything that isn't openssh.  The others just don't feel "battle tested" enough.

PR #4437, opened 2016-07-06
New package: zerotier-one-1.1.6

the-maldridge:
zerotier is a novel p2p networking solution.


Vaelatern:
You chose the name `zerotier` instead of `zerotierone`, what was your reasoning?

the-maldridge:
From the discussion in IRC, it was zerotier when I first started using it.  Now it seems they call the upstream package zerotier-one.

Vaelatern:
Separately from packaging interests, will you be trying to upstream the "patches" necessary here?

the-maldridge:
see zerotier/ZeroTierOne#348

Vaelatern:
Perhaps a patch to allow installation to `/usr/bin` instead of insisting on `/usr/sbin` would be accepted upstream?

srcpkgs/zerotier-one/template
```diff
@@ -0,0 +1,35 @@
...
+	sed -i 's:#include <sys/sysctl.h>::' ./attic/OSXEthernetTap.cpp.pcap-with-bridge-test
+	sed -i 's:#include <sys/sysctl.h>::' ./attic/OSXEthernetTap.cpp.utun-work-in-progress
+	sed -i 's:#include <sys/sysctl.h>::' ./ext/libnatpmp/getgateway.c
+	sed -i 's:#include <sys/sysctl.h>::' ./osdep/BSDEthernetTap.cpp
+	sed -i 's:#include <sys/sysctl.h>::' ./osdep/ManagedRoute.cpp
+	sed -i 's:#include <sys/sysctl.h>::' ./osdep/OSXEthernetTap.cpp
+}
+
+do_build() {
+	make all ZT_\ENABLE_\NETWORK_\CONTROLLER=1
```
Vaelatern: Why all the backslashes?

the-maldridge: They are specified as part of the token in the documentation.  Its probably safe to remove them, but the token is specified in the readme with the slashes in place.

Vaelatern: Unable to see the token at all on the README on github. Can you provide a source?

Vaelatern: Source found.

PR #4440, opened 2016-07-06
New package: plymouth-0.8.8

WilliamO7:
This is a bootsplash application (used in most distros), which works almost surprisingly well on void (probably due to it's great integration with dracut, or whatever.)

It works great in my virtualbox, have not tried on real hardware yet (probably will do that tonight.)

The user would have to enable the functionality his/herself (just making a plymouth.conf in /etc/dracut.conf.d with "add_dracutmodules+="plymouth" is sufficient.)

A transparent distribution logo is included, to add to the overall aesthetic experience, and because I couldn't find it in the void-artwork package (although it could easily be added later.)


srcpkgs/plymouth/template
```diff
@@ -0,0 +1,20 @@
...
+version=0.8.8
+revision=1
+hostmakedepends="pango-devel libpng-devel pkg-config"
+short_desc="Graphical boot animation and logger"
+maintainer="William OD <obirik2005@gmail.com>"
+homepage="https://www.freedesktop.org/wiki/Software/Plymouth/"
+license="GPL-2"
+distfiles="${FREEDESKTOP_SITE}/plymouth/releases/$pkgname-$version.tar.bz2"
+checksum=1bada4e1d3a31a5a99adc2db83f5452d9818839cda51a6e430f044f6281d759b
+only_for_archs="i686 x86_64"
```
Duncaen: Why? Please remove this to let us see what is failing and if its easily fixable.

srcpkgs/plymouth/template
```diff
@@ -0,0 +1,20 @@
...
+build_style=gnu-configure
+configure_args="--with-system-root-install=no --without-rhgb-compat-link \
+--enable-systemd-integration=no --enable-gtk=no --enable-gdm-transition --enable-pango"
```
Duncaen: Please use one space indention here, inside variables and remove the `\` as its not needed between `"`.

srcpkgs/plymouth/template
```diff
@@ -0,0 +1,20 @@
...
+
+
+do_install() {
+	vinstall $FILESDIR/void-logo.png 644 usr/share/plymouth/ bizcom.png
```
Duncaen: this seems wrong, it should be `pre_install, if you overwrite `do_install` `make install` from `build_style=gnu-configure` is never called.

WilliamO7:
Just found out I messed up with the commits on my end...
Trying to squash the two into one now...

srcpkgs/plymouth/template
```diff
@@ -11,10 +11,10 @@ distfiles="${FREEDESKTOP_SITE}/plymouth/releases/$pkgname-$version.tar.bz2"
...
-configure_args="--with-system-root-install=no --without-rhgb-compat-link \
---enable-systemd-integration=no --enable-gtk=no --enable-gdm-transition --enable-pango"
+configure_args="--with-system-root-install=no --without-rhgb-compat-link 
+ --enable-systemd-integration=no --enable-gtk=no --enable-gdm-transition --enable-pango"
 
 
-do_install() {
-	vinstall $FILESDIR/void-logo.png 644 usr/share/plymouth/ bizcom.png
+pre_install() {
+ vinstall $FILESDIR/void-logo.png 644 usr/share/plymouth/ bizcom.png
```
Duncaen: Here you should use tab to indent, just one space inside of variables.

WilliamO7:
All previous edits should be done now.

srcpkgs/plymouth/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'plymouth'
+pkgname=plymouth
+version=0.8.8
+revision=1
+hostmakedepends="pango-devel libpng-devel pkg-config"
```
Duncaen: Those `-devel` packages should go into `makedepends`.

WilliamO7:
Hmm, only i686 and x86_64 glib packages build correctly... Looking into it...

pullmoll:
The culprit seems to be [`rpmatch(3)`](http://linux.die.net/man/3/rpmatch) which musl libc does not have.
A simple, non locale aware replacement could be written to accept e.g. "yes", "y", "no", "n" as valid strings returning 0 for negative and 1 for affirmative strings, or -1 otherwise.

This could be patched into `main.c` itself enclosed in `#if !defined(__GLIBC__)` like this:
```C
#if !defined(__GLIBC__)
int rpmatch(const char* response)
{
    if (!strcasecmp(response, "n") ||
        !strcasecmp(response, "no"))
        return 0;
    if (!strcasecmp(response, "y") ||
        !strcasecmp(response, "yes"))
        return 1;
   return -1;
}
#endif /* !defined(__GLIBC__) */
```

Duncaen:
As long as musl is not locale aware i see no need for it.

WilliamO7:
So should we not build this for musl then?

Vaelatern:
@WilliamO7 I think @Duncaen and @pullmoll are discussing the ideal workaround so we can build it for musl.

pullmoll:
@Duncaen I tend to agree, while musl has `setlocale()` and we could use `gettext-devel`...

Then `gettext()` could used to query the localized strings for e.g. `"yes", "no", "^[yY]" and "^[nN]"` compare their translation to the response.

Just a thought. For now the hack should suffice.

Vaelatern:
@pullmoll Would it be correct to now `makedepends` on `musl-rpmatch-devel` for this package?

pullmoll:
Yes, I think `musl-rpmatch-devel` could now be used.
Instead of trying to patch `configure`, or even `configure.ac` and/or `Makefile.am` in case you wanted to add `automake` and `libtool` to hostdepends, it should be sufficient to have a global scope `LIBS=-lrpmatch` or perhaps `LDFLAGS=-lrpmatch`.

WilliamO7:
Sigh, still doesn't build on musl (and yes, I did move the include up and build again)...
I would look for another file that includes rpmatch, but I really do not have the time right now.

pullmoll:
I'll probably take a look later this afternoon.

pullmoll:
@WilliamO7 LMK if the package works for you. I didn't take the time to test the results.

PR #4446, opened 2016-07-08
perl-DBD-mysql: update to 4.034

diogoleal:


srcpkgs/perl-DBD-mysql/template
```diff
@@ -1,6 +1,6 @@
 # Template file for 'perl-DBD-mysql'
 pkgname=perl-DBD-mysql
-version=4.033
+version=4.034
 revision=2
```
Vaelatern: `1`

diogoleal: Changed!

PR #4452, opened 2016-07-10
run-parts: update to version 4.8

suiginsoft:
debianutils-4.7 has been removed from most Debian repository mirrors, which causes building the packages from source to break unless we use the latest version, debianutils-4.8.


asergi:
Cherry-picked the relevant commit and merged in 58574b3.

PR #4453, opened 2016-07-11
Correct bootstrap build issues when building from Arch linux

wazboy:
I decided to attempt to build void linux from Arch linux.  I ran into a number of bootstrapping issues along the way.  I started from https://github.com/voidlinux/void-packages/blob/master/README.md#using-xbps-src-in-a-foreign-linux-distribution and decided to run "bootstrap" instead of "binary-bootstrap".

I found a number of issues that prevented building all of the bootstrap packages.  The most troubling were builds where include files and libraries on the build host (in this case Arch linux) would be found during bootstrapping and be linked into the bootstrap packages.  In other cases the newer version of gcc shipped with Arch would cause build failures due to -Werror (these don't manifest with gcc 4.9.3).  Finally some packages wouldn't build because the $XBPS_MASTERDIR wasn't populated with the c header files in /usr/include.

With these changes I was able to successfully complete "xbps-src bootstrap" on an x86_64 Arch linux system, overcoming all of the above issues.

### DIFF
--- a/srcpkgs/pkg-config/template
+++ b/srcpkgs/pkg-config/template
@@ -1,7 +1,7 @@
 # Template build file for 'pkg-config'
 pkgname=pkg-config
 version=0.29.1
-revision=1
+revision=2
 bootstrap=yes
 build_style=gnu-configure
 configure_args="--with-internal-glib --disable-host-tool"
--- a/common/environment/configure/bootstrap.sh
+++ b/common/environment/configure/bootstrap.sh
@@ -1,4 +1,5 @@
 if [ -z "$CHROOT_READY" ]; then
 	CFLAGS+=" -isystem ${XBPS_MASTERDIR}/usr/include"
 	LDFLAGS+=" -L${XBPS_MASTERDIR}/usr/lib -Wl,-rpath-link=${XBPS_MASTERDIR}/usr/lib"
+	PKG_CONFIG_LIBDIR="${XBPS_MASTERDIR}/usr/lib/pkgconfig"
 fi
--- a/srcpkgs/gcc/patches/gcc-233720-backport.patch
+++ b/srcpkgs/gcc/patches/gcc-233720-backport.patch
@@ -1,0 +1,117 @@
+--- gcc/cp/Make-lang.in	2016/02/25 15:23:47	233719
++++ gcc/cp/Make-lang.in	2016/02/25 15:33:50	233720
+@@ -111,7 +111,7 @@
+ # deleting the $(srcdir)/cp/cfns.h file.
+ $(srcdir)/cp/cfns.h:
+ endif
+-	gperf -o -C -E -k '1-6,$$' -j1 -D -N 'libc_name_p' -L ANSI-C \
++	gperf -o -C -E -k '1-6,$$' -j1 -D -N 'libc_name_p' -L C++ \
+ 		$(srcdir)/cp/cfns.gperf --output-file $(srcdir)/cp/cfns.h
+ 
+ #
+--- gcc/cp/cfns.gperf	2016/02/25 15:23:47	233719
++++ gcc/cp/cfns.gperf	2016/02/25 15:33:50	233720
+@@ -1,3 +1,5 @@
++%language=C++
++%define class-name libc_name
+ %{
+ /* Copyright (C) 2000-2015 Free Software Foundation, Inc.
+ 
+@@ -16,14 +18,6 @@
+ You should have received a copy of the GNU General Public License
+ along with GCC; see the file COPYING3.  If not see
+ <http://www.gnu.org/licenses/>.  */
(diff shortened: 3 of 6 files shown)


Vaelatern:
Please rebase on top of master.

srcpkgs/libarchive/template
```diff
@@ -12,7 +12,7 @@ hostmakedepends="pkg-config"
 if [ "$CHROOT_READY" ]; then
 	hostmakedepends+="automake libtool"
 fi
-makedepends="zlib-devel bzip2-devel liblzma-devel
+makedepends="glibc-devel zlib-devel bzip2-devel liblzma-devel
```
Vaelatern: This will break all of our cross builds. It seems that perhaps we should change bootstrap to give us a relevant libc by default, first, but that would require building the libc. Besides, we really ought have glibc-devel already in a regular build environment, but not on any musl based builds, so again would break cross compiling.

wazboy: I noticed that requiring glibc-devel would break musl builds.  I was looking for a package that provided c library headers that could be either glibc-devel or the musl equivalent.  I will rebase on master (I noticed run-parts was already updated) and I will remove all of the commits that add "makedepends=glibc-devel .." for now until we can find a better solution.

Vaelatern: Keep them in until @xtraeme or @Gottox weighs in on the matter. Unless you already have removed them.

wazboy:
I have rebased on master as requested.  I performed a full bootstrap build from Arch and it was successful.

I left in the commits that add "makedepends=glibc-devel" as requested.

srcpkgs/pkg-config/template
```diff
@@ -1,10 +1,11 @@
 # Template build file for 'pkg-config'
 pkgname=pkg-config
 version=0.29.1
-revision=1
+revision=2
 bootstrap=yes
 build_style=gnu-configure
 configure_args="--with-internal-glib --disable-host-tool"
+makedepends="glibc-devel"
```
xtraeme: this is really unnecessary, in the bootstrap case, we use the host libc.

srcpkgs/which/template
```diff
@@ -1,9 +1,10 @@
 # Template file for 'which'
 pkgname=which
 version=2.21
-revision=2
+revision=3
 bootstrap=yes
 build_style=gnu-configure
+makedepends="glibc-devel"
```
xtraeme: ditto, unnecessary.

srcpkgs/zlib/template
```diff
@@ -1,8 +1,9 @@
 # Template build file for 'zlib'
 pkgname=zlib
 version=1.2.8
-revision=6
+revision=7
 bootstrap=yes
+makedepends="glibc-devel"
```
xtraeme: ditto, unnecessary.

srcpkgs/xbps/template
```diff
@@ -1,7 +1,7 @@
 # Template file for 'xbps'
 pkgname=xbps
 version=0.51
-revision=9
+revision=10
```
xtraeme: this patch has been applied in git master.

srcpkgs/libarchive/template
```diff
@@ -8,8 +8,11 @@ configure_args="$(vopt_enable acl) $(vopt_enable acl xattr)
  $(vopt_with expat) $(vopt_with lzo lzo2) $(vopt_with lz4)
  $(vopt_with ssl openssl) --without-xml2 --without-nettle
  --disable-rpath ac_cv_func_lchmod=no"
-hostmakedepends="automake libtool pkg-config"
-makedepends="zlib-devel bzip2-devel liblzma-devel
+hostmakedepends="pkg-config"
+if [ "$CHROOT_READY" ]; then
+	hostmakedepends+=" automake libtool"
+fi
+makedepends="glibc-devel zlib-devel bzip2-devel liblzma-devel
```
xtraeme: `glibc-devel` is wrong and unnecessary.

srcpkgs/xbps/template
wazboy: Initially when attempting this I used the Arch aur package for xbps-git.  It built xbps-0.5.2-<git hash>.  For some reason xbps-0.51 inside the chroot couldn't read the metadata created by the xbps-0.52 outside on the host.  I tried to debug xbps-0.51 inside the chroot but I didn't spend the time to install all of the necessary tools.  I'll revert this though.

wazboy:
I've rebased onto master, removed the xbps patch, and removed all "makedepends=glibc-devel..." changes.

xtraeme:
Please rebase again, we've merged most of this stuff already.

wazboy:
I have rebased as requested.

chneukirchen:
Please reopen if this is still relevant.

PR #4467, opened 2016-07-15
postgresql: actually make homedir

Vaelatern:
Fixes the below:

```
bash-4.3$ psql
psql (9.5.3)
Type "help" for help.

postgres=#\q
could not save history to file "/var/lib/postgresql/.psql_history": No such file or directory
bash-4.3$ 
```


Vaelatern:
RFC.

PR #4468, opened 2016-07-15
rmilter: update to 1.8.6

jprjr:
Also enhance the run script

### DIFF
--- a/srcpkgs/rmilter/files/rmilter/run
+++ b/srcpkgs/rmilter/files/rmilter/run
@@ -1,3 +1,8 @@
 #!/bin/sh
 [ -r conf ] && . ./conf
-exec chpst -u rmilter:rmilter rmilter -n -c /etc/rmilter/rmilter.conf ${OPTS}
+: ${RMILTERUSER:=rmilter}
+: ${RMILTERGROUP:=rmilter}
+if [ ! -d /run/rmilter ]; then
+	install -d -m 0755 -o $RMILTERUSER -g $RMILTERGROUP /run/rmilter
+fi
+exec chpst -u $RMILTERUSER:$RMILTERGROUP rmilter -n -c /etc/rmilter/rmilter.conf ${OPTS}
--- a/srcpkgs/rmilter/template
+++ b/srcpkgs/rmilter/template
@@ -1,7 +1,7 @@
 # Template file for 'rmilter'
 pkgname=rmilter
-version=1.8.5
-revision=2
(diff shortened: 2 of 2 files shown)


srcpkgs/rmilter/files/rmilter/run
```diff
@@ -1,3 +1,8 @@
 #!/bin/sh
 [ -r conf ] && . ./conf
-exec chpst -u rmilter:rmilter rmilter -n -c /etc/rmilter/rmilter.conf ${OPTS}
+: ${RMILTERUSER:=rmilter}
+: ${RMILTERGROUP:=rmilter}
+if [ ! -d /run/dovecot ]; then
```
Vaelatern: Mind mentioning why you special case dovecot like this, or why you add a /run/rmilter if it's not there? Comments would be welcome.

jprjr: I don't know how I missed that, that should be /run/rmilter obviously. I'll get that fixed.

The default rmilter config creates a socket at /run/rmilter/rmilter.sock for postfix to connect to. Maybe it would make more sense to just leave that out and have the user make the directory?

Gottox: The way it's done in void is creating these directories *before* the daemon is executed. See https://github.com/voidlinux/void-packages/blob/master/srcpkgs/gdm/files/gdm/run (Personally I prefer install instead, but both is fine.

jprjr: Yeah, I'm not sure how I goofed up so badly on that commit! The latest rebase+push to this branch should have everything fixed.

srcpkgs/rmilter/template
```diff
@@ -11,7 +11,7 @@ maintainer="John Regan <john@jrjrtech.com>"
 license="BSD"
 homepage="https://github.com/vstakhov/rmilter"
 distfiles="$homepage/archive/${version}.tar.gz"
-checksum=e9e1b3c724aa7d9e43f899ad0dbb367d37945e48b742b89188e1a1c256be318f
+checksum=dd00e10a6c5e2a7682bd81155f8cb09fbdbed0416e1c7ccc955a4338c85f144f
 system_accounts="rmilter"
```
Vaelatern: While we are here, can we make this a `_rmilter` as we seem to be doing underscored names for system accounts of late. The runit script will need modification too. Anybody see anything else this might break?

PR #4471, opened 2016-07-16
New package: python-gunicorn-19.6.0

the-maldridge:
Its a nice alternative to uwsgi which requires less understanding of the internals to make work efficiently.


Duncaen:
Seems like it needs python 3.5
https://travis-ci.org/voidlinux/void-packages/jobs/145162265#L786

And you only generate the 2.7 package, 3.4 is missing.

Vaelatern:
See the variable `python_versions`.

the-maldridge:
uwsgi sufficed for my purposes and I would recommend it to others over gunicorn.

PR #4472, opened 2016-07-16
libsasl: break mit-krb5 build dep loop

### LINKED ISSUE #4338: Circular dependency openldap, libsasl, mit-krb5, openldap, ...
There's a circular dependency which shows when you try to build e.g. `openldap` with `./xbps-src -N pkg openldap`. Right now you can only build one of the packages without `-N` to break out of the circle.

the-maldridge:
This should break the loop introduced into the openldap and kerberos5
build path.  This is the currently winning solution to resolve #4338.

This seems to non-deterministically break builds.  I suspect there is a possible race condition in the build, but its difficult to track down as it seems to be a different target that breaks each time I build it.

**Note: I have run no regression testing on this patch to see what, if anything, it breaks.  This would ideally be verified before merging to ensure this doesn't break anything else.**


Vaelatern:
At time of comment, only the armv7hf-glibc build breaks: https://s3.amazonaws.com/archive.travis-ci.org/jobs/145162729/log.txt
Re-running travis on all builds now.

PR #4473, opened 2016-07-16
New package: autojump-22.3.4

### LINKED ISSUE #4438: Package request: autojump
I use this program intensively on my arch setup.
It's ``` cd ``` that easily navigate in directories.
https://github.com/wting/autojump

the-maldridge:
Resolves #4438.

This has got a fairly convoluted installer, thus the need for the custom do_install().  As this isn't a python module I don't really see much point in building it against anything but the default system python.

### DIFF
--- a/srcpkgs/autojump/template
+++ b/srcpkgs/autojump/template
@@ -1,0 +1,23 @@
+# Template file for 'autojump'
+pkgname=autojump
+version=22.3.4
+revision=1
+wrksrc=${pkgname}-release-v${version}
+hostmakedepends="python"
+depends="python"
+noarch=yes
+short_desc="A cd command that learns"
+maintainer="Michael Aldridge <aldridge.mac@gmail.com>"
+license="GPL-3"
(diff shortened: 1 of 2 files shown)


Vaelatern:
`noarch=yes`

Vaelatern:
An `INSTALL.msg` is probably in order for the installer's notification that:
```
Please manually add the following line(s) to ~/.shrc:
	[[ -s /destdir//autojump-22.3.4/etc/profile.d/autojump.sh ]] && source /destdir//autojump-22.3.4/etc/profile.d/autojump.sh
```

Vaelatern:
Actually, possibly not, with the file it puts in /etc/profile.d

Vaelatern:
Stripping out the `/destdir//autojump-22.3.4` from `${DESTDIR}/etc/profile.d/autojump.sh` might be a good idea.

the-maldridge:
Anything else on this?

PR #4487, opened 2016-07-19
New package: apache-maven-3.3.9

the-maldridge:
Proper source version of Apache Maven.


the-maldridge:
bump?

srcpkgs/apache-maven/template
```diff
@@ -0,0 +1,37 @@
+# Template file for 'apache-maven'
+pkgname=apache-maven
+version=3.3.9
+revision=1
+hostmakedepends="apache-ant openjdk"
+depends="virtual?java-runtime"
+noarch=yes
+short_desc="Software project management and comprehension tool"
+maintainer="Michael Aldridge <aldridge.mac@gmail.com>"
+license="Apache 2.0"
```
Vaelatern: `Apache-2.0`

srcpkgs/apache-maven/template
```diff
@@ -0,0 +1,37 @@
...
+	_installRoot=${DESTDIR}/usr/share/${pkgname}/
+	mkdir -p ${_installRoot}
+
+	for _dir in bin boot conf lib ; do
+		cp -r mvn-complete/${_dir} ${_installRoot}/${_dir}
+	done
+
+	mkdir -p ${DESTDIR}/usr/bin
+	for _app in mvn mvnDebug mvnyjp ; do
+		ln -s ${DESTDIR}/usr/share/${pkgname}/bin/${_app} ${DESTDIR}/usr/bin/${_app}
```
Vaelatern: I believe these symlinks to be incorrect. ``${DESTDIR}/usr/share/${pkgname}/bin/${_app}` will not exist in an installed environment.

Vaelatern: Do they need to be symlinked instead of copied?

Vaelatern: Alternately might it be right to provide an `/etc/profile.d/` entry to append to `$PATH`?

the-maldridge: I'll swap this out when I get a free moment.

PR #4488, opened 2016-07-19
New package: apache-tomcat-8.5.4

the-maldridge:


xtraeme:
Please switch to `/usr/share/<pkgname>` rather than `/opt/<pkgname>`.

We've decided time ago that `/opt` shall only be used in some rare cases and only for software provided as binaries.

the-maldridge:
@xtraeme tomcat makes use of a non-reconfigurable temporary directory relative to its binary.  Moving this to /usr/share would place the runtime writable directory there as well.  Is this acceptable?

the-maldridge:
After some more research it seems this can be somewhat clunkily overriden with a combination of debug flags and environment variables.  I'll make those changes in the morning and re-push.  Should the path be changed on both this one and #4487 as well?

chneukirchen:
Can we put binary stuff into `/usr/lib` or /usr/libexec` please? `/usr/share` is classically non-architecture-dependent.

pullmoll:
I think we have quite a few packages installing binaries in `/usr/share/$pkgname` and linking to (some of) them from `/usr/bin`. I know the `mame` package did and still does this.

If there is a consensus to use `/usr/lib/$pkgname` or `/usr/libexec/$pkgname` for binaries, we should probably create a list of packages to modify.

the-maldridge:
@chneukirchen All of the binaries for this are still architecture independant as its Java all the way down.

I'm happy to move it to /usr/share/apache-tomcat as long as people are clear that this will put the webapps directory there and this will mean that in normal operation there may be read/write access open to things in /usr/share/apache-tomcat/webapps.

the-maldridge:
I have made the changes that were discussed above.  Please let me know if there are additional changes needed for this package.

the-maldridge:
bump?

srcpkgs/apache-tomcat/template
```diff
@@ -0,0 +1,70 @@
+# Template file for 'apache-tomcat'
+pkgname=apache-tomcat
+version=8.5.4
+revision=1
+wrksrc=${pkgname}-${version}-src
+hostmakedepends="openjdk apache-ant"
+depends="virtual?java-runtime"
+make_dirs="/usr/share/${pkgname}/webapps 0755 tomcat tomcat \
+	/usr/share/${pkgname}/conf 0755 tomcat tomcat \
+	/usr/share/${pkgname}/logs 0755 tomcat tomcat"
```
Duncaen: usually we indent with one space inside quotes.

Vaelatern:
For now we will call the directory structure as it exists here fine. It's not in `/opt` as @xtraeme wanted and there are no arch-dependent parts in `/usr/share`. The only remaining issue is of rw access in `/usr/share` but that seems unavoidable to maintain consistency with upstream.

PR #4489, opened 2016-07-20
vapoursynth-mvtools: fix library location

lemmi:


Vaelatern:
If nothing is supposed to build off of that library, this move is correct, but the below should be applied.

```
diff --git a/common/shlibs b/common/shlibs
index 38d02e3..dac73da 100644
--- a/common/shlibs
+++ b/common/shlibs
@@ -2449,7 +2449,6 @@ libobs-opengl.so.0 obs-0.14.1_2
 libzimg.so.2 zimg-2.0.4_1
 libvapoursynth.so vapoursynth-R32_2
 libvapoursynth-script.so.0 vapoursynth-R32_1 
-libmvtools.so vapoursynth-mvtools-12_1
 libtwolame.so.0 twolame-0.3.13_1
 libsignon-plugins.so.1 signond-8.58_1
 libsignon-qt5.so.1 signond-8.58_1
```

Vaelatern:
Also fix the whitespace error? This library is not used by any other tool, once the above is applied this can be merged.

PR #4493, opened 2016-07-20
kodi: remove --libdir=/usr/lib32 from configure on i686 to fix #3302

### LINKED ISSUE #3302: Kodi fails to load on i686
Kodi does not start on i686 failing with the error message:
```Unable to find path to Kodi data files!```

Kodi works fine on x64 with no issues.

Hoshpak:
Following up on the suggestion by @xtraeme, I extended do_configure to remove --libdir=/usr/lib32 on i686 from configure_args to fix #3302. Tested on a local i686 VM.


xtraeme:
Well done, thanks :-)

PR #4500, opened 2016-07-24
kodi: fix building against curl 7.50.0 (#4497)

### LINKED ISSUE #4497: Kodi cannot be built with curl 7.50.0
As documented in d725843c9bf04ffce0327b345e0170ccc359a7a3 , kodi builds fail since the update of curl to 7.50. The kodi code most likely needs to be adapted to the code changes regarding CULRM and CURL_HANDLE. Someone already created a [thread](http://forum.kodi.tv/showthread.php?tid=283382&highlight=curl) in the kodi forum about it. Waiting for a statement from the upstream developers now.

Hoshpak:
Patch taken from xbmc/xbmc#10160 (thanks @fritsch for taking care of it). The patch is from kodi master but the relevant code was not changed since the 16.1 release so it applies cleanly against that version as well. Tested building and running kodi with this patch on my Void i686 VM. Fixes #4497

### DIFF
--- a/srcpkgs/kodi/patches/fix-curl-750-compatibility.patch
+++ b/srcpkgs/kodi/patches/fix-curl-750-compatibility.patch
@@ -1,0 +1,44 @@
+diff --git a/xbmc/filesystem/CurlFile.h b/xbmc/filesystem/CurlFile.h
+index b4f5046..81a5ae5 100644
+--- a/xbmc/filesystem/CurlFile.h
++++ b/xbmc/filesystem/CurlFile.h
+@@ -24,11 +24,17 @@
+ #include <map>
+ #include <string>
+ #include "utils/HttpHeader.h"
++#include <curl/curlver.h>
+ 
+ namespace XCURL
+ {
++#if LIBCURL_VERSION_NUM >= 0x073200
++  typedef struct Curl_easy CURL_HANDLE;
++  typedef struct Curl_multi CURLM;
++#else
+   typedef void CURL_HANDLE;
+   typedef void CURLM;
++#endif
+   struct curl_slist;
+ }
+ 
+diff --git a/xbmc/filesystem/DllLibCurl.h b/xbmc/filesystem/DllLibCurl.h
+index db97f92..9e78a91 100644
+--- a/xbmc/filesystem/DllLibCurl.h
++++ b/xbmc/filesystem/DllLibCurl.h
+@@ -52,7 +52,7 @@ namespace XCURL
+     virtual CURLMcode multi_fdset(CURLM *multi_handle, fd_set *read_fd_set, fd_set *write_fd_set, fd_set *exc_fd_set, int *max_fd)=0;
+     virtual CURLMcode multi_timeout(CURLM *multi_handle, long *timeout)=0;
+     virtual CURLMsg*  multi_info_read(CURLM *multi_handle, int *msgs_in_queue)=0;
+-    virtual void multi_cleanup(CURL_HANDLE * handle )=0;
++    virtual CURLMcode multi_cleanup(CURLM * handle )=0;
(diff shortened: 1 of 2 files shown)


fritsch:
Take care of this discussion here: https://github.com/curl/curl/commit/9adf3c473a01b289c781aab111f9ad2fc541ed4e#commitcomment-18369989 - upstream broke our API

fritsch:
I opened a bug with the curl guys: https://github.com/curl/curl/issues/926

Btw. you need to rebuild _every_ application using libcurl. I hope they will release 7.60.0 and bump soname and other things needed.

Gottox:
*sigh*

Gottox:
@fritsch To understand this correctly: The broke the *API* but the *ABI* stays compatible?

fritsch:
Both is broken :-) public types have changed.

xtraeme:
Please do not revbump any package until the main curl author clarifies its status.

xtraeme:
Quoting @bagder

```
We could argue if it breaks the API as in your case an old application no longer builds with the current include files, but then I also think you're pulling some fun tricks out of your sleeves. Your code "trespasses" into libcurl name space territory in a pretty unorthodox way.
```

The kodi code typedef's them again assuming its type was `void`, and now this does not match anymore because those are typedef's to the correct struct type.

So no, there's no need to rebuild all packages linking to libcurl unless they do weird things like kodi.

fritsch:
Do you understand what you are saying or are you just repeating the quote cause it means less work?

Again in short: Every application that forward declared public curl types (typedef void whatever) to not have to include curl.h in a public header is broken now. All such applications that you don't recompile won't run afterwards.

Easy test: Build kodi or anything else against curl 7.49.0 and afterwards just replace curl 7.49.0 with 7.50.0 and see what you get. If you insist of it, I can make you a minimal example.

Vaelatern:
@fritsch That is almost certaintly not every single curl project.
We will not be updating these packages until curl upstream has made themselves clear enough for us to be certain.

Gottox:
I can understand how the API was broken, but I can't understand how the ABI is.

The only thing that a C type cares about is its size. Pointer types - regardless if it's void*, int*, etc - always have the same size. They only differ if you use pointer arithmetic and that was broken in curl due to the usage of void datatypes anyway. As long as this does not change, the ABI stays intact.

fritsch:
Once again, we don't talk about pointers. we talk about void vs struct whatever in public curl.h - but I want to end this discussion here as I think it will lead to nothing, just two questions for you:
what is the sizeof(void) ? and what is the sizeof(struct whatever)?

Vaelatern:
@fritsch Of a void pointer and struct pointer? Identical.
```
$ cat test.c
#include <stdio.h>

struct abc {
	int a;
	float b;
	long c;
	char d;
	struct abc *e;
};

int main()
{
	printf("%d\n%d\n", sizeof(void*), sizeof(struct abc*));
}
$ cc test.c
$ ./a.out
8
8
```

fritsch:
Now for the last time (!)

We talk about void _not_ void* - can't you read the referenced patch? If it was void* it would have been no issue at all.

Gottox:
the discussion aside, the patch is fine.

fritsch:
I changed the patch, one hunk was not needed.

Vaelatern:
Please add a comment in the template file, specifying that any updates to this package should bear in mind the PR against upstream you have linked here.

Hoshpak:
I've amended the pull request to include the current patch from the upstream pull request and added the requested comment to the template file. Could build it again in my VM and got a working binary package. Hope that travis will do his part soon.

fritsch:
Keep an eye of the original bugreport: https://github.com/curl/curl/issues/926#issuecomment-236420206 - most likely other void packages are broken now, too.

Gottox:
Thanks for the hint! I'll keep an eye on the patches.

PR #4508, opened 2016-07-25
New package: dfu-programmer-0.7.1

oholiab:
From https://dfu-programmer.github.io/, tl;dr it's a microcontroller flasher.


srcpkgs/dfu-programmer/template
```diff
@@ -0,0 +1,18 @@
+# Template file for 'dfu-programmer'
+pkgname=dfu-programmer
+version=0.7.1
+revision=1
+build_style=gnu-configure
+hostmakedepends="automake unzip"
+makedepends="libusb-devel"
+depends="libusb"
```
Duncaen: `depends` is not needed in this case, `xbps-src` detects to which shared libraries the binaries are linked, see https://travis-ci.org/voidlinux/void-packages/jobs/147326719#L737

srcpkgs/dfu-programmer/template
```diff
@@ -0,0 +1,18 @@
...
+short_desc="Device Firmware Upgrade class USB driver and flasher"
+maintainer="Matt Carroll <oholiab@gmail.com>"
+license="GPL-2"
+homepage="http://dfu-programmer.github.io"
+distfiles="http://github.com/dfu-programmer/dfu-programmer/archive/v0.7.1.zip"
```
Duncaen: Use the `$version` variable here.

oholiab:
I'm not sure why arm is failing... I'll give it a go on my pi tomorrow - I've already had it working on musl armv7 before I tried packaging it

Duncaen:
Its uses `/usr/include` which is the hosts include directory, the configure.ac script has some `includedir` variables,  it should be set to `${XBPS_CROSS_BASE}/usr/include` which works in all cases, cross and normal compilation.
No idea where to ideally sed/patch/replace it.

Duncaen:
Please rebase, squash and change the commit message as described here https://github.com/voidlinux/void-packages/blob/master/CONTRIBUTING.md#committing-your-changes.

PR #4517, opened 2016-07-28
new package: python-discogs_client-2.2.0

tonkazoid:
Optional dependency for the beets package


srcpkgs/discogs_client/template
```diff
@@ -0,0 +1,19 @@
...
+revision=1
+noarch=yes
+build_style="python-module"
+hostmakedepends="python-setuptools"
+depends="python-requests python-oauthlib"
+pycompile_module="discogs_client"
+short_desc="Official Discogs API client for Python"
+maintainer="Georg Schabel <gescha@posteo.de>"
+homepage="https://github.com/discogs/discogs_client"
+license="Discogs"
```
Duncaen: It looks like 2-clause-BSD

srcpkgs/discogs_client/template
```diff
@@ -0,0 +1,19 @@
...
+noarch=yes
+build_style="python-module"
+hostmakedepends="python-setuptools"
+depends="python-requests python-oauthlib"
+pycompile_module="discogs_client"
+short_desc="Official Discogs API client for Python"
+maintainer="Georg Schabel <gescha@posteo.de>"
+homepage="https://github.com/discogs/discogs_client"
+license="BSD-2"
+distfiles="https://github.com/discogs/${pkgname}/archive/v${version}.tar.gz"
```
Vaelatern: Can we use `${homepage}` here?

Vaelatern:
Why should we include this when it seems to be just a pip module best installed from pip?

Vaelatern:
Can we get a python 3 subpackage?

Vaelatern:
Should be named like the python module it is.

tonkazoid:
Something like `discogs_client-python discogs_client-python3.4`?

Vaelatern:
Sure. But first try building it with python3.4 at all

Vaelatern:
Thank you for your contribution!

PR #4526, opened 2016-07-30
New package: rustfmt-0.6

DirectorX:


DirectorX:
**rebased from** #4527

common/build-style/cargo.sh
```diff
@@ -0,0 +1,11 @@
+#
+# This helper is for templates for Cargo packages.
+#
+
+do_build() {
+	cargo build --release
+}
+
+do_install() {
+	cargo install
```
ebfe: This installs files into ./target/release.

DirectorX: well that weird, i use rustfmt that i build using those scripts
is probably something has changed, again

ebfe: Sorry, not `./target/release` but `$HOME/.cargo/bin`. And as far as i can tell, that's how `cargo install` always worked.

PR #4530, opened 2016-07-31
New package: mergerfs-2.14.0

xinterer:


srcpkgs/mergerfs/template
```diff
@@ -0,0 +1,24 @@
+# Template file for 'mergerfs'
+pkgname="mergerfs"
+version="2.14.0"
+revision=1
+build_style=gnu-makefile
+make_install_args=PREFIX=/usr
```
Duncaen: This is already set in https://github.com/voidlinux/void-packages/blob/master/common/build-style/gnu-makefile.sh#L18

srcpkgs/mergerfs/template
```diff
@@ -0,0 +1,24 @@
...
+nopie=true
+short_desc="FUSE union filesystem"
+maintainer="xinterer <xinterer@users.noreply.github.com>"
+license="ISC License"
+homepage="https://github.com/trapexit/mergerfs"
+distfiles="https://github.com/trapexit/$pkgname/releases/download/$version/$pkgname-$version.tar.gz"
+checksum="54f04cd6f9660f82fd9b3f0e3bedfb22784fcaa82de93c866c1c05940c15fd6f"
+
+do_build() {
+	make
```
Duncaen: Don't do this, let `build_style=gnu-makefile` call it, it has to pass some parameters like our `CFLAGS`.

srcpkgs/mergerfs/template
```diff
@@ -0,0 +1,24 @@
...
+}
+
+do_install() {
+	vlicense LICENSE
```
Duncaen: rename this function to `post_install` `build_style=gnu-makefile` needs `do_install` to call `make install` with predefined parameters.

srcpkgs/mergerfs/template
```diff
@@ -0,0 +1,24 @@
...
+version="2.14.0"
+revision=1
+build_style=gnu-makefile
+make_install_args=PREFIX=/usr
+makedepends="fuse-devel python attr-devel pandoc pkg-config"
+depends="fuse"
+nopie=true
+short_desc="FUSE union filesystem"
+maintainer="xinterer <xinterer@users.noreply.github.com>"
+license="ISC License"
```
Duncaen: Change `license` to `ISC`, to be consistent with all other packages.

srcpkgs/mergerfs/template
```diff
@@ -0,0 +1,24 @@
+# Template file for 'mergerfs'
+pkgname="mergerfs"
+version="2.14.0"
+revision=1
+build_style=gnu-makefile
+make_install_args=PREFIX=/usr
+makedepends="fuse-devel python attr-devel pandoc pkg-config"
+depends="fuse"
```
Duncaen: This is most likely wrong, `xbps-src` detects runtime dependencies by looking which shared libraries binaries need.

srcpkgs/mergerfs/template
```diff
@@ -0,0 +1,24 @@
...
+nopie=true
```
Duncaen: Can you provide some reason, why PIE wouldn't work?

srcpkgs/mergerfs/template
```diff
@@ -0,0 +1,25 @@
...
+nocross=true
```
Duncaen: That is wrong too, without a good reason.

Duncaen:
You are building an empty package, see https://travis-ci.org/voidlinux/void-packages/jobs/148605132#L737 its really not about making travis happy, remove `nocross`, `nopie` etc so we see the problem and maybe find a solution together.

srcpkgs/mergerfs/template
xinterer: I've done this because I think that build_style=gnu-makefile is overriding the Makefile flags and I don't know other way to make it work.
https://forum.voidlinux.eu/t/making-template-for-mergerfs-help-needed/882

srcpkgs/mergerfs/template
xinterer: Gives me this error after compilation:
ERROR: mergerfs-2.14.0_1: non-PIE executable found in PIE build: /usr/bin/mergerfs

srcpkgs/mergerfs/template
```diff
@@ -0,0 +1,22 @@
+# Template file for 'mergerfs'
+pkgname="mergerfs"
+version="2.14.0"
+revision=1
+build_style=gnu-makefile
+makedepends="fuse-devel python attr-devel pandoc pkg-config"
```
Duncaen: pandoc, python and pkg-config should be in `hostmakedepends`

srcpkgs/mergerfs/template
Duncaen: But you need also the flags, that gnu-makefile sets, that would fix cross compiling and the PIE stuff.
Add a patch or use sed, to change the `Makefile` from `CFLAGS =` to `override CFLAGS +=`.

xinterer:
As I see it compiles only on glibc x86_64 and i686. Is it again problem in the template?

Vaelatern:
@xinterer If it's a matter of dependencies not being present, and not actual technical reasons, then do not at all put `nocross` or `only_for_archs` at all. Such things only matter if the software itself can inherently not be cross-built or whatnot.

xinterer:
Guys, thank you very much for the help and the explanations!
So, what remains to be done to the template if it builds only for glibc x86_64 and i686? Its builds, right?

srcpkgs/mergerfs/template
```diff
@@ -0,0 +1,24 @@
...
+version="2.14.0"
+revision=1
+build_style=gnu-makefile
+makedepends="fuse-devel attr-devel "
+hostmakedepends="pandoc python pkg-config"
+short_desc="FUSE union filesystem"
+maintainer="xinterer <xinterer@users.noreply.github.com>"
+license="ISC License"
+homepage="https://github.com/trapexit/mergerfs"
+distfiles="https://github.com/trapexit/$pkgname/releases/download/$version/$pkgname-$version.tar.gz"
```
Vaelatern: Can we use `${homepage}`?

srcpkgs/mergerfs/template
```diff
@@ -0,0 +1,24 @@
+# Template file for 'mergerfs'
+pkgname="mergerfs"
+version="2.14.0"
+revision=1
+build_style=gnu-makefile
+makedepends="fuse-devel attr-devel "
+hostmakedepends="pandoc python pkg-config"
+short_desc="FUSE union filesystem"
+maintainer="xinterer <xinterer@users.noreply.github.com>"
+license="ISC License"
```
Vaelatern: Please do as @Duncaen said and make this consistent with other packages. `license="ISC"`

pullmoll:
You can remove `pandoc` from the `hostmakedepends` and add these lines to global:
```
	*-musl) # Does not have pandoc
		;;
	*)	hostmakedepends+=" pandoc"
		;;
esac
```

The man page is still installed, just not built from the source for *-musl.

pullmoll:
Also `python` can be removed from `hostmakedepends`. It isn't required AFAICT.

xinterer:
Well, I've seen that the template with the changes from @pullmoll  was added and the package are already in the Void repo. So, I assume that this pull request should be closed...
Thank you again, I learned a lot of things here.

pullmoll:
Yea, I made a mistake when I accidentially pushed my own (wrong) version of a template for mergerfs.
I reverted that commit and decided to take your template, apply the fixes and push it.
I hope it works for you @xinterer

PR #4535, opened 2016-07-31
lcms2: distfile sha256sum correction

gotwf:


Vaelatern:
Please don't close and re-open PRs. A force-push will automatically update your PRs.

Vaelatern:
Mind explaining why this is necessary? It is obvious that it works, but the last version bump had a sha256sum update as well. Are we getting a different program now?

gotwf:
As far as I can discern: the pkg would not build locally on a pretty
fresh void install due to a sha256sum mismatch.  I fetched the
distfile manually.  I confirmed the md5 sum of that file with that
from the LSF lcms2 page.  The sha256sum I get from this distfile
matches the one xbps-src downloaded.  The distfile version matches the
template version.  But the sha256sum does not.  So, either maintainer
error or maybe upstream changed distfile w/o bumping rev.  I don't
have an "older" distfile to compare.


On Sun, 31 Jul 2016 17:33:26 -0600,
Toyam Cox wrote:

pullmoll:
The old distfile can be found at https://sources.voidlinux.eu/lcms2-2.8/lcms2-2.8.tar.gz
When extracting both tarballs, renaming the directories to `lcms2-2.8.old` and `lcms2-2.8.new`, then running `diff -r lcms2-2.8.old lcms2-2.8.new` there are *no* differences shown.

The sizes of the tarballs differ a lot: old = 10936320, new = 6687005.
It looks like the change is due to reducing the download size and nothing else.

IMO this PR can be merged.

PR #4537, opened 2016-08-01
New package: pam_ssh-2.1_1

crater2150:


srcpkgs/pam_ssh/template
```diff
@@ -0,0 +1,24 @@
...
+short_desc="PAM module providing single sign-on behavior for SSH"
+maintainer="Alexander Gehrke <void@qwertyuiop.de>"
+license="2-clause-BSD"
+homepage="http://pam-ssh.sourceforge.net/"
+distfiles="${SOURCEFORGE_SITE}/pam-ssh/${pkgname}-${version}.tar.xz"
+configure_args="--with-pam-dir=/usr/lib/security
+ --includedir=${XBPS_CROSS_BASE}/usr/include
+ --libdir=${XBPS_CROSS_BASE}/lib"
+makedepends="libressl-devel pam-devel"
+hostmakedepends="pam-devel openssh"
```
Vaelatern: Do you need `pam-devel` here? If so, document it.

srcpkgs/pam_ssh/template
```diff
@@ -0,0 +1,24 @@
+# Template file for 'pam_ssh'
+pkgname=pam_ssh
```
Vaelatern: Why named `pam_ssh` when upstream calls it `pam-ssh`? Please try to have the naming consistent with upstream unless there is a reason to avoid it.

crater2150: On the project homepage it is called pam_ssh, also the manpage and and the pam module file use underscore. The hyphen in the homepage is probably for sourceforge naming policy reasons.

chneukirchen: Also the tarball is `pam_ssh`, so +1 for keeping it.

PR #4539, opened 2016-08-01
Asciiquarium

the-maldridge:
One of my favorite useless packages, this draws an aquarium drawn entirely from ascii characters on your terminal.

### DIFF
--- a/srcpkgs/asciiquarium/template
+++ b/srcpkgs/asciiquarium/template
@@ -1,0 +1,23 @@
+# Template file for 'asciiquarium'
+pkgname=asciiquarium
+version=1.1
+revision=1
+wrksrc=${pkgname}_${version}
+depends="perl perl-Term-Animation"
+noarch=yes
+short_desc="Enjoy the mysteries of the sea from the safety of your own terminal"
+maintainer="Michael Aldridge <aldridge.mac@gmail.com>"
+license="GPL-2"
+homepage="http://www.robobunny.com/projects/asciiquarium/html/"
+distfiles="http://www.robobunny.com/projects/asciiquarium/asciiquarium_${version}.tar.gz"
+checksum=1b08c6613525e75e87546f4e8984ab3b33f1e922080268c749f1777d56c9d361
+
+do_install() {
+	vbin asciiquarium
+}
+
+# REMARKS: Per request from Vaelatern this comment explains the hard
+# depends on perl.  A direct dependency is noted on perl as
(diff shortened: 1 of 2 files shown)


srcpkgs/asciiquarium/template
```diff
@@ -0,0 +1,16 @@
+# Template file for 'asciiquarium'
+pkgname=asciiquarium
+version=1.1
+revision=1
+wrksrc=${pkgname}_${version}
+depends="perl perl-Term-Animation"
+short_desc=" Enjoy the mysteries of the sea from the safety of your own terminal"
```
Vaelatern: Please remove space at beginning of short_desc.

srcpkgs/asciiquarium/template
```diff
@@ -0,0 +1,16 @@
...
+maintainer="Michael Aldridge <aldridge.mac@gmail.com>"
+license="GPL-2"
+homepage="http://www.robobunny.com/projects/asciiquarium/html/"
+distfiles="http://www.robobunny.com/projects/asciiquarium/asciiquarium.tar.gz"
```
Vaelatern: Please use the versioned tarball, `http://www.robobunny.com/projects/asciiquarium/asciiquarium_1.1.tar.gz`

srcpkgs/perl-Term-Animation/template
```diff
@@ -0,0 +1,13 @@
...
+pkgname=perl-Term-Animation
+version=2.6
+revision=1
+wrksrc=${pkgname//perl-//}-${version}
+build_style=perl-module
+hostmakedepends="perl"
+short_desc="Terminal Animation module for perl"
+maintainer="Michael Aldridge <aldridge.mac@gmail.com>"
+license="Perl"
+homepage="http://robobunny.com/projects/animation/haml/"
```
Vaelatern: Recommend the homepage is not 404: `s/haml/html/` perhaps

Vaelatern:
Please make the perl module depend on perl, and thus remove the `perl` dependency from `asciiquarium`.

the-maldridge:
I agree that the library should depend on perl, however I do not agree that asciiquarium itself shouldn't depend on perl.  From the manual the depends field is specifically for runtime dependencies that are otherwise not picked up by xbps.  Since asciiquarium has a direct dependancy on perl, it should therefore have that in that field.

Vaelatern:
And, finally, both need `noarch=yes`

PR #4540, opened 2016-08-01
mutagen: Revert 1.34

the-maldridge:
The latest version of mutagen broke picard.


Vaelatern:
More information, to document this?

the-maldridge:
Mutagen 1.33->1.34 removes the BitPaddedInt class which is needed for picard to properly function.  To confirm this, simply install picard and try to launch it.

srcpkgs/mutagen/template
```diff
@@ -1,6 +1,7 @@
 # Template file for 'mutagen'
 pkgname=mutagen
-version=1.34
+reverts="1.34_1"
+version=1.33
```
Vaelatern: We were on 1.33.2.

See a25367a490beec3f31e262db451cb7680e7e67df

chneukirchen:
Shouldn't we rather patch Picard?

https://github.com/metabrainz/picard/pull/479/commits/289925e2951b61b06b365d2d031ef0d73f4a33e5

the-maldridge:
@chneukirchen That would be a good solution as far as I'm concerned.

the-maldridge:
meh, at least my local version isn't broken.

PR #4542, opened 2016-08-02
New package: stow-2.2.2

justinbloomfield:


chneukirchen:
We have xstow already, does this add benefit?

justinbloomfield:
Whoops, didn't see that in there. I think it probably adds some extra features, but not enough that are widely used to justify adding it in if something similar exists already.


________________________________

PR #4544, opened 2016-08-02
python-pickleshare: update to 0.7.3

DirectorX:


srcpkgs/python-pickleshare/template
```diff
@@ -14,7 +14,11 @@ maintainer="Alessio Sergi <al3hex@gmail.com>"
 homepage="https://github.com/pickleshare/pickleshare"
 license="MIT"
 distfiles="${PYPI_SITE}/p/pickleshare/pickleshare-${version}.tar.gz"
-checksum=92ee3b0e21632542ecc9a0a245e69a126f62e5114081bdb0d32e0edd10410033
+checksum=b9710d01f777b1bf3b69c8f889b1d05a5145668f79e980cbd0f849e059edd274
+
+post_extract() {
```
Vaelatern: `post_install()` and needs to have an equivalent for the subpackage I believe.

DirectorX: @Vaelatern `post_extract()` -> `post_install()`

Issue #1177, opened 2015-03-21
Package request: sbt


thypon:
http://www.scala-sbt.org/

PR #4547, opened 2016-08-03
New package: scala-2.11.8 and sbt-0.13.12

crater2150:
Using the binary distribution, because building scala from source requires scala and sbt (which in turn requires scala and sbt for building).

Also added sbt, requested in #1177


Vaelatern:
A "bootstrap" flag which gets the binaries, but you can turn off to build off of itself, would be good.

crater2150:
I'm not sure I understand how this works. As far as I can tell, the bootstrap flag causes the package to be installed from the host repository with `./xbps binary-bootstrap`, but how is that initial package in the repository built? For example the template for make does not contain anything about how to build it whithout make.

xtraeme:
Bootstrap pkgs expect software available in the host system. See the
xbps-src script forma details.

Vaelatern:
I was referring to a `build_option` for bootstrap, much like we have for OpenJDK.

PR #4575, opened 2016-08-08
compiler wrapper: add ccache or distcc prefix

### LINKED ISSUE #4574: compiler wrappers disable ccache / distcc
The compiler wrappers created from the `common/wrappers/cross-cc` template disable any ccache and/or distcc setup because the `@BIN@` string is replaced with the absolute path `/usr/bin/<compiler>` of the respective compiler.

The `common/hooks/pre-configure/02-script-wrapper.sh` script probably needs to test for `$XBPS_CCACHE`, `$XBPS_DISTCC` or both and use one of the prefixes `/usr/lib/ccache/bin` or `/usr/lib/distcc/bin` instead of always just `/usr/bin`.

The problem is that e.g. `/usr/lib/ccache/bin/arm-linux-gnueabihf-c++` linking to `../../../bin/ccache` searches for the real compiler on the path and may - or rather will - find the wrapper script again, resulting in an endless loop. So `PATH` needs to be modified, but just for that call to the wrapper... or the ccache or distcc prefix needs to be inserted before the absolute path of the real compiler.

I'm trying to solve this myself, but you're welcome to find a proper solution.

Not using ccache makes cross compiling huge packages with the trial and error approach to squash bugs a tedious job ;-P

pullmoll:
Closes #4574


xtraeme:
:+1:

PR #4580, opened 2016-08-09
New package: arc-themes-20160605

vjon:


srcpkgs/arc-themes/template
```diff
@@ -0,0 +1,26 @@
+# Template file for 'arc-themes'
+pkgname=arc-themes
+version=20160605
+revision=1
+noarch=yes
+build_style=gnu-configure
+configure_script="./autogen.sh"
+configure_args="--prefix=/usr"
```
Vaelatern: not needed

srcpkgs/arc-themes/template
```diff
@@ -0,0 +1,26 @@
...
+hostmakedepends="autoconf automake pkg-config"
```
Vaelatern: `automake pkg-config`

vjon:
I've fixed the unecessary dependancy, but I'm not sure what you mean by that first comment. Are you saying the entire section (first 8 lines) of the template file can be removed?

srcpkgs/arc-themes/template
vjon: In regards to the "configure_script" and "configure_args" lines, xbps-src fails during do_configure() when it attempts to run "./configure" in the project directory.

Vaelatern: I repeat. Do not set `configure_args` unless you can provide a build log showing it does not work.

vjon: Ah, my bad, I was misunderstanding the intention of your original comment. It's fixed now.

PR #4584, opened 2016-08-09
envypn-font: update tarball URL

Ypnose:
Hi,
The upstream URL has changed so it needs to be updated.


Vaelatern:
Please don't bump revision

Vaelatern:
The package doesn't change or need a rebuild. People installing the software don't care that a statistic like the distfiles url changed. If the homepage changed, that'd be a revbump (anything you can see with `xbps-query -RS $pkg` would need a revbump).

Ypnose:
Indeed. I just forgot the detail. I also wanted to change the homepage but I won't do it.

Vaelatern:
If also updating the homepage, then a revbump is fine.

PR #4586, opened 2016-08-09
New package: moka-icon-theme-5.3.2

vjon:


vjon:
The CI builds are failing due to the log length exceeding the 4MB size limit. However, this is due to the file extraction messages (the Moka icon set includes a very large number of icons), and not due to any error in the build itself. What is the recommended method for dealing with this?

pullmoll:
@vjon Ignore travis-ci. It's just a helpful hand which sometimes fails.

vjon:
Cheers, I'll wait for further comments then.

PR #4596, opened 2016-08-12
ldb: update to 1.1.27

the-maldridge:
Currently broken on musl, I'm considering removing this simply because SSSD is such a pain to maintain and use, and its the only thing that depends on this that I know of.


Vaelatern:
How do `rep_stroll` and `stroll` compare? musl has `stroll`

Duncaen:
It uses `stroll` by default but it looks like the tests in `ldb-1.1.27/lib/replace/wscript` are wrong or musls `strtoll` behaves like the bsd variant.

Vaelatern:
I'd believe it behaves like the BSD `stroll` and friends.

Duncaen:
The it should use the internal `rep_strtoll` function from `/ldb-1.1.27/lib/replace/replace.c`.

Vaelatern:
Yet it can't find it.

the-maldridge:
Its worth pointing out that the only consumer of this is sssd and that has insanely tight coupling to glibc.

PR #4599, opened 2016-08-12
python-Django: update to 1.10

the-maldridge:


the-maldridge:
Travis fails the build because it spits out too much log data, the build does actually pass.

ebfe:
Why not 1.10?

the-maldridge:
Didn't notice it was out.

Duncaen:
Why do we have this packaged at all?

the-maldridge:
@ebfe updated
@Duncaen its nice to have it packaged so that I'm not juggling a pip package and xbps packages.

the-maldridge:
Once again its worth pointing out that this does actually build successfully, but travis fails the build for excessive log output due to a number of empty directories.

PR #4609, opened 2016-08-13
New package: gngeo-0.8

mrshampoo:
unsure about the license. the COPYING says that gngeo uses different modules, so there are actually two licenses mentioned there. I hope I wright them in the right way in the template?


mrshampoo:
ops i notised that i forgot to rewright a line!

mrshampoo:
hum. checks failed... It builds just fine on my computer.
fgets.. any ides?

mrshampoo:
probably a dependency. i will google around a bit ;)
"gen68k.c:8:25: fatal error: def68k-iibs.h: No such file or directory

 #include "def68k-iibs.h" "

pullmoll:
Using `disable_parallel_build=yes` should solve the problems.

mrshampoo:
oh, okej, i was just about to test to put a CC=gcc -m32 if it turn out to be a 32bit system.
i will test your thing insted

mrshampoo:
okej that solved most of them. but i686 still fails some where at a fget line :/

mrshampoo:
on MAME.. it was a long time ago sens i used it. i was kinda frustrated of the whey they forced you too change all the filenames in the zip files back then. i wasent that keen on the look and feel of the emulator nether.. but maybe that has changed?
but does it hurt to have them both at hand in xbps?

mrshampoo:
ok, adding nasm seems mostly to generate a lots of warnings and does not solve the i686 compile.
nether did adding "pre_build(){ if ! ((1<<32)) ; then export CC="gcc -m32" fi }"
i removed the pre_build but let the nasm be, just to have it more clean, and it seems as it uses nasm (i doesn't in no other architecture doh :/ )
I'm taking a pause on this, because i am a bit clueless. I might come back later, ether if i found something, or if i just don't give a dam any more, just to return to remove the pull request. (it works for me, so this was mainly to give something more back to the void )

PR #4610, opened 2016-08-13
New package: python-xlwt-1.1.2

diogoleal:


Vaelatern:
Is this a dependency for something else?

diogoleal:
Yes, this and python-xlrd are dependencies for rows.
https://github.com/turicas/rows

Vaelatern:
Why not just `pip install`?

diogoleal:
I made all the changes you suggested.

diogoleal:
Is possible, but I don't like to use pip or other package manager the programming language.

the-maldridge:
@diogoleal can you drop the 3.4 from the function call and the link?  We use only the major version number now for that.  Ping me once you've made those changes, this looks pretty much ready to merge.

diogoleal:
@the-maldridge  Is that what you want?

the-maldridge:
@diogoleal not quite.  You need the python3 subpkg, but not the one that explicitly states 3.4 in the version number.  Add back the python3 subpkg and it should be fine, I think what you had was correct.

asergi:
Obsolete. Feel free to resubmit it.

PR #4623, opened 2016-08-15
emacs: add emacs-gtk3 variant

renatoaguiar:


chneukirchen:
This will be added with #4511.

PR #4625, opened 2016-08-16
wine-gecko: update to 2.47

kurumushi:
The current version of wine in void linux needs this newer version of wine-gecko.
The installer was not available on sourceforge, so I switched to downloading from the wine download site.

### DIFF
--- a/srcpkgs/wine-gecko/template
+++ b/srcpkgs/wine-gecko/template
@@ -1,14 +1,14 @@
 # Template file for 'wine-gecko'
 pkgname=wine-gecko
-version=2.40
+version=2.47
 revision=1
 noarch=yes
 build_style=fetch
-checksum=1a29d17435a52b7663cea6f30a0771f74097962b07031947719bb7b46057d302
-distfiles="${SOURCEFORGE_SITE}/wine/wine_gecko-${version}-x86.msi"
+checksum=3b8a361f5d63952d21caafd74e849a774994822fb96c5922b01d554f1677643a
+distfiles="http://dl.winehq.org/wine/${pkgname}/${version}/wine_gecko-${version}-x86.msi"
 homepage="http://wiki.winehq.org/Gecko"
 license="MPL-2"
-short_desc="Mozilla Gecko Layout Engine for WINE to replace Internet Explorer (32bit)"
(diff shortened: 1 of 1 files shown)


kurumushi:
xlint is complaining about the short_desc size. I did not make any changes there, I just left it as is. Should it be changed?

Vaelatern:
Pleaase avoid closing and re-opening pull requests. It is never necessary.

kurumushi:
Sorry about closing the old one. I realized I made big mistakes and wanted to start over and try again.

Vaelatern:
A force push with git will be able to overwrite anything you have in a git commit. In fact, we request that you do force-push a modified commit instead of adding more commits when there are more comments.

kurumushi:
So would you rather me close this request and open the old one? Also, should I change the description so xlint likes it? The current version in the repo has the exact same description.

Vaelatern:
No, keep this one, now that you are here. Leave the description, an xlint failure isn't that bad unless you can think of a shorter `short_desc` that means the same.

kurumushi:
I updated the shortdesc to remove the part that said "to replace internet explorer" it's probably not really needed in a description.

srcpkgs/wine-gecko/template
```diff
@@ -1,14 +1,14 @@
...
 noarch=yes
 build_style=fetch
-checksum=1a29d17435a52b7663cea6f30a0771f74097962b07031947719bb7b46057d302
-distfiles="${SOURCEFORGE_SITE}/wine/wine_gecko-${version}-x86.msi"
+checksum=3b8a361f5d63952d21caafd74e849a774994822fb96c5922b01d554f1677643a
+distfiles="http://dl.winehq.org/wine/${pkgname}/${version}/wine_gecko-${version}-x86.msi"
 homepage="http://wiki.winehq.org/Gecko"
 license="MPL-2"
-short_desc="Mozilla Gecko Layout Engine for WINE to replace Internet Explorer (32bit)"
+short_desc="Mozilla Gecko Layout Engine for WINE (31bit)"
```
Vaelatern: This is not 31 bit.

kurumushi:
Whoops. sorry for the dumb typo.

PR #4630, opened 2016-08-16
New package: cmst-2016.04.03

vjon:
This is a Qt based GUI with system tray icon functionality, which I believe will prove useful for users constructing a lightweight Qt desktop (e.g. LXQt, Lumina, etc.)

### DIFF
--- a/srcpkgs/cmst/template
+++ b/srcpkgs/cmst/template
@@ -1,0 +1,19 @@
+# Template file for 'cmst'
+pkgname=cmst
+version=2016.04.03
+revision=1
+wrksrc=${pkgname}-${pkgname}-${version}
+build_style=qmake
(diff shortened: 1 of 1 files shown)


srcpkgs/cmst/template
```diff
@@ -0,0 +1,20 @@
+# Template file for 'cmst'
+pkgname=cmst
+version=2016.04.03
+revision=1
+only_for_archs="i686 x86_64 i686-musl x86_64-musl"
```
Duncaen: remove `only_for_archs`.

vjon: That's strange, the build failed on my machine when I tried to cross-compile for arm, but it passed in Travis...

Issue #4606, opened 2016-08-12
flex 2.6.1


chneukirchen:
Project moved to github: https://github.com/westes/flex/releases

PR #4632, opened 2016-08-16
flex: update to 2.6.1.

subsonik:
#4606 

Should there be a convenience variable for GitHub, as listed here: https://github.com/voidlinux/void-packages/blob/master/Manual.md#optional_vars ?


Gottox:
For me github.com is okay. The idea behind those vars was that we could easily switch to a mirror. No idea if there's a central github mirror somewhere ;)

Gottox:
Please fix the arm platforms.

PR #4635, opened 2016-08-17
virtualbox-ose: update to 5.1.2

pullmoll:
This includes the switch to Qt5.


pullmoll:
A somehwat unrelated note:
For me USB does no longer work with VirtualBox ever since I got my new machine (MSI X99A SLI PLUS, Mainboard, Intel Core i7 6800K).

This was the case with `virtualbox-ose-5.0.24` and is still the case with `5.1.2`, so it is most probably not related to VirtualBox, but to the kernel or my setup of Void...

Still, if somebody would test that USB works for him with this update, that could avoid trouble.

PR #4638, opened 2016-08-18
virtualbox-ose: update to 5.1.4

pullmoll:
This one does not work for me with the brand new linux kernel 4.7.1.
After building the modules and when trying to load the `vboxdrv` module I see this in `dmesg` output:
```
[   79.279972] vboxdrv: Unknown symbol mod_timer_pinned (err 0)
[   79.280032] vboxdrv: Unknown symbol __get_vm_area (err 0)
[   79.280066] vboxdrv: Unknown symbol set_pages_nx (err 0)
[   79.280099] vboxdrv: Unknown symbol smp_call_function (err 0)
[   79.280157] vboxdrv: Unknown symbol map_vm_area (err 0)
[   79.280202] vboxdrv: Unknown symbol set_pages_x (err 0)
```
So either the kernel is too new for vbox, or we do not have set some config options which are required.

Do not merge until this is solved.

### DIFF
--- a/srcpkgs/virtualbox-ose/patches/005-gsoap-build.patch
+++ b/srcpkgs/virtualbox-ose/patches/005-gsoap-build.patch
@@ -1,11 +1,11 @@
---- src/VBox/Main/webservice/Makefile.kmk	2013-11-21 19:00:46.812761628 +0100
-+++ src/VBox/Main/webservice/Makefile.kmk	2014-01-04 04:37:05.803599026 +0100
-@@ -683,7 +683,7 @@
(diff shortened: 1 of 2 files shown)


pullmoll:
Note: Both versions of virtualbox-ose (5.1.2 and 5.1.4) work with linux kernel 4.6.7, and both don't work with kernel 4.7.1.

Looks like there is something missing in `srcpkgs/linux4.7/files/config/x86_64-dotconfig`

pullmoll:
Merge after #4640

PR #4640, opened 2016-08-18
linux4.7: disable stripping of ksyms

pullmoll:
Disable the config setting CONFIG_TRIM_UNUSED_KSYMS, or otherwise
the kernel won't export symbols which are required to make
dkms modules built for virtualbox-ose (probably others) load.

This is required for #4638 and also virtualbox-ose-5.1.2 (current).


paoloschi:
>  (probably others) load

to confirm a further case:  my TeVii S464 PCI DVB-S card had stopped working with `4.7` (up to `4.7.1_1`) due to no more loading of cx88xx modules..
the new configuration in `4.7.1_2` reported the card to work again!
thx

PR #4641, opened 2016-08-18
mysql: update to 5.7.14

fishbach:


Duncaen:
I think this should be delayed until boost is merged.
And we should probably check if we have to run something to migrate existing databases or keep 5.6 in the repository and add the version 5.6 or 5.7 to the package name.

fishbach:
I have tried to update boost as well. It has worked quite fine for Boost 1.61.0, which is the current version.
Unfortunately does MySQL need precisely Boost 1.59. I have tried to patch MySQL to use Boost 1.61. But MySQL has modified and copied some of the Boost headers and their cloned headers do not work with with Boost 1.61.
To upgrade Boost to 1.59 just to make MySQL compile, is silly. That's why I have decided to pass the flags to cmake to let it download the preferred version just for compilation.

I like your idea to make an own packet for MySQL 5.7!
Shall I do that?
Just that I do not understand why the download (-DDOWNLOAD_BOOST=1 -DWITH_BOOST=...) did not work for arm and musl.

btw: Many thanks for that great Linux distro! :)

PR #4646, opened 2016-08-20
New package: openbabel-2.3.2

diogoleal:


srcpkgs/openbabel/template
```diff
@@ -0,0 +1,28 @@
+# Template file for 'openbabel'
+pkgname=openbabel
+version=2.3.2
+revision=1
+_ver=2-3-2
+wrksrc=${pkgname}-${pkgname}-${_ver}
+build_style=cmake
+hostmakedepends="cmake pkg-config eigen"
```
Vaelatern: `cmake` not necessary

srcpkgs/openbabel/template
```diff
@@ -0,0 +1,28 @@
...
+makedepends="cairo-devel libxml2-devel wxWidgets-devel"
+short_desc="The Open Source Chemistry Toolbox"
+maintainer="Diogo Leal <diogo@diogoleal.com>"
+license="GPL-2"
+homepage="http://openbabel.org"
+distfiles="https://github.com/openbabel/openbabel/archive/openbabel-${_ver}.tar.gz"
+checksum=cad75883dc2a848fcf1addfce6eeb516d531f783bb81395fc0a41e079b00342c
+nocross=yes
```
Vaelatern: Can you provide a link to a failing test at least? Travis links are good for this. Makes `nocross=link_here`

srcpkgs/openbabel/template
```diff
@@ -0,0 +1,28 @@
+# Template file for 'openbabel'
+pkgname=openbabel
+version=2.3.2
+revision=1
+_ver=2-3-2
```
Vaelatern: Can you replace this with `_ver=${version//./-}`

srcpkgs/openbabel/template
diogoleal: nocross=yes is because:
 [target] wxWidgets-devel-3.0.2_3: not found.
=> ERROR: wxWidgets-devel-3.0.2_3: cannot be cross compiled, exiting...

srcpkgs/openbabel/template
diogoleal: Ah! thanks!

srcpkgs/openbabel/template
Vaelatern: Then don't bother with adding nocross here.
If a dependency can't cross build, who is to say that when we do have `wxWidgets-devel` available for all platforms that this program can't cross build? For now, just remove it, thanks.

diogoleal: Okay, nocross removed.

srcpkgs/openbabel/template
```diff
@@ -0,0 +1,27 @@
...
+makedepends="cairo-devel libxml2-devel wxWidgets-devel"
+short_desc="The Open Source Chemistry Toolbox"
+maintainer="Diogo Leal <diogo@diogoleal.com>"
+license="GPL-2"
+homepage="http://openbabel.org"
+distfiles="https://github.com/openbabel/openbabel/archive/openbabel-${_ver}.tar.gz"
+checksum=cad75883dc2a848fcf1addfce6eeb516d531f783bb81395fc0a41e079b00342c
+
+openbabel-devel_package() {
+		depends="${sourcepkg}>=${version}_${revision}"
```
xtraeme: please fix the indentation.

diogoleal: fixed!

Vaelatern:
Are you really sure you want to put all of `/usr/bin` in the `-devel` subpackage, thus making them unavailable for anybody who expects to run those binaries from the normal package?

srcpkgs/openbabel/template
```diff
@@ -0,0 +1,27 @@
...
+checksum=cad75883dc2a848fcf1addfce6eeb516d531f783bb81395fc0a41e079b00342c
+
+openbabel-devel_package() {
+	depends="${sourcepkg}>=${version}_${revision}"
+	short_desc+=" - development files"
+	pkg_install() {
+		vmove usr/include
+		vmove "usr/lib/*.so"
+		vmove usr/share
+		vmove usr/bin
```
Vaelatern: You sure? Also you may want to also move `usr/lib/cmake`

diogoleal: Ops. fixed. my mistake, sorry!

Vaelatern: When I `xlocate usr/lib/cmake` I find most of them are in `-devel` packages. I suggest you vmove that as well.

diogoleal: done, vmove usr/lib/cmake added in -devel package.

srcpkgs/openbabel/template
```diff
@@ -0,0 +1,27 @@
...
+checksum=cad75883dc2a848fcf1addfce6eeb516d531f783bb81395fc0a41e079b00342c
+
+openbabel-devel_package() {
+	depends="${sourcepkg}>=${version}_${revision}"
+	short_desc+=" - development files"
+	pkg_install() {
+		vmove usr/include
+		vmove "usr/lib/*.so"
+		vmove usr/lib/cmake
+		vmove usr/share
```
Vaelatern: Looking at what's in `usr/share`, it seems static assets and man pages, which should not be in a `-devel` package.

diogoleal: I removed ```vmove usr/share```

PR #4654, opened 2016-08-22
neovim: update to 0.1.5

equalsraf:
- Removed dependency on luajit, it is no longer needed.
- Removed the helptags patch. Instead use host nvim to
  generate helptags when cross compiling.


stpx:
I'm not too crazy about adding neovim as a build dep

Vaelatern:
I'm hotter about that than I am about a big binary patch file.

stpx:
I agree. I wasn't crazy about that patch either. I'm AFK, at work right now. I'll be home in a few.

Issue #4225, opened 2016-05-23
nss update seems to have broken Chromium and Firefox


Rehwanne:
chromium just doesnt start:

```
[23818:23818:0524/010417:ERROR:platform_thread_posix.cc(116)] pthread_create: Invalid argument
[23818:23818:0524/010417:ERROR:chrome_browser_main_posix.cc(290)] Failed to create shutdown detector task.
[23818:23838:0524/010417:ERROR:nss_util.cc(740)] Error initializing NSS with a persistent database (sql:/home/rehwanne/.pki/nssdb): NSS error code: -8023
[23818:23838:0524/010417:ERROR:nss_util.cc(211)] Error initializing NSS without a persistent database: NSS error code: -8023
[23818:23838:0524/010417:FATAL:nss_util.cc(213)] nss_error=-8023, os_error=0
#0 0x55c032143d9e <unknown>
#1 0x55c032157a0e <unknown>
#2 0x55c0321d7d05 <unknown>
#3 0x55c0321d8030 <unknown>
#4 0x55c0323cac15 <unknown>
#5 0x55c03241a62d <unknown>
#6 0x55c031e4448d <unknown>
#7 0x55c0349071ab <unknown>
#8 0x55c034addb5f <unknown>
#9 0x55c032189c85 <unknown>
#10 0x55c0321867fd <unknown>
#11 0x7fd6d9af13f4 start_thread
#12 0x7fd6d7f66bad __clone

```

and firefox cant handle https:// urls anymore:
```

1464044895947	addons.manager	ERROR	Exception loading default provider "resource://gre/modules/addons/XPIProvider.jsm": [Exception... "Component returned failure code: 0x80570016 (NS_ERROR_XPC_GS_RETURNED_FAILURE) [nsIJSCID.getService]"  nsresult: "0x80570016 (NS_ERROR_XPC_GS_RETURNED_FAILURE)"  location: "JS frame :: resource://gre/modules/addons/XPIProvider.jsm :: <TOP_LEVEL> :: line 1679"  data: no] Stack trace: resource://gre/modules/addons/XPIProvider.jsm:1679 < AddonManagerInternal.startup()@resource://gre/modules/AddonManager.jsm:969 < this.AddonManagerPrivate.startup()@resource://gre/modules/AddonManager.jsm:2783 < amManager.prototype.observe()@resource://gre/components/addonManager.js:56
*************************
A coding exception was thrown and uncaught in a Task.

[...]

PR #4655, opened 2016-08-22
[WIP] Firefox 48.0.1

Duncaen:
Firefox depends on `nss>=3.25` which is still broken for some unknown and hard to debug reason...
#4225


asergi:
`nss` updated and fixed(?) in 68c853e.

chneukirchen:
@asergi FREEBL_NO_DEPEND is the magic? can you explain how you found it and what it does?

asergi:
I found it reading the `nss/coreconf/Linux.mk` file. After googling, `FREEBL_NO_DEPEND=1` seems to avoid clashes with glibc, but I didn't have time to dig in and really figure out what's going on.

PR #4664, opened 2016-08-26
mergerfs: update to 2.15.0

xinterer:


Gottox:
Please rebase this PR against current master and squash all commits into one.

Vaelatern:

PR #4674, opened 2016-08-27
New package: nload-0.7.4

jnbrn:


srcpkgs/nload/template
```diff
@@ -0,0 +1,14 @@
+# Template file for 'nload'
+pkgname=nload
+version=0.7.4
+revision=1
+build_style=gnu-configure
+configure_args="--prefix=/usr"
```
Duncaen: You can remove this line, `--prefix=/usr` is default for `build_style=gnu-configure`.

srcpkgs/nload/template
```diff
@@ -0,0 +1,14 @@
...
+makedepends="ncurses-devel"
+depends="ncurses"
```
Duncaen: Is this really needed? xbps-src detects `ncurses-libs` as dependency automatically: https://travis-ci.org/voidlinux/void-packages/jobs/155503667#L794
`ncurses` is only needed if it needs the following binaries `infocmp, clear, tset, tic, tput, tabs, toe, infotocap, reset, captoinfo`.

PR #4676, opened 2016-08-27
New package: avogadro-1.2.0

diogoleal:


srcpkgs/avogadro/template
```diff
@@ -0,0 +1,30 @@
+# Template file for 'avogadro'
+pkgname=avogadro
+version=1.2.0
+revision=1
+build_style=cmake
+configure_args="-DCMAKE_BUILD_TYPE:STRING=Release -DCMAKE_INSTALL_PREFIX:PATH=/usr -DENABLE_TESTING:BOOL=OFF"
```
Duncaen: Did you try it without defining built_type and install_prefix as they are default https://github.com/voidlinux/void-packages/blob/master/common/build-style/cmake.sh#L26

diogoleal: Sorry, removed build_type and install_prefix.

PR #4678, opened 2016-08-28
flex: update to 2.6.1.

dutsky:


dutsky:
Oh, sorry didn't notice #4632 request before :/

Vaelatern:
Why do you make different changes than the ones in #4632 ? And can you fix the arm platforms?

PR #4680, opened 2016-08-28
Add fish's dependency "bc"

awalGarg:
fish depends on bc for the shipped autocompletion files. See https://github.com/fish-shell/fish-shell/blob/master/fish.spec.in#L16

Without bc, typing `sudo x` and hitting tab to autocomplete from the fish-shell with throw an error from the completion file stating "command bc not found". Manually installing bc fixes this - but it should be a package dependency :)

### DIFF
--- a/srcpkgs/fish-shell/template
+++ b/srcpkgs/fish-shell/template
@@ -1,10 +1,11 @@
 # Template file for 'fish-shell'
 pkgname=fish-shell
 version=2.3.1
-revision=1
+revision=2
 build_style=gnu-configure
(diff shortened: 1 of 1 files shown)


Vaelatern:

Vaelatern:
Also please bump revision

Vaelatern:
Seems a travis hiccup. Possible a dependency on python is requried, but since fish doesn't seem to note which version, and we don't yet have virtual packages that I am fully aware of, this will be merged.

PR #4681, opened 2016-08-28
New package: angband-4.0.5

diogoleal:


srcpkgs/angband/template
```diff
@@ -0,0 +1,17 @@
+# Template file for 'angband'
+pkgname=angband
+version=4.0.5
+revision=1
+build_style=gnu-configure
+hostmakedepends="autoconf automake"
```
Vaelatern: `automake` depends on `autoconf`. Only pull in `automake`

srcpkgs/angband/template
```diff
@@ -0,0 +1,17 @@
...
+makedepends="SDL-devel SDL_image-devel SDL_mixer-devel"
```
Vaelatern: `SDL-devel` is redundant.

diogoleal: Removed SDL-devel and autoconf

PR #4685, opened 2016-08-29
[WIP] glib: fix quark init on demand patch

pullmoll:


pullmoll:
Hmm.. it worked for me locally - once - now it does not. Investigating.

pullmoll:
Closing because it does not really work.

PR #4691, opened 2016-08-29
netdata: update to 1.3.0

the-maldridge:
Technically this also depends on python now, but only for additional plugins.  Given that the core system still works without it I have not added python as a hard depends relation to this package.

### DIFF
--- a/srcpkgs/netdata/template
+++ b/srcpkgs/netdata/template
@@ -1,20 +1,40 @@
 # Template file for 'netdata'
 pkgname=netdata
-version=1.1.0
-revision=3
+version=1.3.0
+revision=1
 build_style=gnu-configure
 hostmakedepends="pkg-config"
 makedepends="zlib-devel"
 system_accounts="_netdata"
 conf_files="/etc/${pkgname}/*.conf"
-configure_args="--with-user=_netdata --with-webdir=/usr/share/${pkgname}/webroot"
+configure_args="--with-user=_netdata"
+makedepends="libuuid-devel zlib-devel"
 short_desc="Real-time performance monitoring, done right"
 maintainer="Michael Aldridge <aldridge.mac@gmail.com>"
 license="GPL-3"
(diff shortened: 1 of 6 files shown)


Vaelatern:
Can you speak to this part of the build log output?

```
ATTENTION
/usr/libexec/netdata/plugins.d/apps.plugin requires escalated capabilities:
sudo chown root '/destdir//netdata-1.3.0/usr/libexec/netdata/plugins.d/apps.plugin'
sudo chmod 0755 '/destdir//netdata-1.3.0/usr/libexec/netdata/plugins.d/apps.plugin'
sudo setcap cap_dac_read_search,cap_sys_ptrace+ep '/destdir//netdata-1.3.0/usr/libexec/netdata/plugins.d/apps.plugin'
```

Vaelatern:
Request additional modules, including that install script that seems to be there, be moved to a subpackage, and that package be given a hard python dependency.

the-maldridge:
Modules have been made, charts.d is a core module that doesn't make sense to factor out.  The apps.plugin module needs certain extra permissions to be able to perform certain types of inspection on the system that are normally reserved and such access should not be conferred to the _netdata user.

srcpkgs/netdata/template
```diff
@@ -1,20 +1,40 @@
...
 distfiles="http://firehol.org/download/netdata/releases/v${version}/${pkgname}-${version}.tar.gz"
-checksum=b1967cf2a9f2ddc5798c0576eb6861d3e5bac0665b50077ce784b671c1e55180
+checksum=696772c5761354d3ff395c1fcf5eb17eab83937a959e3ebf945288ca8b6d0ed5
 
 post_install() {
 	vsv netdata
 }
+
+netdata-plugins-node_package() {
+        short_desc+=" - Node Plugins"
```
Vaelatern: Please specify `nodejs` in these instances

PR #4693, opened 2016-08-30
glib: fix on demand init for gquark and gtype

pullmoll:
Besides the quark init on demand, a similar thing is required for `gobject_init()` in `gobject/gtype.c` for the commands like `g-ir-scan` etc. to work right with musl libc.

This fixes NetworkManager `build_option="gir"` for musl libc.

### DIFF
--- a/srcpkgs/NetworkManager/template
+++ b/srcpkgs/NetworkManager/template
@@ -1,7 +1,7 @@
 # Template file for 'NetworkManager'
 pkgname=NetworkManager
 version=1.0.12
-revision=4
+revision=5
 build_style=gnu-configure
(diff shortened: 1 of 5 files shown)


pullmoll:
A question remaining is if these changes should be scoped by `#if !defined(__GLIBC__)` … `#endif` because for glibc the code works without the overhead of on demand initializations.

pullmoll:
The x86_64 and armv[67]l failures on travis don't happen here. They may be related to `LC_ALL ...` warnings.

pullmoll:
No comment means merge?
This fix is required to fix the `gnome-shell` PIE build, because `gnome-shell` depends on `NetworkManager-1.0.gir`.

Vaelatern:
Nothing glaringly wrong that I could see

PR #4697, opened 2016-09-01
Initial pamixer template

teldra:


Vaelatern:

Please fix your xlint errors (see the Travis details for more information)

PR #4698, opened 2016-09-01
New package: pamixer

teldra:


Vaelatern:

**Only one PR, please, unless you are told otherwise. Unrelated features, such as if you fixed something in `firefox` and added a new spotify client, should go in separate PRs. If you wanted to package something, and it depended on a different package not yet in Void, you would put both of those package additions in the *same* PR. Please do not close and open new PRs just because you made changes. Learn enough git commands, specifically 'rebase' and 'push' with the '--force' flag to not need to do this**.

teldra:
Sry, somthing br0ke horrendous.

Vaelatern:
Okay. Rebase to name the commit correctly, then force push? I think a `commit --amend` would do to rename it actually, to match the manual.
typically a rebase on top of the master branch will fix most ills. Rebase a few times and most everything fixes itself.

Vaelatern:
Message of the commit, not name of the issue, needs to conform.

teldra:
Is this correct?

Vaelatern:
Yes

srcpkgs/pamixer/template
```diff
@@ -0,0 +1,16 @@
+# Template file for 'pamixer'
+pkgname=pamixer
+version=1.3.1
+revision=1
+build_style=gnu-makefile
+makedepends="pulseaudio-devel boost-devel"
+short_desc="Pulseaudio command line mixer "
```
Vaelatern: Please don't end in whitespace either.

PR #4699, opened 2016-09-01
New package: py3status

teldra:


Vaelatern:

Please do not close this PR just because you are going to make changes. This loses history of what you were asked to do, and spams the developers.

srcpkgs/py3status/template
```diff
@@ -0,0 +1,16 @@
+# Template file for 'py3status'
+pkgname=py3status
+version=3.0
+revision=1
+short_desc="Alternative i3bar"
+homepage=htpp://i3wm.org
```
Vaelatern: This is not a valid URL. Please also put this just above the `distfiles` variable, near the bottom of the template, for the sake of consistency with other packages.

srcpkgs/py3status/template
```diff
@@ -0,0 +1,16 @@
...
+license=GPL-3
+noarch=yes
+build_style=python-module
+python_versions="2.7"
+pycompile_version="2.7"
```
Vaelatern: Neither this line nor the line before it are necessary

srcpkgs/py3status/template
```diff
@@ -0,0 +1,16 @@
+# Template file for 'py3status'
+pkgname=py3status
+version=3.0
+revision=1
+short_desc="Alternative i3bar"
+homepage=htpp://i3wm.org
+license=GPL-3
```
Vaelatern: `BSD`. Also add the following function to the bottom of the file:
```
post_install() {
    vlicense LICENSE
}

File: srcpkgs/py3status/template (hunk 2 above)
Comment (teldra) [inline srcpkgs/py3status/template]: You mean: python_versions="2.7" and pycompile_version="2.7"?

I am just not sure, if your comment relates to the last lines of shown code.

File: srcpkgs/py3status/template [hunk 4]
```diff
@@ -3,14 +3,16 @@ pkgname=py3status
...
-pycompile_version="2.7"
 pycompile_module="py3status"
 hostmakedepends="python-setuptools python-devel"
 depends="python"
+homepage=https://github.com/ultrabug/py3status
 distfiles="https://github.com/ultrabug/py3status/archive/${version}.tar.gz"
 checksum=f2c0e9c3ecb0bd1c40e952b16255c97bba184de14629b74c0203feb873a3df21
+
+do_install() {
+	vlicense BSD
```
Vaelatern: See the travis builds for why this is wrong. `vlicense path/to/license/file/with/cwd/being/wrksrc` or in this case `vlicense LICENSE`

srcpkgs/py3status/template
Vaelatern: Confirmed.

srcpkgs/py3status/template
```diff
@@ -0,0 +1,18 @@
...
+revision=1
+short_desc="Alternative i3bar"
+license=BSD
+noarch=yes
+build_style=python-module
+pycompile_module="py3status"
+hostmakedepends="python-setuptools python-devel"
+depends="python"
+homepage=https://github.com/ultrabug/py3status
+distfiles="https://github.com/ultrabug/py3status/archive/${version}.tar.gz"
```
Vaelatern: `${homepage}` here.

Vaelatern:

Vaelatern:
Two things. One: That is the name of the issue, not the name of the commit itself. You have 5 commits here. Two: It is still missing the version.

teldra:
ohoh .. :(

Vaelatern:
Please rebase on master

Vaelatern:
1 commit

Vaelatern:
Why does it not install any files except a license?

srcpkgs/py3status/template
```diff
@@ -0,0 +1,19 @@
...
+build_style=python-module
+pycompile_module="py3status"
+hostmakedepends="python-setuptools python-devel"
+depends="python"
+maintainer="xo <oh@rotce.de>"
+homepage=https://github.com/ultrabug/py3status
+distfiles="https://github.com/ultrabug/py3status/archive/${version}.tar.gz"
+checksum=f2c0e9c3ecb0bd1c40e952b16255c97bba184de14629b74c0203feb873a3df21
+
+do_install() {
```
Vaelatern: `post_install()` instead and see what happens

Vaelatern:
Ah.

PR #4700, opened 2016-09-02
New Package: xss-lock-0.3.0

teldra:


srcpkgs/xss-lock/template
```diff
@@ -0,0 +1,22 @@
+pkgname=xss-lock
+version=0.3.0
+revision=2
```
Vaelatern: `revision=1` until the package is merged.

srcpkgs/xss-lock/template
```diff
@@ -0,0 +1,22 @@
+pkgname=xss-lock
```
Vaelatern: First line in file must be a comment. See most other templates for an example.

srcpkgs/xss-lock/template
```diff
@@ -0,0 +1,22 @@
...
+version=0.3.0
+revision=2
+short_desc="Use external locker as X screen saver"
+license=MIT
+build_style=cmake
+hostmakedepends="cmake python-docutils pkg-config libglib-devel xcb-util-devel"
+depends="xcb-util"
+wrksrc=raymonad-xss-lock-1e158fb20108
+homepage=https://bitbucket.org/raymonad/xss-lock
+distfiles="https://bitbucket.org/raymonad/xss-lock/get/master.tar.gz"
```
Vaelatern: No. Only versioned packages are accepted.

teldra: OK.
Just for information: Would it be okay, if i fork that project, make a release and later change back to original source, if there is once a release. Because the last commit ist most relevant.

Vaelatern: We prefer you ask upstream for a release.

PR #4707, opened 2016-09-03
New package: pipes.sh-1.2.0

diogoleal:


Vaelatern:
`noarch=yes`

Vaelatern:
To not waste your time: It's a shell script. It's very easy to install as a user. There is no purpose to having this be a global program except for particularly amused system administrators who can leverage `/usr/local/bin`. Unless a case can be made, this will not at this time be accepted for inlcusion to Void.

PR #4711, opened 2016-09-04
gthumb: add INSTALL.msg file

felix:


Vaelatern:
This should come with a revision bump to `gthumb`.
Why do you not recommend installing `dcraw` as a `gthumb` dependency?

felix:
gthumb does not require dcraw and if the user does not process RAW files then it is an extraneous dependency. I have no idea what Void policy on this is though.

felix:
@Vaelatern do you want it as a required dep?

felix:
@Vaelatern ping

ebfe:
@felix there was still a merge conflict. Rebased locally and applied here e2200c706c91c5ff0db397a21ef52921dc7e8bf5..

PR #4716, opened 2016-09-05
Add package OpenMW and dependencies.

jantatje:
Note: more packages templates following soon.

Required fix for cmake when cmake_builddir="./":
Some applications need to be build from inside the source tree,
instead of a separate build directory, use a absolute path because
the previous approach using sed to build a relative path was broken
when building in-tree, producing "../" for a build dir of "./".

Please give feedback so we can find the best solution for this problem.


ebfe:
Maybe split out the cmake build-style change to a separate PR, so it can be discussed separately? The obvious question is, do the other build_style=cmake templates till work?

PR #4717, opened 2016-09-05
New package: simplenote-1.0.3

rafeyu:
Simplenote is simple application to save important notes, lot of TODOs etc, from Automattic.

I build this from binary version. Build simplenote from source which uses node and Electron isn't useful enough, because users will not get share and publish features. And they should create Simperium account to make it "works".

See https://github.com/Automattic/simplenote-electron/blob/master/README.md

Edit: more clear description


rafeyu:
I just get a clue, use `nopie=yes` instead of skip the strip checks with `nostrip_files="Simplenote"`

the-maldridge:
Thumbs-downed solely because its a binary package that pulls in more node dependencies.  In general we try to avoid binary packages as a whole, but doubly so with anything that touches the nodejs ecosystem.

Vaelatern:
Does anybody want to work on getting this of quality (using non-binary builds, maybe making it PIE, finding an easy way to update in event that it breaks our builds like other Electron apps do)?

pullmoll:
I am clueless regarding nodejs, while it seems to be simple enough for a user to download and install a tarball in /opt. Perhaps the wiki would be a good place to describe the install process and perhaps include a shell script that does it.

I agree with @the-maldridge in that we want to **reduce** the number of binary only packages.

Vaelatern:
We all agree that binary only packages should be minimized, and that this package as binary-only really is not ideal in any way.
I agree with @pullmoll  that for such things as this, in binary form, installing into `/opt`, install guides on the wiki are the best way to go.

rafeyu:
Thanks @the-maldridge, @Vaelatern, and @pullmoll, I agree to reduce binary builds and I just know nodejs/electron apps are hard to maintain without breaks another electron apps. I don't mind if this PR closed, I'll drop this later.

PR #4718, opened 2016-09-05
openvpn: add sv script.

### LINKED ISSUE #4705: OpenVPN packages doesn't provide a service script
When installing openvpn using `xbps-install openvpn` a runinit script is supposed to be installed in `/etc/sv`, but that doesn't happen.

Nothing OpenVPN related gets installed in `/etc/sv` and as such it's impossible to run the service.

Gottox:
fixes #4705


ebfe:
Hmm, there is no /etc/openvpn/openvpn.conf and overwrites existing sv scripts on update. Does more harm than good?

skurtix:
1) This would result in openvpn not being stopped properly resulting in multiple processes on restart
2) It wouldn't automatically create (or remove) a tunnel unless there's a hook for it

SevereOverfl0w:
This might make more sense as an INSTALL message, given that the config location will be decided by the user (although the norm is /etc/openvpn/configname.conf)

the-maldridge:
I'm in favor of merging this, even given the possibility of overwriting an existing file.

Gottox:
Adding an install message?

ebfe:
I still think it's not worth it, since there is no sane default config. At least you could not hardcode the config file.

PR #4728, opened 2016-09-07
New package: heyu-2.10

the-maldridge:


the-maldridge:
Something is very broken at the moment.  Pointers appreciated for what I'm missing.  I know I've seen this error before but don't remember where.

pullmoll:
You need to make sure the $CFLAGS and $LDFLAGS are being passed to configure. By writing your own  `do_configure()` without these flags, you omit this.

You should instead use `configure_script="Configure"` and `configure_args="linux"`.

Also please remove the comment lines with things you don't need.

the-maldridge:
Everything seems to be working fine now, though the install function is not portable should files be added in future versions (though the existence of future versions is somewhat unlikely).

PR #4730, opened 2016-09-07
Update INSTALL due to typos

mpwp:
Wifi wasn't working for me, this change brought it back!


Vaelatern:
Nope. INSTALL paths are all relative. Besides, your commit message would need to be formatted correctly.

mpwp:
Sorry for the formatting, I don't know what happened (first commit ever...).

So can you tell me how to find: why the ownership, group and mode of 'dbus-daemon-launch-helper' were wrong?
It's strange, I'm not a pro at that level (as you may have noticed), it's not due to an unfortunate change for my part.

Vaelatern:
Any number of reasons can make permissions be wrong, but this script doesn't appear to be the right one.

PR #4733, opened 2016-09-08
jbigkit: create shared library + splix: fix PIE

pullmoll:
Created a "fork" of jbigkit-2.1 at github which uses autotools
and libtool to create both, shared and static libraries.

Parallel building fails because test programs which depend on the
libraries are built before these (not yet installed) libs.
Therefore disable_parallel_build=yes

### DIFF
--- a/common/shlibs
+++ b/common/shlibs
@@ -2617,3 +2617,5 @@ libcourier-unicode.so.3 courier-unicode-1.4_1
 libzstd.so.1 zstd-1.0.0_1
 libudis86.so.0 udis86-devel-1.7.2_3
 libsass.so.0 libsass-3.3.6_1
+libjbig.so.0 jbigkit-libs-2.1_2
+libjbig85.so.0 jbigkit-libs-2.1_2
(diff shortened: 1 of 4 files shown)


pullmoll:
cc @thypon - what do you think of this?
We can fork or move the https://github.com/pullmoll/jbigkit-shared repo to voidlinux.

thypon:
I think a patch upon jbigkit is better then a complete jbigkit fork, but it's also ok with the forking approach.

pullmoll:
I tried to patch before, using a modification of Debian's patch set, and it failed for cross builds.
Then I tried with the autotools and got successful cross builds...

thypon:
:+1:

Vaelatern:
Any chance of PRing your improvements upstream?

pullmoll:
I don't believe Markus Kuhn is in any way interested in automake + libtool and there is no repository for JBIG-KIT, so probably no.

Debian has its own patch set on top of jbigkit-2.1 and they certainly had asked upstream for inclusion... the fact that they maintain their own patches somehow suggests we are on our own as well.

OTOH if you `grep -r jbigkit srcpkgs` there is exactly one pkg depending on it :)

PR #4745, opened 2016-09-11
New package: oksh-0.5.9

diogoleal:


srcpkgs/oksh/template
```diff
@@ -0,0 +1,18 @@
...
+revision=1
+build_style=gnu-configure
+register_shell="/bin/ksh"
+short_desc="OpenBSD's version of ksh ported to Linux"
+maintainer="Diogo Leal <diogo@diogoleal.com>"
+license="GPL-3"
+homepage="https://connochaetos.org/oksh/"
+distfiles="http://www.connochaetos.org/oksh/oksh-${version}.tar.gz"
+checksum=2323be6676c88bdc916dd206b9e9bd6c23bbada513edaf71dcfad97e8df3c952
+conflicts="ksh>=0"
```
Vaelatern: Please make it alternatives instead?

diogoleal: yep, changed!
```
alternatives="ksh:/usr/bin/ksh"
```

chneukirchen:
That alternatives= is wrong, install oksh as /usr/bin/oksh and then point the alternatives correctly.

Then we need to adjust the ksh and mksh templates.

chneukirchen:
Other ksh adjusted.

srcpkgs/oksh/template
Vaelatern: Not correct format.

diogoleal:
I fixed alternatives line and name binary e manpage.

srcpkgs/oksh/template
```diff
@@ -0,0 +1,22 @@
+# Template file for 'oksh'
+pkgname=oksh
+version=0.5.9
+revision=1
+build_style=gnu-configure
+register_shell="/bin/ksh"
```
chneukirchen: I think this should be oksh now too? Tricky...

ebfe: To cover all cases it should add both?

ebfe: ... which won't work either since removing `ksh` will also unregister /bin/ksh.

Duncaen: But it switches to the next available alternative in this case.

PR #4757, opened 2016-09-13
libvpx: update to 1.6.0. (soname bump)

ebfe:


pullmoll:
It seems we have a problem with `glibc-locales`, see: https://travis-ci.org/voidlinux/void-packages/jobs/159695462#L588

ebfe:
Unfortunately that's the new normal for travis builds. I never could reproduce locally. Currently running test builds locally. @duncaen managed to get a coredump but that wasn't really enlightening on whats happening.

pullmoll:
Hmm.. the travis-ci builds use proot while locally I use uchroot.
Another difference is the xbps-static being used AFAICT.

pullmoll:
Yup, it's the `XBPS_CHROOT_CMD=proot` which causes the segfault. Confirmed locally.

pullmoll:
The `proot-x86_64` which `common/travis/prepare.sh` downloads is version 5.1.0 (`strings proot-x86_64`), same a our proot pkg. No update since Dec, 18 2014.

Perhaps we can create a `xbps-uchroot-x86_64` static executable and use that instead of proot?

ebfe:
> Perhaps we can create a `xbps-uchroot-x86_64` static executable and use that instead of proot?

Not sure if the kernel used on travis has the required options enabled.

pullmoll:
True. There are some [PRs pending for proot](https://github.com/proot-me/PRoot/pulls) and specifically https://github.com/proot-me/PRoot/pull/91 looks like it would fix a bug that could cause such problems.

pullmoll:
Are the kernel config options accessible for a user somehow? Then we could print info about the (non-)existance of `CONFIG_NAMESPACES, CONFIG_IPC_NS, CONFIG_PID_NS, CONFIG_UTS_NS` in setup.sh or even decide whether or not to use xbps-uchroot.

ebfe:
> Are the kernel config options accessible for a user somehow? 

Maybe zcat /proc/config.gz works? But that depends on the kernel config
;)

pullmoll:
Yeah, or `cat /boot/config-$(uname -r)` .. even less probable :)

pullmoll:
Let's wait for @xtraeme to pop up again and see if he knows what and how to do :)

pullmoll:
This PR lgtm and I didn't mean to wait for Juan to merge this :)

ebfe:
Build tested everything on x86_64.

PR #4767, opened 2016-09-15
opam: unbreak build process and add unzip to dependencies

jnbr:
In opam-1.2.2 there is a broken url in a Makefile, breaking the build process.
This gets fixed in post_extract.

Also, opam depends on unzip, so it is added as a dependency.

### DIFF
--- a/srcpkgs/opam/template
+++ b/srcpkgs/opam/template
@@ -1,12 +1,12 @@
 # Template file for 'opam'
 pkgname=opam
 version=1.2.2
-revision=3
+revision=4
 build_style=gnu-configure
 disable_parallel_build=yes
 make_build_args="lib-ext all"
 makedepends="ncurses-devel ocaml curl"
-depends="curl patch"
+depends="curl patch unzip"
 short_desc="OCaml package manager"
 maintainer="Christian Neukirchen <chneukirchen@gmail.com>"
 license="LGPL-3"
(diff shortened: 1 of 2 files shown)


srcpkgs/opam/template
```diff
@@ -15,3 +15,9 @@ distfiles="https://github.com/ocaml/${pkgname}/archive/${version}.tar.gz"
 checksum=3e4a05df6ff8deecba019d885ebe902eb933acb6e2fc7784ffee1ee14871e36a
 nocross=yes
 nopie=yes
+
+post_extract() {
+	# fix broken url in src_ext/Makefile
+	sed -i -e 's/http:\/\/ocaml-extlib.googlecode.com\/files\/extlib-1.5.3.tar.gz/http:\/\/storage.googleapis.com\/google-code-archive-downloads\/v2\/code.google.com\/ocaml-extlib\/extlib-1\.5\.3\.tar\.gz/' \
```
Vaelatern: This line of sed would be much cleaner thus:
`sed -i -e 's|http://ocaml-extlib.googlecode.com/files/extlib-1.5.3.tar.gz|http://storage.googleapis.com/google-code-archive-downloads/v2/code.google.com/ocaml-extlib/extlib-1\.5\.3\.tar\.gz|`
That said, please instead `mkdir srcpkgs/opam/patches`
and create a patch (using `diff -u`) in that folder, with the filename `fix_url.patch`.
Patches are cleaner than ugly sed.

jnbr:
ok, I changed it from sed to a patch

Vaelatern:
Thank you for your contribution!

PR #4772, opened 2016-09-15
[WIP] New package: coda-6.9.8

Vaelatern:
Somewhat preliminary, just wanted eyes to see if my solution to the only binary that doesn't allow building with CFLAGS is adequate, unless someone has a solution?


Vaelatern:
Fixed that.

the-maldridge:
Is this still in the works or has this been abandoned?

Vaelatern:
In the works.

PR #4778, opened 2016-09-16
pcsclite: add libpcsclite to depends

cavalier911:
The user in this [voidlinux forum post](https://forum.voidlinux.eu/t/opensc-failed-to-establish-context-unable-to-load-external-module/1078/2?u=cardinall) had to install libpcsclite before pcscd would communicate with the card reader. 
Investigation of debian package [pcscd](https://packages.debian.org/stretch/pcscd) which owns the same files as pcsclite indicates it depends on debian package [libpcsclite1](https://packages.debian.org/jessie/i386/libpcsclite1/filelist) which owns same files as libpcsclite


ebfe:
> However, running opensc-tool -v or opensc-explorer -v just gives back the message Failed to establish
> context: Unable to load external module.

Looks more like opensc is missing a dependency.

cavalier911:
This quote is the only content from the  [voidlinux forum post](https://forum.voidlinux.eu/t/opensc-failed-to-establish-context-unable-to-load-external-module/1078/2?u=cardinall) that I am referencing in regard to this pull request.

> Apparently I just had to install libpcsclite and it works. Shouldn't pcsclite depend on it?

PR #4781, opened 2016-09-16
readline: update to 7.0.000. (soname bump)

ebfe:


ebfe:
clisp broken

chneukirchen:
For clisp, try
`        sed -i '/rl_readline_state/s/type int/type ulong/' modules/readline/read\line.lisp    `

ebfe:
Thanks, that worked. Everything builds on x86_64 now.

ebfe:
Spoke too soon. There are still wired build errors with gdb, iverilog, es. Not related to readline though.
It seems the first build works fine, but when trying to rebuild it make randomly? throws errors of the form:

```make[8]: *** No rule to make target '../.././../gnulib/import/wctype-h.c', needed by 'wctype-h.o'.  Stop.```

```make[8]: *** No rule to make target '../.././../gnulib/import/malloca.c', needed by 'malloca.o'.  Stop. ```

I'd say this is a parallel build issue but `disable_parallel_build` is already set. Disabling ccache seems to avoid this but is not an option for the builders.

pullmoll:
FWIW I also saw random hiccups of this kind when trying to build pkgs with gcc6.
It seems to be related to ccache, yes, because here also it worked after disabling ccache.
Examples are the linux kernels, where the `fixdep` binary seems to find dependencies in the wrong place.

Perhaps we need an extra option for ccache. It had several updates recently, so time for RTFM.

ebfe:
Downgraded ccache locally to 3.2.8. Can't reproduce the issue anymore.

pullmoll:
> Downgraded ccache locally to 3.2.8. Can't reproduce the issue anymore.

Then we should downgrade ccache until we know those issues are fixed.

ebfe:
x86_64 builds ok now, on x86_64-musl clisp and mdbtools are broken. not related to the readline update tough.

ebfe:
57bd0102b1b32ea09d46eeba4b64b41e5a2480ae

PR #4786, opened 2016-09-17
[RFC] common/hooks: print per package ccache stats

ebfe:


pullmoll:
Of course it would be nice if the global stats were preserved.
I think backing up `$CCACHE_DIR/?/stats` could do that, but then the new stats would not be included. Too bad there is non `ccache --delta-stats` :)

ebfe:
I guess we could compute it ourself by comparing `ccache -s` before and after the build, but my sh is to weak for that ;)

pullmoll:
Hmm.. perhaps adding the values from the new stats would update some info, but you can't know the number of files removed from the cache when it becomes full. Thus only adding the number of files (one of those numbers) would be wrong.

I guess we can just discard the global info, though.

ebfe:
Personally i'm mostly interested in the cache hit % per build. That could be computed either way.

pullmoll:
If what you need or want to see can be computed (easily) without zapping the stats first, that would be nicer.

pullmoll:
Oh, wait, forgot this: We currently use a shared cache for all architectures, so this won't work for parallel builds. If we want to apply it, we need to have separate `ccache-$arch` directories again.

PR #4787, opened 2016-09-17
slstatus: update to 2.0

ribmo:


srcpkgs/slstatus/INSTALL.msg
```diff
@@ -1,7 +1,3 @@
 To use slstatus, put the following code in your ~/.xinitrc (or similar):
 
-  while true; do
-    slstatus
-  done &
-
-The loop is needed so that the program runs after suspend to ram.
+  slstatus &
```
Vaelatern: Does this work after suspend to RAM now?

srcpkgs/slstatus/template
```diff
@@ -9,10 +9,9 @@ license="ISC"
 homepage="https://git.nulltime.net/slstatus"
 distfiles="${homepage}/snapshot/${pkgname}-${version}.tar.gz"
 makedepends="alsa-lib-devel libX11-devel"
-checksum=77c42876571c7b1d850325e8d09a37ca9264031e4e6237195480d718a9a584f6
+checksum=a7fd452940d4136376df3f9887e231fc183f4cbe8473c6c63a0687905406351f
 
 pre_build() {
-	cp ${FILESDIR}/config.h config.h
```
Vaelatern: Do they provide a decent `config.h` now?

ribmo: [Yep][1]. The simplified `config.def.h` in v2.0 means the application works out of the box without any additional config.

However, perhaps I should include that file in this package and keep this line in template, so it's easier for users to configure with xbps-src. [I've done this on my own fork][2]. What do you recommend?

[1]: https://git.nulltime.net/slstatus/tree/config.def.h?h=v2.0&id=2db21462bd2768f4632d52296ce688476c1a3287
[2]: https://github.com/ribmo/void-packages/tree/25789f908e333385a276ee4dccd1d3ebb57e90cf/srcpkgs/slstatus

srcpkgs/slstatus/INSTALL.msg
ribmo: Yep, it seems to work after suspend to RAM now. The usage instructions in the [upstream readme][1] were changed to reflect this as well.

[1]:https://git.nulltime.net/slstatus/commit/?h=v2.0&id=2db21462bd2768f4632d52296ce688476c1a3287

srcpkgs/slstatus/template
Vaelatern: Do as you did in your own fork.

ribmo: Thanks. Done.

ribmo:
Anything else needed to get this merged?

PR #4790, opened 2016-09-17
New package: qt-webkit-2.3.4

pullmoll:
Instead of the webkit shipped with `qt-4.8.7` (which is impossible to build with gcc6), add a `qt-webkit-2.3.4` package and revbump packages depending on `libQtWebKit.so.4`, adding the new dependency `qt-webkit-devel`.


Gottox:
@pullmoll: qt-webkit needs qt-designer? Is this true?

pullmoll:
Well, yes, but it probably should not.

qt-webkit also generates the webview qtdesigner plugin `libqwebview.so`, which depends on `libtQtDesigner.so`. I guess the best way to solve this is to split off a `qt-designer-libs` which qt-webkit could then depend on.

PR #4799, opened 2016-09-18
New package: openblas-0.2.19

juliendehos:


srcpkgs/OpenBLAS/template
```diff
@@ -0,0 +1,17 @@
...
+revision=1
+build_style=gnu-makefile
+hostmakedepends="perl"
+short_desc="OpenBLAS is an optimized BLAS library"
+maintainer="Julien Dehos <dehos@lisic.univ-littoral.fr>"
+homepage="http://www.openblas.net/"
+license="BSD"
+distfiles="https://github.com/xianyi/${pkgname}/archive/v${version}.tar.gz"
+checksum="9c40b5e4970f27c5f6911cb0a28aa26b6c83f17418b69f8e5a116bb983ca8557"
+nocross=yes
```
Vaelatern: Why?

srcpkgs/OpenBLAS/template
```diff
@@ -0,0 +1,17 @@
+# Template file for 'openblas'
+pkgname="OpenBLAS"
+version="0.2.19"
+revision=1
+build_style=gnu-makefile
+hostmakedepends="perl"
+short_desc="OpenBLAS is an optimized BLAS library"
```
Vaelatern: A better description without `OpenBLAS` or `BLAS` mentioned?

juliendehos:
I've changed the description and found a config which cross-compiles.

Vaelatern:
Please match the case of the package in the commit and template. Capital letters are fine, consistency is requested.

juliendehos:
fixed

srcpkgs/openblas/template
```diff
@@ -0,0 +1,31 @@
...
+case "$XBPS_TARGET_MACHINE" in
+	armv6*) make_build_args+=" TARGET=ARMV6" ;;
+	armv7*) make_build_args+=" TARGET=ARMV7" ;;
+esac
+post_install() {
+	vlicense LICENSE
+}
+
+openblas-devel_package() {
+	depends="openblas-devel-${version}_${revision} ${sourcepkg}-${version}_${revision}"
```
Vaelatern: Not a good idea. To install a package you need to install the package? `depends="${sourcepkg}>=${version}_${revision}"` would be better.

juliendehos: Oops... Fixed.

srcpkgs/openblas/template
```diff
@@ -0,0 +1,31 @@
...
+	armv7*) make_build_args+=" TARGET=ARMV7" ;;
+esac
+post_install() {
+	vlicense LICENSE
+}
+
+openblas-devel_package() {
+	depends="${sourcepkg}>=${version}_${revision}"
+	short_desc+=" - development files"
+	pkg_install() {
```
pullmoll: Add `vmove usr/include` here, because the header files belong into the devel subpkg, not the main pkg.

juliendehos: Ok. Done.

PR #4800, opened 2016-09-19
Remove seemingly useless patch sets

Vaelatern:
With these changes I was able to build a kernel for armv7. Not yet tested, we don't yet officially support the Odroid C1+.

Point is, the patch sets seem wrong and never apply any more. Should we stop pretending to apply them?


chneukirchen:
Obsolete.

PR #4802, opened 2016-09-19
New package: codelite-9.2

thypon:


srcpkgs/codelite/template
```diff
@@ -0,0 +1,25 @@
...
+build_style=cmake
+configure_args="-DENABLE_CLANG=1 -DENABLE_LLDB=1 -DWITH_MYSQL=1"
+hostmakedepends="pkg-config cmake clang"
+makedepends="wxWidgets-devel libmariadbclient-devel sqlite-devel libssh-devel
+gtk+-devel"
+short_desc="Open-source, cross platform IDE for the C/C++ programming languages"
+maintainer="Andrea Brancaleoni <miwaxe@gmail.com>"
+license="GPL-3"
+homepage="http://www.codelite.org"
+distfiles="https://github.com/eranif/${pkgname}/archive/${version//_/-}.tar.gz"
```
pullmoll: I think `${version}` is `9.2` and thus there is no underscore to replace?

srcpkgs/codelite/template
```diff
@@ -0,0 +1,25 @@
+# Template file for 'codelite'
+pkgname=codelite
+version=9.2
+revision=1
+build_style=cmake
+configure_args="-DENABLE_CLANG=1 -DENABLE_LLDB=1 -DWITH_MYSQL=1"
+hostmakedepends="pkg-config cmake clang"
```
Vaelatern: It is

srcpkgs/codelite/template
```diff
@@ -0,0 +1,26 @@
...
+makedepends="wxWidgets-devel libmariadbclient-devel sqlite-devel libssh-devel
+gtk+-devel"
+short_desc="Open-source, cross platform IDE for the C/C++ programming languages"
+maintainer="Andrea Brancaleoni <miwaxe@gmail.com>"
+license="GPL-3"
+homepage="http://www.codelite.org"
+distfiles="https://github.com/eranif/${pkgname}/archive/${version}.tar.gz"
+checksum="2e82e26d55e9884cd5c1061efd3ab4d8bd81dc6e095c5b811d0a3cba395e6118"
+
+only_for_archs="i686 x86_64"
```
Vaelatern: Reasoning for the restrictions?

srcpkgs/codelite/template
```diff
@@ -0,0 +1,26 @@
...
+nopie=yes
```
Vaelatern: And reasoning here?

thypon:
Codelite PHP parser can't be compiled with musl, nor it can be cross compiled since it depends on nocross libraries.

nopie because i can't reproduce a pie artifact.

As far as I know we have not strict policies on nocross/nopie, plus it's awkward you run a C++ integrated development environment on arm.

srcpkgs/codelite/template
thypon: fix later

PR #4818, opened 2016-09-22
New package: hitch-1.4.1

jannisch:
Please tell me in which of the 3 ways the example config file should be installed:

1. (currently used): default path `/usr/share/doc/hitch/hitch.conf.example`
2. with `vsconf`
3. with `vconf` directly into /etc

The user needs to edit the config anyway because of the custom TLS certificate.


jannisch:
1.4.1 was released in the meantime, please don't merge yet.

jannisch:
Updated, I hope that I did not make a mistake when changing the commit message.

Vaelatern:
I'd be happy with a `vconf` is the config can be used right out of the box.
Otherwise the default is just fine.

jannisch:
Thanks, so no change is required I guess (does not work out of the box).

srcpkgs/hitch/template
```diff
@@ -0,0 +1,20 @@
...
+build_style=gnu-configure
+hostmakedepends="pkg-config"
+makedepends="libev-devel libressl-devel"
+short_desc="Libev-based high performance SSL/TLS proxy"
+maintainer="Jannis Christ <hello@jannis.ovh>"
+license="2-clause-BSD"
+homepage="https://hitch-tls.org"
+distfiles="https://hitch-tls.org/source/hitch-${version}.tar.gz"
+checksum=d6118a3bf0325fc7d7635eec2c467364ac4bac3c7584b2f194067714ba36453d
+system_accounts="hitch"
```
Vaelatern: We seem to be trending to using `_` to denote system level fields, so `system_accounts="_hitch"` and other relevant changes.

jannisch: So in this case just the `system_accounts` field, right? I'm not completely sure what you mean.

Vaelatern: Looks like you made the right changes to me.

PR #4820, opened 2016-09-23
nodejs: add -devel package

the-maldridge:
I'm working on some software that against all wishes needs nodejs headers.  That software may later be PR'd if I can find a way around the curl|bash in the middle of it.


the-maldridge:
That entire ecosystem is unworkable.  This is not needed, nor should it ever be.

PR #4826, opened 2016-09-23
New package: tup-0.7.5

juliendehos:


chneukirchen:
I'd really prefer if this used our lua and sqlite, but that seems to require some patching.

The build system doesn't use our CFLAGS at all.

juliendehos:
Thanks for the CFLAGS; I've fixed that.

Tup aims to be fast, so I think Lua and sqlite are embedded into tup to maximize performance (moreover lua and sqlite are designed to be embedded). So I don't think patching is a good idea here.

chneukirchen:
It will make no measurable speed difference, and we don't end up having to patch 12 instances of sqlite in our packages.

juliendehos:
I've found an option for compiling tup with the system sqlite.
For lua, the embedded code is modified so we have to use it.

srcpkgs/tup/template
```diff
@@ -0,0 +1,33 @@
+# Template file for 'tup'
+pkgname="tup"
+version="0.7.5"
+revision=1
+hostmakedepends="fuse-devel pkg-config"
+makedepends="fuse-devel sqlite-devel"
+depends="sqlite"
```
Duncaen: `xbps-src` already detects `sqlite` as a dependency see https://travis-ci.org/voidlinux/void-packages/jobs/162641007#L761.

srcpkgs/tup/template
```diff
@@ -0,0 +1,33 @@
+# Template file for 'tup'
+pkgname="tup"
+version="0.7.5"
+revision=1
+hostmakedepends="fuse-devel pkg-config"
```
Duncaen: Is `fuse-devel` really necessary here?

srcpkgs/tup/template
```diff
@@ -0,0 +1,33 @@
...
+makedepends="fuse-devel sqlite-devel"
+depends="sqlite"
+short_desc="File-based build system"
+maintainer="Julien Dehos <dehos@lisic.univ-littoral.fr>"
+homepage="http://gittup.org/tup/index.html"
+license="GPL-2"
+distfiles="https://github.com/gittup/${pkgname}/archive/v${version}.tar.gz"
+checksum="361b3e069308ce1d9505d1cb927999ac448811a3425c724123e0c48602a9d1e4"
+nostrip_files="tup"
```
Duncaen: Any reason to not strip `tup`?

srcpkgs/tup/template
```diff
@@ -0,0 +1,33 @@
...
+
+do_build() {
+	# build tup for host arch using a script
+	CC="gcc" CFLAGS="-O0" LDFLAGS="-fPIC" ./build.sh
+	# build tup for target arch using the previous build (boostrapping)
+	sed -i 's/git describe/echo v'${version}'/' Tupfile
+	sed -i 's/CC = gcc/CC = '${CC}'/' Tuprules.tup
```
Duncaen: maybe just use double quotes if you want to expand the variables.

srcpkgs/tup/template
juliendehos: Yes. Tup's building is boostrapped: a shell script builds tup on the host then the host uses this build to cross-build tup for the target. So fuse-devel is necessary for both builds.

srcpkgs/tup/template
juliendehos: Right. Fixed.

srcpkgs/tup/template
juliendehos: Ok. Fixed.

srcpkgs/tup/template
juliendehos: Building fails without this line ("ERROR: tup-0.7.5_1: non-PIE executable found in PIE build: /usr/bin/tup").
Honestly, I don't understand that perfectly; if you have any idea...

Duncaen: Then it doesn't use our `CFLAGS`, and/or `LDFLAGS`.
`nostrip` is not the solution.

juliendehos: I've added LDFLAGS and it indeed works without nostrip. Thanks for the explanation.

chneukirchen:
Applied with a few more tweaks and an aarch64 patch.

Thanks :tada:

srcpkgs/tup/template
cr6git: Seems like that changed in the latest release: https://github.com/gittup/tup/releases/tag/v0.7.6
````
 - Added a bootstrap-nofuse.sh script that avoids using FUSE during bootstrap for restricted environments
````

PR #4827, opened 2016-09-24
New package: libpng12-1.2.56

lemmi:
This is needed to get into steam's BigPicture mode without the steam runtime `STEAMRUNTIME=0 steam`. Currently this is the only way to start steam on my machine. Someone also mentioned the need for libgcrypt15 to start Gary's Mod. 

Maybe we could consider switching to native runtime completely, as the steam runtime tends to break constantly. It should only require a few more entries to the steam's dependencies and maybe an older version of a package. What do you think? 

Alternatively, we could have a meta package for native runtime dependencies, to make this less of a hassle.


thypon:
https://github.com/thypon/void-packages/commits/steam-libs these alredy includes all the library needed to run steam with native libraries

lemmi:
@thypon ah nice, but they are not very useful as long as they are only in your repo, hm :)?

chneukirchen:
If it is just these 5 backports, I can live with them being merged.

thypon:
I didn't merge the steam-libs branch because there is another library needed by steam: LibreSSL is not completely binary compatible, therefore we need a bincompat openssl library with steam.

thypon:
We could compile openssl or ship steam with openssl and LD_PRELOAD on startup

thypon:
https://m.reddit.com/r/linux_gaming/comments/559imj/arch_now_has_all_packages_for_the_native_steam/

lemmi:
@thypon that's a brilliant idea, we should do that, too!

Wait. I already said that.

But seriously. I'll compile a list of what we need to actually package once I get to it.

ebfe:
Apart from steam, requests for libpng12 have come up several times now. I think @the-maldridge wanted it for something i can't remember right now.

thypon:
@lemmi i'll repeat the package needed for steam are all in place besides openssl in my steam-libs branch :)

I'll rebase them and see how it works with the compilation

PR #4832, opened 2016-09-25
New package: plantuml-8047

juliendehos:


ayghor:
its been a whiel since i tried, but have you got pdf generation working?

juliendehos:
I didn't even know plantuml can generate PDF and I still don't see any benefit since we can convert EPS to PDF easily using epstopdf. Adding PDF generation to plantuml requires more than 100 MB (!!!) of supplementary jar files (apache-batik, apache-fop) and I don't how to package that correctly.

However I can modify the plantuml package to automatically generate EPS and convert to PDF. I found no package for epstopdf but I can package it. What do you think about it ?

ayghor:
omg i wrote a reply days ago and didnt hit teh send button!

It would be awesome to has PDF generation, but I wouldn't bother too much! Matter of fact, i gave up tiem ago because teh deps stink! ;o3

thypon:
:+1:

PR #4837, opened 2016-09-26
New package: seriespl-1.3.3

tastytea:
Extracts stream-URLs for entire seasons of tv series from bs.to


tastytea:
The version changed meanwhile, hope that's okay.

PR #4838, opened 2016-09-26
New package: lightdm-gtk-greeter-settings-1.2.1

Alventoor:
I accidentally deleted my previous commit, so I recreated a new.

@Duncaen all changes are done, except for `do_build()`, because without him the sofware is build in the `build3.4` folder instead of `build`. Which causes an error during `do_install` step.


srcpkgs/lightdm-gtk-greeter-settings/template
```diff
@@ -0,0 +1,20 @@
...
+short_desc="Settings editor for LightDM GTK+ Greeter"
+maintainer="Julien Marconnet <julienm99@gmail.com>"
+license="GPL-3"
+homepage="https://launchpad.net/lightdm-gtk-greeter-settings"
+distfiles="https://launchpad.net/${pkgname}/${version%.*}/${version}/+download/${pkgname}-${version}.tar.gz"
+checksum=ab91027b56ca0e0c98a38169acc453c08801bc4937dc1a53bd6e98b7aaceea96
+
+do_build() {
+	python3 setup.py build
+}
```
Gottox: do_build can be removed

Gottox: That means the buildsystem is somewhat broken. Unfortunately using do_build directly breaks various features of xpbs. @asergi Do you have an idea how to fix this?

srcpkgs/lightdm-gtk-greeter-settings/template
```diff
@@ -0,0 +1,20 @@
+# Template file for 'lightdm-gtk-greeter-settings'
+pkgname=lightdm-gtk-greeter-settings
+version=1.2.1
+revision=1
+build_style=python-module
```
asergi: `build_style=python3-module`

srcpkgs/lightdm-gtk-greeter-settings/template
```diff
@@ -0,0 +1,20 @@
...
+python_versions="3.4"
+pycompile_version="3.4"
```
asergi: You can remove both `python_versions` and `pycompile_version`. They are no longer needed.

srcpkgs/lightdm-gtk-greeter-settings/template
```diff
@@ -0,0 +1,20 @@
...
+pycompile_module="lightdm_gtk_greeter_settings"
+hostmakedepends="python3.4-distutils-extra intltool"
+depends="lightdm-gtk3-greeter python3.4-gobject"
```
asergi: We switched to `python3-*` packages in the meantime, so: `'s/python3\.4/python3/g'`.

srcpkgs/lightdm-gtk-greeter-settings/template
Gottox: @asergi Could we fix this by globaly changing the build directory name?

asergi: This `setup.py` is weird because it creates files at build time in a directory outside the classical `build-X.Y`.
I wouldn't lose too much time at it. Skipping the `do_build()` is fine, imo.

Alventoor:
Thanks for your help @Gottox and @asergi , it works.

PR #4841, opened 2016-09-26
tcpflow: using github for distfiles.

Gottox:
@ebfe It's a workaround, but it fixes the template.


ebfe:
If we only had something that archived sources ;)

Gottox:
@ebfe yea, but this would break update check. :)

PR #4844, opened 2016-09-26
New package: epstopdf-2.25

juliendehos:


ayghor:
yes ghostscript and perl > java bullshit

juliendehos:
Indeed, I think epstopdf is a good alternative for implementing PDF generation in plantuml (#4832).
Moreover, it can be useful in itself.

PR #4849, opened 2016-09-27
New package: CellWriter-1.3.5

the-maldridge:
Very useful for pen-driven systems.


the-maldridge:
The build works fine on my travis runs, this seems to have failed because XBPS blew up while installing items from makedepends.

PR #4853, opened 2016-09-28
New package: mbpfan-1.9.1

Pizza-Boy:


srcpkgs/mbpfan/files/mbpfan/run
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+exec 1>&2
```
Vaelatern: Why this direction?

srcpkgs/mbpfan/template
```diff
@@ -0,0 +1,21 @@
+# Template file for 'mbpfan'
+pkgname=mbpfan
+version=1.9.1
+revision=1
+build_style=gnu-makefile
+short_desc="Macbook Pro Fan Control Daemon"
+maintainer='noah <nsawyer1993@gmail.com>'
+license="GPL-3"
+homepage="https://github.com/dgraziotin/${pkgname}"
```
Vaelatern: Please don't use `$pkgname` here.

srcpkgs/mbpfan/template
```diff
@@ -0,0 +1,21 @@
...
+distfiles="https://github.com/dgraziotin/${pkgname}/archive/v${version}.tar.gz"
```
Vaelatern: Please use `${homepage}` here.

srcpkgs/mbpfan/template
```diff
@@ -0,0 +1,21 @@
...
+checksum="a7cf850a393ebfce21427b992436b84cc4b20e1cb8d673d45d2c8b991c69e68c"
+replaces="mbpfan>=0"
```
Vaelatern: This line is not necessary.

srcpkgs/mbpfan/template
```diff
@@ -0,0 +1,21 @@
...
+
+do_build() {
+	sed -i 's/sbin/bin/' Makefile
+}
```
Vaelatern: This change should be done in a function called `pre_build` not `do_build`. As it stands this is wrong.

srcpkgs/mbpfan/template
```diff
@@ -0,0 +1,21 @@
...
+
+do_install() {
+	make install
+	vsv mbpfan
+}
```
Vaelatern:
```
post_install() {
    vsv mbpfan
}
```
not a `do_install` since the `build_style` you selected provides a good one that respects our variables.

srcpkgs/mbpfan/template
```diff
@@ -0,0 +1,24 @@
+# Template file for 'mbpfan'
+pkgname=mbpfan
+version=1.9.1
+revision=1
+only_for_archs="x86_64 i686"
```
Vaelatern: Why this restriction?

Pizza-Boy: The package is meant to control fans on Apple computers and, to my knowledge, there aren't any Apple computers running on that architecture. Just seemed easier.

Vaelatern: Also -musl

srcpkgs/mbpfan/files/mbpfan/run
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+exec 2>&1
+mbpfan || exit 1
+exec chpst -b mbpfan pause
```
Duncaen: Use `exec mbpfan -f ` to keep it from daemonizing, this avoids having to use `pause` which is just wrong.
Using pause this breaks all functionallity you want to have with supervised processes.
You cant send signals to mbpfan through `sv` because it would just send them to pause, showing that everything worked but in fact it just send them to `pause` and starts another `mbpfan` process in case pause exits on this signal.
And if `mbpfan` crashes or goes down for whatever reason `sv` would think its still running.

srcpkgs/mbpfan/template
```diff
@@ -0,0 +1,24 @@
+# Template file for 'mbpfan'
+pkgname=mbpfan
+version=1.9.1
+revision=1
+only_for_archs="x86_64 i686 x86_64-musl i686-musl"
```
Vaelatern: Please add a comment explaining your reasoning for the restriction (about those being the only arches close to possible on the macbookpro).

srcpkgs/mbpfan/template
```diff
@@ -0,0 +1,24 @@
...
+distfiles="${homepage}/archive/v${version}.tar.gz"
+checksum="a7cf850a393ebfce21427b992436b84cc4b20e1cb8d673d45d2c8b991c69e68c"
+
+pre_build() {
+	sed -i 's/sbin/bin/' Makefile
+}
+
+do_install() {
+	make install
+}
```
Vaelatern: Not sure why you still have a `do_install()`, can you either remove it or explain why it is necessary?

Pizza-Boy:
Alright, I've managed to really goof up the commits on this package somehow, and I don't really know how to clear it.

If necessary I could just delete this branch and push the changes again in a new pull request.

Vaelatern:
Please don't.

Vaelatern:
The way to fix is likely to rebase on voidlinux/void-packages master branch

Pizza-Boy:
Looks like the rebase fixed everything.

Pizza-Boy:
Sorry for the confusion, I'm still figuring out git.

pullmoll:
The `make install` does not honor the value of `$DESTDIR`. It installs into `/usr/bin` instead of `$DESTDIR/usr/bin`. See https://travis-ci.org/voidlinux/void-packages/jobs/164653827#L661

Pizza-Boy:
Alright, I think I've got it installing to the right location. Whole thing was a real pain in the ass because the Makefile doesn't have any way to change the destination directory by default--just patched it in myself through `sed` commands. Let me know if there's an easier way to do that.

Pizza-Boy:
Corrected the file to use `vbin` and `vinstall` instead--think this should be simpler.

Pizza-Boy:
So after doing some testing on my system, there appears to be a bug in the 1.9.1 version of mbpfan that causes a buffer overflow error when running tests, doing other things, etc.

It looks like it's been reported and fixed upstream in the master git repository, but I didn't want to use that because it seems that stable versions of software are preferred.

I know the flag that needs to be changed for gcc (_FORTIFY_SOURCE=2 to _FORTIFY_SOURCE=1), but I'm not sure what the easiest way to do this'd be. Alternatively, we could pull the master repository (but that's not a stable release, technically), which should have this issue patched out, I think. Or maybe write a patch for it.

Pizza-Boy:
Alright, setting `nopie` to something turns off the hardening features that cause the issue and make the program install correctly. Would it be alright for a non-hardened package to be installed, or should we pursue a different avenue?

pullmoll:
Hardened packages are, of course, preferred, but if it's too much of a hassle I guess we can live with it.
There's always a chance that someone takes a look and improves things, once a package is merged.

Pizza-Boy:
I'm sure that someone who knows C++ could write a patch integrating the changes upstream, but I'm not that guy (because I know nothing about C++).

srcpkgs/mbpfan/template
```diff
@@ -0,0 +1,22 @@
...
+short_desc="Macbook Pro Fan Control Daemon"
+maintainer='noah <nsawyer1993@gmail.com>'
+license="GPL-3"
+homepage="https://github.com/dgraziotin/mbpfan"
+distfiles="${homepage}/archive/v${version}.tar.gz"
+checksum="a7cf850a393ebfce21427b992436b84cc4b20e1cb8d673d45d2c8b991c69e68c"
+
+do_install() {
+	vbin bin/mbpfan
+	vinstall mbpfan.conf 644 etc
```
Vaelatern: This implies you need a `conf_files="/etc/mbpfan.conf"`
This line should also be `vconf` unless you have a reason.

PR #4856, opened 2016-09-29
lmdb: fix soname generation

dutsky:
It needed for kde kf5 packages


Vaelatern:
Please also move the `.so` file to the subpackage for `-devel`

dutsky:
Thank you. I'm too stupid

PR #4857, opened 2016-09-29
New package: hopper-3.11.20.1

thypon:


ebfe:
Hmm, that's a demo version?

dutsky:
Oh :/

thypon:
The demo version it's equal to the complete version: you only need the full license to unlock 15mins+ sessions and save states

PR #4873, opened 2016-10-02
Enable building of browser plugin for icedtea-web

Hoshpak:
CC @xtraeme and @felix as maintainers of the affected packages.

With this PR I propose the following changes to the icedtea-web and related packages to enable the building of the NPAPI browser plugin:
- New package "npapi-sdk" to provide the header files necessary to build the browser plugin
- move binaries in the package to /usr/lib/jvm/icedtea-web to avoid conflicts with alternatives from other packages
- add alternatives to the package for all binaries
- enable building of the npapi plugin
- add .desktop file to associate .jnlp files with javaws
- move javaws from the "java" into the "javaws" alternatives group in oracle-jdk and orcale-jre so you can select individually which javaws you want to use without having to switch over all other java binaries

I tried it with firefox(-esr) and epiphany so far and with the resulting packages both running java applets in the browser and automatically opening .jnlp files with javaws and executing the applications works.


Vaelatern:
Why the `.desktop` file? Can't that come from somewhere else?

srcpkgs/npapi-sdk/template
```diff
@@ -0,0 +1,19 @@
...
+version=0.27.2
+revision=1
+noarch=yes
+
+build_style=gnu-configure
+short_desc="Netscape Plugin API (NPAPI)"
+maintainer="Helmut Pozimski <helmut@pozimski.eu>"
+license="MPL-1.1"
+homepage="https://bitbucket.org/mgorny/npapi-sdk"
+distfiles="https://bitbucket.org/mgorny/npapi-sdk/downloads/npapi-sdk-$version.tar.bz2"
```
Vaelatern: `distfiles="${homepage}/downloads/${pkgname}-${version}.tar.bz2"`

srcpkgs/npapi-sdk/template
```diff
@@ -0,0 +1,19 @@
...
+checksum=01bf34bdd0fcaa546ba3d5c96f5d91f91dd117d4ed5ffe13f56bbd5aa566ae77
+
+do_install() {
+	make DESTDIR=$DESTDIR install
+	cd ${DESTDIR}/usr/lib/pkgconfig
+	ln -sf npapi-sdk.pc mozilla-plugin.pc
+}
```
Vaelatern: Can you not provide a `do_install()`  but do any custom actions in a `post_install()`?

Hoshpak:
@Vaelatern Thanks for reviewing. I am still getting used to the way xbps templates are built. Doing everything in do_install is something I took from another existing template. I will test it be back with a new version of this pull request soon.

Regarding .desktop file: I think it's definitely beneficial to have a file association for .jnlp files so these files are automatically opened with the right command. Since other packages also ship their own .desktop files (an this is the way it is done with javaws in other distributions too) I don't know where else the file would be supposed to come from unless we would create a separate package just for it.

Hoshpak:
amended and pushed.

Vaelatern:
I do not have the time to review this right now, if anybody else would like to, then by all means.

Gottox:
I believe there are still some issues with moving files from one alternative to another. I'd like to consider @xtraeme here. Otherwise it looks good to me.

PR #4881, opened 2016-10-04
openimageio: update to 1.7.7. + rebuilds

lemmi:


lemmi:
OpenCL-CLHPP: luxmark won't compile without it. And while the repo does include a LICENSE.txt, the tarball of the release doesn't.

ebfe:
I'd vote for lowercasing `OpenCL-CLHPP`. 
As for the license `sed -n '1,\,*/,p' input_cl.hpp`?

lemmi:
i chose to download the license file from their git master. on the next update this can be removed again and replaces with a simple vlicense LICENSE.txt

PR #4702, opened 2016-09-02
New Package: volctl-0.4

teldra:

PR #4888, opened 2016-10-05
New package: volctl-0.4

teldra:


srcpkgs/volctl/template
```diff
@@ -0,0 +1,20 @@
...
+maintainer="xo <oh@rotce.de>"
+license="GPL-2"
+homepage="https://buzz.github.io/volctl/"
+#distfiles="https://github.com/buzz/volctl/archive/${version}.tar.gz"
+#checksum=31b7079767172c03653af380181272ade50f04f7ca32c69725239ea8116f0640
+
+wrksrc=volctl
+
+do_fetch() {
+	git clone https://github.com/buzz/volctl.git volctl
```
Duncaen: Please use releases, `git clone` can't be cached and without a specific rev it would be just master which is different each build, a package/template should represent a specific release/state of the code.

ebfe:
#4702

srcpkgs/volctl/template
teldra: oh, yes. sry. i just forgot.
should i close the pr and wait for release, just wait for release or use the old version, with major bug?

PR #4900, opened 2016-10-05
New package: kde-l10n

Hoshpak:
KDE 4 localization packages:
- one binary package for each source tar ball provided by the KDE project
- one meta package to install all languages together

I know that these are a lot of packages and commits. I would have preferred to build all packages from a single template file but since I couldn't find a way to do it with xbps-src, I went for separate template files instead.


pullmoll:
You can take a look at how the `firefox-i18n` template handles a list of languages, their download locations and checksums.

There are several `*-i18n` packages doing it in a similar way.

Hoshpak:
I already did that and tried for several hours to build a package follwoing the same pattern. However the existing -i18n-packages use build-style=meta and then just do a vinstall for the other packages defined in a loop. kde-l10n needs to be built using cmake so it would need to be one package with build-style=meta defining several other packages with build-style=cmake and that doesn't seem to work or did I miss anything?

pullmoll:
I think you can set the *global* `build-style=meta` and have `build-style=cmake` defined locally in the `_pkgtmpl() { ...`. Then the `pkg_install()` inside that one would then compile and install in one go , assuming that a `make install` also builds the pkg.

Hoshpak:
Thanks. I have already tried combining global build_style=meta with build_style=cmake in the sub-package with every combination I could think of an it just wouldn't work. The only thing that did work was doing eveything (configure, make, make install) manually in pkg_install. However I think that would be wrong since the function is only intended to install stuff, not configure and build it.

Doing the configure and build steps in the functions of the main (meta)-package and installing in pkg_install works and is just a bit less convenient since I need to configure things like configure_args manually. I think this might be the best solution available right now. I will update the pull request with a single file doing it this way tomorrow.

Hoshpak:
Pushed the new version. Now it's only one commit and one template file all necessary packages are build from.

pullmoll:
I'm glad you made this single template, even if it's a little more awkward than other *-i18n templates. It's definitely more convenient to handle a new version by modifying just one template's checksums.

Issue #4891, opened 2016-10-05
qt5 host tools


chneukirchen:
The arrangement of qt5 tools is bad for cross-compilation, while qmake is in qt5-qmake, e.g. uic is only in qt5, so you have to install whole qt5 on the host.

pullmoll:
Agree. Furthermore we should have `cross-qt5-qmake-$arch` which exists while building qt5 for a cross arch, but is not packaged. This is a qmake for the host which has all values set for the cross arch.

I don't know if then these qmakes should become a cross-vpkg-dummy entry, or what else could be done to automagically use the required `cross-qt5-qmake-$arch` package in the qmake build style.

pullmoll:
The packages that should go into `qt5-tools` or perhaps a new `qt5-host-tools` are the ones which are compiled twice, once for the host and once for the target: `qmake`, `moc`, `rcc`, `uic`, `qlalr`, `qdbuscpp2xml` and `qdbusxml2cpp`.

chneukirchen:
+1 for making build_style=qmake add the correct hostmakedepends.

pullmoll:
I'll try to package the target specific qmakes and then create a `common/environment/build-style/qmake.sh` which adds the required one to the hostmakedepends.

This will take some time because it requires testing all arches...

pullmoll:
I just realized this can't work because when compiling for `-a aarch64` I can't create host packages :frowning_face: 

The only way I see is to create a number of `cross-$XBPS_CROSS_TRIPLET-qt5` or `qt5-qmake-$XBPS_CROSS_TRIPLET` packages which create the target specific qmake via the bootstrap stage of qt5, just like `srcpkgs/qt5/template` does now. This involves creating a `mkspecs` entry with a `qmake.conf` for the cross compiler, linker, flags etc.

The packages would then have `conflicts="qt5-qmake>=0"` and be used instead of this in the build style qmake.

I'll see if I can make these templates use the version, distfile, checksum etc. info from the qt5 template.

chneukirchen:
Aren't these tools portable enough that we can run the host version when compiling for the target?

pullmoll:
The problem is with specifying the `-xplatform` or the `-device some-device-name` and `-device-options CROSS_COMPILE=${XBPS_CROSS_TRIPLET}-`. I have not found a way to tell these to a qmake built for e.g. x86_64.

These options are passed in the `./configure` before qmake is built.

[...]

PR #4901, opened 2016-10-06
New package: kdbusaddons-5.26.0

dutsky:


Gottox:
What packages need this?

dutsky:
kservice, kwrited and polkit-kde-agent

Gottox:
Is there any chance to make crossbuild work before we merge?

dutsky:
I don't know why it's not build in voidlinux. Maybe it's qmake issue #4891

pullmoll:
Qt5 cross compiling is being worked on, but it seems to be difficult to get done.
The culprit is qmake and its (default) settings, which need to be overriden for cross compiling and some of these settings seem to be compile-time only.

Still just let the packages fail to cross build for now, perhaps we manage to solve the issue later on.

PR #4903, opened 2016-10-06
New package: ccextractor-0.82

styrsven:


srcpkgs/ccextractor/template
```diff
@@ -0,0 +1,36 @@
+# Template file for 'ccextractor'
+
+pkgname=ccextractor
+version=0.82
+revision=1
+wrksrc=${pkgname}.${version}
+only_for_archs="i686 x86_64"
```
ebfe: Remove this so it's build for arm* and *-musl too.

srcpkgs/ccextractor/template
```diff
@@ -0,0 +1,36 @@
...
+hostmakedepends="unzip"
+makedepends="leptonica-devel tesseract-ocr-devel"
+short_desc="Extract subtitles from video"
+maintainer="Styrsven <sven.a.jonsson@bahnhof.se>"
+license="GPL-2"
+homepage="http://www.ccextractor.org/doku.php"
+distfiles="${SOURCEFORGE_SITE}/${pkgname}/${pkgname}.src.${version}.zip"
+checksum=890e7786256c74c7e4850592784da027451dd7c3e3a353c9bad3bea5467b7b77
+
+pre_build() {
```
ebfe: Maybe use post_extract here

srcpkgs/ccextractor/template
```diff
@@ -0,0 +1,36 @@
...
+	sed -i -e 's/\(^LDFLAGS \)=\(.*$\)/\1+=\2/' \
+	       -e 's/\(^CFLAGS \)=\(.*$\)/\1+=\2/' linux/Makefile
+}
+
+do_build() {
+	pushd linux
+	make
```
ebfe: make CC=${CC} fixes the cross builds.

srcpkgs/ccextractor/template
styrsven: fixed

srcpkgs/ccextractor/template
styrsven: fixed

PR #4914, opened 2016-10-08
New package: xca-1.3.2

dutsky:
easy-rsa replacement with gui


srcpkgs/xca/template
```diff
@@ -0,0 +1,25 @@
...
+short_desc="X Certificate and Key Management"
+maintainer="Denis Revin <denis.revin@gmail.com>"
+license="BSD"
+homepage="http://xca.sourceforge.net/"
+distfiles="${SOURCEFORGE_SITE}/${pkgname}/${version}/${pkgname}-${version}.tar.gz"
+checksum=04f11f2148c8369c1a1b4e109956ed8cc807ec266d0d936410624ba8ee4b5ce4
+
+do_build() {
+	make all
+}
```
Vaelatern: Can you replace this with a simple `make_build_target="all"` ?

srcpkgs/xca/template
```diff
@@ -0,0 +1,25 @@
...
+
+do_install() {
+	make destdir=${DESTDIR} install
+}
```
Vaelatern: And I think this might not be necessary with `make_install_args="destdir=${DESTDIR}"`

PR #4935, opened 2016-10-10
New package: 9mount-1.3

CMB:


ebfe:
Consider adding the follwing upstream patch http://sqweek.net/hg/9mount/index.py/rev/c6da9cac8d01. Otherwise working with hosts that have a AAAA record is a bit annoying (v9fs only supports ipv4).

CMB:
Ok, I added that patch.
It's really too bad that V9FS doesn't do IPv6.  Do you know if there are any
plans to change that in the near future?

ebfe:
> It's really too bad that V9FS doesn't do IPv6.  Do you know if there are any
> plans to change that in the near future?

Don't think so. In some way it actually supports ipv6 and a whole lot of
other things via the fd transport (rfdno/wdfno) but that isn't supported
by 9mount.

srcpkgs/9mount/INSTALL
```diff
@@ -0,0 +1,5 @@
+case "${ACTION}" in
+post)
+	chown root:users usr/bin/9{mount,umount,bind}
```
ebfe:
```
9mount-1.3_1: configuring ...
chown: cannot access 'usr/bin/9{mount,umount,bind}': No such file or directory
chmod: cannot access 'usr/bin/9{mount,bind,umount}': No such file or directory
```

ebfe: That's due to the 'sh' alternative being set to 'dash' (which seems to be the default). For reasons unknown it's set to 'bash' in the masterdir, so to miss things like that.

CMB:
Yeah apparently usr/bin/9{bind,mount,umount} was a bashism and your
/bin/sh is dash (or busybox sh for that matter).
Fixed.  Thanks for catching.

ebfe:
Thanks.

PR #4938, opened 2016-10-10
geary: update to 0.11.2.

jannisch:
I made quite a few changes as well:

- homepage now redirects to the gnome wiki page
- license was wrong
- gnome-keyring: I noticed that it's not linked against libgnome-keyring at all so libgnome-keyring-devel is not required, but `gnome-keyring` itself for storing passwords (otherwise I ran into [this](https://forums.freebsd.org/threads/50155/) issue). I guess it's handled by libsecret instead.

### DIFF
--- a/srcpkgs/geary/template
+++ b/srcpkgs/geary/template
@@ -1,17 +1,17 @@
 # Template file for 'geary'
 pkgname=geary
-version=0.11.0
+version=0.11.2
 revision=1
 build_style=cmake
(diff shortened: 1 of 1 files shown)


jannisch:
After spending some time with geary and [on the gnome bugtracker](https://bugzilla.gnome.org/show_bug.cgi?id=772721), I can confirm that `gnome-keyring` is just a backend for the new `libsecret` by now.

Issue #4670, opened 2016-08-26
Package request: kmymoney


Hoshpak:
Personal finance manager:

https://kmymoney.org/

Hoshpak:
merged in #4939 .

PR #4939, opened 2016-10-10
New package: kmymoney-4.8.0

Hoshpak:
This pull request introduces the new package kmymoney requested by me in #4670 . I also added the new package libalkimia which is a necessary dependency for kmymoney and added the libraries provided by both packages to common/shlibs. After testing the build I noticed that the binary packages couldn't be installed because libgwengui-qt4.so.0 couldn't be resolved to package so I moved the entry in common/shlibs to the -qt package the contains the .so file.

### DIFF
--- a/common/shlibs
+++ b/common/shlibs
@@ -2618,3 +2618,11 @@ libqmmp.so.1 qmmp-1.1.4_1
 libqmmpui.so.1 qmmp-1.1.4_1
 libksignalplotter.so.4 kde-workspace-4.11.22_1
 libalkimia.so.5 libalkimia-5.0.0_1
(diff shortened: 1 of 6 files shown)


Hoshpak:
pull request updated with a backported patch to fix the previous build failure on i686.

pullmoll:
You could squash the common/shlibs changes into their packages to reduce the number of commits.

Hoshpak:
sure, I just thought that it would be more 	comprehensible if I put them into separate commits. Squashed and pushed.

Hoshpak:
branch rebased on top of current master

Issue #4754, opened 2016-09-12
Package request: knemo


Hoshpak:
A network monitorfor KDE.

web site: https://www.kde.org/applications/internet/knemo/
git: https://quickgit.kde.org/?p=knemo.git

Hoshpak:
package added to the repository in #4941 .

PR #4941, opened 2016-10-11
New Package: knemo-0.7.7

Hoshpak:
This pull request adds the new package knemo-0.7.7 requested by me in #4754 . Since the package couldn't be installed after my first build attempt because libksignalplotter.so.4 couldn't be resolved, I also added that library to common/shlibs.

### DIFF
--- a/common/shlibs
+++ b/common/shlibs
@@ -2614,3 +2614,4 @@ libz80ex.so.1 z80ex-1.1.21_1
 libz80ex_dasm.so.1 z80ex-1.1.21_1
 libqmmp.so.1 qmmp-1.1.4_1
 libqmmpui.so.1 qmmp-1.1.4_1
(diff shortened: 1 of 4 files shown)


pullmoll:
You could perhaps include [this patch from Arch Linux](https://git.archlinux.org/svntogit/community.git/tree/trunk/knemo-gcc6.patch?h=packages/knemo) to be prepared for a change to gcc6.

Hoshpak:
That's certainly a good idea. I incorporated one of the two patches from the Arch package into the pull request. The other one doesn't apply since Arch uses a newer git revision and version 0.7.7 doesn't even include the file trayicon.cpp. I was able to build knemo version 0.7.7 in a quickly set up Arch VM with gcc6 with the patch for only that single file so it should be fine.

PR #4944, opened 2016-10-11
New package: reaver (t6x fork)

felix:


srcpkgs/pixiewps/template
```diff
@@ -0,0 +1,24 @@
...
+makedepends="libressl-devel"
+short_desc="Brute force attack against Wifi Protected Setup mod with Pixie Dust Attack"
+maintainer="Felix Hanley <felix@userspace.com.au>"
+homepage="https://github.com/wiire/${pkgname}"
+license="GPL"
+distfiles="https://github.com/wiire/${pkgname}/archive/v${version}.tar.gz"
+checksum=450127429976329b55d6b4e8510c91577ff51e66d43ac1677acd9bcf586c180e
+
+do_build() {
+    cd src && make CCFLAGS="$CCFLAGS -std=c99 -fPIC" LDFLAGS="-lssl -lcrypto $LDFLAGS"
```
pullmoll: This should probably be `CFLAGS="$CFLAGS -std=c99 -fPIC"`, i.e. CFLAGS not CCFLAGS.

**But** then you should not have to write a `do_build()` at all!
You can globally define `CFLAGS="-std=c99"`, and omit the `-fPIC`, because it's the default.
Also a global `LDFLAGS="-lssl -lcrypto"` is sufficient, or perhaps even `LIBS="-lssl -lcrypto"` if the `Makefile` uses `$(LIBS)`.

If the work source directory is `${wrksrc}/src`, as it seems, you should define `build_wrksrc="src"`.

Note that in this case the `$PWD` when entering a functions is `${wrksrc}/src` already.

srcpkgs/reaver-wps-fork-t6x/template
```diff
@@ -0,0 +1,30 @@
...
+depends="libpcap sqlite pixiewps"
+makedepends="libpcap-devel sqlite-devel"
+short_desc="Brute force attack against Wifi Protected Setup mod with Pixie Dust Attack"
+maintainer="Felix Hanley <felix@userspace.com.au>"
+homepage="https://github.com/t6x/reaver-wps-fork-t6x"
+license="GPL2"
+distfiles="https://github.com/t6x/${pkgname}/archive/v${version}.tar.gz"
+checksum=b68d988ba7546949b74f3d89ebf9efbd745806906b717f3db7644ea4b12c63b6
+
+CFLAGS="-fPIC $CFLAGS"
```
pullmoll: You can remove that line, because `"-fPIC"` is the default.

srcpkgs/reaver-wps-fork-t6x/template
```diff
@@ -0,0 +1,30 @@
...
+LDFLAGS="$LDFLAGS"
```
pullmoll: This line is wrong because it does nothing.

felix:
I can't seem to get -fPIC into the CCFLAGS, hence the do_build()

srcpkgs/pixiewps/template
felix: Makefile uses CCFLAGS

PR #4949, opened 2016-10-12
New package: kcov-31

NoctuaNivalis:
I needed [kcov](http://simonkagstrom.github.io/kcov/index.html) ([GitHub](https://github.com/SimonKagstrom/kcov)) for checking my coverage, so I had to make a package anyway.

### DIFF
--- a/srcpkgs/kcov/template
+++ b/srcpkgs/kcov/template
@@ -1,0 +1,12 @@
+pkgname=kcov
+version=31
+revision=1
+build_style=cmake
+hostmakedepends="python pkg-config"
+makedepends="binutils-devel libcurl-devel elfutils-devel"
+short_desc="Code coverage tester for compiled programs, Python and shell"
+maintainer="Felix Van der Jeugt <felix.vanderjeugt@gmail.com>"
+license="GPL-2"
+homepage="http://simonkagstrom.github.io/kcov/index.html"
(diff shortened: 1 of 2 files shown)


srcpkgs/kcov/template
```diff
@@ -0,0 +1,11 @@
+pkgname=kcov
+version=31
+revision=1
+build_style=cmake
+makedepends="python pkg-config binutils-devel libcurl-devel elfutils-devel"
```
ebfe: move python and pkg-config to `hostmakedepends`, should fix the non-musl cross builds.

NoctuaNivalis: Thanks for the hint, I hadn't though of testing the build for other architectures. I did now (using your hint) and everything seems to work on my laptop. Let's hope for Travis.

ebfe:
Putting the following pach into `srcpkgs/kcov/patches` fixed the -musl builds

```diff
--- src/engines/ptrace.cc.orig	2016-10-12 16:59:39.621744581 +0200
+++ src/engines/ptrace.cc	2016-10-12 17:00:04.938745887 +0200
@@ -31,6 +31,10 @@
 #define str(s) #s
 #define xstr(s) str(s)
 
+#ifndef __GLIBC__
+typedef int __ptrace_request;
+#endif
+
 enum
 {
 	i386_EIP = 12,
```

NoctuaNivalis:
Thanks a lot, @ebfe.

PR #4950, opened 2016-10-12
KF5 additional packages

dutsky:
Together with my previous PR it's KDE Framework 5 basis to build Plasma and applications packages. I hope you will not kill me for this. God save us.


Vaelatern:
I hate to tell you this... but we do ask that each template altered gets its own commit...

PR #4953, opened 2016-10-13
New package: pycharm-professional-2016.2.3

DirectorX:


srcpkgs/pycharm-professional/template
```diff
@@ -0,0 +1,38 @@
+# Template file for 'pycharm-professional'
+pkgname=pycharm-professional
+version=2016.2.3
+revision=1
+wrksrc=pycharm-${version}
+only_for_archs="i686 x86_64"
+depends="virtual?java-environment giflib libXtst python"
```
DirectorX: TODO: add `pydevd` as dependency

DirectorX: `pydevd` should stay in the helpers directory because pycharm try to upload it to remote machines (like `vagrant`) while remote debugging

PR #4965, opened 2016-10-14
pysolfc: enable sound

styrsven:


ebfe:
I doubt this will work without additional `depends=`.

ebfe:
Ah, overlooked the line

PR #4966, opened 2016-10-14
freetype, cairo, fontconfig: drop ultimate buildoption

ebfe:
Obsolete.


Duncaen:
Can we drop it from `cairo` and `fontconfig` too?

thypon:
OK for me

PR #4971, opened 2016-10-15
New package: roccat-tools-5.5.0 

Somnium42:
This will add the Opensource configuration tool for roccat mice/keyboards and its dependency libgamminggear-0.15.0.

I did split the pkg into multiple Subpkgs to avoid that Users install 21 devices they don't need. This also prevents 21 useless Desktop entries from beeing made. Each device has its configuration tool so they had to choose the right programm after the installation anyway.

### DIFF
--- a/srcpkgs/beets/template
+++ b/srcpkgs/beets/template
@@ -1,6 +1,6 @@
 # Template file for 'beets'
 pkgname=beets
-version=1.4.1
+version=1.4.2
 revision=1
 noarch=yes
 build_style=python2-module
@@ -13,7 +13,7 @@ maintainer="Juan RP <xtraeme@voidlinux.eu>"
 homepage="http://beets.radbox.org/"
 license="MIT"
 distfiles="${PYPI_SITE}/b/${pkgname}/${pkgname}-${version}.tar.gz"
-checksum=373a8723c9e38f5b2933948ee6a4ff8df2ca4156050916c76bd746963b42d693
+checksum=b54c72e220d7696740823d0a4e4f38d57d1e463daaf06da5194a358d3a14ca6a
 
 post_install() {
 	vman man/beet.1


Somnium42:
while crossbuilding for aarch64 etc it aborts with an error.
`CMake Error at cmake_modules/INSTALL_CMAKE_MODULESDIR.cmake:22 (MESSAGE):
  Don't know where to install cmake modules.  Please set
  INSTALL_CMAKE_MODULESDIR manually.
Call Stack (most recent call first):
  CMakeLists.txt:26 (INCLUDE)
`
Someone from the AUR fixed the problem by 
"-DINSTALL_CMAKE_MODULESDIR="/usr/share/cmake-3.3/Modules"
although i have still no idea if this program was ever compiled/intended to on any other aarch than x86

srcpkgs/roccat-tools/REMOVE
```diff
@@ -0,0 +1,5 @@
+case "${ACTION}" in
+purge)
+	rm -rf /var/lib/roccat
+	;;
+esac
```
Vaelatern: No absolute paths. Fix end of file error.

srcpkgs/roccat-tools/template
```diff
@@ -0,0 +1,269 @@
+# Template file for 'roccat-tools'
+
+pkgname="roccat-tools"
+version="5.4.0"
+revision=1
+build_style=cmake
+wrksrc="${pkgname%}-${version%}"
+hostmakedepends='cmake pkg-config'
```
Vaelatern: cmake redundant

Somnium42:
Fixed small typos like newline.
Added glib-devel to hostmakedepends since it prevented building libgaminggear on other arches den X86_64 i686 and X86_64 musl.
roccat-tools isn't crossbuildable, since it checks TryRun. I wrote a small fix for the TryRun problem but it still won't crossbuild so i made it nocross=yes. Not 100% sure if it will build eg aarch64 nativly.

Somnium42:
Im going to be abroad with only my laptop for the next 2 weeks and i was not able to make roccat-tools crossbuilld.
Everything else is well tested and i can't find any obvious errors, although there may be some more format/typo mistakes i missed.

Somnium42:
roccat-tools does crossbuild now but aborts because it isn't able to find .../90-roccat-arvo.rules.  This is the same error libgaminggear had and should be fixed easily by the configure arg. INSTALL_UDEVDIR=...

PR #4976, opened 2016-10-16
New package: extremetuxracer-0.7.3

venv21:


srcpkgs/extremetuxracer/template
```diff
@@ -0,0 +1,21 @@
+# Template file for 'extremetuxracer'
+pkgname=extremetuxracer
+version=0.7.3
+revision=1
+wrksrc=etr-${version}
+only_for_archs="i686 x86_64"
```
Vaelatern: Please explain this limitation.

srcpkgs/extremetuxracer/template
```diff
@@ -0,0 +1,21 @@
...
+build_style=gnu-configure
+hostmakedepends="pkg-config gcc-c++"
```
Vaelatern: `gcc-c++` is already available. Don't list here.

srcpkgs/extremetuxracer/template
```diff
@@ -0,0 +1,21 @@
...
+makedepends="SFML-devel glu-devel"
+depends="SFML glu"
```
Vaelatern: Remove this line, xbps-src is smart enough to detect it:

```
   SONAME: libGLU.so.1 <-> glu>=9.0.0_1
   SONAME: libGL.so.1 <-> libGL>=7.11_1
   SONAME: libsfml-system.so.2.3 <-> SFML>=2.3_1
   SONAME: libsfml-audio.so.2.3 <-> SFML>=2.3_1
   SONAME: libsfml-window.so.2.3 <-> SFML>=2.3_1
   SONAME: libsfml-graphics.so.2.3 <-> SFML>=2.3_1
   SONAME: libstdc++.so.6 <-> libstdc++>=4.4.0_1
   SONAME: libm.so.6 <-> glibc>=2.8_1
   SONAME: libgcc_s.so.1 <-> libgcc>=4.4.0_1
```

srcpkgs/extremetuxracer/template
venv21: I've just used it because it was "suggested" by xnew. I've removed it and it seems to be build on most arches except for arm with musl

srcpkgs/extremetuxracer/template
venv21: I didn't know that, fixed it, thanks

Issue #4436, opened 2016-07-05
Package request: Google Noto Fonts


aeosynth:
https://www.google.com/get/noto/

> Google has been developing a font family called Noto, which aims to support all languages with a harmonious look and feel.

i use noto fonts for the unicode support. on arch there are three packages: `noto-fonts`, `noto-fonts-cjk`, `noto-fonts-emoji`.

Vaelatern:
Is it possible to install in your home directory on a base install (no compilation or make tools available)?
If so, it won't be accepted.

aeosynth:
will try in my vm

aeosynth:
wouldn't this fall under quality requirement 1?

> 1. System: The software **should** be installed system-wide, not per-user.

this is a system font

Vaelatern:
xtraeme refuses to allow fonts that are not compiled in some way into the repo. I can't find the PR where he was harsh on this.

ribmo:
@xtraeme Could exceptions be made for Noto? It is more valuable than just another typeface IMO.

> When text is rendered by a computer, sometimes characters are displayed as “tofu”. They are little boxes to indicate your device doesn’t have a font to display the text.
> 
> Google has been developing a font family called Noto, which aims to support all languages with a harmonious look and feel. Noto is Google’s answer to tofu. The name noto is to convey the idea that Google’s goal is to see “no more tofu”. Noto has multiple styles and weights, and freely available to all. The comprehensive set of fonts and tools used in our development are available in our GitHub repositories. 

I had planned to put together a package for this soon. I will probably still maintain a package locally for my own use but I certainly think others would benefit from it being in the repo.

teh-jazzman:
I'm curious. What makes the Noto fonts more valuable than any other font? Just the fact that they, amongst themselfs, provide most glyphs in the unicode standard? There are other fonts that do that, Unifont for example.

ribmo:
@teh-jazzman, yeah that's exactly what I liked about Noto. I wasn't aware of others so I'll check out Unifont, thanks.

Also, Noto's emoji support is nice. It's nice to be able to read emails from friends and family without squares or empty spaces.

kurumushi:
notosans and notoserif are already in the void package tree. They are included in google-fonts-ttf.

gotwf:
[...]

PR #4977, opened 2016-10-16
New package: noto-fonts-ttf 20161003

pbui:
This is a package for [Google Noto Fonts](https://www.google.com/get/noto/).  While the google-fonts-ttf includes a version of these fonts, it is not complete (e.g. is missing Noto Sans).  This package is a more complete version of the font family.


dutsky:
Sorry, i'm neither dev nor important guy here, but why no Sans? I'm just interested because Noto Fonts needed for KDE5

pbui:
Ah, sorry, I meant that the google-fonts-ttf package, which is already in void-packages, includes Noto Sans and Noto Serif, but not Noto Mono.

This new package includes all three and only the Noto family (while the google-fonts-ttf package includes many more fonts).

aeosynth:
there's an unresolved issue (#4436) about whether noto should be packaged. apparently void is weird about packaging fonts.

srcpkgs/noto-fonts-ttf/template
```diff
@@ -0,0 +1,22 @@
...
+license="OFL-1.1"
+homepage="https://www.google.com/get/noto/"
+distfiles="https://github.com/googlei18n/noto-fonts/archive/${_githash}.tar.gz"
+checksum=02075aaee1f83c876f0d9ef66563218be9871980e642aaf3beb037bee22005d2
+conflicts="google-fonts-ttf>=0"
+
+do_install() {
+	vmkdir usr/share/fonts/TTF
+	install -m644 hinted/Noto*.ttf ${DESTDIR}/usr/share/fonts/TTF
+	vlicense LICENSE
```
asergi: `SIL` license is part of `base-files`. So, `vlicense` is unnecessary.

pbui:
@asergi I've updated the template to no longer include `vlicense`.

PR #4982, opened 2016-10-17
linux4.4: update to 4.4.25

pbui:


asergi:
Already updated in 4556962. Thanks anyway.

PR #4984, opened 2016-10-17
New package: ferm-2.3

jnbr:
I removed the systemd unit file from the installer but I did not include a runit script.
I don't think that a runit script makes sense, beacause people probably want to start their iptable rules before setting up the network, so adding it to the dhcpcd script should be sufficient.


jnbr:
I missed to add perl as a dependency. Now fixed.

Vaelatern:
I think perl should be added as a `hostmakedepends`

jnbr:
You're right. I added perl to `hostmakepends` as well.

jnbr:
I got the last commit wrong and missed to add the template file, so perl was still missing in `hostmakedepends`. 
@Vaelatern can you please check the package again?

jnbr:
updated from 2.3 to 2.3.1

PR #4985, opened 2016-10-17
all: drop hostmakedepends provided by build_style

ebfe:


Vaelatern:
I'd rather not deal with the fallout, or make this sweeping change.

Perhaps the script you used to do this can augment `xlint` so at some point people will see it and fix it on smaller levels. The concern came up when the change for `cmake` was proposed, that this would be a bear to take care of. I think it's cleaner to fix things as we come across them, rather than all at once.

If you disagree with the above, then I'm happy with the changes.

ebfe:
There should be no fallout. The argument about `xlint` is in theory correct, in practice nothing is ever fixed (see the 700 packags that have failing license lints).

pullmoll:
I don't see a problem merging this now. The packages aren't rebuilt and should they fail on the next update, they can still be fixed.

Gottox:
please rebase and merge

PR #4996, opened 2016-10-18
New Package: openra-20161019

Hoshpak:
adds openra and also nuget as a build dependency.


Hoshpak:
I could use some help understanding what travis is doing and what could possibly cause the errors reported during the travis build. The build on x86_64 works which was kind of expected since I tested extensively on that platform myself.

However the build on i686 fails with an error during the call to NuGet.exe during the build phase. I encountered a similar error while creating the template because mono didn't have the CA in its store of trusted root certificates. This is why I added the import of the Mozilla CA file in pre_build. The travis build log shows that the certificates were imported but the command still fails. I am completely unable to reproduce the error myself, I have built the package with the same template about 20 times in a i686 VM and never got any failures whatsoever.

The other build jobs act even more strange. They build nuget which is correct since it's a build dependency, register the package but then try to build the package again as a build dependency for openra. This fails because the build environment is seemingly not cleaned before and the source code is still checked out there.

In theory (that's to say if both xbps-src and mono do their job properly) it should all be architecture independent code and behave the same across processor architectures.

Hoshpak:
I rebased on top of master to get the fix from #5013 and replaced the calls to install with vinstall. The build failure on i686 now doesn't occur any more, I don't understand why, though. the arm builders still try to build the nuget package twice and at least two of the musl builders are for some reason missing "proot-x86_64".

Hoshpak:
i686 works now, rebuilding of nuget on the other builders remains.

Hoshpak:
Binary packages are available at https://repo.cyber-everything.de in case anyone is interested.

PR #4997, opened 2016-10-19
gegl3: addition of lib package to support gimp 2.9.x

antonio-malcolm:
Added the lib package to gegl3 as a dependency for Gimp v2.9.x, for which I also have a template.


Gottox:
* use the commit message format as described at https://github.com/voidlinux/void-packages/blob/master/CONTRIBUTING.md#committing-your-changes
* increase the revision of this package.

antonio-malcolm:
Done.

Gottox:
okay, after looking into the package: What's the reason for the split?

Gottox:
Also note: this change will break all packages using gegl3, including gegl3-devel.

antonio-malcolm:
Wait- how is this going to break all packages using gegl3?

antonio-malcolm:
How does this not break with other packages which have lib subpackages?
What would be the preferred approach?

Gottox:
A rebuild of all dependants would fix this situation. But I don't think this is worth the effort. Just depend on gegl3-devel in the template and you'll be fine.

antonio-malcolm:
Ok, that would mean it will need it in both hostmakedepends and depends.
Otherwise, simple is my preferred approach to, thanks. 
Will close this pull request and delete the branch.

PR #4998, opened 2016-10-19
New package: gimp2.9-2.9.2_1

antonio-malcolm:
This is the new 2.9.x Gimp, with improved bit depth and other features.


ebfe:
Why a new package rather than updating the existing one?

Gottox:
keep in mind that 2.9.x is the development branch of gimp 2.10.

antonio-malcolm:
We can certainly do that, but while 2.9.x is reportedly stable (used in production by a few), it is still a developer's release (as it will be, until 2.10.x or 3.x), and I thought it would be good to give folks the choice of using this new version, or staying with 2.8.x.

If folks would rather have this as an update, I can move the code over.

antonio-malcolm:
(BTW, this is for i686 and x86_64 only, so it won't pass the ARM tests.)

Gottox:
I'd like to wait untill gimp releases a stable version.

antonio-malcolm:
Good luck. That will be a long-time coming, knowing the Gimp team (esp. if they decide to skip 2.10, for a 3.x release). There's a lot of capability in 2.9, which you won't get with 2.8. I'll keep my branch up,and keep it updated, with the template, for anyone who wants it.

PR #5002, opened 2016-10-19
fdupes: update to 1.6.1

yopito:
Let's see if Tracis CI builds are fine with musl libc.


yopito:
builds with musl are fine. 2nd try: fixed xlint checks.

PR #5004, opened 2016-10-20
otter-browser: update to 0.9.11

diogoleal:


srcpkgs/otter-browser/template
```diff
@@ -1,6 +1,6 @@
 # Template file for 'otter-browser'
 pkgname=otter-browser
-version=0.9.10
+version=0.9.11
 revision=2
```
Vaelatern: Reset revision

PR #5008, opened 2016-10-20
New package: acoustibrainz-client

felix:


srcpkgs/acousticbrainz-client/template
```diff
@@ -0,0 +1,38 @@
+# Template file for 'acousticbrainz-client'
+pkgname=acousticbrainz-client
+version=0.1
+revision=1
+only_for_archs="i686 x86_64"
```
Duncaen: Why?

ayghor:
just noticed a small typo on teh subject:

[voidlinux/void-packages] New package: acoustibrainz-client (#5008)

is missing teh last c of acoustic

-- 
Igor

pullmoll:
If you really need to install one of two binary packages, you need to add a case/esac statement to set both, distfiles and checksum for the binary.

I'd still prefer to have this built from source. At the very least this binary package should have a `-bin` extension, if some day the same package is built from source. This already happened for some packages in Void, e.g. lightzone and libreoffice.

felix:
Okay, leave this for now and I will play with a local build from source.

Issue #4627, opened 2016-08-16
'xbps-src show' should work on broken packages


ebfe:
Short of reimplementing a shell, 'xbps-src show' is the only way to get metadata about the templates in (a somewhat reasonable) machine readable format.

PR #5014, opened 2016-10-20
xbps-src: allow show of broken pkgs

Vaelatern:
#4627 also.


Vaelatern:
No comments, assuming OK.

PR #5019, opened 2016-10-21
broadcom-wl-dkms: add patch to fix dkms on 4.8

Nimloth:


Vaelatern:
From the reference, it seems the patch is not necessary? Please explain further.

Nimloth:
Hi there, how do you mean: from the reference?
If you refer to the wording

"We believe that the bug you reported is fixed in the latest version of
broadcom-sta"

on the Debian bug report, that is just their template when closing a bug. The new version they mention includes this very patch.
This can be seen further down in the message:

"Changes:
 broadcom-sta (6.30.223.271-4) unstable; urgency=medium
 .
   * Added patch from Koos Vriezen to work around kernel crash with version
     4.8+ (THANKS!) (closes: #839629)"

Issue #4909, opened 2016-10-07
travis builds "randomly" failing


ebfe:
See https://travis-ci.org/voidlinux/void-packages/jobs/165454444 for an example.

```
=> schismtracker-20160913_1: running pre_configure ...
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
    LANGUAGE = (unset),
    LC_ALL = (unset),
    LC_COLLATE = "C",
    LANG = "en_US.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
    LANGUAGE = (unset),
    LC_ALL = (unset),
    LC_COLLATE = "C",
    LANG = "en_US.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
    LANGUAGE = (unset),
    LC_ALL = (unset),
    LC_COLLATE = "C",
    LANG = "en_US.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
sh: git: command not found
/usr/bin/m4: stack overflow
```

I think @pullmoll tracked this down to proot? somewhere but i can't find the issue atm.

pullmoll:
It happens with `proot` **and** with `uchroot`, not with `uunshare`.
With the former two the reconfigure of the glibc-locales segfaults.

ackalker:
Perhaps related, `proot` segfaults for me when I try and `proot` into a mounted x86_64-musl installation from  x86_64 host:
```sh
$ proot -S mnt
proot info: pid 4251: terminated with signal 11
$ gdb -q proot
Reading symbols from proot...Reading symbols from /usr/lib/debug//usr/bin/proot...done.
done.
(gdb) run -S mnt
Starting program: /usr/bin/proot -S mnt
proot info: pid 4526: terminated with signal 11
[Inferior 1 (process 4522) exited with code 0377]
(gdb) bt
No stack.
```
Huh? Oh... could be that it is actually a child process spawned by `proot` which is segfaulting. Let's try again:
```sh
$ gdb -q proot
Reading symbols from proot...Reading symbols from /usr/lib/debug//usr/bin/proot...done.
done.
(gdb) set follow-fork-mode child
(gdb) run -S mnt
Starting program: /usr/bin/proot -S mnt
[...]

PR #5030, opened 2016-10-23
Fix various packages that conflict with inetutils alternatives.

CMB:
Here are descriptions of the conflicts and my fixes.

inetutils-ifconfig and net-tools:
Both provide the ifconfig alternative group, but the manpage symlink
in net-tools is broken.  It is named ifconfig.1, and it points to
net-tools-ifconfig.8 in /usr/share/man/man1.
Fixed by pointing the ifconfig.1 symlink to
/usr/share/man/man8/ifconfig.8.

inetutils-syslog:
Provides the logger group, which will clobber logger from util-linux when
installed.  Added the logger alternative group to util-linux.

inetutils-traceroute:
Provides the traceroute group, which clobbers things installed by
the traceroute package.
Renamed traceroute's binary and manpage, and the
traceroute package now provides the alternative group.

### DIFF
--- a/srcpkgs/net-tools/template
+++ b/srcpkgs/net-tools/template
@@ -1,7 +1,7 @@
 # Template file for 'net-tools'
 pkgname=net-tools
 version=1.60.20150526git
-revision=2
+revision=3
 hostmakedepends="git"
 short_desc="Basic networking tools"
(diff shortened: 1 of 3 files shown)


ebfe:
The glibc build failures are a case of #4909.

PR #5034, opened 2016-10-23
New Package: xsane-0.999

Hoshpak:


Hoshpak:
Is the musl failure also a case of #4909  ? I fails to even extract the archive however my local x86_64-musl VM not only extracts it without complaining but also builds the whole package without issues.

Hoshpak:
seems to have been, it works now on musl.

srcpkgs/xsane/template
```diff
@@ -0,0 +1,25 @@
+# Template file for 'xsane'
+pkgname=xsane
+version=0.999
+revision=1
+build_style=gnu-configure
+hostmakedepends="pkg-config sane-devel"
```
Vaelatern: I think this variable is both missing some packages and should not have `sane-devel` without a good reason given in a comment.

Hoshpak: That must have been a leftover of a misguided attempt to get the package to crosscompile. (which is doomed anyway as long as gimp can't be crosscompiled) I removed it. Whatdo you believe to be missing? Everything else should be provided by build_style.

asergi:
https://github.com/voidlinux/void-packages/commit/d106e114930e5a990ab8d9adddc7d101ffc8c9e7

PR #5036, opened 2016-10-23
New package: vscode-1.6.1

### LINKED ISSUE #3208: Package Request: Visual Studio Code
Source releases available here: https://github.com/Microsoft/vscode/releases

it depends on electron and node

jannisch:
I made these decisions which could be discussed on:
- named it vscode not _visual-studio-code_. This is also used for the [source repo](https://github.com/Microsoft/vscode)
- copy the whole installation to /usr/lib/vscode because it does not follow the UNIX directory convention and symlink executable to /usr/bin
- name the executable in /usr/bin `vscode` as well, not _code-oss_ as or _code_ as generated because I found this confusing

Fixes #3208

### DIFF
--- a/srcpkgs/vscode/patches/enable_extensions_store.patch
+++ b/srcpkgs/vscode/patches/enable_extensions_store.patch
@@ -1,0 +1,39 @@
+--- product.json	2016-10-13 17:20:53.000000000 +0200
++++ product.json.with_extensions	2016-10-24 02:04:55.813371552 +0200
+@@ -1,6 +1,6 @@
+ {
+ 	"nameShort": "Code - OSS",
+-	"nameLong": "Code - OSS",
++	"nameLong": "Visual Studio Code - OSS",
+ 	"applicationName": "code-oss",
+ 	"dataFolderName": ".vscode-oss",
+ 	"win32MutexName": "vscodeoss",
+@@ -12,5 +12,26 @@
+ 	"win32AppUserModelId": "Microsoft.CodeOSS",
+ 	"darwinBundleIdentifier": "com.visualstudio.code.oss",
+ 	"reportIssueUrl": "https://github.com/Microsoft/vscode/issues/new",
+-	"urlProtocol": "code-oss"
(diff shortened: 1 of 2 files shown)


jannisch:
Just discovered that it can't connect to the internet. Do you have any idea why it works on the prebuilt binary?

Sorry!

ebfe:
> Error: API rate limit exceeded for 52.22.60.255. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)

jannisch:
where did you read that? :confused:

ebfe:
In the build log...

jannisch:
Ah, thanks. But previous builds worked.

The AUR maintainers have written [this patch](https://aur.archlinux.org/cgit/aur.git/tree/product_json.patch?h=visual-studio-code-oss) that fixes the problem.

Would this conflict with the licensing guidelines? The package itself does not ship nonfree code but nonfree extensions could be installed by the user I guess.

jannisch:
The latest build issue seems to be a common problem with version incompatibilities caused by a recent update. I need some time to fix that.

jannisch:
The problem does not occur when building from the bleeding edge, so should I just wait for the next release or try to patch it some way?

pullmoll:
[halloween mode one]@pullmoll shudders in anticipation of big, untracable, unexpected and unavoidable failures arising from adding this package to the Void repo[/halloween mode off]

Just joking around, being drunken and fubar.

jannisch:
Happy halloween @pullmoll :jack_o_lantern: :smile: 
But please don't blame me for creating this PR, neither am I a fan of this nodejs hype nor could I expect the breakage :no_mouth:

pullmoll:
I was really just kidding! I don't blame anyone for creating PRs. Happy :jack_o_lantern: to you too.

jannisch:
1.7.1 is out now but I have to carefully check the complicated license terms again and if it's possible to compile electron instead of downloading any binary code. Didn't knew this before

PR #5038, opened 2016-10-24
libu2f-host: fix udev rules

the-maldridge:
anything less than 777 doesn't seem to work.


ebfe:
Whatever the problem is, making security related devices world writable doesn't look like the right solution.

ebfe:
The question is why the `TAG+=uaccess` doesn't work. The old version used `GROUP="plugdev", MODE="0660"` https://github.com/Yubico/libu2f-host/blob/master/70-old-u2f.rules.

ebfe:
#4909

ebfe:
Apparently `TAG+=uaccess` is systemd thing.

the-maldridge:
yes, tags are only applied if the system was booted with systemd, otherwise they are silently ignored.  I can revert to the old group, but users seemed more applicable.

the-maldridge:
Anyone else want to weigh in on what the system group should be?  The build passes now that the proot issues have been dealt with.

PR #5041, opened 2016-10-24
python-pycparser: fix permissions

jnbr:
With version 2.16, files in `pycparser.egg-info` are not world readable anymore.
This breaks the library for all users except members of group root.


asergi:
You're right. Actually, all files in the tarball have bad permissions.

srcpkgs/python-pycparser/template
```diff
@@ -15,6 +15,11 @@ license="3-clause-BSD"
 distfiles="${PYPI_SITE}/p/pycparser/pycparser-${version}.tar.gz"
 checksum=108f9ff23869ae2f8b38e481e7b4b4d4de1e32be968f29bbe303d629c34a6260
 
+post_extract() {
+	# fix permissions
+	chmod 0644 pycparser.egg-info/*
```
asergi: Just to be sure: `chmod -R go+rX .`

PR #5044, opened 2016-10-24
busybox: Added poweroff/halt/reboot/init to be compiled

jcjordyn130:


chneukirchen:
What is this good for? (I guess `halt -f` can make sense, but the one in runit-void isn't exactly large).

Also, why all the other changes?

NACK for now.

jcjordyn130:
I don't know where the other changes came from, I only enabled init/poweroff/reboot/halt in menuconfig.

Vaelatern:
Why do you want busybox's version of init? Would that play well at all with anything else?

jcjordyn130:
I tested it on a fresh install of void with xfce in a vm and on my personal void system, and it didn't conflict with anything as it doesn't override init/poweroff/reboot/halt from runit

pullmoll:
There can't be a conflict because nothing is installed. Take a look at https://travis-ci.org/voidlinux/void-packages/jobs/170271420#L563 and expand the line.

Vaelatern:
So the advantage is...?

jcjordyn130:
None really, its just for people who want to use busybox init. I do get your point though

PR #5048, opened 2016-10-25
Poezio

delehef:
Adds [poezio](http://poez.io/en/) and its dependencies.


srcpkgs/python3-aiodns/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'python-aiodns'
```
chneukirchen: Wrong name.

srcpkgs/python3-slixmpp/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'python3-slixmpp'
+pkgname=python3-slixmpp
+version=1.2.1
+revision=1
+noarch=yes
+wrksrc="slixmpp-${version}"
+short_desc="Slixmpp is an elegant Python library for XMPP"
```
chneukirchen: Don't use "XXX is..."

chneukirchen:
Cross-build needs fixing, else LGTM.

chneukirchen:
The culprit is `-I/usr/include/python3.5m`.

delehef:
Thanks, I fixed the two typos.

However, I don't have an ARM machine to experiment further with the ARM build and I'm only using the standard `python3-module`, so I don't see how I could fix it.

Should I limit architectures to i686 & x86_64?

chneukirchen:
Fixed.

PR #5066, opened 2016-10-28
New package: FreeRADIUS-3.0.12

the-maldridge:
Currently marked nocross as the build is very much convinced it needs to execute build artifacts while configuring that have been cross compiled.

### DIFF
--- a/srcpkgs/FreeRADIUS/template
+++ b/srcpkgs/FreeRADIUS/template
@@ -1,0 +1,64 @@
+
+# Template file for 'FreeRADIUS'
+pkgname=FreeRADIUS
+version=3.0.12
+revision=1
+wrksrc=freeradius-server-${version}
+build_style=gnu-configure
+nocross=yes # Not supported by upstream
+makedepends="talloc-devel libressl-devel mit-krb5-devel pam-devel \
+ libmariadbclient-devel postgresql-libs-devel json-c-devel"
+system_accounts="_freeradius"
+make_dirs="/etc/raddb 0750 _freeradius _freeradius"
+short_desc="The world's most popular RADIUS Server"
+maintainer="Michael Aldridge <aldridge.mac@gmail.com>"
+license="GPL-2"
+homepage="http://freeradius.org"
+distfiles="ftp://ftp.freeradius.org/pub/freeradius/freeradius-server-${version}.tar.bz2"
+checksum=fe4e1f52cc2873f6aee2b12b0f03236978e4632f2acf298f834686b240c4183d
+
+pre_install() {
+	# FreeRADIUS uses 'R' instead of 'DESTDIR'
+	export R=${DESTDIR}
+}
+
+post_install() {
+	# Remove the default generated certificates, it would be nice to just
+	# not build these, but the Makefiles are inordinately complex and
+	# cannot be easily modified to not build the certificate files.
+	_CERTDIR=${DESTDIR}/etc/raddb/certs
+	rm ${_CERTDIR}/index*
+	rm ${_CERTDIR}/*.crt
+	rm ${_CERTDIR}/*.csr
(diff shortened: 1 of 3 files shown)


Vaelatern:
I feel like there should be something in `conf_files`, but I do not know what other than `/etc/raddb/*` would match what it should and not too much else.

the-maldridge:
given that conf_files doesn't seem to take a wildcard, what is the suggestion for handling the many many files in /etc/raddb?

Vaelatern:
I would not be opposed to a subpackage just for the modules, but nor do I think it's a clean solution. A find(1) call could provide a listing, but it does not seem clean either.

Vaelatern:
The wildcard `*` is accepted, just not recursively.

Vaelatern:
I'm not convinced that the way OpenBSD does it, which is install the config in /usr/share/examples and tell the user to install it themselves, is all wrong. Considering the number of ways to customize how FreeRadius acts, it may be the fastest and cleanest.

srcpkgs/FreeRADIUS/template
```diff
@@ -0,0 +1,62 @@
...
+	# Install the service
+	vsv FreeRADIUS
+
+	#  While the config installed by the makefiles is usable, its
+	# also very complex and not likely to be a good solution for
+	# most sites installing FreeRADIUS, so the existing config
+	# will be moved out of the way here so that experienced admins
+	# can install a configuration set that is appropriate for the
+	# specific site.
+	mkdir -p ${DESTDIR}/usr/share/examples/${pkgname}/
```
Vaelatern: `vmkdir`

srcpkgs/FreeRADIUS/template
```diff
@@ -0,0 +1,62 @@
...
+
+	# Install the service
+	vsv FreeRADIUS
+
+	#  While the config installed by the makefiles is usable, its
+	# also very complex and not likely to be a good solution for
+	# most sites installing FreeRADIUS, so the existing config
+	# will be moved out of the way here so that experienced admins
+	# can install a configuration set that is appropriate for the
+	# specific site.
```
Vaelatern: Please add a note to `INSTALL.msg` basically saying `No configuration is installed to /etc/raddb/, look in /usr/share/examples/FreeRADIUS/ and copy them to /etc/raddb/ as you see fit`

srcpkgs/FreeRADIUS/template
```diff
@@ -0,0 +1,62 @@
...
+	mkdir -p ${DESTDIR}/usr/share/examples/${pkgname}/
+	mv ${DESTDIR}/etc/raddb/ ${DESTDIR}/usr/share/examples/${pkgname}
```
Vaelatern: This creates `${DESTDIR}/usr/share/examples/${pkgname}/raddb/`, if this was intentional please document it as such.

srcpkgs/FreeRADIUS/template
```diff
@@ -0,0 +1,64 @@
+# Template file for 'FreeRADIUS'
+pkgname=FreeRADIUS
+version=3.0.12
+revision=1
+wrksrc=freeradius-server-${version}
+build_style=gnu-configure
+nocross=yes # Not supported by upstream
+makedepends="talloc-devel libressl-devel mit-krb5-devel pam-devel \
+ libmariadbclient-devel postgresql-libs-devel json-c-devel"
+system_accounts="freeradius"
```
Vaelatern: Finally, can we have the system account start with an underscore, to set it apart from user groups

srcpkgs/FreeRADIUS/template
```diff
@@ -0,0 +1,65 @@
+
```
Vaelatern: Trim this line?

PR #5068, opened 2016-10-28
Adapta: update to 3.22.2.22.

demiaster:


chneukirchen:
Is it really necessary to update this package every single day?

demiaster:
I could update only when the minor c (a.b.c.d) releases comes out, if you think it is better. Anyway, since it is the first time for me maintaining a package, may I ask why is not ok to update so frequently?

pullmoll:
And I thought `lollypop` was the most frequently updated package. :lollipop: 

In theory it's ok to update every time, but then there may be changes which are not **that** important, i.e. not that kind of important as in *live saving important*.

For lollypop I always wait until after the weekend, which usually allows me to skips 3-4 builds. :smile:

demiaster:
@pullmoll thank you for the clarification, I will follow your advice :smile:

PR #5070, opened 2016-10-28
mono: update to 4.6.2.6

Hoshpak:
While I was working on another .Net based software I noticed that the mono package in Void is relatively outdated (version 4.0.1.44 was released more than a year ago) and is possibly vulnerable to one ore more of the [security vulnerabilties](http://www.mono-project.com/docs/about-mono/vulnerabilities/) published since then. This pull request updates the package to the current version 4.6.1.5. Another issue I ran across was that by default there are no CA certificates installed in the mono trust store which lead to failures establishing any TLS connection. So I added an install script that syncs the certificates with the ones installed on the system using cert-sync as recommended in the [Security FAQ](http://www.mono-project.com/docs/faq/security/). Since the package failed to build due to a missing pointer definition, I created a small patch to fix that and reported the issue to [upstream](https://bugzilla.xamarin.com/show_bug.cgi?id=46159).


Hoshpak:
great, I guess that stack overflow on x86_64 means I definitely managed to run into #4909 . I wonder if travis will build on any architecture...

pullmoll:
It's about time to track down the reason why `proot` and `uchroot`  segfault here.
It's just that I have no (initial) idea what to `gdb` to find out what's going nuts there.

Hoshpak:
rebased on top of master to get the changes of #5186 , the travis build succeeded.

asergi:
https://github.com/voidlinux/void-packages/commit/6aaa3f55aff5621328c540d8543d8db886dd3372
https://github.com/voidlinux/void-packages/commit/653309c1880069bf17b8e679c8986112f203bc4c

PR #5075, opened 2016-10-29
lemonbuddy: update to 2.1.10

jaagr:


srcpkgs/lemonbuddy/template
```diff
@@ -1,22 +1,27 @@
...
-_i3ipcpp_version=0.3.0
-_xpp_version=1.0.0
+version=2.1.9
+revision=1
+wrksrc="${pkgname}-${version}"
+_i3ipcpp_version=0.5.1
+_xpp_version=1.3.0
 build_style=cmake
 configure_args="
+	-DCMAKE_CXX_COMPILER=clang++
```
Gottox: Any reason why not to use the default CC?

srcpkgs/lemonbuddy/template
```diff
@@ -25,14 +30,15 @@ distfiles="
 	https://github.com/jaagr/xpp/archive/${_xpp_version}.tar.gz
 	${build_option_i3:+https://github.com/jaagr/i3ipcpp/archive/v${_i3ipcpp_version}.tar.gz}"
```
Gottox: use ```$(vopt_if '...')``` instead of ```${build_option_i3:+...}```

srcpkgs/lemonbuddy/template
Gottox: Ah, I see, C++14 support.

PR #5076, opened 2016-10-29
WIP: common/shlibs -> common/shlibs.d

rngd:
this is a quick proof-of-concept to see if the idea gets any traction:  i'd like to see common/shlibs split into multiple files, one per package.

expected benefit: reduced probability of merge conflicts.


Gottox:
Do you had any merge conflicts with that file?

Vaelatern:
Expected loss: many many many files where one will do just fine. Vote to close.

pullmoll:
I thought about this approach several months ago and tried myself, but now have to agree with @Vaelatern . Having to scan several thousand files for every build, perhaps even several times, is not worth the rare trouble caused by common/shlibs merge conflicts.

Gottox:
Merge conflicts shouldn't happen to that file anyway due to .gitattributes.

PR #5078, opened 2016-10-29
New package: tinyssh-20160812

CMB:


srcpkgs/tinyssh/template
```diff
@@ -0,0 +1,30 @@
+# Template file for 'tinyssh'
+pkgname=tinyssh
+version=20160812
```
chneukirchen: This is the upstream numbering scheme, probably ok then.  Usually we use 0.0.yyyymmdd for date-releases.

srcpkgs/tinyssh/template
```diff
@@ -0,0 +1,30 @@
...
+revision=1
+build_style=gnu-makefile
+make_dirs="/etc/tinyssh 0755 root root"
+short_desc="A minimalistic SSH server"
+maintainer="Christopher Brannon <chris@the-brannons.com>"
+license="public domain"
```
chneukirchen: use "Public Domain"

srcpkgs/tinyssh/template
```diff
@@ -0,0 +1,30 @@
...
+	# Build system needs a way to disable tests. Here's a dirty hack.
+	echo 'int main(void) { return 0; }' > cmbbogustest.c
+	for i in tinyssh-tests/*test.c crypto-tests/*test.c; do
+		cp cmbbogustest.c "$i"
+	done
+}
+
+post_install() {
+	vlicense LICENCE
+	vsv tinysshd
```
chneukirchen: AFAIU, with our tcpserver, this will listen on IPv4 and IPv6 yes?

CMB:
> This is the upstream numbering scheme, probably ok then. Usually we use 0.0.yyyymmdd for date-releases.

Yes, this is their versioning scheme, but I don't know whether they want
to change it someday, since it's technically alpha software.
So now I wonder if I shouldn't use 0.0.yyyymmdd after all, to be more
future proof?  Another thing I forgot about is that apparently tags
don't necessarily correspond to releases.
E.G., upstream tagged 20161001, but this isn't a release.  It isn't
mentioned in the homepage or in the github releases page for the
project.  That'll give false positives for update-check, so I need to
add tinyssh/update and repush.

CMB:
It's nocross because it runs target binaries at build time.
One group is the tests, which can't be fully disabled.  I replaced them
so they always pass, but the build system is still going to try and run
them.  No way around this other than patching the build system.
It also computes sysdeps at build time using target binaries, and
there's no way around that either.  I can't even give it a prebuilt
sysdep directory, AFAIK.

CMB:
Yes, it'll listen on both v4 and v6.

CMB:
Fixed "Public Domain".  Added an update file that doesn't give false
positives.  Don't want to mess with changing the version number, since
you said this would be ok, and making 0.0.yyyymmdd play nicely with
update-check doesn't seem possible, since upstream does use yyyymmdd as
the version number.

ebfe:
> Don't want to mess with changing the version number, since
> you said this would be ok, and making 0.0.yyyymmdd play nicely with
> update-check doesn't seem possible, since upstream does use yyyymmdd as
> the version number.

You could override version in the `update` file. Stripping the `0.0.`
should work?

CMB:
Oh nice, that had not occurred to me.  Thank you.
Updated to use 0.0.yyyymmdd format.

ebfe:
@chneukirchen good to go now?

Issue #4825, opened 2016-09-23
WIP: Void built with gcc-6.2.1


pullmoll:
#### I started over to build Void Linux with `gcc-6.2.1`

The hint came from @chneukirchen that  [Arch Linux](https://www.archlinux.org/packages/core/x86_64/gcc/) uses gcc-6.2.1.

The compiler was cloned with

```
git clone git://gcc.gnu.org/git/gcc.git gcc-6.2.1
cd git-6.2.1
git checkout gcc-6-branch
cd ..
```

Then a tarball was created with

```
find gcc-6.2.1 \! -type d -print0 | \
    sort -z | \
    grep -z -v \.git/ | \
    tar cvf - -T - | \
    xz -v -9 -T 0 > gcc-6.2.1.tar.xz
```

I created an `etc/conf` with the following contents

```
XBPS_MASTERDIR=/tmp/gcc6
XBPS_HOSTDIR=/home/void/void-gcc6
XBPS_CHROOT_CMD=uunshare
XBPS_MAKEJOBS=12
XBPS_CCACHE=yes
XBPS_DISTFILES_MIRROR="https://sources.voidlinux.eu/ http://mirror.netcologne.de/gentoo/distfiles"
```

and then did run `./xbp-src zap` and `./xbps-src bootstrap`.
#### The gcc-6.2.1 homegrown tarball of 2016-11-11 is available at

```
distfiles="http://distfiles.voidlinux.de/gcc-6.2.1/gcc-6.2.1.tar.xz"
checksum=6bc5af0d2138bb8492c62ced00776b76bbc98b7297287aaa55adbe7d07786ac8
```
#### The repository including the patches required to bootstrap with gcc-6.2.1 is available at

https://github.com/pullmoll/void-packages-gcc6.2.1
#### Packages built will be uploaded to

| Everything built | Fast, but only a subset of pkgs [<a href="#space">1</a>] |
| --- | --- |
| http://voidlinux.myds.me/repo/current/ | http://repo.voidlinux.de/current |

**Note:** The musl libc pkgs are all in the same directory so as to save space for noarch pkgs.
#### Live ISOs created from these pkgs will be uploaded to

http://voidlinux.myds.me/repo/live/
#### Current package stats - updated  every now and then

| Architecture | Package count |
| --- | --- |
| x86_64 | 7028 |
| x86_64-musl | 6877 |
| i686 | 7043 |
#### Issues: open and solved
[...]

PR #5086, opened 2016-10-30
gcc: update to gcc-6.2.1

pullmoll:
Review is required.

Meanwhile testing the (new) bootstrap.


Gottox:
Is the list of failing packages in https://github.com/voidlinux/void-packages/issues/4825 up to date?

pullmoll:
Yep,  the list was created with the 2nd update I tried.

pullmoll:
Oops: The patches for gcc6 fail here with the latest.

pullmoll:
It looks as if this patch is already included in macro.c, so we just can remove the patch.

pullmoll:
More failures when bootstrapping this one. There's a problem with `libatomic` now.
Investigating.

Gottox:
This is todays snapshot release.

pullmoll:
Obviously upstream is adding fixes and changes, so we need to cope with it. Most or all of them will be in the official release anyway. You or anyone else could test the bootstrap with this one as well and see what needs to be fixed.

pullmoll:
Reverting to an earlier snapshpt? Hm, I'd rather move on forward to the release version.
I don't expect complicated fixes to be required now, but just minor things.

The libatomic failure may be just a missing configure option.

pullmoll:
Here's the current problem with libatomic:
http://misc.voidlinux.de/gcc-6.2.1-build.txt

pullmoll:
Haha.. now I got it. You checked out the latest gcc-7.0.0 branch, not gcc6!

Gottox:
argh! let me fix.

pullmoll:
Darn. Then I need to restore all the `libcpp-source_date_epoch.patch` stuff, but of course I force pushed the changes :(

pullmoll:
Let's drop this one and try again later.

PR #5096, opened 2016-11-01
New package: elvish-0.5

diogoleal:


srcpkgs/elvish/template
```diff
@@ -0,0 +1,13 @@
+# Template file for 'elvish'
+pkgname=elvish
+version=0.5
+revision=1
+build_style=go
+go_import_path="github.com/elves/elvish"
+short_desc="A novel Unix shell"
+maintainer="Diogo Leal <diogo@diogoleal.com>"
+license="GPL-3"
```
Vaelatern: BSD unless you specifically know differently

diogoleal: @Vaelatern changed.

PR #5097, opened 2016-11-01
ntpd: redirect output to /dev/null

Hoshpak:
ntpd currently prints it output to tty which is not necessary since ntpd
itself already logs to syslog if a syslog daemon is running. By redirecting
output of the process to /dev/null (stdout and stderr), this is avoided.


Vaelatern:
Package for the commit message is `ntp` not `ntpd`. Please fix.

Hoshpak:
fixed and pushed

PR #5099, opened 2016-11-01
plantuml: prevent unsupported PDF generation

juliendehos:
PDF generation is not enabled in plantuml since it requires heavy supplementary jar files (#4832).
This PR prevents PDF generation and suggests using EPS + epstopdf (instead of displaying a java callstack).
Switching "PDF" to "EPS+epstopdf" automatically is not that easy because of the "-output" option in plantuml; so just displaying a warning message (this PR) may be a reasonable compromise.


ayghor:
<3

Vaelatern:
Please do this differently by adding a file to the files dir for this package, and moving that into position.

juliendehos:
Oops. I forgot this PR. I'll try to fix that later...

Issue #2066, opened 2015-07-22
Package Request: 0 AD


johnku1:
http://play0ad.com/

thypon:
Same for unvanquished. I'll package whenever a stable release is out. Wasting repo space with alpha nonfree packages it's not a priority.

PR #5102, opened 2016-11-01
New Package: 0ad-0.0.21

Hoshpak:
requested in #2066, also contains gloox as build dependency.


Hoshpak:
Again a travis failure:

It seems that building the big -data package uses too much RAM for xbps-create so travis kills the process. I don't think there is anything I could do about it, the packages build fine locally (and probably also will on the build server which should have more available RAM than a travis container).

CherryMan:
Why was this closed?

Hoshpak:
I usually close PRs that aren't being merged after some time. In this spefic case I don't think the template will ever meet the quality standards to be included in the main repository since the build system is a mess, it includes a lot of bundled third-party libraries and won't cross-compile.

If you are interested in using the template to build 0ad yourself or improve on the template, I currently maintain it in my private fork of void-packages at https://git.cyber-everything.de/Hoshpak/void-packages/src/unmerged-changes/srcpkgs/0ad . I haven't updated the template to Alpha 22 but will do so  within the next 2-3 days.

Issue #2387, opened 2015-08-27
Package request: palemoon


skrzyp:
[Pale Moon](https://www.palemoon.org/) browser, a conservative Firefox fork (yes, now it's a pure fork, not only a rebranding - it has own release cycle and different codebase), with classic UI, lack of "modern" "features" like Pocket integration, and some other improvements.

Duncaen:
Until browser developers get their act together and only work on ONE browser I refuse to waste my time patching their software.
https://github.com/MoonchildProductions/Pale-Moon/issues/366

crocket:
one browser? Do you mean palemoon and basilisk?

Gottox:
Nope, not in the mood to get threats..

skrzyp:
Just for reference, can I get a backstory about that? :)

crocket:
palemoon developers refused to support musl, so void-packages maintainers refused to include palemoon, too.

Hoshpak:
They also refuse to support current gcc versions which makes it unfit to be included in a rolling release distribution.

Gottox:
For reference: https://github.com/jasperla/openbsd-wip/issues/86

Gottox:
Void packages lots of stuff that refuses to support musl ( == proper standards).

In contrast to openbsd-wip, which only distributes build instructions, (this is why the linked issue isn't reasonable) we ship binaries. That means the threats that were done to the linked repository actually apply to Void if we would add palemoon with system libraries.

PR #5103, opened 2016-11-01
New Package: palemoon-27.0.1

Hoshpak:
as requested in #2387


Hoshpak:
I'd appreciate it if someone would advise on how to proceed from here. The package builds fine locally. It seems to me like "g++: internal compiler error: Killed (program cc1plus)" is caused by travis or the travis environment.

srcpkgs/palemoon/template
```diff
@@ -0,0 +1,87 @@
...
+checksum="9715e4ca1240826cbf0168ae85062596d9542a777fc0123c99ad706ad38ce4e8"
+nopie="yes"
+lib32disabled=yes
+only_for_archs="i686 i686-musl x86_64 x86_64-musl"
+
+case "$XBPS_TARGET_MACHINE" in
+	*-musl)
+		# palemoon needs fts.h to compile
+		makedepends+=" musl-fts-devel"
+		broken="needs __uint16_t which is not defined in musl"
```
Vaelatern: Why not patch in a typedef from `__uint16_t` to `uint16_t`? That or add a patch if it's not everywhere.

srcpkgs/palemoon/template
```diff
@@ -0,0 +1,87 @@
...
+
+do_extract() {
+	7za -o${wrksrc} x "${XBPS_SRCDISTDIR}/${pkgname}-${version}/${pkgname}-${version}-source.7z"
+	chmod u+x ${wrksrc}/build/autoconf/mozconfig-find
+	chmod u+x ${wrksrc}/build/autoconf/config.guess
+	chmod u+x ${wrksrc}/build/autoconf/mozconfig2client-mk
+}
+
+do_configure() {
+	cd ${wrksrc}
```
Vaelatern: Please remove this wasted line.

srcpkgs/palemoon/template
```diff
@@ -0,0 +1,87 @@
...
+makedepends="python-devel gtk+-devel dbus-glib-devel alsa-lib-devel pulseaudio-devel gst-plugins-base-devel libXt-devel nspr-devel libevent-devel libjpeg-turbo-devel hunspell-devel libvpx-devel sqlite-devel"
+homepage="http://www.palemoon.org"
+short_desc="Gecko-based web browser, fork of Firefox"
+maintainer="Helmut Pozimski <helmut@pozimski.eu>"
+license="MPL-1.1, GPL-2, LGPL-2.1"
+distfiles="ftp://source:current@ftp2.palemoon.org/${pkgname}-${version}-source.7z"
+checksum="9715e4ca1240826cbf0168ae85062596d9542a777fc0123c99ad706ad38ce4e8"
+nopie="yes"
+lib32disabled=yes
+only_for_archs="i686 i686-musl x86_64 x86_64-musl"
```
Vaelatern: Please no.

srcpkgs/palemoon/template
```diff
@@ -0,0 +1,87 @@
...
+hostmakedepends="p7zip autoconf213 python perl unzip zip pkg-config yasm"
+makedepends="python-devel gtk+-devel dbus-glib-devel alsa-lib-devel pulseaudio-devel gst-plugins-base-devel libXt-devel nspr-devel libevent-devel libjpeg-turbo-devel hunspell-devel libvpx-devel sqlite-devel"
+homepage="http://www.palemoon.org"
+short_desc="Gecko-based web browser, fork of Firefox"
+maintainer="Helmut Pozimski <helmut@pozimski.eu>"
+license="MPL-1.1, GPL-2, LGPL-2.1"
+distfiles="ftp://source:current@ftp2.palemoon.org/${pkgname}-${version}-source.7z"
+checksum="9715e4ca1240826cbf0168ae85062596d9542a777fc0123c99ad706ad38ce4e8"
+nopie="yes"
+lib32disabled=yes
```
Vaelatern: Why?

srcpkgs/palemoon/template
```diff
@@ -0,0 +1,87 @@
...
+revision=1
+hostmakedepends="p7zip autoconf213 python perl unzip zip pkg-config yasm"
+makedepends="python-devel gtk+-devel dbus-glib-devel alsa-lib-devel pulseaudio-devel gst-plugins-base-devel libXt-devel nspr-devel libevent-devel libjpeg-turbo-devel hunspell-devel libvpx-devel sqlite-devel"
+homepage="http://www.palemoon.org"
+short_desc="Gecko-based web browser, fork of Firefox"
+maintainer="Helmut Pozimski <helmut@pozimski.eu>"
+license="MPL-1.1, GPL-2, LGPL-2.1"
+distfiles="ftp://source:current@ftp2.palemoon.org/${pkgname}-${version}-source.7z"
+checksum="9715e4ca1240826cbf0168ae85062596d9542a777fc0123c99ad706ad38ce4e8"
+nopie="yes"
```
Vaelatern: Why? Please try to fix this instead.

srcpkgs/palemoon/template
```diff
@@ -0,0 +1,87 @@
+# Template file for 'palemoon'
+pkgname=palemoon
+version=26.5.0
+revision=1
+hostmakedepends="p7zip autoconf213 python perl unzip zip pkg-config yasm"
```
Vaelatern: `unzip` AND `zip`?

srcpkgs/palemoon/template
```diff
@@ -0,0 +1,87 @@
...
+makedepends="python-devel gtk+-devel dbus-glib-devel alsa-lib-devel pulseaudio-devel gst-plugins-base-devel libXt-devel nspr-devel libevent-devel libjpeg-turbo-devel hunspell-devel libvpx-devel sqlite-devel"
```
Vaelatern:
```
makedepends="python-devel gtk+-devel dbus-glib-devel pulseaudio-devel
 gst-plugins-base-devel libXt-devel nspr-devel libevent-devel
 hunspell-devel libvpx-devel sqlite-devel"
```

srcpkgs/palemoon/template
```diff
@@ -0,0 +1,87 @@
...
+	echo "ac_add_options --with-system-zlib" >> .mozconfig
+	echo "ac_add_options --with-system-libevent" >> .mozconfig
+	echo "ac_add_options --with-system-libvpx" >> .mozconfig
+	echo "ac_add_options --enable-system-pixman" >> .mozconfig
+	echo "ac_add_options --enable-system-hunspell" >> .mozconfig
+	echo "ac_add_options --enable-system-sqlite" >> .mozconfig
+	echo "ac_add_options --enable-system-ffi" >> .mozconfig
+	echo "ac_add_options --with-system-icu" >> .mozconfig
+	echo "ac_add_options --enable-pie" >> .mozconfig
+	echo "ac_add_options --prefix=/usr" >> .mozconfig
```
Vaelatern: Please replace this block of echo with something more readable such as a single echo over multiple lines. Is it not possible to compress the ac_add_options to something else?

Is there no configure script?

srcpkgs/palemoon/template
```diff
@@ -0,0 +1,87 @@
...
+			# XXX see https://sourceware.org/ml/binutils/2014-09/msg00230.html
+			# XXX disable jemalloc.
+			echo "ac_add_options --disable-jemalloc" >> .mozconfig
+			echo "ac_add_options --enable-gold=no" >> .mozconfig
+		;;
+	esac
+}
+
+do_build() {
+	make -f client.mk build
```
Vaelatern: Please remove this function in favor of a build style.

srcpkgs/palemoon/template
```diff
@@ -0,0 +1,87 @@
...
+}
+
+do_install() {
+	make DESTDIR=$DESTDIR -f client.mk install
+
+	# We don't want the development stuff
+	rm -rf ${DESTDIR}/usr/{include,lib/palemoon-devel,share/idl}
```
Vaelatern: Perhaps this stuff belongs in a `-devel` subpackage. Is there any reason people might want this stuff later?

srcpkgs/palemoon/template
```diff
@@ -0,0 +1,87 @@
...
+		;;
+	esac
+}
+
+do_build() {
+	make -f client.mk build
+}
+
+do_install() {
+	make DESTDIR=$DESTDIR -f client.mk install
```
Vaelatern: Please make this a `post_install` and use a build style instead of this line.

srcpkgs/palemoon/files/palemoon.desktop
```diff
@@ -0,0 +1,121 @@
+[Desktop Entry]
```
Vaelatern: Where is this from?
Can we get it from there instead of keeping it in the repo?

Vaelatern:
The Travis errors are seemingly just Travis related. Don't worry about them unless they tell you something otherwise.

srcpkgs/palemoon/template
Hoshpak: That's indeed correct. The build scripts check for the existance of both binaries and fail if one of them is missing. I can only guess that they are usually used to build xpi-files which are zip archives though no xpi files are in the binary package.

srcpkgs/palemoon/template
Hoshpak: Taken from the firefox template. I will remove it if it's unnecessary.

srcpkgs/palemoon/template
Hoshpak: OK, I have absolutely no idea how to get this to cross-compile though.

srcpkgs/palemoon/template
Hoshpak: There's no configure script in the source package. The script is created and executed by the make file though. I will see if I can generate that manually somehow which would also make it easier to fit everything into an build_style.

srcpkgs/palemoon/template
Hoshpak: I can't think of any. This line was also copied from the firefox template. Having an additional -devel package wouldn't hurt of course. I am going to create one.

srcpkgs/palemoon/template
Vaelatern: Does the makefile have a configuration step? Maybe calling that would work.

srcpkgs/palemoon/template
Vaelatern: It's a matter of respecting the $CC variable, in most cases.

srcpkgs/palemoon/template
Hoshpak: I can generate a configure script using autoconf which is identical to the one created and executed by the make file. However executing this configure script only does part of the job and is not sufficient to compile the application afterwards.

So I think there is no way around using the client.mk file which does everything. This seems to be based on the old Mozilla build system. While Mozilla changed it so that firefox can now be compiled with the usual ./configure; make; make install palemoon still contains the old one.

I am currently working into fitting it at least a bit better into build_style=gnu-configure by specifiying a make_cmd and putting the ac_add_options in configure_args. This way I will no longer need to specify do_build and do_install. The do_configure will shrink a bit since I can iterate over configure_args and greatly reduce the number of echo commands. If the build system for palemoon changes in a future version, it could then easily be converted into a "regular" gnu-configure by removing the do_configure function and make_cmd.

pullmoll: You could perhaps use the slightly less emanative `cat`'s way of creating `.mozconfig`
```
        cat > .mozconfig << EOF
ac_add_options --with-pthreads
ac_add_options --disable-strip
ac_add_options --enable-optimize="${CFLAGS}"
ac_add_options --with-system-nspr
ac_add_options --with-system-bz2
ac_add_options --with-system-jpeg
ac_add_options --with-system-zlib
ac_add_options --with-system-libevent
ac_add_options --with-system-libvpx
ac_add_options --enable-system-pixman
ac_add_options --enable-system-hunspell
ac_add_options --enable-system-sqlite
ac_add_options --enable-system-ffi
ac_add_options --with-system-icu
ac_add_options --enable-pie
ac_add_options --prefix=/usr
EOF
```

Hoshpak:
@Vaelatern  I just pushed my last changes. If I haven't missed anything, almost everything you mentioned is now fixed. The package now builds with pie and also on musl-based systems. The only thing I still can't figure out is how to get the package to cross-compile. While changing CC might work for most packages, I hit quite a few roadblocks trying to cross-compile palemoon, e.g:

* It compiles some C code to check for sqlite functionality which would fail with a cross-compiler. To "fix" this, the make file checks if it is cross-compiling and just assumes that sqlite isn't compiled with the needed functionality which makes the whole build fail.
* It compiles a binary called nsinstall which is executed in the build process. When using a cross-compiler, this gives me an arm binary which cannot be executed on the local system.
* nss assumes (for some reason, I can just guess that people didn't think about other architectures when writing these scripts) that all 32 bit architectures should be built with "-m32"

While it would be nice to have every package compiled for every architecture, I don't believe that I would be able to resolve these issues in the foreseeable future. Currently I set "nocross=yes" in the template which from my understanding is at least a bit less restrictive than "only_for_archs". Would that be acceptable?

Vaelatern:
Okay, for now `nocross="please put a reason here"` is acceptable. A simple `nocross=yes` provides no documentation for why the variable is set.

Hoshpak:
changed and pushed.

Hoshpak:
palemoon 27.0.0 has been released. I updated the pull request and refreshed all the patches to make sure it still builds. 27.0.0 uses gstreamer 1.0 now so I could update the build dependency and move away from the to be removed 0.10 version.

Hoshpak:
Package updated to 27.0.1 and rebased on top of master. @Vaelatern Is there anything I still need to do in order for this PR to be ready to be merged?

PR #5106, opened 2016-11-02
New package: font-linux-libertine-ttf-5.3.0

delehef:


m-cz:
It feels wrong to just bump this, but I think the Travis CI build fail (due to connection timeout) is the only reason why this is still waiting for action from maintainers.

PR #5108, opened 2016-11-02
Add package: capnproto

the-eater:


srcpkgs/capnproto/files/LICENSE
```diff
@@ -0,0 +1,22 @@
+Copyright (c) 2013-2014 Sandstorm Development Group, Inc. and contributors
```
Vaelatern: They really provide no copy of the license with their software?

srcpkgs/capnproto/template
```diff
@@ -0,0 +1,51 @@
...
+pkgname=capnproto
+version=0.5.3
+revision=1
+wrksrc="${pkgname}-c++-${version}"
+build_style=gnu-configure
+depends="libcapnproto"
+short_desc="Cap'n Proto serialization/RPC system"
+maintainer="eater <hello@eaterofco.de>"
+license="MIT"
+homepage="https://capnproto.org/"
```
Vaelatern: Drop the end slash here

srcpkgs/capnproto/template
```diff
@@ -0,0 +1,51 @@
...
+distfiles="${homepage}capnproto-c++-${version}.tar.gz"
```
Vaelatern: And add it here

srcpkgs/capnproto/template
```diff
@@ -0,0 +1,51 @@
...
+checksum="cdb17c792493bdcd4a24bcd196eb09f70ee64c83a3eccb0bc6534ff560536afb"
+
+post_install() {
+	 vlicense "${FILESDIR}/LICENSE"
+}
+
+libcapnproto_package() {
+	short_desc+=" - shared libraries"
+	depends="libkj"
+
```
Vaelatern: No blank line here

srcpkgs/capnproto/template
```diff
@@ -0,0 +1,51 @@
...
+	pkg_install() {
+		vmove usr/include/capnp
+		vmove usr/lib/pkgconfig
+	}
+}
+
+libkj_package() {
+	short_desc+=" - utility shared libraries"
+
```
Vaelatern: You get the picture.

srcpkgs/capnproto/files/LICENSE
the-eater: ah it was called `LICENSE.txt` while in the github repo they called it `LICENSE` fixed it

the-eater:
Fixed commit message and set nocross to yes as it tries to run a binary it build just before in the same process :(

atm musl doesn't work, I will take a look tomorrow if I can make it work or not (I have no C/C++ exp.)

Vaelatern:
https://www.gnu.org/software/gnulib/manual/html_node/execinfo_002eh.html

Not present on musl...

the-eater:
@Vaelatern sorry for the late reaction, but yea I saw that, now included a patch which checks explicitly for GLIBC :)

srcpkgs/capnproto/template
```diff
@@ -0,0 +1,48 @@
...
+wrksrc="${pkgname}-c++-${version}"
+build_style=gnu-configure
+depends="libcapnproto"
+short_desc="Cap'n Proto serialization/RPC system"
+maintainer="eater <hello@eaterofco.de>"
+license="MIT"
+homepage="https://capnproto.org"
+distfiles="${homepage}/capnproto-c++-${version}.tar.gz"
+checksum="cdb17c792493bdcd4a24bcd196eb09f70ee64c83a3eccb0bc6534ff560536afb"
+nocross=yes
```
Vaelatern: Why `nocross`?

Vaelatern:
Okay, please rebase and solve the common/shlibs conflict (tip, go a few lines up from the bottom so additions merge smoother). Also please explain the nocross

srcpkgs/capnproto/template
Vaelatern: Please set `nocross="relies on tools built at compilation to complete"` or something to that effect.

PR #5115, opened 2016-11-04
openvpn: update to 2.3.13.

NimGangstaCat:


Vaelatern:
https://swupdate.openvpn.org/community/releases/openvpn-2.3.13.tar.gz: Not Found
?

NimGangstaCat:
It is found for me.

Vaelatern:
Re-triggering job to see if it was a momentary glitch.

Hoshpak:
404 for me from several locations. The sub domain resolves to two IP addresses and both return 404 for me. Since they both point to cloudflare, it could be that the file is on the openvpn servers but the caching on some of the cloudflare proxies is broken or some of them are unable to access the upstream server to retrieve the file. Or maybe they pulled the file again for some reason and some of the proxy servers still have it.

NimGangstaCat:
Or we can use http://build.openvpn.net/downloads/releases/openvpn-2.3.13.tar.gz - what do you think?

PR #5118, opened 2016-11-04
xorg-server: enable security extension

zev0:
This option used to be in the template, but was removed again. I wasn't able to find out why.

afaict it's default on most distros, and I use it a whole lot myself.

### DIFF
--- a/srcpkgs/xorg-server/template
+++ b/srcpkgs/xorg-server/template
@@ -1,7 +1,7 @@
 # Template build file for 'xorg-server'.
 pkgname=xorg-server
 version=1.18.4
-revision=3
+revision=4
 build_style=gnu-configure
 configure_args="--enable-ipv6 --enable-record --enable-xorg
  --enable-xnest --enable-xephyr --enable-composite --enable-xvfb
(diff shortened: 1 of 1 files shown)


Duncaen:
2b108b00d5411d30cc352eb02fb59a30e068afc4

zev0:
Thanks for the link Duncaen.

I'm still not sure what the exact reason was for removing it.

Though I do see that it failed to build on aarch64, according to Travis CI. But I don't know if that was the case before this PR too.

xa0:
@Duncaen Not sure if you're the right person to ask, but any chance of merge? :smiley:

Duncaen:
If its tested and other agree with it sure.

xa0:
Alright. Pinging @pullmoll @ebfe for review.

I've used this PR for months now without any noticeable breakages. And I'm not sure about the travis error, but it looks like a false positive.

Duncaen:
Did you test it on a musl system using the mesa drivers, thats why it was deactivated?

xa0:
I tested in a x86_64 musl VM using mesa drivers, and it still launches fine.

PR #5127, opened 2016-11-05
argyllcms: update to 1.9.2.

lemmi:


lemmi:
So cross compile breaks because in a few places the build system has some autodetect features that only work if headers and libs are in "normal" places. 
I'd like to mark this nocross for now (tested on armv7) until I can dig deeper.

PR #5133, opened 2016-11-06
libcgroup: add upstream patch that fixes infinite loop in cgrulesengd

styrsven:


Vaelatern:
Please add the origin of the patch

styrsven:
Origin of patch added in commit message

PR #5144, opened 2016-11-09
New package: libui-alpha3.1

gangsterakato:
I based my work mainly on the [Arch recipe](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=libui-git). A header file is commented out because it will be available in a future release, this way I just need to comment out when a future version is released.

libui is required for Nim as the possibly future official UI package of the programming language (i.e. libui wrapper at the moment).
See https://github.com/nim-lang/ui/tree/add-readme for more information.

Also, libui might be required by others programs as their UI library in the future, depending of the adoption of this library.

Thanks to @Vaelatern for his precious help. :+1:


srcpkgs/libui/template
```diff
@@ -0,0 +1,37 @@
+# Template file for 'libui'
+pkgname=libui
+version=alpha3.1
+revision=1
+build_style=cmake
+hostmakedepends="pkg-config"
+makedepends="gtk+3-devel"
+depends="libX11 libxcb libffi"
```
Vaelatern: `libxcb` is redundant here.
```
$ xmindep libX11 libxcb libffi
libX11 libffi
```

PR #5151, opened 2016-11-09
New package: create_ap-0.4.3

drkhsh:


xtraeme:
This is just a script... is there a need for a pkg? you can download, extract this into ~ and exec it.

drkhsh:
@xtraeme yeah, i would really like it :smile: i mean, there are many programs written in shell?

Vaelatern:
Very few in the repositories, but normally because shell is so portable it almost belongs in dotfiles.

drkhsh:
@Vaelatern nevertheless, I think it would fit as package

Vaelatern:
To repeat what @xtraeme asked, why should it be a package beyond "it can be"?

drkhsh:
@Vaelatern it is packaged in a lot of distros and it is "cleaner" as package

drkhsh:
So will this be merged or should I close the PR?

Vaelatern:
I don't think this will be merged. It isn't in many official distro repositories. The final call is for @xtraeme to make, but if not in 2 days then it's a no.

drkhsh:
@Vaelatern OK. It is in the official Arch repos.

Vaelatern:
Yes, it is in their Community repo (not the AUR, the more maintained one).

chneukirchen:
It could be useful in hrmpf, so :+1: from me.

drkhsh:
I think that this won't be merged, so i will close it. 😞

Vaelatern:
@chneukirchen You could just do the download in hrmpf's creation, I suppose...

PR #5152, opened 2016-11-10
Fix cinnamon

the-maldridge:
Adds a new package to satisfy a refactor dependency and then adds the python3 versions of certain packages to the runtime dependency list for cinnamon.


Vaelatern:
Malformed commit msg: `Update libgnomekbd to 3.22.0.1`

Vaelatern:
Please fix the xlint error, which is indentation with tabs in `xapps`

srcpkgs/xapps/template
```diff
@@ -0,0 +1,26 @@
...
+pre_configure() {
+        ./autogen.sh
+}
+
+xapps-devel_package() {
+        short_desc+=" - development files"
+        depends="${sourcepkg}>=${version}_${revision}"
+        pkg_install() {
+                vmove usr/include
+                vmove usr/lib/pkgconfig
```
Vaelatern: Also `/usr/lib/*.so` ?

PR #5153, opened 2016-11-10
boswars: fix broken png + missing files

Piraty:
unit files were missing.
few png are reported to be incorrect by libpng (see here: https://bugs.gentoo.org/show_bug.cgi?id=475764).


srcpkgs/boswars/template
```diff
@@ -26,14 +26,29 @@ do_install() {
 	vmkdir usr/share/boswars
 	vcopy fbuild/release/boswars usr/share/boswars/
 	vcopy campaigns usr/share/boswars/
-	vcopy engine  usr/share/boswars/
```
Piraty: that folder only contains .cpp, .h files, as well as a tolua folder. i figure that this is only needed at compile time. the game runs fine without it.

PR #5154, opened 2016-11-10
Add files via upload

traeak:
urbanterror-data: update to 4.3.1


Hoshpak:
shouldn't this package be in nonfree? While the license of the engine might be GPL, the data files are not under any free culture license.

thypon:
fixed upstream

PR #5161, opened 2016-11-12
lemonbuddy: update to 2.2.5

jaagr:


srcpkgs/lemonbuddy/template
```diff
@@ -14,9 +14,9 @@ configure_args="
 	-DENABLE_I3=$(vopt_if i3 ON OFF)
 	-DENABLE_MPD=$(vopt_if mpd ON OFF)
 	-DENABLE_NETWORK=$(vopt_if network ON OFF)"
-depends=("libXft" "xcb-util-wm")
+depends=("libXft" "xcb-util-wm" "xcb-util-image")
```
Duncaen: This should be a normal space separated string, no idea why it was a bash array before.

jaagr: Write a functional linter if it keeps you awake.

PR #5166, opened 2016-11-13
python-google-api-python-client: fix permissions error

pbui:
Some files are not world/all readable when extracted, so chmod them.


Vaelatern:
What bug is there presently if not for this change?

pbui:
@Vaelatern for the record, non-root users could not use the package because all the files in the `/usr/lib/python2.7/site-packages/google_api_python_client-1.5.5-py2.7.egg-info/` directory were not readable by the world/all.  So importing the module would yield a permission denied error.

Issue #5165, opened 2016-11-13
efibootmgr creating unwanted side-effects


Neffistopheles:
efibootmgr package has just updated to automatically register the kernel as a bootable option with the system's EFI boot manager,

#5085 

I feel as though this is an invasive choice. The process of booting the system is a decision which should be left to the system administrator. Virtually nobody explicitly installs efibootmgr. 

It's especially worrisome because the quality of EFI implementations still varies greatly, and this may cause unexpected problems for some users, who may have trouble pinpointing it to this package. For instance, on my local machine I use full disk encryption with an external, removable bootloader to prevent bootkitting. Unfortunately my EFI considers removable devices to be ephemeral and their boot selection order will never persist. So despite this re-assuring message I get after updating efibootmgr:

`The bootorder itself is not changed, so your previous boot loader will
stay enabled until you can edit the order in your firmware interface or
using "efibootmgr -o <hexnum>"`

In fact this _will_ change my boot order. I will now have to manually remove the boot entry on every kernel update to avoid my local EFI from circumventing my preferred boot device, or having to manually choose it on each bootup. 

Moreoever, I don't see a correct way to persistently defeat these hooks, which is stable across package updates, so this is specifically working against the intentions of the local system administrator in a very troubling way. We are given no correct means to opt-out of this silly behavior. 

Please revert this absolutely ridiculous commit. This is the type of stuff that belongs in local user scripts, not at the packaging level.

CMB:
I'm also a little worried about boot entries changing out from under me.
But I think I found a compromise which might make everyone happy.
Check out the efiboot branch on CMB/void-packages.
https://github.com/CMB/void-packages/tree/efiboot

Neffistopheles:
Indeed, I have no quarrels with the change if it's optional and disabled by default. I also like your moving the config to `-kernel-hook.conf`, because otherwise it seems as though this behavior is a feature of efibootmgr proper, and not our specific packaging. I'm still wary of _packagers_ adding features like this; I think it's the job of upstream to add features like this to the project proper. But it's at least lot saner than just taking control over local system policies without consent. 

For now I've just reverted the commit in my local void-packages repo. But I really hate having to keep packages on hold. I hope to see your changes pushed soon.

CMB:
Unfortunately my changes really haven't been tested in the wild, as
that would require a reboot of my only EFI machine.
Seems good in theory though.

CMB:
I think this can be closed.  Fixed with commit
[...]

PR #5168, opened 2016-11-13
efibootmgr: make modifying of boot records optional, disabled by default.

CMB:
Also changed the name of the config file, since the syntax
is much different.

See issue #5165 for the discussion.

### DIFF
--- a/srcpkgs/efibootmgr/INSTALL.msg
+++ b/srcpkgs/efibootmgr/INSTALL.msg
@@ -1,11 +1,12 @@
-From now on, efibootmgr automatically generates efi boot entries to
+efibootmgr can automatically generate EFI boot entries to
 directly boot the kernel after every kernel update or installation.
+Enable this by editing /etc/default/efibootmgr-kernel-hook.
 
 The efi variables need to be accessible in order for this to work, so
 add this line to your fstab and reboot or mount manually:
   efivarfs /sys/firmware/efi/efivars efivarfs 0 0
 -----------------------------------------------------------------------
-Kernel cmd options can be configured in /etc/efibootmgr-options.conf,
+Kernel cmd options can be configured in /etc/default/efibootmgr-kernel-hook,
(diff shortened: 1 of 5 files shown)


srcpkgs/efibootmgr/INSTALL.msg
```diff
@@ -1,11 +1,12 @@
-From now on, efibootmgr automatically generates efi boot entries to
+efibootmgr can automatically generate EFI boot entries to boot
 directly boot the kernel after every kernel update or installation.
```
aeosynth: > to boot directly boot

the added 'boot' is unnecessary

srcpkgs/efibootmgr/template
```diff
@@ -23,7 +23,7 @@ do_install() {
 post_install() {
 	vlicense COPYING
 	vmkdir etc/
-	touch ${DESTDIR}/etc/efibootmgr-options.conf
+	vinstall "${FILESDIR}/efibootmgr-kernel-hook.conf" 644 etc/
```
Vaelatern: Please use `vconf`

srcpkgs/efibootmgr/files/efibootmgr-kernel-hook.conf
```diff
@@ -0,0 +1,6 @@
+# Options for the kernel hook script installed by the efibootmgr package.
+modify_efi_entries=0
```
Vaelatern: Please change these options to be all-caps.

CMB:
Fixed to use vconf.
Options in the config file are all-caps.
Also added braces around variable names in a couple places.
Typo in the install message is fixed.

Thanks for the reviews.

Vaelatern:
@Duncaen Should the conf file be moved? To `/etc/default/` seems reasonable, but I am not sure.
the problem is once this gets merged, moving a file like this will be more problematic.

PR #5171, opened 2016-11-14
texlive-bin: tlarch x86_64 to PATH in texlive.sh for x86_64-musl

cavalier911:
XBPS_ARCH=x86_64-musl

x86_64) tlarch=x86_64;; 
No match, No ${tlarch} is added
tar xf  texlive-bin-2016_1.x86_64-musl.xbps 
cat etc/profile.d/texlive.sh
````
#location of the TeXLive binaries
export PATH=$PATH:/opt/texlive/2016/bin/-linux
````
x86_64*)tlarch=x86_64;;
Has match
tar xf  texlive-bin-2016_2.x86_64-musl.xbps 
cat etc/profile.d/texlive.sh

````
#location of the TeXLive binaries
export PATH=$PATH:/opt/texlive/2016/bin/x86_64-linux
````
Reference: https://forum.voidlinux.eu/t/texlive-bin-auctex-ess-on-x86-64/1331


chneukirchen:
That won't make the binaries magically work on musl.

cavalier911:
If texlive binaries don't work on musl then why not remove texlive-bin and texlive20{14,15,16}-bin musl versions from void-packages?

PR #5176, opened 2016-11-15
New Package: ksnapshot-4.14.3

Hoshpak:


srcpkgs/ksnapshot/template
```diff
@@ -0,0 +1,13 @@
+# Template file for 'ksnapshot'
+pkgname=ksnapshot
+version=4.14.3
+revision=1
+build_style=cmake
+hostmakedepends="qt-qmake automoc4"
+makedepends="kdelibs-devel qt-devel perl phonon-devel libXfixes-devel libkipi-devel libX11-devel"
+homepage="://www.kde.org/applications/graphics/ksnapshot"
```
Vaelatern: `="://www.kde.org` Missing something, I think ;)

Hoshpak: Thanks, I can't believe that slipped through my own review. It's fixed now.

PR #5178, opened 2016-11-15
Manual: build option commands were incomplete

Piraty:


Vaelatern:
But it also works for other commands.

Piraty:
so maybe add sth. like: (pkg|extract|...) ?
If you don't agree, please close it.

ebfe:
Maybe ```$ ./xbps-src -o option,option1 <cmd>``` avoids the confusion

Piraty:
in fact, i call `./xbps-src <cmd> <some more flags> -o  <pkgname>`.
but in general i agree.

Vaelatern:
@Piraty Until recently, your syntax would not have worked.
I'm happy with this PR as `<cmd>` instead of `pkg`

Piraty:
Well, it's been corrected.

asergi:
Please squash your commits together next time.

PR #5180, opened 2016-11-16
New packages: Bear-2.2.1, sound-icons-0.1, dex-0.7

ackalker:
These are my first few contributions to Void Linux, I hope that these packages are welcome. Any comments or suggestions are very welcome.

### DIFF
--- a/srcpkgs/dex/template
+++ b/srcpkgs/dex/template
@@ -1,0 +1,19 @@
+# Template file for 'dex'
+pkgname=dex
+version=0.7
+revision=1
+noarch=yes
+build_style=gnu-makefile
+make_install_args="MANPREFIX=/usr/share/man"
+python_version=3
+depends="python3"
+short_desc="Generate and execute DesktopEntry files of the Application type"
+maintainer="Alain Kalker <a.c.kalker@gmail.com>"
+license="GPL-2"
+homepage="https://github.com/jceb/dex"
+distfiles="https://github.com/jceb/dex/archive/v${version}.tar.gz>${pkgname}-${version}.tar.gz"
+checksum=2b0c98832c4b894486807a9ac0417ca127425013e0f07da352621065cb36adc6
+
+post_install() {
+	rm ${DESTDIR}/usr/share/doc/${pkgname}/LICENSE
+}
--- a/srcpkgs/Bear/template
+++ b/srcpkgs/Bear/template
@@ -1,0 +1,12 @@
+# Template file for 'Bear'
+pkgname=Bear
+version=2.2.1
+revision=1
+build_style=cmake
+depends="python"
+short_desc="A tool that generates a compilation database for clang tooling"
+maintainer="Alain Kalker <a.c.kalker@gmail.com>"
+license="GPL-3"
+homepage="https://github.com/rizsotto/${pkgname}"
+distfiles="https://github.com/rizsotto/${pkgname}/archive/${version}.tar.gz>${pkgname}-${version}.tar.gz"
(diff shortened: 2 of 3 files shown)


ackalker:
Regarding `sound-icons` package: this enhances `speech-dispatcher` by adding 'sound icons', a kind of [earcon](https://en.wikipedia.org/wiki/Earcon)s which are specifically for use with text-to-speech software. `speech-dispatcher` can use these sound icons 'out-of-the-box' as packaged, no change in configuration files necessary.
Regarding `dex`: although it doesn't need compilation as required by `Manual.md`, I think it is useful for Void Linux users who use minimalist window managers or desktop environments, because it allows one to easily run 'autostart' desktop applications from `*.desktop` files in `/etc/xdg/autostart` and `${HOME}/.config/autostart` directories:
```sh
$ dex -a # Run everything autostartable
$ dex -ae i3 # Do not run any DE-specific file unless it specifies `OnlyShowIn=i3`
# dex -ae GNOME # Fake running GNOME desktop
```

ebfe:
Could you rebase this onto master? That should fix the travis builds/results.

ackalker:
@ebfe I've rebased to master and pushed again, thanks.

ebfe:
Installed /usr/bin/dex has a /usr/bin/python2 shebang but package depends on python3? 

Otherwise looks ok, though i'd vote for `s/Bear/bear/`.

ackalker:
Thank you for your comments.
Regarding Bear's Python dependency, the documentation states that Bear supports Python versions >= 2.7, so I made the package depend on `python`, which in Void Linux is now at version 2.7.12_4. Please let me know if I should make the package depend on `python3` instead, in which case I will also change the shebang.
Regarding the package name `Bear`, I was in doubt about whether to capitalize the name as does the documentation does, or to simply name it `bear` after the name of the script. Since the release tarballs don't have the package name, there's no obvious choice. I'll gladly rename the package to `bear` if that is more to your liking.

ebfe:
> Regarding Bear's Python dependency, the documentation states that Bear
> supports Python versions >= 2.7, ...
The python dependency problem is with `dex` not bear. The package
depends on python3 but there is a hook that rewrites unversioned python
shebangs:

```
=> dex-0.7_1: running pre-pkg hook: 03-rewrite-python-shebang ...
   Shebang converted to '#!/usr/bin/python2': /usr/bin/dex
```

> I'll gladly rename the package to `bear` if that is more to your liking.

Your choice really, i just think uppercase in package names are a bit annoying.

ackalker:
Ah, my fault, I'm so sorry. I'll have a look at the `dex` package tomorrow. BTW, my `/usr/bin/dex` has `#!/usr/bin/env python3` as shebang, as it should (it is Python 3 code).

ebfe:
29c37543cf75436b4c602fd1a0f16a11b3c05539 might have changed the default.

srcpkgs/dex/template
```diff
@@ -0,0 +1,18 @@
+# Template file for 'dex'
+pkgname=dex
+version=0.7
+revision=1
+noarch=yes
+build_style=gnu-makefile
+make_install_args="MANPREFIX=/usr/share/man"
```
asergi: `python_version=3`

ackalker:
@ebfe , @asergi Thanks for your tip and review. I've fixed `dex` package with the change.
I've decided to stick with the name `Bear` for that package, it is the name of the project on Github and is used consistently in the documentation, and several other packages in Void also follow this lead. Also, `xbps-query -Rs bear` has no problem finding the package regardless of whether it's name is capitalized.
Thanks for your time, I hope to be able to provide Void Linux with other useful packages in the future.

ackalker:
Thanks to all :smile:

Issue #4798, opened 2016-09-18
Removing gstreamer 0.10


chneukirchen:
I'd like to remove gstreamer 0.10, its ancient and buggy code.

Outside dependencies:
- [x] gnac-0.2.4.1_1, upstream dead, even removed from Debian. (Arch has a patch for 1.0.)
- [x] gst123-0.3.4_1, apparently supports 1.0
- [x] libnice-0.1.13_1: apparently supports 1.0
- [x] parole-0.8.1_1: apparently supports 1.0
- [ ] perl-GStreamer & perl-GStreamer-Interfaces: only used by gmusicbrowser, which support 1.0 via Glib::Object::Introspection it seems
- [ ] pocketsphinx-0.8_4: <s>apparently supports 1.0</s> only CMU Sphinx has gstreamer.
- [x] spectrum3d-2.5.0_1: apparently supports 1.0
- [x] wxGTK-2.8.12.1_7: I wanted to kill this all the time, moving diff-pdf and erlang? (last deps) to wx 3.0
- [x] wxWidgets-3.0.2_3: perhaps use --enable-mediactrl=no ? Do we have an app that uses this? (only wxPython uses wxMedia...)
- [ ] xfce4-mixer-4.11.0_2: upstream dead

chneukirchen:
In case anyone needs more arguments: http://scarybeastsecurity.blogspot.de/2016/11/0day-exploit-compromising-linux-desktop.html

ebfe:
835b40ca9258b1f7e465b104cde3b16b4171a1fa

PR #5181, opened 2016-11-16
gst-plugins-bad: disable building of the nsf plugin

Hoshpak:
disable building of the nsf plugin to prevent the vulnerability
currently known as CESA-2016-0001. Full description available at
http://scarybeastsecurity.blogspot.de/2016/11/0day-exploit-compromising-linux-desktop.html


Vaelatern:
#4798

Hoshpak:
I'm well aware of that. I just thought it would be better to fix that while the package is still around.

Vaelatern:
@chneukirchen  ?

chneukirchen:
We are still breaking a few apps if we remove gstreamer 0.1

PR #5182, opened 2016-11-17
chromium: update to 54.0.2840.100.

Duncaen:
`x86_64` and `x86_64-musl` working (compiled and tested).
New patch `gn_bootstrap.patch` to make the `GN` bootstrap work with musl libc and to use `libevent` from the system to avoid another issue with libatomic and musl while bootstrapping.
`last-commit-position.patch` and some bits from linux from scratch.

I had some problems with `NaCL` and deactivated it for now.


srcpkgs/chromium/files/musl-patches/musl-hacks.patch
```diff
@@ -1,14 +1,3 @@
---- ./third_party/usrsctp/usrsctplib/usrsctplib/netinet/sctp_userspace.c.orig
-+++ ./third_party/usrsctp/usrsctplib/usrsctplib/netinet/sctp_userspace.c
-@@ -72,7 +72,7 @@
- 	pthread_setname_np(name);
- #endif
- #if defined(__Userspace_os_Linux)
--	pthread_setname_np(pthread_self(), name);
-+	//pthread_setname_np(pthread_self(), name);
```
ebfe: I think this is available on musl with -D_GNU_SOURCE.

Duncaen: This part of the patch was removed, it uses prctl SET_NAME now.

ebfe: Euh, too early...

chneukirchen:
Just cosmetics, but I think `conf+='foo="bar"'` is nicer than quoting `\"`.

chneukirchen:
Drop `# XXX: trigger rebuild`.

Also, I feel like the gn configuring and pre-ninja part should be in `do_configure`.

Gottox:
ready to merge?

Duncaen:
Merged in 86a2c949d8ab5d39f080d98be21a50091a7f1186 and fixed musl gn bootstrap in 59294bb56176524a99914b47e11afbb2efeeda92.

PR #5192, opened 2016-11-19
New package: polybar-3.0.5

jaagr:


srcpkgs/polybar/template
```diff
@@ -0,0 +1,55 @@
+# Template file for 'polybar'
+pkgname=polybar
+version=2.3.5
+revision=1
+wrksrc="${pkgname}-${version}"
```
Vaelatern: Redundant

mixedCase:
@Vaelatern Any news on this?

Vintodrimmer:
How does the process work?
I see a lot of pulls that have been reviewed and passed the tests but are not in the "repo".
What is the final wait factor?

Duncaen:
@Vintodrimmer That really depends on the package, if its really reviewed and everything fine it can be merged, but someone has to do it.
But this one is special, renaming packages is not something that should happen. 
People who have `lemonbuddy` installed would have to manually switch to the renamed package or they are stuck for ever with some old version.
Adding a meta sub package with the old name that depends on the new package would avoid this.

Vaelatern:
Many times people 'cheat' with things like nopie to get the build to pass.
Or they get the build to pass but got something else wrong.
This is why it must be reviewed.

But in this case I think the template is right but (I've tested it and) the upgrade process doesn't work at all.

Vintodrimmer:
@Duncaen That's reasonable. But isn't there some way to provide the end user with the information that package is deprecated? Maybe a new version so during the system update xbps will tell that this package has been replaced?

The problem is that even if people do upgrade from 'lemonbuddy', their config will almost certainly not work, due to the number of changes in syntaxis and behaviour of the panel. It is close to a new product, in my opinion.

@Vaelatern Can you provide more information on the error, please? I think it will speed up the process and help a lot.

Vaelatern:
Perhaps the solution is to provide a subpackage for lemonbuddy that just has a "lemonbuddy has been depreciated in favor of <pkgname>" or no message just do the switch.

Error: I build and install lemonbuddy. I then -Su. This does not show up. I then install this, old one is not removed.
The upgrade path sucks.

Vintodrimmer:
@Vaelatern So, if I ask author to treat it as a completely different package, instead of update. this will be satisfactory and the package might appear in the repo faster?

Because I personally really miss this package and would be glad to see it in the repos, even if it is not update of lemonbuddy, which, as I have mentioned, will most probably not work with the old config files.

asergi:
@jaagr https://gist.github.com/asergi/a0ebc9aaf18ac6f9a7cedd18e2322198

srcpkgs/polybar/template
```diff
@@ -0,0 +1,57 @@
+# Template file for 'polybar'
+pkgname=polybar
+version=3.0.3
+revision=2
```
asergi: `revision=1`

srcpkgs/polybar/template
```diff
@@ -0,0 +1,57 @@
...
+_i3ipcpp_version=0.6.2
+_xpp_version=1.3.6
+build_style=cmake
+configure_args="
+ -DENABLE_ALSA=$(vopt_if alsa ON OFF)
+ -DENABLE_CURL=$(vopt_if curl ON OFF)
+ -DENABLE_I3=$(vopt_if i3 ON OFF)
+ -DENABLE_MPD=$(vopt_if mpd ON OFF)
+ -DENABLE_NETWORK=$(vopt_if network ON OFF)"
+depends="cairo xcb-util-wm xcb-util-image xcb-util-xrm"
```
asergi: `depends` line is redundant. See https://travis-ci.org/voidlinux/void-packages/jobs/195969639#L917-L935.

srcpkgs/polybar/template
```diff
@@ -0,0 +1,57 @@
...
+distfiles="
+ https://github.com/jaagr/polybar/archive/${version}.tar.gz
+ https://github.com/jaagr/xpp/archive/${_xpp_version}.tar.gz
+ $(vopt_if i3 https://github.com/jaagr/i3ipcpp/archive/v${_i3ipcpp_version}.tar.gz)"
+checksum="
+ c6a19345b728c2f270b816f87b06201bb86ff8a1aa52140a1340409b8bf9f2b1
+ b81993893190bf7ba19529e0da6e8b069e13e833f6a62874830b46785bee3349
+ $(vopt_if i3 7ad9c0c1e1cb9dfd07ecc2a7769eae9ac962c22db2e07be08b2186b7fd85189d)"
+
+build_options="alsa curl i3 mpd network"
```
asergi: The above build options are missing the description. Example: `desc_option_i3="Enable support for i3"` and so on.

Epictek:
Is there any progress on getting this merged?

Vintodrimmer:
I don't think it will be merged, to be honest.

They want the author to make lemonbuddy upgradeable to polybar, which is not going to happen, due to a variety of reasons.

There is a more up to date template on his github and it builds perfectly (except on musl). You can also try to add it to VUR.

Duncaen:
Just fix the issues and it gets merged.
There is a lot of confusion in this PR because of the naming, a renamed package should be upgradable.
If its a "new" package call it as this and fix all the issues and get all architectures working.
If you drop the old template we have no chance to increase the revision with a message that the package is deprecated.

srcpkgs/polybar/template
Vintodrimmer: Can you explain the output, please?

I'd like to try and properly format the template

srcpkgs/polybar/INSTALL.msg
```diff
@@ -0,0 +1,9 @@
+
+  Get started with the example configuration:
+
+    $ install -Dm644 /usr/share/doc/polybar/config $HOME/.config/polybar
+    $ polybar example
+
```
Vaelatern: Way too many newlines in here. Please cut out empty lines.

Vintodrimmer:
It's not always pleasant to be proven wrong, but today is one of those exceptions.

Thanks for pulling it.

Now I can finally stop building it with every update.

PR #5193, opened 2016-11-19
xbps-src: new vopt_onoff helper

jaagr:
CMake helper function used to set bool argument value for cache entries.

Usage example:

    configure_args+=" $(vopt_onoff logging WITH_LOGGING)"

    configure_args+=" -DWITH_LOGGING=OFF"


Anachron:
I like it, can we search and replace pkgs automatically so that they use this?

jaagr:
We can at least replace the ones that are using the same format: `-D<prop>:BOOL={ON,OFF}`

xtraeme:
You can skip completely the ":TYPE" part... simply "-DPROP=VALUE".

2016-11-19 14:59 GMT+01:00 Michael Carlberg <notifications@github.com>:

Vaelatern:
@xtraeme this looks good to merge?

maxice8:
@Vaelatern i don't think you're getting an answer

ping

Vaelatern:
lol

PR #5202, opened 2016-11-20
New package: speedtest-cli-1.0.0

drkhsh:


srcpkgs/speedtest-cli/template
```diff
@@ -0,0 +1,13 @@
+# Template file for 'speedtest-cli'
+pkgname=speedtest-cli
+version=1.0.0
+revision=1
+build_style=python3-module
+makedepends="python3-setuptools"
```
Vaelatern: I do think you mean `hostmakedepends` here

srcpkgs/speedtest-cli/template
```diff
@@ -0,0 +1,13 @@
...
+depends="python3-setuptools"
```
Vaelatern: You sure you depend on that after build time?

srcpkgs/speedtest-cli/template
drkhsh: sure

srcpkgs/speedtest-cli/template
drkhsh: i dont think so, i will test

drkhsh:
@Vaelatern please check now

PR #5204, opened 2016-11-20
xbps: use dash to exec scripts

ebfe:
bash requires readline, INSTALL/REMOVE scripts are executed via /bin/sh. When
updating bash and readline at the same time, bash is unpacked before readline.
If /bin/sh is linked to bash and a package with a INSTALL/REMOVE script is
updated before unpacking readline, xbps aborts (as it can't exec /bin/sh due to
missing shlib) which results in a non-functional /bin/sh. Explicitly use
/bin/dash to work around this for now.


ebfe:
Doesn't work, patch is against xbps master :/

ebfe:
Also backported 430a25f422bd3abfe9901c45ba069e633e1803a9 to make this work.

ebfe:
This is still incomplete. Triggers or similar scripts that are executed from INSTALL/REMOVE and use /bin/sh still fail.

ebfe:
Ping @xtraeme.

xtraeme:
This is completely wrong... xbps must not have any knowledge
about what shell is /bin/sh really... and hardcoding it to /bin/dash
is even worse imho.

-1

ebfe:
Any suggestion on how to ensure /bin/sh -> /bin/bash keeps working _during_ the transaction? Even if we had separate packages for readline6 and readline7 there seems no way to ensure readline7 is unpacked before bash if they are update in the same transaction. Which results in xbps aborting and leaving a broken /bin/sh.

xtraeme:
I think I can change the sorting order in the transaction dictionary
returned by xbps_transaction_prepare().

xtraeme:
proplib arrays are unsorted, so that after processing all pkgs they should be already sorted properly.

I think right now this returns a dictionary... which is sorted alphanumerically.

xtraeme:
A new test case must be written to verify this works as expected...

ebfe:
I'm not sure we can determine the right ordering if we have no idea what the
current /bin/sh actually is.  Consider 2 packages sh1, sh2 (both sh
alternatives) that depend on package libX: libX has a soname bump what is the
correct order to update them?

For the concrete case the only sh that depends on readline is bash. A
workaround to the immediate problem would be to statically link
readline to bash (assuming bash is always updated before readline with the
current ordering inside a transaction).

pullmoll:
I think a proper solution would be to make xbps depend on its very own shell, be it dash or bash.

It could be statically linked, in case bash is required for any scripts, and installed as `/usr/bin/xbps-sh` or similar together with the xbps binaries. Then xbps internal `execv()`s should explicitly use this shell. Think of a system without either dash or bash, but just ksh or zsh or something else. Would xbps work at all in that case?

ebfe:
> I think a proper solution would be to make xbps depend on its very own shell, be it dash or bash.
> 
> It could be statically linked, in case bash is required for any scripts, and installed as `/usr/bin/xbps-sh` or similar together with the xbps binaries. Then xbps internal `execv()`s should explicitly use this shell. Think of a system without either dash or bash, but just ksh or zsh or something else. Would xbps work at all in that case?

That would only be a partial solution. If we has that, and all
INSTALL/REMOVE scripts had a `#!/usr/bin/xbps-sh`, xbps would not fail
to exec the INSTALL and not abort. All packages would be unpacked and
after it is done, /bin/sh should be working again. So it's an
improvement, but still /bin/sh is broken during the transaction and all
stuff called from INSTALL that depends on /bin/sh would fail.

pullmoll:
Hm... since `/usr/bin/sh` is a symlink and `/bin -> /usr/bin`, perhaps xbps-install, xbps-remove etc. could redirect `/usr/bin/sh -> xbps-sh` while they are running. In an `atexit()` function the previous symlink would be restored.

I know, an awkward thought and probably too much of a hack for being considered useful.

ebfe:
This doesn't only affect '/bin/sh' but pretty much everything that's called from INSTALL scripts during unpacking. The only way i can up with to solve this in general is to not execute the scripts during unpacking.

PR #5207, opened 2016-11-21
breeze-icons: update to 5.28.0.

RunningDroid:
I had to update extra-cmake-modules so I could build breeze-icons, is it OK to have one PR to update both of them or should I separate them?


Vaelatern:
Huh, only failure is in `extra-cmake-modules`. Can you fix that one, by any chance, while you are here?

Vaelatern:
Okay, @dutsky intends to update this with the rest of that system, but thank you for your work.

ebfe:
28ba2228f024c77293b9173c78b160de18b43cf3

PR #5209, opened 2016-11-21
New package: xwinmosaic-0.4.2

stobenski:
X11 window switcher that draws windows list as a colourful mosaic.
https://github.com/soulthreads/xwinmosaic


stobenski:
Sorry, my first void package, I dont understand what I need to do next.

srcpkgs/xwinmosaic/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'xwinmosaic'
+pkgname=xwinmosaic
+version=0.4.2
+revision=1
+build_style=cmake
+hostmakedepends="pkg-config"
+configure_args="-DCMAKE_INSTALL_PREFIX=/usr"
```
Vaelatern: This line is redundant and is provided by the build style.

Vaelatern:
Cross builds fail because it is looking for include files in the wrong place, it seems.

PR #5211, opened 2016-11-21
bash: use static readline

ebfe:
This is a workaround for the problems described in #5204 assuming the bash update is always ordered before the readline update.

### DIFF
--- a/srcpkgs/bash/template
+++ b/srcpkgs/bash/template
@@ -5,12 +5,13 @@ _bash_patchlevel=048
 # Dracut breaks with bash 4.4. See https://github.com/dracutdevs/dracut/issues/118
 reverts=4.4.0_1
 version=${_bash_distver}.${_bash_patchlevel}
-revision=2
+revision=3
 wrksrc=${pkgname}-${_bash_distver}
 build_style=gnu-configure
-configure_args="--without-bash-malloc --with-curses --with-installed-readline"
+configure_args="--without-bash-malloc --with-curses --without-installed-readline"
+make_build_args="TERMCAP_LIB=${XBPS_CROSS_BASE}/usr/lib/libncursesw.a"
 hostmakedepends="bison"
-makedepends="ncurses-devel readline-devel"
+makedepends="ncurses-devel"
 conflicts="chroot-bash>=0"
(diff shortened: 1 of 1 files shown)


chneukirchen:
makedepends can be cleaned up too then.

I'd also propose to link ncurses statically, but it may be overcautious.

:+1: from me, let's have bash have the same solib-depends as the other /bin/sh we support (dash, mksh, busybox), it will save a lot of hassle.

ebfe:
updated makedepends. looks like static curses needs more patching.

chneukirchen:
Very easy it turns out:

```
make_build_args="TERMCAP_LIB=${XBPS_CROSS_BASE}/usr/lib/libncursesw.a"
```

ebfe:
Ah, thanks!

```
$ ldd /bin/bash
        linux-vdso.so.1 (0x00007ffe6c3cf000)
        libdl.so.2 => /usr/lib/libdl.so.2 (0x00007ff2b6bdf000)
        libc.so.6 => /usr/lib/libc.so.6 (0x00007ff2b6840000)
        /lib/ld-linux-x86-64.so.2 (0x00007ff2b6de3000)
$ ldd /bin/dash
        linux-vdso.so.1 (0x00007ffc1552f000)
        libc.so.6 => /usr/lib/libc.so.6 (0x00007f78e15c9000)
        /lib/ld-linux-x86-64.so.2 (0x00007f78e1968000)
$ ldd /bin/mksh
        linux-vdso.so.1 (0x00007ffc5c1df000)
        libc.so.6 => /usr/lib/libc.so.6 (0x00007fd58148b000)
        /lib/ld-linux-x86-64.so.2 (0x00007fd58182a000)
$ ldd /bin/busybox
        linux-vdso.so.1 (0x00007ffc95d0e000)
        libm.so.6 => /usr/lib/libm.so.6 (0x00007f140c334000)
        libc.so.6 => /usr/lib/libc.so.6 (0x00007f140bf95000)
        /lib/ld-linux-x86-64.so.2 (0x00007f140c63a000)
```

chneukirchen:
FWIW, this bloats bash from ~800k to ~1200k. I can live with that.

PR #5212, opened 2016-11-21
linux4.4: update to 4.4.34

pbui:


chneukirchen:
Was already queued, thanks.

Issue #4746, opened 2016-09-11
proposal: use shfmt for all shell files


mvdan:
I've recently written shfmt, the equivalent of gofmt for bash/shell programs. void-packages is a pretty active project with lots of contributors, so I think it would benefit from its use.

More info: https://github.com/mvdan/sh

I can tag a release and/or upload binaries to github if that would make it easier to integrate in Travis.

I ran `shfmt -l -w .` and these are the diffs. This does all the files that either have a bash/shell extension or have no extension and a valid bash/shell shebang.

By default it indents with tabs, but it can also use any number of spaces. For now, nothing else is configurable, and I hope to keep it that way. I used tabs in this example since that's what you guys seem to prefer.

`git diff` - https://clbin.com/kPVkt
`git diff -w` - https://clbin.com/M2Rhl

If you have any questions or issues, I'm happy to help.

the-eater:
This is currently ignoring all template files correct? these also have bash in them, but don't have a hashbang

pullmoll:
Personally I don't like the pipe symbol `|` and logic operators `&&`, `||` being moved to the next line. Almost all shell scripts right now use `| \', '&& \', and '|| \'  for continuations, i.e. have the next command as the first word on a new line.

But that's just my .02€

Duncaen:
I think the spaces around `|` in case statements are confusing.

mvdan:
> This is currently ignoring all template files correct? these also have bash in them, but don't have a hashbang

Ah, correct. The command would probably be `shfmt -l -w . srcpkgs/*/template` then. The diffs I posted should still work OK as an example of the changes.

> Personally I don't like the pipe symbol | and logic operators &&, || being moved to the next line.

I initially thought the same. Even more, the `\` can be ignored as these tokens accept a newline after:

```sh
foo &&
    bar
```

I went with this format because it's simpler to see what commands are chained together and how. If the `&&` is at the beginning of the line, one can quickly see it which might not be the case if the lines are long or complex.

This is also mentioned in the Google shell style: https://google.github.io/styleguide/shell.xml#Pipelines

> I think the spaces around `|` in case statements are confusing.

I'm not sure about this one. I've spaced `;;` in single-line case statements for the same reason that `&&` and others are spaced. In this case, sometimes switch cases can become complex:

```sh
case $a in
    b|c) foo ;;
    *"some long string"*|"some other long string") bar ;;
esac
```

[...]

PR #5230, opened 2016-11-25
Run shfmt on all scripts except srcpkgs/

mvdan:
As proposed in https://github.com/voidlinux/void-packages/issues/4746#issuecomment-257049926.

This diff is huge because a lot of files were inconsistently being indented with spaces, so use `git diff -w` for your own sanity. Here: https://clbin.com/e2liY

I added it to the xlint script to avoid introducing new scripts.

Also, since Go isn't available in the image (I assume) this does a download of a binary. Other options could be taken.

CC @ebfe @Vaelatern @dominikh


mvdan:
Whoops, the travis scripts aren't bash. But most of the files on the repo are. What's the policy here?

Vaelatern:
- Wow that is a lot of changes.
 - From what I saw, they seemed to not change anything at all functionality-wise, but I couldn't be sure (see above)

mvdan:
> Wow that is a lot of changes.

`git diff -w` helps, as well as the fact that srcpkgs is excluded here.

> From what I saw, they seemed to not change anything at all functionality-wise, but I couldn't be sure (see above)

In theory this is just a formatter, so it should not change the program at all. I have a large number of unit tests for all of it and a few other projects are using it, so in practice that should be true too.

> The style seems okay.

To clarify, the style hasn't changed much since the time I opened the issue. Notably, `` `foo` `` is now replaced with `$(foo)`. But otherwise it should be the same.

chneukirchen:
Can you make it indent
```
FOO=BAR \
        foo --bar
```
as-is?

chneukirchen:
There still are spaces around `|` in case, which we don't want.

chneukirchen:
Also I don't think
```
foo \
        || bar
```
is a good idea.

mvdan:
> Can you make it indent
> 
> FOO=BAR \
>         foo --bar
> as-is?

That's a bug, will fix it. Thanks for noticing!

> There still are spaces around `|` in case, which we don't want.

I don't want the tool to have a dozen knobs, so I tried to make the style as consistent as possible. See the discussion in the original issue. Words are always spaced at either side, so I don't see why this one should be an exception.

> Also I don't think
> 
>    foo \
>            || bar
>
> is a good idea.

Also mentioned in the original issue - I'm not too fussed either way, but this form seems to be preferred by some popular styles like Google's. I haven't heard any point against this style other than personal preference, so I haven't changed it.

As before, the less knobs, the better. In any case, I think thinking too much about what style to enforce is bikeshedding - and I've certainly put a lot of thought into it so far :)

mvdan:
@chneukirchen actually, I can't reproduce your first issue:

```
 $ cat f.sh
FOO=BAR \
        foo --bar
 $ shfmt f.sh
FOO=BAR \
        foo --bar
```

chneukirchen:
I don't see anything on wrapping in the Google style guide line, and I think having to add `\` is not intuitive.

chneukirchen:
Ah, what I meant actually happened in `common/build-style/python-module.sh`, and it's

```
env FOO=BAR \
       FOO=BAR \
                foo --bar
```

Guess that's a semantical difference then...

mvdan:
The page isn't rendering for me on chrome, so linking to the source xml directly:

https://github.com/google/styleguide/blob/7ebdcc4f949907b08694dc6010b518bc303bb373/shell.xml#L333

As for:

```
env FOO=BAR \
       FOO=BAR \
                foo --bar
```

The second level of indentation is nonexistent here, so it's on purpose.

mvdan:
I'll be closing this, since there doesn't seem to be enough interest.

PR #5233, opened 2016-11-26
New gcc 6.2.1

Gottox:
ToDo:

* [ ] improve commit messages


pullmoll:
A new official gcc-6.2.1.tar.xz should be created and the PRs updated according to https://github.com/voidlinux/void-packages/issues/4825#issuecomment-263078295

Gottox:
Currently building the tar

srcpkgs/kernel-libc-headers/template
```diff
@@ -11,7 +11,7 @@ short_desc="The Linux API headers for userland development"
 license="GPL-2"
 homepage="http://www.kernel.org"
 distfiles="http://www.kernel.org/pub/linux/kernel/v4.x/linux-${version}.tar.xz"
```
pullmoll: $KERNEL_SITE

srcpkgs/gcc/template
```diff
@@ -74,19 +74,20 @@ case "$XBPS_TARGET_MACHINE" in
...
-		*-musl) patch -Np1 -i ${FILESDIR}/gcc-4.9.2-musl.diff;;
+		*-musl) patch -Np1 -i ${FILESDIR}/gcc-${version}-musl.diff;;
 	esac
 	# _FORTIFY_SOURCE needs an optimization level.
 	sed -i "/ac_cpp=/s/\$CPPFLAGS/\$CPPFLAGS -O2/" {gcc,libiberty}/configure
-	# As specified in the LFS book, disable installing libiberty.
-	sed -i 's/install_to_$(INSTALL_DEST) //' libiberty/Makefile.in
-	# Do not run fixincludes
-	sed -i 's@\./fixinc\.sh@-c true@' gcc/Makefile.in
+	mkdir gcc_build
```
pullmoll: This line can be removed. The build directory is `build` and created in `do_configure()`.

Gottox:
@pullmoll recheck

pullmoll:
LGTM. I'll restart bootstrapping locally now for a try.

pullmoll:
https://repo.voidlinux.eu/distfiles/gcc-6.2.1.tar.xz results in a 403. Access rights?

Gottox:
403 is fixed.

srcpkgs/chromium/template
```diff
@@ -30,7 +30,8 @@ depends="libexif hwids desktop-file-utils hicolor-icon-theme"
 nostrip_files="nacl_irt_x86_64.nexe nacl_irt_x86_32.nexe"
 
 if [ -z "$CROSS_BUILD" -a "$XBPS_MACHINE" = "x86_64" ]; then
-	hostmakedepends+=" gcc-c++-multilib"
+	# Is this required?
```
Gottox: No idea, let's keep it better safe then sorry :)

pullmoll:
When building `-N pkg base-chroot` for `i686` here, a failure to build `perl` happened again. I've seen that before and it does not work to continue (retry) the pkg, but only to clean and rebuild `perl` until it succeeds.

I think we need `disable_parallel_build=yes` for perl.

PR #5236, opened 2016-11-27
New package: sklogw

xant0me:
A simple cache for dns query.

### DIFF
--- a/srcpkgs/sklogw/files/sklogw/log/run
+++ b/srcpkgs/sklogw/files/sklogw/log/run
@@ -1,0 +1,5 @@
+#!/usr/bin/sh
+
+#log script
+[ -d /var/log/sklogw ] || mkdir -p /var/log/sklogw
+exec svlogd -ttt /var/log/sklogw
--- a/srcpkgs/sklogw/files/sklogw/run
+++ b/srcpkgs/sklogw/files/sklogw/run
@@ -1,0 +1,7 @@
+#!/usr/bin/sh
+
+# copy the persistant cache to the in memory tmpfs
+rm -rf /tmp/cache_dns
+cp -rp /var/cache/dns /tmp/cache_dns
+chmod 700 -R /tmp/cache_dns
+exec sklogw
--- a/srcpkgs/sklogw/template
+++ b/srcpkgs/sklogw/template
@@ -1,0 +1,17 @@
+# Template file for 'sklogw'
+pkgname=sklogw
+version=0.1
+revision=1
+short_desc="A simple cache for dns query"
+maintainer="xant0me <xant0me@openmailbox.org>"
+license="MIT"
+homepage="https://framagit.org/xant0me/cache_dns"
+distfiles="https://framagit.org/xant0me/cache_dns/repository/archive.tar.gz?ref=v${version}>sklogw-${version}.tar.gz"
+checksum="7f28ab7f6b9a1cee4f16a49c3314b72f6bb192c3e4a119d4a3aa0de300e80b04"
+build_style="gnu-makefile"
+wrksrc="cache_dns-v0.1-e3aff30a1afa1819384cb66ac0252f472d45bb9a"
+
+post_install() {
+	vlicense LICENSE
+	vsv sklogw
+}
--- a/srcpkgs/sklogw/files/sklogw/finish
+++ b/srcpkgs/sklogw/files/sklogw/finish
@@ -1,0 +1,6 @@
+#!/usr/bin/sh
+
+#finish script
+
+rm -rf /var/cache_dns
+cp -rp /tmp/cache_dns /var/cache_dns


srcpkgs/sklogw/template
```diff
@@ -0,0 +1,28 @@
...
+maintainer="xant0me <xant0me@openmailbox.org>"
+license="MIT"
+homepage="https://framagit.org/xant0me/cache_dns"
+distfiles="https://framagit.org/xant0me/cache_dns/repository/archive.tar.gz?ref=v${version}>sklogw-${version}.tar.gz"
+checksum="7f28ab7f6b9a1cee4f16a49c3314b72f6bb192c3e4a119d4a3aa0de300e80b04"
+
+post_extract() {
+	mv \
+	/builddir/cache_dns-v0.1-e3aff30a1afa1819384cb66ac0252f472d45bb9a\
+	/${wrksrc}
```
Vaelatern: Please instead set `$wrksrc` to be correct.

srcpkgs/sklogw/template
```diff
@@ -0,0 +1,28 @@
...
+}
+
+do_build() {
+	make CFLAGS="-fPIC"
+}
```
Vaelatern: Please do not do this. Set a `build_style`, perhaps `gnu-makefile`

srcpkgs/sklogw/template
```diff
@@ -0,0 +1,28 @@
...
+
+do_install() {
+	make install
+	[ ! -d /etc/sv/sklogw ] && mkdir -p /etc/sv/sklogw/log
+	cp /${wrksrc}/sklogw_sv/{finish,run} /etc/sv/sklogw/
+	cp /${wrksrc}/sklogw_sv/log/run /etc/sv/sklogw/log/
```
Vaelatern: These last three lines are totally wrong. Please look at `vsv` in the Manual.

srcpkgs/sklogw/template
```diff
@@ -0,0 +1,28 @@
...
+	vlicense LICENSE
```
Vaelatern: Once you set a build_style, you will want to put this line in a `post_install()`

xant0me:
Thanks for the review. The Build seems to fail because of an error 500 when fetch the distfiles, but I can't reproduce the error. How can I ask travis to retry ? Sorry if this a trivial question.

Duncaen:
@xant0me force push to trigger travis again.

srcpkgs/sklogw/template
```diff
@@ -0,0 +1,21 @@
...
+version=0.1
+revision=1
+short_desc="A simple cache for dns query"
+maintainer="xant0me <xant0me@openmailbox.org>"
+license="MIT"
+homepage="https://framagit.org/xant0me/cache_dns"
+distfiles="https://framagit.org/xant0me/cache_dns/repository/archive.tar.gz?ref=v${version}>sklogw-${version}.tar.gz"
+checksum="7f28ab7f6b9a1cee4f16a49c3314b72f6bb192c3e4a119d4a3aa0de300e80b04"
+build_style="gnu-makefile"
+make_build_args="CFLAGS=-fPIC"
```
Vaelatern: This really should be already set.

xant0me: I put this option because I have an error when I `./xbps-src pkg` saying I should try with -fPIC.
If It's a mistake to do this way I'll add it in the Makefile.

xant0me:
I'll remove the post-extract and set wrksrc. I didn't want to do this way because I would have change it every new version. But the way I do now I still have to change every new version.

srcpkgs/sklogw/files/sklogw/log/run
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/sh
+
+#log script
+[ -d /var/log/sklogw ] || mkdir -p /var/log/sklogw
+exec chpst -u root svlogd -ttt /var/log/sklogw
```
Duncaen: I don't think `chpst` is usefull here.

srcpkgs/sklogw/files/sklogw/run
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/sh
+
+# copy the persistant cache to the in memory tmpfs 
+rm -rf /tmp/cache_dns
+cp -rp /var/cache/dns /tmp/cache_dns
+chmod 700 -R /tmp/cache_dns
+exec /usr/local/bin/sklogw
+
+
```
Duncaen: Please remove all trailing empty lines from all files.

srcpkgs/sklogw/files/sklogw/run
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/sh
+
+# copy the persistant cache to the in memory tmpfs 
+rm -rf /tmp/cache_dns
+cp -rp /var/cache/dns /tmp/cache_dns
+chmod 700 -R /tmp/cache_dns
+exec /usr/local/bin/sklogw
```
Duncaen: The path is wrong, just use `sklogw`.

xant0me:
Thanks @Duncaen, I change the scripts.

srcpkgs/sklogw/template
Vaelatern: The makefiles should respect our CFLAGS and LDFLAGS.
-fPIC is set there.

Vaelatern: Seems it is done.

Vaelatern:
Please have only one commit, with the message:
`New package: sklogw-0.1`

PR #5237, opened 2016-11-27
New package: grabc-1.1

drkhsh:


srcpkgs/grabc/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'grabc'
+pkgname=grabc
+version=1.1
+revision=1
+wrksrc=grabc${version}
+build_style=gnu-makefile
+makedepends="libX11-devel"
+depends="libX11"
+nopie=1
```
Vaelatern: Why? Please fix the cause of this, instead of setting `nopie`

srcpkgs/grabc/template
```diff
@@ -0,0 +1,19 @@
...
+short_desc="A utility for reporting the color of a pixel"
+maintainer="Aaron Marcher <info@nulltime.net>"
+license="GPL-3"
```
Vaelatern: Chances are very good this is incorrect because the version listed here was released only 10 years before GPL-3 came out. While time travel is possible, I doubt it.

srcpkgs/grabc/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'grabc'
+pkgname=grabc
+version=1.1
+revision=1
+wrksrc=grabc${version}
+build_style=gnu-makefile
+makedepends="libX11-devel"
+depends="libX11"
```
Vaelatern: No.
https://travis-ci.org/voidlinux/void-packages/jobs/179184021#L606-L608

srcpkgs/grabc/template
drkhsh: No clue... hope i fixed it now

drkhsh:
@Vaelatern I hope I fixed everything

srcpkgs/grabc/template
```diff
@@ -0,0 +1,18 @@
+# Template file for 'grabc'
+pkgname=grabc
+version=1.1
+revision=1
+wrksrc=grabc${version}
+build_style=gnu-makefile
+makedepends="libX11-devel"
+nostrip_files="grabc"
```
Vaelatern: Please don't do this either.

Vaelatern: The issue is that it is not respecting our `$LDFLAGS` in addition to its own (`-L/usr/X11R6/lib -lX11`)

drkhsh: Oh ok, i will see

drkhsh: yes

Vaelatern:
c08127a

PR #5245, opened 2016-11-28
gcc: gcc-go conflicts with go

ebfe:
$ xbps-query -o /usr/bin/go
    gcc-6.2.1_2: /usr/bin/go (regular file)
    go-1.7.3_1: /usr/bin/go (regular file)


chneukirchen:
Also go.1, and gofmt, gofmt.1, and i'd prefer renaming over a conflict.

ebfe:
/usr/lib/gcc/x86_64-unknown-linux-gnu/6.2/cgo
/usr/lib/libnetgo.a
/usr/lib/libgolibbegin.a

Rename /usr/bin/go to `?`

chneukirchen:
`gcc-go`, `gcc-gofmt`?

ebfe:
This is getting ugly...
```
$ strace -e exec -f go fmt
[...]
[pid 14740] execve("/usr/bin/gofmt", ["/usr/bin/gofmt", "-l", "-w", "main.go"], [/* 54 vars */]) = 0
[...]
```

chneukirchen:
Then keep the conflict, Arch does too.

pullmoll:
BTW: Perhaps someone finds the cause why linking go fails for musl. I disabled go for musl, because I could fix the compiling errors, but not the linking stage.

thypon:
why not having alternatives?

chneukirchen:
We could make a `gcc-go-tools` which conflicts with `go`... I think it could be useful to have `gccgo` installed in parallel to go.

ebfe:
Feel free to take over the PR

chneukirchen:
Adopted.

@pullmoll gccgo doesn't work on musl anyway due to lack of setcontext etc.

Issue #4762, opened 2016-09-14
$pkgname.googlecode.com dowload locations failing


pullmoll:
It looks like code.google.com is removing archives.
The list of packages which (may) need `distfiles` updates:
- [ ] blazeblogger
- [x] bmkdep
- [x] chroot-distcc
- [ ] clamz
- [x] distcc
- [ ] dnsmap
- [x] fbterm
- [ ] gmock
- [ ] gmtk
- [ ] gnome-colors-icon-theme
- [ ] gnome-mplayer
- [ ] gperftools
- [x] gtest
- [x] gtkdialog
- [ ] ibus
- [ ] key-mon
- [x] libmp4v2
- [ ] libpam-google-authenticator
- [x] libproxy
- [ ] librlog
- [ ] lua-luadbi
- [ ] m2m-aligner
- [ ] mitlm
- [ ] mongoose
- [x] mp3check
- [x] pingus
- [ ] radamsa
- [ ] serf
- [ ] touchegg
- [ ] wmii
- [x] wvdial
- [ ] wvstreams
- [x] xar

jcjordyn130:
True

PR #5256, opened 2016-11-30
wvdial: fetch tarball from Debian (#4762)

dutsky:


jcjordyn130:
What was wrong with the old link?

pullmoll:
*.googlecode.com packages are vanishing. See #4762

PR #5267, opened 2016-12-01
New package: ghq-0.7.4

SevereOverfl0w:


SevereOverfl0w:
Just realised I left `git` as a `hostmakedepends`, I mirrored this from `hub`, it's possibly more appropriate to have this as a `depends` (in both projects?)

The program does `os.exec` using the `git` command.

Could somebody please advise?

Vaelatern:
It can be in both.

SevereOverfl0w:
Added to both.

PR #5271, opened 2016-12-02
gegl3: add gobject-introspection

antonio-malcolm:


thypon:
`gobject-introspection` cannot be cross compiled. I think it's the rease was explicitly disabled.

A solution could be disabling selectively when cross compiling.

antonio-malcolm:
Yup, that makes sense. I'll update the template.

Gottox:
Closing for now. Reopen if you're ready.

PR #5275, opened 2016-12-02
New packages: pam_yubico-2.24 and ykclient-2.15

crater2150:
pam_yubico provides a PAM module for authenticating with a Yubikey. It requires the yubikey client library ykclient.


crater2150:
Hm, not sure what's the problem on ARM, the build seems fine but the help2man call at the end fails...

thypon:
The problem help2man tries to execute ykclient to create the manual, unfortunatly when crosscompiling the target architecture isn't the same as the host and you cannot execute ykclient. Maybe it's possible to disable help2man.

crater2150:
There is no configure switch to disable it, so I'm now simply removing the relevant part of Makefile.am. The manpages are there anyway, they are included in the distfile.
Now it only fails on aarch64, where it can't find autoreconf o.O

PR #5276, opened 2016-12-02
New package: python-yamllint-1.6.0

lafrenierejm:


Vaelatern:

srcpkgs/python3.4-yamllint
```diff
@@ -0,0 +1 @@
+python-yamllint
```
Vaelatern: Remove this symlink

Vaelatern:
`noarch=yes`
`alternatives="yamllint:yamllint:/usr/bin/yamllint2`
in subpkg
`alternatives="yamllint:yamllint:/usr/bin/yamllint3`

PR #5281, opened 2016-12-03
New package: attract

revington:
Attract for void https://github.com/mickelson/attract

### DIFF
--- a/srcpkgs/attract/template
+++ b/srcpkgs/attract/template
@@ -1,0 +1,18 @@
+# Template file for 'attract'
+pkgname=attract
+version=2.2.0
+revision=1
+build_style=gnu-makefile
+hostmakedepends="pkg-config"
+makedepends="xine-lib glu-devel libjpeg-turbo-devel MesaLib-devel libGL SFML-devel libopenal-devel zlib-devel freetype-devel ffmpeg-devel libavcodec fontconfig-devel libarchive-devel"
+depends="git"
+short_desc="Attract-Mode is a graphical frontend for emulators"
+maintainer="Pedro Narciso García Revington <p.revington@gmail.com>"
+license="GPL-3"
+nopie=yes
+homepage="http://attractmode.org/"
+distfiles="https://github.com/mickelson/attract/archive/v${version}.tar.gz"
+checksum=d350a078dec7d97498846adb56fb54210e2d4bab484a8c2cd558ce197194b07b
+do_build() {
+	CPATH="${XBPS_CROSS_BASE}/usr/include/freetype2/" make -j 3
+}
--- a/srcpkgs/attract/patches/Makefile.patch
+++ b/srcpkgs/attract/patches/Makefile.patch
@@ -1,0 +1,11 @@
+--- Makefile	2016-11-08 08:06:21.000000000 +0100
++++ Makefile	2016-12-04 20:18:08.087376672 +0100
+@@ -84,7 +84,7 @@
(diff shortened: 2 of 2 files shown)


srcpkgs/attract/template
```diff
@@ -0,0 +1,17 @@
...
+makedepends="glu-devel libjpeg-turbo-devel MesaLib-devel libGL SFML-devel libopenal-devel zlib-devel freetype-devel ffmpeg-devel libavcodec fontconfig-devel libarchive-devel"
+depends="git"
+short_desc="Attract-Mode is a graphical frontend for emulators"
+maintainer="Pedro Narciso García Revington <p.revington@gmail.com>"
+license="GPL-3"
+homepage="http://attractmode.org/"
+distfiles="https://github.com/mickelson/attract/archive/v${version}.tar.gz"
+checksum=d350a078dec7d97498846adb56fb54210e2d4bab484a8c2cd558ce197194b07b
+do_build() {
+	CPATH="/usr/include/freetype2/" make -j 3
```
Duncaen: Tough would be easier to export `CPATH` in `pre_build` and let xbps-src handle `do_build`.
You need to pass more variables and use `${makejobs}` if you use your `do_build`.

revington:
@Duncaen: 
Oh! Sorry. I thought it worked because `./xbps-src build attract` did not throw any error. Now I can see it failing by running `./xbps-src pkg attract`. I exported `CPATH` as suggested but same problem. It is my first package. Any idea?

revington:
Oh I see. /usr/local is hardcoded on the makefile. I'll checkout how to workaround that

Vaelatern:
Sed or a patch. Patches preferred, but sed works.

pullmoll:
Now squash those 4 commits using `git rebase -i HEAD~5`, set the commit message to `New package: attract-2.2.0` and force push with `git push -f`.

srcpkgs/attract/template
```diff
@@ -4,14 +4,16 @@ version=2.2.0
...
+makedepends="xine-lib glu-devel libjpeg-turbo-devel MesaLib-devel libGL SFML-devel libopenal-devel zlib-devel freetype-devel ffmpeg-devel libavcodec fontconfig-devel libarchive-devel"
 depends="git"
 short_desc="Attract-Mode is a graphical frontend for emulators"
 maintainer="Pedro Narciso García Revington <p.revington@gmail.com>"
 license="GPL-3"
+nopie=yes
 homepage="http://attractmode.org/"
 distfiles="https://github.com/mickelson/attract/archive/v${version}.tar.gz"
 checksum=d350a078dec7d97498846adb56fb54210e2d4bab484a8c2cd558ce197194b07b
+patch_args="-p1"
```
pullmoll: Instead of using patch_args="-p1" you could remove the a/ and b/ from the patch itself.
Self created patches should not set patch_args, but use the default, which is -p0.

pullmoll:
When modifying your branch and adding changes, you can use `git commit --amend` to avoid creating new commits. Then do a `git push -f` to overwrite the HEAD of your branch.

pullmoll:
I took your template and modified the Makefile.patch to use Void's CC, CXX, CFLAGS and LDFLAGS. No need to set nopie=yes. Cross compiling works, and the musl libc works with a small patch.

revington:
Awesome @pullmoll thank you very much! GIT is being used at the Makefile to determine some flags. Sorry, this was my first package :(

PR #5282, opened 2016-12-03
New package: ccid-1.4.25

qbit:
ccid is a usb chip/smart card device driver for use with pcsclite.
With this package you can use YubiKeys in CCID mode.


qbit:
Wooops - pcsc-ccid exists.

PR #5294, opened 2016-12-05
New package: tanidvr-1.4.1

ym00n:


srcpkgs/tanidvr/template
```diff
@@ -0,0 +1,13 @@
+# Template file for 'tanidvr'
+pkgname="tanidvr"
+version="1.4.1"
+revision=1
+build_style=gnu-configure
+configure_args=" --prefix=/usr"
```
Vaelatern: Redundant, remove this line.

ym00n: Line removed. Should I do anything else?

Vaelatern:
You may notice I did a squash and merge. This is because every PR needs to only have 1 commit per template. This time it was fine, in more complicated cases it may not be fine.

PR #5313, opened 2016-12-07
dma: fix permissions

ackalker:


srcpkgs/dma/template
```diff
@@ -11,7 +12,7 @@ license="BSD"
 homepage="https://github.com/corecode/dma"
 distfiles="https://github.com/corecode/dma/archive/v${version}.tar.gz"
 checksum=288bd57f17aba696b6423b1fa0cd0f7d9b7228d32811f1cd7b821d5f020051b8
-system_accounts="mail"
+system_groups="mail"
```
Vaelatern: Why remove the mail user?
While we are here, can we make it `_mail` instead?

ackalker: Ah, I'll reinstate the user, I had downgraded it to be only a group for testing. Why rename it to `_mail`, to prevent conflict with other mailers? Wouldn't that create confusion?

chneukirchen:
Group mail is defined in `/etc/groups` already, no need to declare it.

ackalker:
@chneukirchen Right...
@Vaelatern  Why would you like me to change the user name to `_mail`?

Vaelatern:
We are trying, slowly, to change system users and groups to start with an underscore. If that doesn't happen now, then it doesn't happen now, but xt seemed quite in favor of it. 78548dba2a5a4c74d5940107a6b2c9a0aa8e1c9f

Vaelatern:
There are reasons to not change this group now, as discussed in IRC, so by all means, leave it as it is.

ackalker:
Thank you @chneukirchen and @Vaelatern , I will change from group back to user `mail`, rebase and force push the change.

PR #5321, opened 2016-12-09
google-chrome: update to 55.0.2883.75-1

the-maldridge:
This PR updates google chrome to the latest stable version.


srcpkgs/google-chrome/template
```diff
@@ -1,5 +1,6 @@
+
```
Vaelatern: Spurious addition of a line at the beginning, can it be removed?

PR #5322, opened 2016-12-09
wps-office: refactor and update

the-maldridge:
Update wps-office to use the packaged libpng instead of the builtin one.


the-maldridge:
the patchlevel variable is retained so that this doesn't need constant refactoring when new versions are released and then subsequent patches to those versions are released.

PR #5326, opened 2016-12-10
abcm2ps: fix destdir

lemmi:


lemmi:
I was just testing this and I get strange results, so please don't merge yet.

lemmi:
so 8.13.0 seems to have a few issues. but we should at lest get out that fix for 8.12.9

ebfe:
711d5b4370617797a45c1b4aee668f7a64f5c2ce

PR #5332, opened 2016-12-10
perl-Exception-Class: update to 1.41.

enguerrand:


ebfe:
1c94595c1009f0b35e4dae629ac5b7c8cdd47c57

PR #5339, opened 2016-12-12
linux4.9: kernel hardening

thypon:
following https://kernsec.org/wiki/index.php/Kernel_Self_Protection_Project


pullmoll:
Do you have a pointer for the source of the idea / changes, or did you devise all of them yourself?

thypon:
As i alredy write in the first commit message Kees Cook's Kernel Self Protection Project is the main source. https://kernsec.org/wiki/index.php/Kernel_Self_Protection_Project

This is one of his talks
https://www.youtube.com/watch?v=aMkCKeZ8xZw

chneukirchen:
I'd prefer putting all security sysctl into a separate .conf files for easier disabling.

I think page poisoning is too expensive to enable by default. It also breaks hibernate afaiu.

Not sure CONFIG_HIGHMEM64G=y makes sense for i386?
Why change CONFIG_INET_DIAG?

thypon:
> I'd prefer putting all security sysctl into a separate .conf files for easier disabling.

Me too, let's hear someone else.

> I think page poisoning is too expensive to enable by default. It also breaks hibernate afaik.

I'll do performance regression tests inside a virtual machine if you think they would impact the performances too much.

> Not sure CONFIG_HIGHMEM64G=y makes sense for i386?

I think it's needed with PAE to render RANDOMIZE_BASE more effective.

> Why change CONFIG_INET_DIAG?

The wiki advise a kernel pointer leak prior 4.1, but i think you are right and could stay enabled.

thypon:
@chneukirchen @Duncaen ... others

http://desktop.pompel.me/compound.results

The results on the left belongs to the non hardened kernel while the one on the right belongs the hardened one. To sum up, only the process creation is affected more than 10%, but i hardly fork bomb on my computer.

I obviously reenabled CONFIG_INET_DIAG.

thypon:
Another thing i should disable is livepatch given we have no real usage right now.

chneukirchen:
Note that running configure and building software is bound limited by process creation.

thypon:
Ok then a good benchmark could be compiling a linux kernel or chromium with the two configurations.

thypon:
@chneukirchen 

Without hardening:
```
bash-4.3# time make -j16
real    27m22.752s
user    96m49.877s
sys     8m0.508s
```

With hardening and poisoning enabled at startup:
```
bash-4.3# time make -j16
real    28m29.004s
user    98m21.723s
sys     10m19.178s
```

20% overhead on systime less then 1% overhead on usertime.

thypon:
The patches is okay. I'm not quite sure how to split the sysctl.conf file instead @chneukirchen for other infos.

chneukirchen:
More fallout: https://groups.google.com/forum/#!topic/voidlinux/imKgCybt6Q4

PR #5342, opened 2016-12-12
New package: typespeed-0.6.5

kalichakra:


srcpkgs/typespeed/template
```diff
@@ -0,0 +1,14 @@
+# Template file for 'typespeed'
+pkgname="typespeed"
+version="0.6.5"
+revision=1
+build_style=gnu-configure
+makedepends="ncurses-devel"
+short_desc="Test your typing speed"
+only_for_archs="x86_64 i686 aarch64 armv7hf"
```
Vaelatern: Please don't do this. For example, you have just unintentionally banned `armv6hf`, for which this probably builds just fine. Remove this line, let someone else decide how to fix `-musl`, or if to do so at all.

srcpkgs/typespeed/template
```diff
@@ -0,0 +1,14 @@
...
+maintainer="David <kalichakra@zoho.com>"
+license="GPL-2"
+homepage="http://typespeed.sourceforge.net"
+distfiles="${homepage}/${pkgname}-${version}.tar.gz"
```
Vaelatern: We have a more "official" way to download from sourceforge, documented in the Manual. Please use it.

srcpkgs/typespeed/template
```diff
@@ -0,0 +1,14 @@
...
+checksum="5c860385ceed8a60f13217cc0192c4c2b4705c3e80f9866f7d72ff306eb72961"
+nocross=yes
```
Vaelatern: Please remove this line too.

kalichakra:
Fixed

Vaelatern:
https://travis-ci.org/voidlinux/void-packages/jobs/183317693#L644
Seems like clock_t is defined in muslc, have to patch to include that header file in the right place. (I might suggest a patch later)

Vaelatern:
https://travis-ci.org/voidlinux/void-packages/jobs/183317692#L791-L794
It's calling `gcc` instead of `$(CC)`, which will cause problems. This should also be patched.

kalichakra:
I've looked into clock_t and muslc, indeed it's defined in "bits/alltypes.h", and "time.h" would need to be included before as well, and these should be included from "src/typespeed.h". Don't waste your time on this i'll provide a patch.

As for gcc, src/Makefile.am uses "CC = gcc", which could just be commented i suppose ?

kalichakra:
Updated. Fixed compilation on -musl arches. CC=gcc is stripped from the main makefiles.

PR #5343, opened 2016-12-12
vapoursynth: revbump for libmagick-6.9.6.8

lemmi:


pullmoll:
https://build.voidlinux.eu/changes/33567
It's just that it is not yet built because openshot fails. Will fix this soon.

pullmoll:
It's bumped now.

PR #5347, opened 2016-12-12
New package: open-vm-tools-10.1.0

ottbot:
This adds open-vm-tools with pam and X options enabled by default, the dependent `libdnet` is also added.


jsumners:
Does the open-vm-tools package require that X be installed?

Also, there's an actually maintained version of libdnet at https://github.com/nmap/nmap/tree/master/libdnet-stripped . I haven't had a chance to see if open-vm-tools builds against it though.

PR #5353, opened 2016-12-13
New package: teamspeak3-3.0.19.4

raedwulf:
This is a new package template for Teamspeak3.  The version is included in the name to avoid confusion with an earlier, incompatible, Teamspeak 2 which sometimes available as a package as "teamspeak" in other distributions.

Some comments in no particular order:
- This package is proprietary; so I've tried to follow the `nvidia` and `steam` package.
- It is popular with gamers (so a useful complement to `steam` and users that game with Windows players in some games).
- It has an update script for checking the version.


raedwulf:
Unfortunately, there's a blocker before this can be accepted... Teamspeak3 don't allow redistribution without explicit consent.  They have granted it in the past for Arch Linux: https://git.archlinux.org/svntogit/community.git/tree/trunk/PERMISSION.eml?h=packages/teamspeak3

We'd probably need to do the same - perhaps someone who represents Void Linux, or I can attempt it myself.  Could someone advise on this matter?

raedwulf:
As there are packages in a similar situation in the repository, I've marked it as "restricted=yes".

anschch:
Thank you for this! I hope it gets merged to master.

PR #5354, opened 2016-12-13
New package: linphone-3.10.2

kalichakra:
This is linphone-3.10.2 with all up-to-date dependencies. Belle-sip requires antlr3 to build ; i've reused the package "antlr3-bin" written by "duncaen", my previous template had belle-sip and antlr3's jar packaged together but because they have different licenses it seems better to package them separately.


kalichakra:
bctoolbox doesn't compile right now on *-musl because musl doesn't have mallinfo() which is used by src/tester.c in test_complete_message_handler() to monitor the space allocated by the process. This functions calls mallinfo() if max_vm_kb is set with --max-alloc, by default it is not called.

If we want to this to compile with musl there are a few solutions ; we can use getrusage() only on -musl to provide pretty much the same functionality as mallinfo(). We could also remove this part of the function and disable the --max-alloc argument but this is less acceptable.

kalichakra:
I propose this patch to replace mallinfo() with getrusage() in bctoolbox, this way we get the same behaviour whatever libc is being used.
[rusage.patch.txt](https://github.com/voidlinux/void-packages/files/650515/rusage.patch.txt)

Vaelatern:
The patch looks okay, but I'd have to read the documentation on both approaches before saying "absolutely", but by all means add the patch to your PR

kalichakra:
I included the patch in the bctoolbox package. As a result, bctoolbox, mediastreamer and linphone compile as well on *-musl.

First i'd like to say that this patch doesn't change anything critical IMOO.
Bctoolbox provides two libraries, bctoolbox and bctoolbox-tester, the latter
can be used to validate test suites (belle-sip has a "tester" program). When
initiating the bctoolbox-tester library, it is possible to limit the memory
allocated by the process and threads through the --max-alloc argument (in
kilobytes). Not one of the "tester" programs in belle-sip, mediastreamer or
linphone uses this functionality, and none of them is packaged, they are just
here for debugging purposes.

I don't know how getrusage() behaves on musl compared to glibc's mallinfo()
IMOO, we could just as well disable --max-alloc since it's not used at all here

srcpkgs/belle-sip/template
```diff
@@ -0,0 +1,27 @@
...
+revision=1
+build_style=gnu-configure
+short_desc="SIP (RFC3261) implementation written in C"
+maintainer="David <kalichakra@zoho.com>"
+license="GPL-2"
+homepage="http://www.linphone.org"
+distfiles="http://download.savannah.gnu.org/releases/linphone/${pkgname}/${pkgname}-${version}.tar.gz"
+checksum="4bc44ee83f8455fb6248d3b3045c38ab6e71e6699cde3f0cc5d67493bb38a43a"
+
+makedepends="antlr3-bin mbedtls-devel libantlr3c bctoolbox-devel"
```
Vaelatern: This really requires a java program to be able to build?

srcpkgs/cunit/template
```diff
@@ -0,0 +1,21 @@
...
+short_desc="Unit testing framework for C"
+build_style="gnu-configure"
+maintainer="David <kalichakra@zoho.com>"
+license="GPL-2"
+homepage="http://cunit.sourceforge.net"
+distfiles="${SOURCEFORGE_SITE}/project/${pkgname}/CUnit/${version}-3/CUnit-${version}-3.tar.bz2"
+checksum="f5b29137f845bb08b77ec60584fdb728b4e58f1023e6f249a464efa49a40f214"
+wrksrc="CUnit-${version}-3"
+
+hostmakedepends="pkg-config libtool autoconf automake"
```
Vaelatern: Remove autoconf from here. automake pulls it in.

kalichakra:
belle-sip requires antlr3 to build, and antlr3 depends on java.

kalichakra:
Updated.

Vaelatern:
shlibs conflicts.

the-maldridge:
Has there been any update on this PR since February?  This would currently close the second oldest issue in the repo ( #155 )

kalichakra:
I haven't had the occasion to make the changes (changing the shlibs file ?). Can't it be fixed at merge time ?

NeoChontrix:
@kalichakra I have updated all the Linphone related templates in my repo so you can pull them for there
https://github.com/NeoChontrix/void-packages/commits/linphone

You can also drop the cunit package since it isn't needed for Linphone anymore.

asergi:
#7490

PR #5359, opened 2016-12-14
Preparatory work for monodevelop

t0b1nux:
All courtesy for mono update goes to @Hoshpak (cf. https://github.com/voidlinux/void-packages/pull/5070).
This is just a minor update (because #5070 was closed and not merged). Most explanations for the first commit are given at #5070.
All of this is just some prep work to add monodevelop.

### DIFF
--- a/common/shlibs
+++ b/common/shlibs
@@ -2788,3 +2788,8 @@ libmutter-cogl-path.so mutter-3.22.0_1
 libmutter-cogl.so mutter-3.22.0_1
 libgeoclue-2.so.0 geoclue2-2.4.4_1
 libgepub.so.0 libgepub-0.4_1
+libmono-profiler-gui-thread-check.so.0 gtk-sharp-3.20.4_1
+libatksharpglue-3.so gtk-sharp-3.20.4_1
+libgtksharpglue-3.so gtk-sharp-3.20.4_1
+libgiosharpglue-3.so gtk-sharp-3.20.4_1
+libpangosharpglue-3.so gtk-sharp-3.20.4_1
--- a/common/shlibs
+++ b/common/shlibs
@@ -2793,3 +2793,9 @@ libatksharpglue-3.so gtk-sharp-3.20.4_1
 libgtksharpglue-3.so gtk-sharp-3.20.4_1
 libgiosharpglue-3.so gtk-sharp-3.20.4_1
 libpangosharpglue-3.so gtk-sharp-3.20.4_1
+libpangosharpglue-2.so gtk-sharp2-2.12.42_1
+libatksharpglue-2.so gtk-sharp2-2.12.42_1
+libgladesharpglue-2.so gtk-sharp2-2.12.42_1
+libglibsharpglue-2.so gtk-sharp2-2.12.42_1
+libgtksharpglue-2.so gtk-sharp2-2.12.42_1
+libgdksharpglue-2.so gtk-sharp2-2.12.42_1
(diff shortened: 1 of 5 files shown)


t0b1nux:
After a thought, I checkout and it seems to be not a single program relying on gtk-sharp-2.99 (contained in the package gtk-sharp).
I also added a fsharp recipe. I have some mixed feeling about this one (it need to have internet access during the build to fetch some NuGet packages, so it might breaks without reason) but it is a requirement for monodevelop. I yet have to investigate if it is possible, feasible and interesting to integrate the NuGet packages locally to have a reproducible build proccess.

t0b1nux:
Getting monodevelop to compile on voidlinux is not going to be easy as it requires gnome-sharp, which in turn requires libgnomeui (otherwise it only compile gconf-sharp), which requires itself libbonoboui, etc...
I do not have the time currently to make all theses packages.
So I thinkg upgrading at least mono is still wise, even just from the security point of view. As for gtk-sharp & fsharp, I don't know what would be best.

t0b1nux:
All right, so what I've done so far: update mono (of course), create a new package called gtk-sharp2, add fsharp (needed for monodevelop as stated before) and update gtk-sharp to 3.20.4. I should say that this isn't the version from the mono repository but a fork that is up to date and widely used.

I know the convention used by most gtk packages is : 
- 2.x release: no suffix;
- 3.x release: append a '3' to the package name.
In this case however, gtk-sharp was already C# bindings for gtk3+ (despite being published as 2.99.3 by mono developers). So I tought it was logical to update it to 3.20.4 and create another package called gtk-sharp2 which would contain the release 2.12.42 (that is, the one with gtk+-2.x bindings).

t0b1nux:
All right, so I guess it's time to explain the situation:
to get the current monodevelop version working would require making a gnome-sharp package, which will in turn have other dependencies. Quite frankly, I'm not ready to do this work and further maintain theses required (yet outdated) packages.
However, the development branch of monodevelop has dropped (mono/monodevelop#1661 for reference and the rationale) its requirement for gnome-sharp. As such, I think it's better to wait a bit until the 6.2 branch is out with theses patches.

Gottox:
Sorry for the long delay, void devs were busy moving the server. :D

Okay, let's wait. Does this affect this preparation PR in any way? Can we still merge?

t0b1nux:
No problem for the delay, I do not expect volunteers to be available 24/7, even more if they have way more important things going on.
As for theses packages, they should be fine (except the internet requirement during compilation explained above, I don't know voidlinux policy about it). However, I can not get monodevelop to work until they release 6.2.
This work can be merged if it meet your quality requirements.

asergi:
In the meantime, I already pushed the mono update (https://github.com/voidlinux/void-packages/commit/653309c1880069bf17b8e679c8986112f203bc4c), so that this PR can be merged at a later time.

Gottox:
please resolve the merge conflicts and rebase :D

Vtec234:
Any progress on this? It would be nice to have for Unity3D development.

t0b1nux:
Sorry, I can't handle this for the moment (this week), I'll try to compile and package monodevelop next week when I will have a bit more time.

t0b1nux:
Sorry, but if someone could handle this (@Vtec234 ?), it would be great ! (I tried to compile it but the latest version from git is somewhat incompatible with the Microsoft.Builder package of mono in the distribution). I'm not using monodevelop anymore (I just needed it for a project and we migrated to another language for easier cross-platform simplicity), so this PR is now orphan. Shall I close it ?

Vtec234:
I forked your branch, so you can close this if you want. Hopefully I can get it to work.

t0b1nux:
Ok, thanks and sorry for the disconvenience.

PR #5368, opened 2016-12-16
terraform: update to version 0.8.1

ahjones:
Since version 0.7 Terraform comes with plugins built in. See
https://www.terraform.io/docs/internals/internal-plugins.html

This change also removes the `terraform-` executables from the path,
and stops Terraform from warning about overridden plugins

### DIFF
--- a/srcpkgs/terraform/template
+++ b/srcpkgs/terraform/template
@@ -1,6 +1,6 @@
 # Template file for 'terraform'
 pkgname=terraform
-version=0.8.0
+version=0.8.1
 revision=1
 build_style=go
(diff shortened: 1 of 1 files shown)


srcpkgs/terraform/template
```diff
@@ -23,8 +23,6 @@ do_install() {
 	do
 		if [ "$(basename $line)" = terraform ]; then
 			vbin $line
-		else
-			vbin $line terraform-$(basename $line)
```
thypon: This is not the correct fix, but thank you :)

PR #5371, opened 2016-12-17
New package: ncdns-0.0.2

ananteris:


srcpkgs/ncdns/template
```diff
@@ -0,0 +1,20 @@
+# Template file for 'ncdns'
+pkgname=ncdns
+version=0.0.2
+revision=1
+only_for_archs="x86_64-musl x86_64 i686"
+build_style=go
+hostmakedepends="git libcap-devel"
```
Duncaen: Move `libcap-devel` to `makedepends`.

srcpkgs/ncdns/template
```diff
@@ -0,0 +1,20 @@
...
+go_get=yes
```
Duncaen: You already provide a distfile, is `go get` really needed?
If not I guess you can drop the `git` `hostmakedepends` too.

srcpkgs/ncdns/template
```diff
@@ -0,0 +1,20 @@
...
+go_import_path="github.com/hlandau/ncdns"
+short_desc="Nameserver for Namecoin .bit domains"
+maintainer="ananteris <ananteris@mailinator.com>"
+license="GPL-3"
+homepage="https://github.com/hlandau/ncdns"
+distfiles="https://github.com/hlandau/${pkgname}/archive/v${version}.tar.gz"
+checksum="a9457216f5bbc3c3b41e0538b78f8f67ad9a04304743a352a7325eef6af4a439"
+
+do_install() {
+	vbin ${GOPATH}/bin/ncdns
```
Duncaen: The `go` `build_style` already searches and installs binaries in `$GOPATH` by default (https://github.com/voidlinux/void-packages/blob/master/common/build-style/go.sh#L17).
You can remove it here, and rename the function to `post_install` to just install the config file.
I think if you drop `go_get=yes` you will be in the `ncdns/` directory and you dont need the long path.

ananteris: All of these have been resolved.

PR #5374, opened 2016-12-17
New package: premake5-5.0.0alpha10

raedwulf:
~Bootstrap.mak is unfortunately necessary from this alpha because the buildscripts provided by the package do not seem to work with libressl unless you regenerate them.  This bootstrap script comes from the premake5 git repository.~

~The bootstrap mechanism doesn't appear to work because it needs to be compiled on the host architecture as well... how do you make it cross-compile aware?~
~See if my solution is acceptable, I override CC for the bootstrap makefile so that it uses `CC=$XBPS_TRIPLET-gcc` .~

~I can't get the bootstrap mechanism working properly like this... I'll create a bootstrap package that needs to be built/installed on the host.~

Managed to make a couple of simple patches to fix the build scripts that come with premake5-5.0.0-alpha10.

This should be ready.


raedwulf:
I'm confused as to why the `hostmakedepends` is being compiled for the target architecture as opposed to the host architecture? Can someone shed some light on this? Am I doing this correctly?

There's a third option - I could manually patch the build scripts so that I don't need a bootstrap package and build it directly - although this will be more prone to breakage in the future.

asergi:
7515c189fbd7096f86129a0338572ba3f304ca32 (#10771)

PR #5376, opened 2016-12-17
Oracle-JDK provides JRE (prevents breakage)

skurtix:
While Oracle-JDK is installed packages that depends on virtual?java-runtime will also try to install Oracle-JRE which introduces an additional hook in /etc/profile.d/ which sets JAVA_HOME to the JRE instead of the JDK which breaks applications that depend on the JDK. Since Oracle-JDK contains the JRE, there is no need to install the Oracle-JRE and so **I've added java-runtime to provides on the package**.

**Additional notes (not included in this commit):**
The precedence order of Oracle-JDK and Oracle-JRE's /etc/profile.d hooks is also erroneous:
Oracle-JDK installs 11_oracle-jdk.sh (should be 10_oracle-jdk.sh)
Oracle-JRE installs 10_oracle-jre.sh (should be 11_oracle-jre.sh)
(for reference OpenJDK installs 10_openjdk.sh)

In case one would want to maintain the capability of having both installed on the same system the above should change. But I fail to see a reason for having both Oracle-JRE and Oracle-JDK installed simultaneously as Oracle-JDK provides JRE and there should probably be a 'replaces' addition for it.


ebfe:
10_oracle-jdk.sh should probably set `JDK_HOME` instead of/in addition to `JAVA_HOME`. Anything that actually requires the full jdk should give preference to `JDK_HOME` (as for example intellij idea does).

skurtix:
I'm not sure that JRE/JDK_HOME is as universally respected as it should be. But since Oracle-JDK actually _includes_ Oracle-JRE I don't see any reason for allowing both to begin with. Let one _replace_ the other and set JAVA_HOME=${JDK_HOME=/usr/lib/jvm/oracle-jdk}

PR #5378, opened 2016-12-18
New package: ccl-1.11

lipidity:
The Clozure CL implementation of the Common Lisp programming language


chneukirchen:
According to the website, the license is Apache-2.0?

If the examples are not too big, you can install them too.

Please use a `case` in line 18/19.

Else :+1:

Vaelatern:
Nopie?

chneukirchen:
Nopie is ok, as CCL is a native compiler that probably doesn't support pie.

lipidity:
Thanks @chneukirchen and @Vaelatern for the feedback on my first submission.

PR #5384, opened 2016-12-18
encompass and dependencies 

ananteris:


ananteris:
Working commits for encompass-0.5.0, although encompass-0.6.0rc1 should be used instead. For some odd reason a required py file is not getting packaged or noticed by xbps-src (lib/chains/bitcoin_chainkey.py). If this file is manually installed into the package 0.6.0rc1 works as expected.

PR #5385, opened 2016-12-18
Fix handling of package options in packages with '+' character in name.

hadrianw:
Substitute '+' character with '_', because eval treats '+' character
as not a part of variable name.

Edit: It is strange that I can like my own comment. I just wanted to see who liked it.


hadrianw:
Without this patch:

```
$ ./xbps-src -o ~wayland show-options gtk+3
=> gtk+3-3.22.5_1: the following build options are set:
   broadway: Enable support for the HTML5 Broadway backend (ON)
   colord: Enable support for colord (ON)
   cups: Enable support for CUPS (ON)
   gir: Enable support for gobject introspection data (ON)
   wayland: Enable support for the Wayland backend (ON)
   x11: Enable support for X11 (ON)
```

With this patch:
```
$ ./xbps-src -o ~wayland show-options gtk+3
=> gtk+3-3.22.5_1: the following build options are set:
   broadway: Enable support for the HTML5 Broadway backend (ON)
   colord: Enable support for colord (ON)
   cups: Enable support for CUPS (ON)
   gir: Enable support for gobject introspection data (ON)
   wayland: Enable support for the Wayland backend (OFF)
   x11: Enable support for X11 (ON)
```

hadrianw:
If I add `-o` argument it should append a rule to whatever comes from `etc/conf`? If this is the case the patch is not enough, because it ignores that I disabled broadway in `etc/conf` globally and for package itself.

PR #5386, opened 2016-12-19
New packages: aspcud-1.9.1, clingo-5.1.0, clasp-3.2.1; update opam

edwintorok:
Opam [works best with the `aspcud` solver](http://opam.ocaml.org/doc/Install.html#ExternalSolvers) , but this was not available in Void Linux.
So I created packages for `aspcud` and its dependencies (`clingo` and `clasp`), and updated the `opam` package to use `aspcud`.

I've made each package build and run its tests in `post_build` too.
(This is quite important for `opam`, there was a bug in the Ubuntu package in the past which would've been detected by the built-in tests, except these weren't run.  And then the package broke during an automated rebuild when a newer version of its dependency got installed)

For `clingo` I made those changes to `CXXFLAGS` because those are the default ones from `SConscript`, except `void`'s build system overrides them and then the packages defaults are not used.
Similarly for `clasp` those are the recommended defaults for a release build, except I also needed to add `-fPIC`, so I needed to specify the full set of flags.


edwintorok:
All builds are green (after I force pushed some patches), please let me know if there is anything else I should change on this PR, so it can be considered for merging.

srcpkgs/aspcud/template
```diff
@@ -0,0 +1,23 @@
+pkgname="aspcud"
```
Vaelatern: comment on first line

srcpkgs/aspcud/template
```diff
@@ -0,0 +1,23 @@
...
+revision=1
+build_style=cmake
+short_desc="A solver for package problems in CUDF format using ASP"
+maintainer="Török Edwin <edwin@etorok.net>"
+license="GPL-3"
+homepage="https://potassco.org/aspcud/"
+wrksrc="${pkgname}-${version}-source"
+distfiles="${SOURCEFORGE_SITE}/potassco/${wrksrc}.tar.gz"
+checksum="e0e917a9a6c5ff080a411ff25d1174e0d4118bb6759c3fe976e2e3cca15e5827"
+hostmakedepends="cmake re2c"
```
Vaelatern: Drop `cmake` here

srcpkgs/clasp/template
```diff
@@ -0,0 +1,38 @@
+pkgname="clasp"
```
Vaelatern: comment

srcpkgs/clasp/template
```diff
@@ -0,0 +1,38 @@
...
+
+post_extract() {
+	# musl libc doesn't have it
+	sed -i -e 's/xlocale.h/locale.h/g' libprogram_opts/src/string_convert.cpp
+}
+
+do_configure() {
+	CXXFLAGS="${CXXFLAGS} -O3 -DNDEBUG -fPIC"
+	./configure.sh --config=custom --prefix /usr ${configure_args} --with-mt=std
+}
```
Vaelatern: Please try to use the build style. Possible `gnu-configure`

srcpkgs/clasp/template
```diff
@@ -0,0 +1,38 @@
...
+
+do_build() {
+	: ${make_cmd:=make}
+	cd build/custom_mt
+
+	${make_cmd} ${makejobs} ${make_build_args} ${make_build_target}
+}
```
Vaelatern: Same. Build styles are good.

srcpkgs/clasp/template
```diff
@@ -0,0 +1,38 @@
...
+
+post_build() {
+	if [ -z "$CROSS_BUILD" ]; then
+		${make_cmd} ${makejobs} test LDFLAGS="${LDFLAGS} -pthread"
+	fi
+}
+
+do_install() {
+	vbin build/custom_mt/bin/clasp clasp
+}
```
Vaelatern: Makefile really doesn't offer an option?

srcpkgs/clingo/template
```diff
@@ -0,0 +1,51 @@
+pkgname="clingo"
```
Vaelatern: comment

srcpkgs/clingo/template
```diff
@@ -0,0 +1,51 @@
...
+do_configure() {
+	: ${make_cmd:=scons}
+
+	${make_cmd} ${makejobs} CC=$CC CXX=$CXX CCFLAGS="$CFLAGS" \
+		CXXFLAGS="$CXXFLAGS" LINKFLAGS="$LDFLAGS" \
+		prefix=/usr destdir=${DESTDIR} DESTDIR=${DESTDIR} \
+		${make_build_args} configure
+	if [ -z "$CROSS_BUILD" ]; then
+		sed -i -e "s|^CPPPATH.*|CPPPATH=['/usr/include/lua-5.3', '/usr/include/python2.7']|" build/release.py
+	fi
```
Vaelatern: Please try to use the build style.

srcpkgs/clasp/template
edwintorok: It has an `install` target, but no support for `DESTDIR` (it uses `INSTALL_DIR`, and doesn't know how to mkdir "${DESTDIR}/usr/bin"). Seemed more readable to just use `vbin` than working around the broken install target.

srcpkgs/aspcud/template
edwintorok: Thanks for the review, I pushed some fixups, indeed the templates could be simplified.

edwintorok:
Please let me know if I should make any other changes to this PR, or if I should put the change to opam in a separate PR.

silvernode:
Found this by typing random words into the issue tracker. We should decide the fate of this PR since it has been lingering around for quite awhile. At least that's my 2 cents on old stuff.

edwintorok:
opam `2.0.0~rc` embeds the `mccs` solver, so an external solver like aspcud will no longer be required.
Don't know why this PR has lingered for so long, but I think it is no obsolete.

PR #5390, opened 2016-12-20
python-qrcode: update to 5.3_2

ananteris:


Vaelatern:
Please see 3700e00 for a better example of how to name these commits.

PR #5391, opened 2016-12-20
New package: adobe-flash-installer-24.0.0.186

masato9000:
This works like now-broken chromium-pepper-flash package, except downloads from adobe instead and installs according to their instructions.  NPAPI version.


Vaelatern:
Is there any way to merge your packages into a single template and use subpackages? Or are they actually very different?

masato9000:
@Vaelatern: perhaps, I'd need to look.  However, I'm withdrawing these PRs for now (but keeping my branches) as I have come up with something that seems a bit better based on some input I received on the forum and IRC.

PR #5412, opened 2016-12-24
New package: lzlib-1.8 and plzip-1.5

bra1nwave:


srcpkgs/plzip/template
```diff
@@ -0,0 +1,16 @@
...
+build_style=gnu-configure
+makedepends="lzlib-devel"
+short_desc="Multi-threaded compressor using the lzip file format"
+maintainer="bra1nwave <brainwave@openmailbox.org>"
+license="GPL-2"
+homepage="http://www.nongnu.org/lzip/plzip"
+distfiles="${NONGNU_SITE}/lzip/plzip/${pkgname}-${version}.tar.gz"
+checksum=cd401683c2c57ecf170c99a873e41e9e5fee0e9680341646e31032e31489cc8a
+
+do_configure() {
```
chneukirchen: Why is this explicit?

srcpkgs/lzlib/template
```diff
@@ -0,0 +1,25 @@
...
+configure_args="--enable-shared --disable-static"
+short_desc="Compression library for the lzip file format"
+maintainer="bra1nwave <brainwave@openmailbox.org>"
+license="GPL-2"
+homepage="http://www.nongnu.org/lzip/lzlib"
+distfiles="${NONGNU_SITE}/lzip/lzlib/${pkgname}-${version}.tar.gz"
+checksum=41bfa82c6ee184ed0884437dc4074ad505e64cb747432cefa97976b89045cbad
+
+do_configure() {
+	./configure --prefix=/usr $configure_args CC="$CC" CFLAGS="$CFLAGS" LDFLAGS="$LDFLAGS"
```
chneukirchen: Same.

srcpkgs/plzip/template
bra1nwave: The global variables don't get invoked, which makes cross-compiling not possible. 
I don't think a patch is necessary for this.

Also the [lzip template](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/lzip/template) does exactly the same.

chneukirchen:
Ah yes, these are stupid configure scripts. ok.

bra1nwave:
Thanks and Happy holidays :)

PR #5414, opened 2016-12-24
sbcl: enable musl build

lipidity:
* Patch for header differences

* Contrib modules sb-bsd-sockets and sb-simple-streams fail
  Hopefully this is fixed at some point otherwise no quicklisp

* Threading support only with glibc

* We also cannot use the binary sbcl for bootstrapping
  so use ccl (clisp would also work, but probably slower)


chneukirchen:
Bootstrapping from ccl is a nice idea, better than running sbcl with fake-glibc.

I'll look into whether I can fix the outstanding things soon.

chneukirchen:
I found my old SBCL patches, integrating them.

PR #5419, opened 2016-12-25
darktable: update to 2.2.0. [NOMERGE]

lemmi:
This is a test whether things compile fine on the supported archs. A lot of went into this release. Even the developers partially think of this of the final release candidate, so i'd like to wait for 2.2.1.


lemmi:
I think this can be merged. Build only fails on aarch64 because of dependencies.

ebfe:
With 
```diff
diff --git a/srcpkgs/darktable/template b/srcpkgs/darktable/template
index d369deed8c..a8f2aa6179 100644
--- a/srcpkgs/darktable/template
+++ b/srcpkgs/darktable/template
@@ -9,7 +9,7 @@ maintainer="lemmi <lemmi@nerd2nerd.org>"
 homepage="http://www.darktable.org/"
 license="GPL-3"
 short_desc="Virtual lighttable and darkroom for photographers"
-hostmakedepends="pkg-config intltool"
+hostmakedepends="pkg-config intltool libxslt-devel"
 makedepends="gtk+3-devel glib-devel exiv2-devel libglade-devel libxslt-devel
  dbus-glib-devel libcurl-devel libgphoto2-devel libwebp-devel libsoup-devel
  lensfun-devel sqlite-devel librsvg-devel lua52-devel json-glib-devel
```

configure works on aarch64, but the build then hits the `#include_next <stdlib.h>` issue.

pullmoll:
With cmake packages it seems you can (have to) patch the `flags.make` files where the `-isystem` occurs. I don't know where this comes from, because qt5-5.7.1_2 now doesn't have `-isystem` anymore.

Patching `flags.make` worked to get `kwindowsystem` cross compiled.

lemmi:
oh, i didn't even bother to dig deeper since zlib seemed to have issues on aarch64. 
@pullmoll what changes are that in flags.make specifically?

pullmoll:
@lemmi Take a look at the `srcpkgs/kwindowsystem/template`. There's a `post_configure()` function which patches all `flags.make` files replacing `-isystem` with just `-I`.

PR #5431, opened 2016-12-28
dhcpcd: fix typo/broken syntax for runit systems (dhcpcd-run-hooks)

masato9000:
This bug in dhcpcd upstream was preventing it from restarting services that were configured to use its own internal start-stop functions.  Looks like 50-ntp.conf might have been the only hook affected, so perhaps this is why it's gone unnoticed.

There's a typo and the runit syntax was wrong.

Bug reported [upstream](http://roy.marples.name/projects/dhcpcd/tktview?name=62772337e0), but until we see a fixed release, this ought to do.


masato9000:
Note: Travis XLINT failure - it was like that when I found it; I only updated the package revision.

Issue #5445, opened 2016-12-31
Package request: acpi_call (& tpacpi-bat)


damadmai:
Could you add [acpi_call](https://github.com/mkottman/acpi_call) so that it is not necessary to install it manually after each kernel update?
If it's not too much hassle please also consider adding [tpacpi-bat](https://github.com/teleshoes/tpacpi-bat)
See also http://www.thinkwiki.org/wiki/Tpacpi-bat <- includes [install.pl](https://github.com/teleshoes/tpacpi-bat/blob/master/install.pl) which does the installation of acpi_call.
Happy New Year!

thypon:
https://github.com/voidlinux/void-packages/blob/master/srcpkgs/exfat-dkms/template follow this template and you can contribute the package yourself.
This is an example from archlinux aur repository https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=acpi_call-dkms :)

asergi:
#3724

Issue #3724, opened 2016-02-29
Package request: acpi_call


Varkolac:
https://github.com/mkottman/acpi_call

kernel module for Thinkpad advanced battery features

Last commits (+ release) were years ago.

Vaelatern:
So you know, we do have `tp_smapi-dkms` that should do similar.

Varkolac:
`tlp` recommends installing both. If I understand it correct `acpi-call` adds features for Sandy Bridge or newer processors

Vaelatern:
Appears your assumption is correct, `acpi-call` is necessary for newer hardware.

Vintodrimmer:
Any progress?

damadmai:
As #5452 (which is based on mkottman/acpi_call & all additions to it I have found so far) is now merged, this package request is fullfilled & the issue can be closed.

PR #5455, opened 2017-01-03
New package: tpacpi-bat-3.1

damadmai:
See #5445


damadmai:
As acpi_call-dkms is now merged the depency requirements for tpacpi-bat are now fullfilled. See #3724 & #5452

thypon:
perl should be in depends.

damadmai:
you're right, I added it now.

PR #5461, opened 2017-01-04
New package: monero-0.10.1

ananteris:


srcpkgs/monero/template
```diff
@@ -0,0 +1,38 @@
+# Template file for 'monero'
+pkgname=monero
+version=0.10.1
+revision=1
+build_style=cmake
+makedepends="libressl-devel pkg-config boost-devel libldns-devel libunwind"
```
Vaelatern: no, `pkg-config` is a `hostmakedepends`

Vaelatern: `libunwind` and not `libunwind-devel`?

srcpkgs/monero/template
```diff
@@ -0,0 +1,38 @@
...
+maintainer="ananteris <ananteris@protonmail.ch>"
+license="Proprietary license"
+homepage="https://monero.org"
+distfiles="https://github.com/${pkgname}-project/${pkgname}/archive/v${version}.tar.gz"
+checksum=7fec47217c5515d5fe943dc04854cd3157f09da64e98d3ccdf3c19c7da5cf8d1
+repository=nonfree
+
+case "${XBPS_TARGET_MACHINE}" in
+	*-aarch64)
+			broken="https://travis-ci.org/voidlinux/void-packages/jobs/189296294" ;;
```
Vaelatern: Wrong.

srcpkgs/monero/template
```diff
@@ -0,0 +1,38 @@
...
+	*-armv7hf)
+			broken="https://travis-ci.org/voidlinux/void-packages/jobs/189296294" ;;
```
Vaelatern: Wrong

srcpkgs/monero/template
```diff
@@ -0,0 +1,38 @@
...
+	*-armv7hf-musl)
+			broken="https://travis-ci.org/voidlinux/void-packages/jobs/189296299" ;;
+esac
```
Vaelatern: Just remove the whole case statement. It's because `pkg-config` needs to be run, and makes no sense as a `makedepends`

srcpkgs/monero/template
```diff
@@ -0,0 +1,38 @@
...
+
+do_install() {
+	# these fail to get install automatically
```
Vaelatern: Really?
Try a post_install and see what the makefile actually produces. By overriding `do_install` you don't give it a chance

ananteris: using a cmake-build style do_install does NOT work, it is better to have a explicit do_install than and manually declare vbins and remove extraneous miniupnpc files that are included by default than having a post_install which manages only the delta.

srcpkgs/monero/template
```diff
@@ -0,0 +1,28 @@
+# Template file for 'monero'
+pkgname=monero
+version=0.10.1
+revision=1
+build_style=cmake
+hostmakedepends="pkg-config"
+makedepends="libressl-devel boost-devel libldns-devel libunwind-devel"
+depends="libunwind"
```
Vaelatern: https://travis-ci.org/voidlinux/void-packages/jobs/191756748#L1289
Remove this line

ananteris:
This does not compile on the other archs and has been stale for a month now.

Duncaen:
Can't you fix the other architectures by passsing `ARCH` to cmake? https://github.com/monero-project/monero/blob/master/CMakeLists.txt#L54
i686 should work, just a minor issue https://travis-ci.org/voidlinux/void-packages/jobs/199830582#L1068.

asergi:
637c8a1cd5d0bd9eda5489b5b7f5561accf04a77

PR #5472, opened 2017-01-05
openimageio: update to 1.7.10.

lemmi:


srcpkgs/openimageio/template
```diff
@@ -17,7 +17,7 @@ maintainer="lemmi <lemmi@nerd2nerd.org>"
 license="BSD"
 homepage="https://sites.google.com/site/openimageio/home"
 distfiles="https://github.com/OpenImageIO/oiio/archive/Release-${version}.tar.gz"
-checksum=c186f5662b6c1e75c89ae7d97e91b342adc8dfaf694402f9954f7f288e79457e
+checksum=ec5c567b6b4078d17df22262b764ebb7c09ce5b69f0bc2095b51e56c7070f5b7
 nocross=yes # -> ilmbase
```
ebfe: ilmbase is cross compiled now for a while.

lemmi:
i think this should work. aarch64's only trouble seems to be that a lot of signatures are wonky.

PR #5478, opened 2017-01-06
New package: pychess-0.12.4

cipr3s:


Vaelatern:
Can't help but think this should be `python2-module`, and `pycompile_module="pychess"`

cipr3s:
As agreed yesterday and since pychess can use python2 or python3 indifferently, i updated the template to make it use python3 and the matching dependencies.

pullmoll:
There are no target specific binaries in this pkg, so it again can be `noarch=yes`.

cipr3s:
Updated.

pullmoll:
I don't know too much about build style python{2,3}-module, but doesn't this still require the `pycompile_module="pychess"` that @Vaelatern mentioned? The module is what needs to be byte-compiled when the package is installed: https://github.com/voidlinux/void-packages/blob/master/Manual.md#pkgs_python

cipr3s:
pycompile_module="pychess" wouldn't do anything since the pychess module is inside the lib/ directory. I updated the template to use pycompile_module="lib/pychess" but it's not strictly required like in a "pure-library" case, as when you first run pychess, python will byte-compile the modules anyway.

cipr3s:
Thanks again for reviewing this, i think it's good to merge as it is but i'm happy to make any changes.

PR #5481, opened 2017-01-07
racket: install zsh completion

okapia:
Would be nice to have the zsh completion function in a directory where zsh will find it.


Vaelatern:
`[mmv: cannot stat '/destdir//racket-6.7/share/racket/pkgs/shell-completion/racket-completion.zsh': No such file or directory`

okapia:
My computer has too little memory to actually build the package so I'm giving up on trying to work out how to actually fix this properly.

PR #5484, opened 2017-01-07
New package: m17n-db-1.7.0

hyonhyoro:


srcpkgs/m17n-db/template
```diff
@@ -0,0 +1,23 @@
+# Template file for 'm17n-db'
+pkgname=m17n-db
+version=1.7.0
+revision=1
+noarch=yes
+only_for_archs="aarch64 armv7hf i686 x86_64"
```
Vaelatern: why?

Vaelatern: No, really, why? Especially having both `only_for_archs` and `noarch`. Why?

srcpkgs/m17n-db/template
```diff
@@ -0,0 +1,23 @@
...
+homepage="http://www.nongnu.org/m17n/"
+distfiles="${NONGNU_SITE}/m17n/${pkgname}-${version}.tar.gz"
+checksum=a2ba9f80161433d5c06e57915a9cd51f26c6df4a8909723f952cdbb9b48508f0
+
+m17n-db-devel_package() {
+	noarch=yes
+	depends="${makedepends} ${sourcepkg}>=${version}_${revision}"
+	short_desc+=" - development files"
+	pkg_install() {
+		vmove usr/bin
```
Vaelatern: What is this program for?

srcpkgs/m17n-db/template
```diff
@@ -0,0 +1,23 @@
+# Template file for 'm17n-db'
+pkgname=m17n-db
+version=1.7.0
+revision=1
+noarch=yes
+only_for_archs="aarch64 armv7hf i686 x86_64"
+build_style=gnu-configure
+short_desc="Multilingual text processing library - database"
```
Vaelatern: Remove the `-` here

srcpkgs/m17n-db/template
Vaelatern: Probably want only the `noarch` line

hyonhyoro:
The m17n-db is needed to build the uim, a multilingual input method framework, with the m17n-lib. Ordinarily, it has noarch dependency but makedepends on glibc-locales. I put a limit on *-musl because I didn't know whether I could used them.

asergi:
ab6d6be9931c9c5c854bfcaa181ba353a9831b75 (#7587)

PR #5485, opened 2017-01-07
New package: reaver-1.4

cipr3s:


srcpkgs/reaver/template
```diff
@@ -0,0 +1,29 @@
...
+makedepends="libpcap-devel sqlite-devel"
+short_desc="Brute force attack tool against Wifi Protected Setup (WPS)"
+maintainer="cipr3s <cipr3s@gmx.com>"
+license="GPL-2"
+homepage="https://code.google.com/archive/p/reaver-wps/"
+distfiles="https://storage.googleapis.com/google-code-archive-downloads/v2/code.google.com/reaver-wps/${pkgname}-${version}.tar.gz"
+checksum="add3050a4a05fe0ab6bfb291ee2de8e9b8a85f1e64ced93ee27a75744954b22d"
+
+do_configure() {
+	cd src; ./configure ${configure_args}
```
pullmoll: Please use `build_wrksrc="src"` and remove the `do_configure()` and `do_build()` functions instead.

srcpkgs/reaver/template
```diff
@@ -0,0 +1,29 @@
...
+do_build() {
+	cd src; make ${makejobs}
+}
+
+do_install() {
+	# The Makefile's install target is broken (ignores destdir)
+	vmkdir etc/${pkgname}
+	vmkdir usr/bin
+	vinstall src/${pkgname}.db 644 etc/${pkgname}
+	vcopy src/wash usr/bin
```
pullmoll: Use `vbin wash`. With `build_wrksrc="src"` the script is executed while in `./src`.
You can also remove the `vmkdir usr/bin`.

srcpkgs/reaver/template
```diff
@@ -0,0 +1,29 @@
...
+	vcopy src/${pkgname} usr/bin
```
pullmoll: Use `vbin reaver`.

cipr3s:
Thanks, i updated the template.

pullmoll:
Fixing the cross builds requires a patch that uses the provided `$LD` to link. It looks like it's trying to link the e.g. aarch64 object files using the x86_64 `ld`.

Perhaps you can change the patch to make use of `$(LD)` or just `$(CC)` as for linking as well. That should fix cross builds.

cipr3s:
I've looked into the problem and i now know how to fix it. When creating "iwlib.a" in the "lwe" directory, it runs:

ar cru libiw.a iwlib.so
ar: `u' modifier ignored since `D' is the default (see `U')
ranlib libiw.a

It's wong, it uses "ar" (because of AR=ar in lwe/Makefile) instead of the proper arch's ar. I'll fix that and upload a patch. As far as libwps goes, i see in the log (for armv7 in this case):

make[1]: Entering directory '/builddir/reaver-1.4/src/libwps'
armv7l-linux-gnueabihf-gcc -Wall -fPIC libwps.c -c

which seems correct, because i patched libwps/Makefile (see patches/libwps.makefile.patch) not to use CC=gcc

cipr3s:
It now builds on aarch64 and armv7 but on *-musl it "fails" after making libtls.a and leaving src/tls, no apparent error message, any idea ?

pullmoll:
The `*-musl` error is in the log output. Just search for `" error:"`
```
libwps.h:50:32: error: unknown type name 'u_char'
```
You could perhaps add a sed to replace `u_char` with e.g. `unit8_t` or simply `unsigned char` in all `*.h` and `*.c` files in `post_extract()` or `pre_configure()`:
```
    find -type f -exec sed -i "{}" -e"s;u_char;unsigned char;g" \;
```

Or alternatively, you could define a global `CFLAGS="-Du_char=uint8_t"` in the template for `*-musl`:
```
        *-musl) CFLAGS="-Du_char=uint8_t" ;;
esac
```
Using `uint8_t` of course assumes that `#include <stdint.h>` is there in all affected files.

pullmoll:
And BTW you should not use `CFLAGS=` as well. Either add extra cflags with `CFLAGS+=` or leave it to the build system to define the required CFLAGS. We have hardening, PIC/PIE etc. in the default `CFLAGS`.

pullmoll:
I forgot one detail. I think `/etc/reaver` can be modifed by the user and so there should be a line `conf_files="/etc/reaver"` or `conf_files="/etc/${pkgname}"`.

cipr3s:
Thanks for your input! Just updated the package, it builds on all arches i believe. I removed all CFLAGS= declarations in the makefiles, and used the sed replace only on "*-musl". Just added the conf_files also.

PR #5486, opened 2017-01-07
New package: scapy-2.3.3

cipr3s:


pullmoll:
I think you can add `noarch=yes` for scapy and perhaps also python-pyx. There are no architecture specific binaries in the packages, no?

cipr3s:
Updated with noarch=yes on both packages.

On a side note, scapy has various optional features that are enabled by the use of 3rd-party programs, this is documented in the file doc/scapy/installation.rst. I just added a dependency on python-pyx for PDF rendering and sox for VOIP decryption but there are others like graphviz, imagemagick... I didn't want to overload the package with too many dependencies and they can be installed manually if needed...

PR #5493, opened 2017-01-09
New package: dsniff-2.4b1

cipr3s:
PR for libnids-1.24 and dsniff-2.4b1

Most of the patches for dsniff come from the debian's dsniff package, except for 09_sshcrypto.libressl.patch and 25_void.patch


cipr3s:
Right now only builds on x86_64. I believe it doesn't build on i686 because in libnids, src/checksum.c uses #if ( __i386__ || __i386 ) and some functions (ip_compute_csum, my_tcp_check, ...) end up not being included in the .so but i'm not sure.

dsniff requires an RPC library and there's no native RPC support on musl, i tried using libtirpc but it still doesn't find the includes probably because dsniff's configure doesn't use pkg-config.. It would probably be quite some work to make it build on musl.

PR #5498, opened 2017-01-10
Libgit2 0.25

Gottox:


chneukirchen:
I'd propose to mark libgit2-glib as broken (or remove it), then rebump. Else +1.

ebfe:
8a69c0f556795d51094707946a9335bbaf2f1cf0

PR #5501, opened 2017-01-10
libressl: add patch for CVE-2016-7056

Duncaen:
CVE-2016-7056 ECDSA P-256 timing attack key recovery
http://seclists.org/oss-sec/2017/q1/52


Gottox:
#yolo

PR #5503, opened 2017-01-10
New package: python-chess-0.16.1

cipr3s:


asergi:
You can remove all about `python3.4`.

PR #5504, opened 2017-01-10
RFC: steam-native-runtime

thypon:
The package is not completed but it only needs the libappindicator-gtk2 to made
steam work on voidlinux open source drivers (mesa).

Working vulkan library is also a nice addition, but it's currently unneded.


srcpkgs/curl/template
```diff
@@ -22,7 +22,7 @@ distfiles="$homepage/download/$pkgname-$version.tar.bz2"
 checksum=d16185a767cb2c1ba3d5b9096ec54e5ec198b213f45864a38b3bda4bbf87389b
 
 build_options="gnutls ldap rtmp ssh ssl"
-build_options_default="ssh ssl"
+build_options_default="ssh gnutls" # use gnutls by default to avoid conflicts with steam runtime
```
ebfe: :-1:

thypon: this one or we have to have two different libcurl implementations :(

Vaelatern:
We now have libappindicator-gtk3 as of 1a095b7 , is that alright for this?
If we need to, we can also build it for gtk2.

thypon:
I think we also need the gtk2 version

Vaelatern:
Oh alright.

PR #5507, opened 2017-01-11
GNUpod and its dependencies

evandeaubl:


evandeaubl:
Weird, cross build works for me (at least armv7hf).

PR #5518, opened 2017-01-12
New package: btcd-0.12.0

ananteris:


Vaelatern:
Please look at other examples of go templates.
`prometheus` is a good one.
`acmetool` is also a good example.

ananteris:
Is there a specific problem with this template that you are referring to?

ananteris:
This PR is over a month stale. What is the hold up?

ananteris:
@Vaelatern This package is also stale still. I fail to see any noticeable difference in style/quality from Prometheus or acmetool

ebfe:
At a glance: It doesn't use the release tarball, version claims it's 0.12.0 but there is no 0.12.0 only 0.12.0-beta, go_get fetches master...

PR #5521, opened 2017-01-12
New package: glide-0.12.3

ananteris:


srcpkgs/glide/template
```diff
@@ -0,0 +1,17 @@
+# Template file for 'glide'
+pkgname=glide
+version=0.12.3
+revision=2
```
Vaelatern: `=1` not `=2`

srcpkgs/glide/template
```diff
@@ -0,0 +1,18 @@
+# Template file for 'glide'
+pkgname=glide
+version=0.12.3
+revision=
```
Vaelatern: =1 though

PR #5528, opened 2017-01-12
New package: zcash-1.0.4

ananteris:


srcpkgs/zcash/template
```diff
@@ -0,0 +1,27 @@
+# Template file for 'zcash'
+pkgname=zcash
+version=1.0.4
+revision=1
+makedepends="glibc boost"
```
Vaelatern: Never depend on a libc.

srcpkgs/zcash/template
```diff
@@ -0,0 +1,27 @@
...
+hostmakedepends="zlib-devel libgomp-devel wget autoconf perl m4 autoconf automake libtool pkg-config python"
```
Vaelatern: PLEASE read the Manual sections about `hostmakedepends` vs. `makdepends`.

ananteris:
m4 has been added to hostmakedepends. Are there any other outstanding issues?

Duncaen:
This downloads a whole bunch of dependencies at compile time and builds them, can we use the system wide ones instead?

ananteris:
No, you cannot. Removing any single one will break the install. That is why they are in there. Curiously enough with the most recent version of void-packages even more hostdepends (autoconf, m4, automake, libtool) are needed.

Vaelatern:
Does the build try to use its own tools? Or does it use ours?

ananteris:
It uses exactly the tools the developers of zcash intended for it to. I am not aware of any script that is pulling its own version of m4/automake/autoconf or libtool but if it did there would likely be a reason for it. At the same time, this are listed as hostmakedepends for literally thousands of other programs in the repo (in the case of automake), why is my package held to a different standard than any of them?

Duncaen:
> At the same time, there are literally thousands of other programs in the repo (in the case of automake) that list these as hostmakedepends. Why is my package held to a different standard than any of them?

I don't see the problem, what do you mean by different standards?

Can you at least disable the ccache that it downloads?
And maybe instead of fetching the archives at build time, provide them as `distfiles` to be able to cache them?

Vaelatern:
This is now obsolete, several newer versions of zcash have been released.

PR #5533, opened 2017-01-13
New package: OpenRCT2-0.0.5

lemmi:


srcpkgs/OpenRCT2/template
```diff
@@ -0,0 +1,27 @@
+# Template file for 'OpenRCT2'
+pkgname=OpenRCT2
+version=0.0.5
+revision=1
+nopie=yes
```
Vaelatern: Why?

srcpkgs/OpenRCT2/template
```diff
@@ -0,0 +1,27 @@
...
+hostmakedepends="pkg-config"
+makedepends="SDL2-devel SDL2_ttf-devel jansson-devel libcurl-devel
+ speexdsp-devel fontconfig-devel libzip-devel"
+short_desc="OpenRCT2 recreates RollerCoaster Tycoon 2 into opensource code"
+maintainer="lemmi <lemmi@nerd2nerd.org>"
+license="GPL-3"
+homepage="https://openrct2.com"
+distfiles="https://github.com/OpenRCT2/OpenRCT2/archive/v${version}.tar.gz"
+checksum=68a88f9fd6e818c69d8e30cbfb30f2e900b22bcda3b912f0f0907e95102999df
+only_for_archs="i686 i686-musl x86_64 x86_64-musl"
```
Vaelatern: Please no

srcpkgs/OpenRCT2/template
```diff
@@ -0,0 +1,27 @@
+# Template file for 'OpenRCT2'
+pkgname=OpenRCT2
+version=0.0.5
+revision=1
+nopie=yes
+build_style=cmake
+hostmakedepends="pkg-config"
+makedepends="SDL2-devel SDL2_ttf-devel jansson-devel libcurl-devel
```
Vaelatern:
```
$ xbps-query -Rx SDL2_ttf-devel       
freetype-devel>=0
MesaLib-devel>=0
SDL2-devel>=0
libSM-devel>=0
SDL2_ttf>=2.0.14_1
```
remove `SDL2-devel`

srcpkgs/OpenRCT2/template
lemmi: Earlier versions needed bits from the original binary, which isn't the case anymore. Seems to with PIE now.
An install message is a good idea.

srcpkgs/OpenRCT2/template
lemmi: What are you referring to?

lemmi:
@Vaelatern please have a look again. I added a -data package. Musl is a bit of a mess because of fts unfortunately, but I'm open to suggestions.

lemmi:
ok, next version is out and i need to rework the template for that anyhow. until i'm getting to that i'll close the pull request in the meantime.

PR #5535, opened 2017-01-13
inkscape: update to 0.92.0.

o1ux:


srcpkgs/inkscape/template
```diff
@@ -1,27 +1,28 @@
...
 pkgname=inkscape
-version=0.91
-revision=21
+version=0.92.0
+revision=1
 build_style=gnu-configure
 configure_args="--enable-lcms --enable-poppler-cairo
  --without-gnome-vfs --disable-static"
-hostmakedepends="pkg-config intltool"
+hostmakedepends="autoconf automake gettext-devel glib-devel pkg-config
```
Vaelatern: Remove `autoconf`

Vaelatern: And the `-devel` packages

PR #5552, opened 2017-01-15
ccnet: update to 6.0.2

yopito:
and also `seafile-libclient` to 6.0.2.  
Runtime of  `seafile-client-qt` is fine with these updates.


yopito:
will open another PR instead with seafile-client-qt update to 6.0.2 too.

PR #5553, opened 2017-01-15
New package: flatpak-0.8.0

t0b1nux:
Hello,
this is a PR adding two packages: flatpak and ostree (needed to get flatpak to compile and work).
N.B: sorry for the  successive pushes, which triggered rebuilds, but I omitted to push the local changes before submitting the PR.


t0b1nux:
Sorry, there is unexpected failures, I will work on this, test more thoughtfully, and come back when I will have found a solution.

PR #5554, opened 2017-01-15
New package: vimpager-2.06

ivierlin:
How do I tell xbps-src this package depends on either vim or neovim?

Should this provide an alternative to less or should it be the users job to alias it?

Thanks!


Vaelatern:
Providing an alternative seems reasonable.

I don't recall if we have real virtual packages yet.

Duncaen:
I wouldn't link it to less, you should just set `$PAGER` if you want to use it by default.

ivierlin:
So should I leave it like that?

It's kind of obvious `vimpager` needs some form of `vim`. On the other hand resolving dependencies is half the fun in using a package manager at all.

pullmoll:
You can just add an `INSTALL.msg` file telling the user to install either `vim` or `neovim`, too.

Anachron:
I think an "dependsonany='vim neovim ...'" would not be a bad idea after all.

If someone uninstalls vim the pkg-manager could then pick up the useless parent-pkg if none dependsonany are remaining.

This could be done to many more packages.

ivierlin:
Or we simply use `build_options` for that? There is a `$(vopt_if <option> <pkg>)` parser anyway.
In this case set `build_options_default` to `vim` and provide `neovim` as option.

That's the beauty of a binary/source model. Provide sensible default binaries and make it easy to change them.

Anachron:
Not a bad idea. XBPS would still not realize that vimpager is useless without vim/neovim though, right?

ivierlin:
It would. Have a look in the template I just pushed. Should be self explaining.

The package now depends on vim unless you rebuild it with `-o neovim`.
I think that's the most sane option?

Anachron:
Ah, now that is a good idea! I somehow didn't realize you could just `vopt` on `depends` too.

Duncaen:
I think its totally useless. Now a user needs to rebuild the complete package to not depend on vim if he uses neovim.
Just remove it and let the user find out that he needs to install vim, if its not already installed.

PR #5560, opened 2017-01-16
kcov: update to v33

NoctuaNivalis:


NoctuaNivalis:
Related AArch64 issue: SimonKagstrom/kcov#178

PR #5562, opened 2017-01-16
Oce 0.18 and freecad

yopito:
Travis CI will probably fails since long build and huge build log.  
These updates are fine @home (runtime on x86_64/glibc).


Vaelatern:
Please add a related `broken` for `aarch64*`

Vaelatern:
running an x86_64 native build on my own gear.

PR #5563, opened 2017-01-16
New package: lemonbar-xft-1.0.20170116

ivierlin:
Not a tagged release.

### DIFF
--- a/srcpkgs/lemonbar-xft/template
+++ b/srcpkgs/lemonbar-xft/template
@@ -1,0 +1,24 @@
+# Template file for 'lemonbar-xft'
+pkgname=lemonbar-xft
+version=1.0.20161107
+revision=1
+_commit=043ad4757cc079666f50212ee0a2ef0729ecac6b
+build_style=gnu-makefile
+wrksrc="bar-${_commit}"
+hostmakedepends="perl libXft-devel libxcb-devel"
+makedepends="libxcb-devel libXft-devel"
+short_desc="A lightweight xcb based bar with xft support"
+maintainer='ivierlin <github@ivierlin.de>'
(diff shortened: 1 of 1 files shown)


Anachron:
There seems one file to be missing:
https://travis-ci.org/voidlinux/void-packages/jobs/192615499
/usr/aarch64-linux-gnu/usr/include/X11/Xft/Xft.h:39:22: fatal error: ft2build.h: No such file or directory

pullmoll:
For cross builds you may need the `libXft-devel` and/or `libxcb-devel` in the host. Or you'd have to patch the `Makefile` for cross builds to `-I${XBPS_CROSS_BASE}/usr/include/<whatever>`.

For `gnu-configure` packages this is handled by a special `pkg-config` wrapper which prints these paths with `${XBPS_CROSS_BASE}`, where the target dependencies are, prepended.

srcpkgs/lemonbar-xft/template
```diff
@@ -0,0 +1,24 @@
+# Template file for 'lemonbar-xft'
+pkgname=lemonbar-xft
+version=20161107
```
pullmoll: Since there is a 1.0 release, I'd suggest to give this one the version `1.0.20161107`, just in case a `1.1` is released which otherwise would be numerically less than your current `20161107`.

ivierlin: I am patching out `1.1` in the Makefile. Maybe we can just use that instead of the date?
Just thought it would be bad style since there is no release on github yet.

pullmoll: If you use `1.1` now, you can't update to `1.1` when it is official :)
Thus my suggestion to use `1.0.<datestamp>` instead.

PR #5564, opened 2017-01-17
New package: sauerbraten-2013.02.03

the-maldridge:


srcpkgs/sauerbraten/template
```diff
@@ -0,0 +1,50 @@
...
+		vcopy $wrksrc/data usr/share/$sourcepkg/data
+		vcopy $wrksrc/packages usr/share/$sourcepkg/packages
+	}
+}
+
+sauerbraten-server_package() {
+	pkg_install() {
+		vbin $wrksrc/bin_unix/native_server sauer_server
+		vmkdir etc/$sourcepkg
+		vcopy $wrksrc/server-init.cfg etc/$sourcepkg/server-init.cfg
```
Vaelatern: Shouldn't this be a conf_file? It seems like it is meant to be changed.

PR #5565, opened 2017-01-17
libui: add gnome-themes-standard in depends

gangsterakato:
`gnome-themes-standard` is needed in order to get the icons on widgets used by libui (so by GTK+3), like the little arrow on lists for example (which is blank otherwise).


Vaelatern:
Not sure why not just depend on an icon theme, but then it's a question "Which icon theme."

gangsterakato:
Isn't here an icon theme by default? I'm using LXQt and there is visibly no icon theme installed by default at all, and I got these problems in all GTK apps using some non-custom icons like Audacious or GEdit.

Here is an official screenshot from the libui developer:
https://github.com/andlabs/libui/blob/master/examples/controlgallery/unix.png

So I guess the icon theme which must be used is present in the `gnome-themes-standard` package and is probably the default one too on this screenshot.

Vaelatern:
I'm not sold.
If you want an icon theme, you can install one.

gangsterakato:
Without any theme you don't have any arrow on lists or default square icons on some buttons on LXQt, this is clearly a problem, not really related to libui anyway.

Vaelatern:
Yes, but we don't have a virtual package for "theme for gnome like projects" since there are a lot of options.

gangsterakato:
This [file](https://github.com/voidlinux/void-packages/blob/d19eefb899eb023fc7d8ae28860579f5ba690f56/srcpkgs/gtk%2B/files/gtkrc) indicate that the default choosed GTK+ icon theme is `Adwaita`. I confirm that installing `adwaita-icon-theme` fixes the issue. Maybe this package should be installed by default with GTK+ but shouldn't replace the choosed one by the user?

PR #5573, opened 2017-01-18
rstudio: add update file

ivierlin:
Maybe wait to merge this until there is a new version anyway? It's just a convenience for me not the users.


Vaelatern:
Users won't notice, it's not a revision bump.

PR #5575, opened 2017-01-18
nvidia340: add patch to add DRIVER_LEGACY to nv-drm.c

Hoshpak:
The nvidia340 legacy driver fails to load on the 4.9 kernel.
This commit adds a patch written by ejmarkow, taken from
https://devtalk.nvidia.com/default/topic/982052/linux/latest-nvidia-driver-340-101-builds-compiles-properly-but-fails-to-load-has-errors-with-linux-kernel-4-9-resolved-with-patch-/
that adds DRIVER_LEGACY to the driver features in nv-drm.c.


srcpkgs/nvidia340/template
```diff
@@ -35,6 +35,9 @@ do_extract() {
 	install -m755 ${XBPS_SRCDISTDIR}/${pkgname}-${version}/${_pkg}.run ${wrksrc}
 	cd ${wrksrc}
 	./${_pkg}.run --extract-only
+	cd ${_pkg}
+	patch -p0 < ${FILESDIR}/patch_nv1.diff
```
Vaelatern: Why is this not done as a sub of `patches/` instead of manually running the patch?

Hoshpak: That's because the patch has to be applied to the source which sits in ``_pkg`` which depends on the version and architecture, applying it directly in ``wrksrc`` would not  work because then I would have to put the sub-directory in the patch file and it would only work for one architecture and only until the version number changes.

asergi: Move `patch -p0 < ${FILESDIR}/patch_nv1.diff` to `do_configure()`. See `nvidia304` as example. Thanks.

Hoshpak:
done.

PR #5580, opened 2017-01-19
libGL: make wayland optional

hadrianw:
Add wayland build option to libGL.


hadrianw:
Underlying error: srcpkgs/libGL/template:14: license 'MIT', but no use of vlicense

It should fail regardless of my change.

hadrianw:
I messed up with all this forking business sorry for this. I created a new pull request as it will not update this one.

Here's the new pull request: https://github.com/voidlinux/void-packages/pull/5941

PR #5582, opened 2017-01-20
New package: 9base

raedwulf:
Alternative plan9 userspace port to plan9port from suckless.  Advantage of this version is that it compiles with musl and can cross-compile using bison (I supply a couple of patches to make the YACC grammars compilable with bison).

### DIFF
--- a/srcpkgs/9base/template
+++ b/srcpkgs/9base/template
@@ -1,0 +1,54 @@
+# Template file for '9base'
+pkgname=9base
+version=6.0.20160912
+revision=1
+_githash=de27e3fba58c3dd467c29ea9a1aea6f9e65a5299
+build_style=gnu-makefile
+# we use byacc instead of internal version
+hostmakedepends="byacc"
+short_desc="Revived minimalist port of Plan 9 userland to Unix"
+maintainer='Tai Chi Minh Ralph Eastwood <tcmreastwood@gmail.com>'
+license="custom"
+homepage="http://git.suckless.org/9base"
+distfiles="http://git.suckless.org/9base/snapshot/9base-${_githash}.tar.bz2"
+provides="plan9port-20160418_4"
+conflicts="plan9port"
+checksum=a6eda0673ee38fb9eb48921a7d6f0cfaa2012020dde81d10d7be5980e2af4e32
+wrksrc="9base-${_githash}"
+
+pre_build() {
+	# These are platform specific, but aarch64 and mips have their
+	# program counter in the same place; we'll just use the ARM
+	# implementation.
+	cp lib9/getcallerpc-arm.c lib9/getcallerpc-aarch64.c
+	cp lib9/getcallerpc-arm.c lib9/getcallerpc-mips.c
+
+	# config.mk has conflicting flags to the build-style; we enforce
+	# specific ones in config.mk to make sure they don't get changed
+	# in do_build.
+	sed -i 's|^PREFIX \+= /usr/local/plan9|override PREFIX = /usr/lib/plan9|g' config.mk
+	sed -i 's|^CFLAGS \++=|override CFLAGS += -D_DEFAULT_SOURCE |g' config.mk
+	sed -i 's|^LDFLAGS.*||g' config.mk
+	sed -i 's|^CC.*||g' config.mk
+	# Add rc to the AR flag as the 9base build-scripts require AR to
+	# be declared with the 'rc' argument.  However, AR can be
+	# different to 'ar' if declared differently within xbps e.g.
+	# gcc-ar which is necessary for link-time optimisation builds.
+	sed -i 's|^AR.*|override AR += rc|g' config.mk
+	sed -i 's|^YACC \+=.*|YACC = byacc|g' config.mk
+
+	# This allows 9base's build-style know which getcallerpc
+	# implementation to use.
+	case "$XBPS_TARGET_MACHINE" in
+	i686*)    sed -i 's|^OBJTYPE \+=.*|OBJTYPE = 386|g' config.mk ;;
+	x86_64*)  sed -i 's|^OBJTYPE \+=.*|OBJTYPE = x86_64|g' config.mk ;;
+	arm*)     sed -i 's|^OBJTYPE \+=.*|OBJTYPE = arm|g' config.mk ;;
+	aarch64*) sed -i 's|^OBJTYPE \+=.*|OBJTYPE = aarch64|g' config.mk ;;
+	mips*)    sed -i 's|^OBJTYPE \+=.*|OBJTYPE = mips|g' config.mk ;;
+	esac
+}
+
+post_install() {
+	vinstall $FILESDIR/plan9.sh 644 etc/profile.d
+	vlicense LICENSE
+}
--- a/srcpkgs/9base/files/plan9.sh
+++ b/srcpkgs/9base/files/plan9.sh
@@ -1,0 +1,2 @@
+export PLAN9=/usr/lib/plan9
+export PATH=$PATH:$PLAN9/bin


Vaelatern:
Can it cross compile using yacc?

raedwulf:
Technically - yes if I make a bootstrap package from yacc because it needs to be installed on the host

Vaelatern:
We  have byacc packaged, just add that to hostmakedepends?

raedwulf:
Ah I just noticed that! It seems like it should work - I'll test it.

raedwulf:
Seems to work! Thanks for the byacc suggestion!

srcpkgs/9base/files/getcallerpc-mips.c
```diff
@@ -0,0 +1,8 @@
+#include <lib9.h>
+
+ulong
+getcallerpc(void *x)
+{
+	return ((ulong*)x)[-2];
```
Vaelatern: Can you please document exactly why these files are provided?

Vaelatern:
Now that is a lot less to maintain :)

srcpkgs/9base/files/getcallerpc-mips.c
raedwulf: They get platform-specific program counter from the stack.  The contents are identical across arm, mips and aarch64 because that's where they put their program counter.

http://git.suckless.org/9base/tree/lib9/getcallerpc-arm.c

plan9port does the same - 9base just doesn't have any platform-specific files for aarch64 and mips.

raedwulf: Should I add a comment for this in the template?

raedwulf: Another reference: https://swtch.com/plan9port/man/man3/getcallerpc.html

Vaelatern: A comment would be welcome.
Could you just copy the files for arm to mips and aarch64?

srcpkgs/9base/template
```diff
@@ -0,0 +1,58 @@
+# Template file for '9base'
+pkgname=9base
+version=7.0.20160912
+revision=1
+_githash=de27e3fba58c3dd467c29ea9a1aea6f9e65a5299
+build_style=gnu-makefile
+if [ "$CROSS_BUILD" ]; then
+	hostmakedepends="byacc"
```
Vaelatern: Why is this not a hostmakedepends for all builds?

srcpkgs/9base/template
```diff
@@ -0,0 +1,58 @@
+# Template file for '9base'
+pkgname=9base
+version=7.0.20160912
```
Vaelatern: can you use a version starting with 6? Just in case 7.0 exists in the future

srcpkgs/9base/template
```diff
@@ -0,0 +1,58 @@
...
+	arm*)
+		sed -i 's|^OBJTYPE \+=.*|OBJTYPE = arm|g' config.mk
+		;;
+	aarch64*)
+		sed -i 's|^OBJTYPE \+=.*|OBJTYPE = aarch64|g' config.mk
+		;;
+	mips*)
+		sed -i 's|^OBJTYPE \+=.*|OBJTYPE = mips|g' config.mk
+		;;
+	esac
```
Vaelatern: Please write each case statement as a single line, to make it more readable.

srcpkgs/9base/template
raedwulf: It doesn't need to be because yacc is included in 9base and used by default in normal/non-cross builds.  But changing to byacc would not cause problems either.

srcpkgs/9base/template
```diff
@@ -0,0 +1,58 @@
...
+}
+
+do_build() {
+	make ${makejobs}
+}
```
Vaelatern: Any reason you can't use the makefile build style?

srcpkgs/9base/files/getcallerpc-mips.c
raedwulf: Yes that would work.

srcpkgs/9base/template
raedwulf: The `do_build` passed some additional variable overrides that conflicted with the build files.

raedwulf: I can't quite recall which they were (I'll check shortly)

raedwulf: The build scripts redefine `AR` to be `ar rc`.  I've modified the `config.mk` with sed to make sure that it does not get changed by the build style `do_build`.

srcpkgs/9base/template
raedwulf: Done

srcpkgs/9base/template
raedwulf: Changed

raedwulf:
Added a few comments throughout the template for non-obvious choices.

srcpkgs/9base/template
```diff
@@ -0,0 +1,56 @@
+# Template file for '9base'
+pkgname=9base
+version=6.0.20160912
+revision=1
+_githash=de27e3fba58c3dd467c29ea9a1aea6f9e65a5299
+build_style=gnu-makefile
+# byacc is only necessary in cross-compiles.  On native-builds, the
+# internal version of yacc is used instead.
```
Vaelatern: Internal version? As in, a version distributed with the binary?
For more consistent builds, please just install it unconditionally.

srcpkgs/9base/template
```diff
@@ -0,0 +1,56 @@
...
+	# in do_build.
+	sed -i 's|^PREFIX \+= /usr/local/plan9|override PREFIX = /usr/lib/plan9|g' config.mk
+	sed -i 's|^CFLAGS \++=|override CFLAGS += -D_DEFAULT_SOURCE |g' config.mk
+	sed -i 's|^LDFLAGS \++=|override LDFLAGS +=|g' config.mk
+	sed -i 's|^CC \+=.*||g' config.mk
+	sed -i 's|^AR|override AR|g' config.mk
+	# In cross compiles, use the hosts' byacc package.
+	if [ "$CROSS_BUILD" ]; then
+		sed -i 's|^YACC \+=.*|YACC = byacc|g' config.mk
+	fi
```
Vaelatern: This would fail if the build occurs natively on armv6, correct?

Vaelatern:
Your use of comments is much appreciated, though your commit message is no longer correct (for the correct version).

srcpkgs/9base/template
raedwulf: It should be fine - the yacc that comes with 9base is a relative to byacc, but arguably byacc is better as it has had much more love over the years.

srcpkgs/9base/template
raedwulf: Okay done - after thinking about it, it does make a lot more sense!

srcpkgs/9base/template
```diff
@@ -0,0 +1,54 @@
...
+	# in do_build.
+	sed -i 's|^PREFIX \+= /usr/local/plan9|override PREFIX = /usr/lib/plan9|g' config.mk
+	sed -i 's|^CFLAGS \++=|override CFLAGS += -D_DEFAULT_SOURCE |g' config.mk
+	sed -i 's|^LDFLAGS \++=|override LDFLAGS +=|g' config.mk
+	sed -i 's|^CC \+=.*||g' config.mk
+	# Add rc to the AR flag as the 9base build-scripts require AR to
+	# be declared with the 'rc' argument.  However, AR can be
+	# different to 'ar' if declared differently within xbps e.g.
+	# gcc-ar which is necessary for link-time optimisation builds.
+	sed -i 's|^AR.*|override AR += rc|g' config.mk
```
raedwulf: Normally, `AR = ar`.  9base wants the declaration to be `AR = ar rc`.
With gcc link-time optimisation, it needs to be `AR = gcc-ar`.  So, the simplest way to incorporate both requirements is to append the extra argument `rc` to whatever `xbps` declares `AR` to be in the environment.

Vaelatern:
So this creates static executables only, bypassing any conversation of if it uses our CFLAGS.
This incidentally means that most libc rebuilds would mean we would want to rebuild this as well. I don't think we want that.

raedwulf:
Unfortunately that's true - I can try and see if I can build it non-statically.

raedwulf:
I dropped the LDFLAGS from config.mk and it builds non-static binaries

Vaelatern:
I'll give this a spin

raedwulf:
Thanks! Let me know if there's any further issues.

PR #5586, opened 2017-01-20
New package: kdenlive-17.08.0

jnbr:


jnbr:
- musl patch for `mlt` is broken
- building `krash` fails in do_configure on arm

Vaelatern:
Okay, you have 3 things in one PR, making it hard to review. For example, I'd love to review and merge the ffmpeg commit, the mlt commit seems to add a lot of patches that do nothing for us, and I didn't even see the kdenlive template in my scan of the changes.

jnbr:
I split the PR and moved the change in ffmpeg to  #5588.
The mlt commit does  not add a lot of patches, in fact it removes one patch which is not needed anymore (`mlt-0.9.8-ffmpeg3.patch`) and updates the other patch to the new mlt version.

srcpkgs/kdenlive/template
```diff
@@ -0,0 +1,23 @@
...
+depends="breeze-icons frei0r-plugins kinit ffmpeg dvdauthor"
+short_desc="Non-linear video editor"
+maintainer="johannes <johannes.brechtmann@gmail.com>"
+license="GPL-2"
+homepage="https://kdenlive.org"
+
+do_fetch() {
+	local _url="git://anongit.kde.org/kdenlive.git"
+	git clone ${_url} ${wrksrc} -b v${version}
+}
```
Vaelatern: I don't like using `git clone` to get tarballs, it prevents any checksum to confirm there was no MitM or upstream quietly changing a branch/tag.
But it doesn't seem like kde wants to have tarballs of their git master.

Vaelatern: https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/kdenlive Turns out there is a way to get a tarball.

pullmoll:
You can get the tarball here: http://download.kde.org/stable/applications/16.12.1/src/kdenlive-16.12.1.tar.xz

jnbr:
Thank you @Vaelatern and @pullmoll for the hint to the tarball.

jnbr:
I added `pulseaudio-devel` to `makedepends` for `mlt`. mlt is now building on all platforms.
Still issues with musl and arm for `kdenlive`

jnbr:
I set kdenlive to `nocross`.
At least `kdoctools` and possible other dependencies can not be cross compiled.

jnbr:
I can not reproduce the error with x86_64-musl failing to find `pkg-config` in `configure` of kdenlive.
For now, I disabled musl and arm.

jnbr:
Found the problem: some dependencies (`kinit`, `knewstuff` and `knotifyconfig`) could not be simply installed but have to be built for musl. Building them happens after installing `pkg-config` for `kdenlive`.
After they are built and packaged, `pkg-config` gets removed and is not available when configuring kdenlive.

This is now avoided by revbumps for the packages (and setting nocross=yes, since this does not work at the moment).

dominikh:
Needs to install qt5-quickcontrols, else it segfaults on startup.

dominikh:
You'll also want to backport https://cgit.kde.org/kdenlive.git/commit/?id=8a20fca86e9e2ce3d04b3e024b0752e19f9d8c8e and https://cgit.kde.org/kdenlive.git/commit/?id=754b8eeed8c1a874a5535eb5136b02841a812180 or else kdenlive 16.12.1 won't start at all for users with nvidia cards.

jnbr:
Thank you, @dominikh !
I added `qt5-quickcontrols` to the dependencies and included the patches.

Sir-Boops:
Bump, Current version of kdenlive is now 16.12.3

jnbr:
Bumped version to 16.12.3 and removed backported patches.

Sir-Boops:
:+1:

Sir-Boops:
Current version is now 17.04.0

jnbr:
I updated to 17.04.0.
This version uses more compile flags. The result is a massive amount of warnings during compilation, exceeding the maximum log file size for Travis-Ci, so the test fails.
Local builds work.

jnbr:
Updated to 17.04.3.
Cross compilation is now possible, changes on KDE packages are not necessary anymore.
Travis-CI builds still fail because of to many warnings, local builds work (tested with armv7hf and x86).

jnbr:
updated to 17.08.
`xlocale.h` was removed from glibc, so a patch for mlt was needed to use `locale.h` instead.

newbluemoon:
I was trying to make a template for Shotcut (#6005) which also requires the MLT framework and started to look into updating it. Then I found out you already did it :) I already created templates for the missing dependencies ‘movit’ and ‘vid.stab’. I found that vid.stab is already in the repo as ‘libvidstab’ (I opened a PR to update it, #7483). For movit I created PR #7484. Maybe you can add it to your mlt template (if/when accepted)?

Is it possible to detach the update of MLT from kdenlive? Maybe that way it gets accepted sooner and would be available for other packages (like Shotcut)?

jnbr:
@newbluemoon of course I can add `vid.stab` and `movit` to MLT as soon as your PRs are merged.

@Vaelatern, @pullmoll can one of you review this PR again and eventually merge it? Does it make sense to split this PR? I avoided it because kdenlive doesn't build without the updated version of MLT.

newbluemoon:
@jnbr just wanted to let you know that `movit` is now in the repository. (Just in case you didn’t notice.) :-)

jnbr:
mlt is now built with `movit` and `libvidstab`. Also a patch from upstream is now used for the deprecated `xlocale.h` in glibc (Same as #7530).

lemmi:
@jnbr can you rebase? mlt has conflicts atm.

lemmi:
@jnbr almost. If possible try to do it via a rebase to keep a clean history. If you have a kdenlive branch, you can use something like:
```
# git checkout master
# git pull upstream master
# git checkout kdenlive
# git rebase master
```
Git rebase will stop on conflicts to let you fix that commit.
Alternatively, you might be able to salvage this with a
```
# git rebase -i HEAD~4
```

jnbr:
@lemmi Thanks. I know this wasn't how to do it. I'm on vaccation and my laptop died, so I tried to use the github web interface on a tablet. Will fix this as soon as possible.

lemmi:
@jnbr Alright, I'll wait for that then, no worries.

jnbr:
@lemmi fixed

lemmi:
@jnbr kdenlive
Otherwise it seems to work. Musl compiles aswell.

PR #5592, opened 2017-01-24
elasticsearch: update to 5.2.1

ayghor:
EHLO!

could someone review this?

some points you may yell at me for:

 * split into elasticsearch5 instead of updateing elasticsearch because it may break stuff
 * kinda followed teh directory structure from teh debian package

im in a hurry these days but will try to fix asap.

### DIFF
--- a/srcpkgs/elasticsearch/template
+++ b/srcpkgs/elasticsearch/template
@@ -1,41 +1,48 @@
 # Template file for 'elasticsearch'
 pkgname=elasticsearch
-version=1.7.3
-revision=2
+version=5.1.2
+revision=1
+wrksrc="elasticsearch-${version}"
 only_for_archs="i686 x86_64"
 depends="virtual?java-runtime"
 short_desc="A distributed, open source search and analytics engine"
-maintainer="bougyman <bougyman@voidlinux.eu>"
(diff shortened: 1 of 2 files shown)


ebfe:
Unless someone complains i don't think we need to keep the old version.

ebfe:
cc @bougyman

ayghor:
i will apply over teh old pkg then!

ayghor:
wat bout this?

srcpkgs/elasticsearch/template
```diff
@@ -1,41 +1,48 @@
...
-version=1.7.3
-revision=2
+version=5.1.2
+revision=1
+wrksrc="elasticsearch-${version}"
 only_for_archs="i686 x86_64"
 depends="virtual?java-runtime"
 short_desc="A distributed, open source search and analytics engine"
-maintainer="bougyman <bougyman@voidlinux.eu>"
+maintainer="dota? =op <rogi@skylittlesystem.org>"
```
Gottox: I'm excited to see if that will break something😆

ayghor: lint didnt complain its perfectly safe

ayghor: =O3

PR #5593, opened 2017-01-24
pdfchain: fix update check

yopito:
according to http://pdfchain.sourceforge.net/download.html  
bump package is usefull in this case ?


Vaelatern:
No, revision bumps would not be useful here.
The binary package is not changed.

PR #5600, opened 2017-01-24
vapoursynth: update to R36.

lemmi:


srcpkgs/zimg/template
```diff
@@ -3,7 +3,7 @@ pkgname=zimg
 # 2.1 needs a patch to cross compile.
 # Please remove on update
 # https://github.com/sekrit-twc/zimg/issues/47
```
Vaelatern: This comment and the patch can probably be removed now. Else a new issue should be opened.

ebfe:
3acdaa231c72008905c2270e12d6d0079ca9650c

PR #5609, opened 2017-01-25
New package: python-smmap-2.0.1

lafrenierejm:


asergi:
@lafrenierejm https://pypi.python.org/pypi/smmap2
@Vaelatern `yaml`?

lafrenierejm:
@asergi The PyPi package does not include the license file, which xlint requires for BSD. Hence my using the GitHub release.

The `python-yaml` dependency should not be there.  Thank you for catching that.

PR #5614, opened 2017-01-26
thunderbird: fix nss>=3.28

masato9000:
Same issue that firefox-esr had with nss>=3.28 (which upstream fixed in a version bump, which was committed here the other day).  Gmail affected, possibly others.

Since we don't seem to be using the betas (t-bird upstream has the fix in HG, and probably in the last beta releae) here's a backported fix.  Hopefully this is all put together correctly (gentoo took this from upstream, and I **shamelessly** copied them).

The other patch is to correct a build failure on systems with sed>=4.3.


asergi:
[45.7.0](https://www.mozilla.org/en-US/thunderbird/45.7.0/releasenotes/) is out.

masato9000:
@asergi 
aha.  If I had just been a bit more patient 😃 (it definitely wasn't there when I started this 24 hours ago).
Then this ain't needed, as it looks like the new version's already been bumped in void-packages.

I imagine the nss fix is in this version, but I'll double-check before closing this PR.

masato9000:
I didn't download the tarball to confirm, but seems to have been resolved in the package.
Closing.
Thanks @asergi for the heads-up.

PR #5618, opened 2017-01-26
New package: atomicparsley-0.9.0

svenper:
Sorry about closing and re-opening, I misunderstood the instructions.


srcpkgs/atomicparsley/template
```diff
@@ -0,0 +1,28 @@
...
+makedepends="unzip"
+short_desc="Lightweight program for manipulating MPEG-4 metadata"
+maintainer="svenper <svenper@users.noreply.github.com>"
+license="GPL-2"
+homepage="http://atomicparsley.sourceforge.net"
+distfiles="${SOURCEFORGE_SITE}/${pkgname}/AtomicParsley-source-${version}.zip"
+checksum=de83f219f95e6fe59099b277e3ced86f0430ad9468e845783092821dff15a72e
+wrksrc=AtomicParsley-source-${version}
+nopie=yes
+only_for_archs="x86_64 i686"
```
pullmoll: This is probably wrong, unless there is assembler code involved which does not build on `aarch64*`, `arm*` or `mips*`.

Instead `./build` should be convinced to make use of the environment variables `$CC`, `$CFLAGS` and `$LDFLAGS`. Perhaps it's sufficient to add these to the command like
```
        ./build CC="$CC" CFLAGS="$CFLAGS" LDFLAGS="$LDFLAGS"
```
In any case, if it fails to cross compile, that does not necessarily mean it can't be built on e.g. native armv7l.

svenper: The build script is a mess, but I did what I could in commits https://github.com/voidlinux/void-packages/pull/5618/commits/3dedcffc32f9d14149498352408ffae4680ea8ad and https://github.com/voidlinux/void-packages/pull/5618/commits/ac4360392ceb8cf16a9dee6f3dfd096529103c93. Travis still can't get it to build on other archs. Am I missing something or should I leave it as it is?

srcpkgs/atomicparsley/template
```diff
@@ -0,0 +1,28 @@
+# Template file for 'atomicparsley'.
+pkgname=atomicparsley
+version=0.9.0
+revision=1
+makedepends="unzip"
```
pullmoll: Use `hostmakedepends="unzip"`.

pullmoll:
Ok, now looks good to me. You can leave the *-musl errors, I think I can take care of that.
Please squash your commits into one using `git rebase` and `git push -f`.

srcpkgs/atomicparsley/template
```diff
@@ -2,7 +2,7 @@
 pkgname=atomicparsley
 version=0.9.0
 revision=1
-hostmakedepends="unzip"
+hostmakedepends="unzip libgcc"
```
pullmoll: This addition is wrong. The compiler and its libraries are always installed anyway.

pullmoll:
When running `git rebase -i HEAD~6` just drop the `atomicparsley: add libgcc dependency`

pullmoll:
When updating a branch / fork from upstream you can always use `git pull --rebase …` to avoid the merge commit message :smile:

svenper:
Yeah, I'm having some issues figuring this out. I think it's good by now, but my commit includes all the files that changed the last 20 hours.

pullmoll:
Yeah, I remember the time when I struggled with git and github pull requests. Just remember to always use `git pull --rebase …` and get familiar with `git rebase -i`, then your PRs will be merged rather sooner than later.

I got the patches for *-musl done and will push now.

pullmoll:
Hmm.. it seems I got the fpos_t patch wrong. Need to take another look and the struct and the difference between gnu and musl libc.

PR #5621, opened 2017-01-27
jeti-filemanager: update to 2.0.1

mrshampoo:
fast buggfixes.


mrshampoo:
wtf?!?!

mrshampoo:
I dont fuly understand this.
I only fixed a few typos from last time. (and then it worked fine)
I tested also to download and install it on my own computer right now, and it compiled without a problem.
Can someone fill me in, ihave i done something wrong?!?

mrshampoo:
I updated my void (xbps-install -Su) and now I get the same issue on my computer too.
running a normal make; make install still works as it should.
I have tried removing
pre_build() {
        export LIBS="-lncurses $LDFLAGS"
}
same result. I have also removed $LDFLAGS alone. no further result.
I changed back the original makefile to what it was before so this commit must be updated. but I dont see any point in it before I can run it on my own system first...
I'm just so clueless of what to do to make it work. :/

pullmoll:
The shipped Makefile is wrong. Linking stage is missing `$(LDFLAGS)` and `$(LIBS)` should be afte the object files, i.e. after `$^`.

zqqw:
||  cmd[i] == '\''

autocomplete.c
Did you intend 2 single quotes at end?

mrshampoo:
oh, that perhaps needs a \ ? :P

zqqw:
Oh, you are probably escaping the single quote character here. Guess that works then, haven't done that myself so didn't recognise it. It was the only thing that looked odd in the commit I saw.

mrshampoo:
oh, it was the single quote you ment.. now i put a \ on the double quote " ( \" ) no complaints from the compiler there eather... :P 
(I didn't work whitout the \ on the single quote when i tested on a file, but with the slash it did ;) )

mrshampoo:
christ! this is making me insane.. i haven't eaten anything since lunch...
It always is something like this. you think it is going to be easy, just a quik upload aaaand. nothing.

zqqw:
I looked it up and that is the right way to escape a single quote, it was only when I first saw it I thought perhaps it was intended to be a back slash but with an extra quote. I didn't spot any other obvious problems in the diffs. It'll probably be great when you get it sorted.

mrshampoo:
:) well, i think i have to continue tomorrow with this.. (need some sleep and Makefiles is kinda my nemesis)

PR #5625, opened 2017-01-28
kauth: fix cross

jnbr:


srcpkgs/kauth/template
```diff
@@ -12,6 +12,10 @@ homepage="https://projects.kde.org/projects/frameworks/kauth"
 distfiles="http://download.kde.org/stable/frameworks/${version%.*}/${pkgname}-${version}.tar.xz"
 checksum=f2c51dfb42196fe5d604883c9f28bda7fd3f40d308144f521f4090a941c2091d
 
+if [ -n "$CROSS_BUILD" ]; then
+	hostmakedepends+=" qt5-host-tools qt5-qmake qt5-tools-devel polkit-qt5-devel"
+fi
```
Vaelatern: I'm not sold on this being conditional. @pullmoll ?
This would be relevant to #5626 and #5627 as well.

pullmoll: I always made this conditional for build style cmake, yes.
It isn't required for native builds because cmake there finds the required bits and pieces in the host.

PR #5629, opened 2017-01-28
jeti-filemanager: update to 2.0.1

mrshampoo:
there, finally!
dam git has changed too. I cant use anything, had to use git-cola to rewind my head

### DIFF
--- a/srcpkgs/jeti-filemanager/template
+++ b/srcpkgs/jeti-filemanager/template
@@ -9,11 +9,8 @@ short_desc="Total Commander filemanager ncurses clone"
 maintainer="Harri Leino <mr.leino@gmail.com>"
 license="GPL-3"
 distfiles="https://github.com/mrshampoo/jeti-filemanager/archive/${version}.tar.gz"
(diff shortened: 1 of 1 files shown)


mrshampoo:
and thank you again @pullmoll for helping me with my Makefile!

pullmoll:
The package is already at 2.0.1. If you modify the template for 2.0.2, after you changed the Makefile in the archive, please base it on how it is in the xbps repository now, i.e. don't mix `LIBS` and `LDFLAGS` but just set `LIBS` in the main section.

mrshampoo:
what? why is it alredy 2.0.1 in the archive, it didn't pass the Travis test then!?

mrshampoo:
umm okej, it explains the weird changes in the template... it seems to work too.
Then I don't really see the point in making an update for something that will result in the same output.
I would rather track down a bug or something before making a update to 2.0.2

mrshampoo:
(when changing version i need to change the version in include/projecttypes.h too, so the right version number pops up when writing "jeti -v")

PR #5632, opened 2017-01-29
NetworkManager: update to 1.6.2

yopito:
build and runtime on x86_64/glibc of NetworkManager and network-manager-applet is ok.  
Let's see if cross-building is fine too.


Gottox:
Nice! 

Please see my changes to make 1.4 build with musl. Maybe we can merge our effort.

yopito:
thanks, on reworking my PR to take care of your changes.
FYI, I was able to achieve build NetworkManger 1.6.2 with musl libc.
stay tuned.

yopito:
I wish I was ready for review ... but don't understand why Travis CI fails on only one cross-arch and not all of them.
Please notice that doc generation is disabled since cross-builds do not like it.

PR #5641, opened 2017-01-29
mumble: update to 1.2.19

Hoshpak:
additionally adds a patch fixing the include of the speech-dispatcher
headers to enable building with text to speech enabled and another one
linking the overlay against libGL, fixing the OpenGL overlay functionality.


Hoshpak:
59358871d6520750d4f50f017178e440b11fcdf6

PR #5649, opened 2017-01-30
cups: build with gssapi support

Hoshpak:
The cupsd.conf currently provided by the cups package contains a Kerberos
section. Since cupsd is built without gssapi support, the daemon fails
to verify it's configuration and refuses to start. This commit enables
building the package with gssapi support to fix this.


xtraeme:
Well, I'd like to make this a build option instead.

Hoshpak:
More like that? Normally I would have put it in configure_args but the position of the argument matters to the configure script and it doesn't pick it up otherwise. So I defined a new variable and appended it at the end of the configure arguments.

Duncaen:
I would say, a `build_option` but disabled by default while fixing the config file.
No one requested this gssapi support and just adding it to make the default config file work seems a but useless.

srcpkgs/cups/template
```diff
@@ -49,7 +58,7 @@ do_configure() {
 		--enable-libpaper --enable-avahi --enable-pam --enable-ssl \
 		--with-menudir=/usr/share/applications --with-xinetd=/etc/xinetd.d \
 		--with-optim="${CFLAGS}" --without-systemd \
-		--without-perl --without-php --without-python
+		--without-perl --without-php --without-python ${_extra_args}
```
Duncaen: Just use `$(vopt_enable gssapi)` instead of `${_extra_args}` here?

srcpkgs/cups/template
```diff
@@ -23,11 +23,20 @@ make_dirs="
 	/var/spool/cups 	0755 cups lp
 	/etc/cups/ssl 		0700 cups lp"
 
+build_options="gssapi"
+desc_option_staging="Enable gssapi support"
```
Duncaen: `desc_option_staging` is wrong, but `desc_option_gssapi` is already defined in `common/options.description`.
Just remove it.

Hoshpak:
I made the changes you requested. Building the package with gssapi support was the whole point of the pull request, the build option was added because @xtraeme requested it. So adding it just not to use it would be useless from my point of view. I see no harm in building cups with gssapi support, it is done this way by many distributions and is apparently now considered to be the default by upstream.

But I agree that the most important thing is to get cups working again so I don't want to argue to much about that. I've ammended the PR again to remove the option as default and pull in a patch taken from alpine to the cupsd.conf which removes the need to have the package built with gssapi enabled to start the daemon.

srcpkgs/cups/template
```diff
@@ -23,11 +23,17 @@ make_dirs="
...
+
 # Package build options
 system_accounts="cups"
 cups_pgroup="lp"
 cups_homedir="/var/spool/cups"
 
+if [ ${build_option_gssapi} ]; then
+	makedepends+=" mit-krb5-devel"
+fi
+
```
Duncaen: Forgot about this, you can use `$(vopt_if gsapi mit-krb5-devel)` in `makdepends=`.

Hoshpak:
done

asergi:
Remove the old `--disable-gssapi` flag. Thanks.

Hoshpak:
done. Thanks for pointing that out, I didn't notice it.

PR #5650, opened 2017-01-30
sigil: add necessary python3 dependency; fix cross

jannisch:
The application throws a ton of errors before startup and then crashes without this python3 package being installed.

### DIFF
--- a/srcpkgs/sigil/template
+++ b/srcpkgs/sigil/template
@@ -1,19 +1,27 @@
 # Template file for 'sigil'
 pkgname=sigil
 version=0.9.7
-revision=1
+revision=2
 create_wrksrc=yes
 build_style=cmake
-hostmakedepends="pkg-config unzip"
+hostmakedepends="pkg-config unzip qt5-qmake"
 makedepends="zlib-devel minizip-devel pcre-devel qt5-tools-devel qt5-svg-devel
  qt5-webkit-devel qt5-xmlpatterns-devel qt5-multimedia-devel qt5-sensors-devel
  qt5-declarative-devel qt5-location-devel qt5-webchannel-devel
  qt5-plugin-odbc qt5-plugin-mysql qt5-plugin-pgsql qt5-plugin-sqlite qt5-plugin-tds
  boost-devel hunspell-devel python3-devel"
-depends="desktop-file-utils"
+depends="desktop-file-utils python3-lxml"
 short_desc="A multi-platform EPUB ebook editor"
 maintainer="Juan RP <xtraeme@voidlinux.eu>"
 license="GPL-3"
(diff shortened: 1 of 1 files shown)


jannisch:
Might the failed build show a common issue? I wonder why `qmake` can't be found although I only added a runtime dependency.

Duncaen:
I guess it did not cross compile before, because support for cross compiling was added to the qt5 stuff recently.
`qt5-qmake` needs to be a `hostmakedepends`.

jannisch:
Thanks for the info, hope it builds now!

pullmoll:
For cross builds you probably need to add a statement like this:
```
        hostmakedepends+=" qt5-host-tools qt5-tools-devel qt5-svg-devel"
        hostmakedepends+=" qt5-webkit-devel qt5-xmlpatterns-devel qt5-multimedia-devel"
        hostmakedepends+=" qt5-sensors-devel qt5-declarative-devel qt5-location-devel"
        hostmakedepends+=" qt5-webchannel-devel"
fi
```

jannisch:
Really didn't expect those issues, will need some time...

pullmoll:
Wild guess: The linker giving errors for missing symbols when cross compiling may be a case of `gcc-6.3.0` eliminating unused code. You could try, if `CXXFLAGS="-fno-lifetime-dse"` in the global section helps.

jannisch:
Unfortunately that does not work (I hope _global_ means at the end of the template, tried it in `configure_args` as well)

I could not find anything similar in the [Arch PKGBUILD](https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/sigil) or in the [Gentoo ebuild](https://gitweb.gentoo.org/repo/gentoo.git/tree/app-text/sigil/sigil-0.9.6.ebuild) as well 😞

pullmoll:
Arch doesn't try to cross compile sigil, neither does gentoo AFAICT.
I'll take a closer look and come back to you.

jannisch:
Oops that's obvious for Arch...
Thank you!

pullmoll:
Just adding `python3-devel` to the list of `hostmakedepends+="…"` inside the `if` statement solves cross building, i.e.
```
        hostmakedepends+=" qt5-host-tools qt5-tools-devel qt5-svg-devel"
        hostmakedepends+=" qt5-webkit-devel qt5-xmlpatterns-devel qt5-multimedia-devel"
        hostmakedepends+=" qt5-sensors-devel qt5-declarative-devel qt5-location-devel"
        hostmakedepends+=" qt5-webchannel-devel python3-devel"
fi
```

jannisch:
That's it 😃

PR #5655, opened 2017-02-01
Include zathura as dependency for zathura-*

Vaelatern:
@lemmi seem okay?


lemmi:
@Vaelatern of course, go ahead. thanks for the heads up 👍

PR #5665, opened 2017-02-02
New package: python-sh-1.12.9

pulux:


srcpkgs/python-sh/template
```diff
@@ -0,0 +1,30 @@
...
+checksum=579aa19bae7fe86b607df1afaf4e8537c453d2ce3d84e1d3957e099359a51677
+
+post_install() {
+	vlicense LICENSE.txt
+}
+
+python3-sh_package() {
+	noarch=yes
+	pycompile="sh.py"
+	short_desc="${short_desc/Python2/Python3}"
```
Vaelatern: Maybe changing the depends here too, to `python3`

srcpkgs/python-sh/template
```diff
@@ -0,0 +1,33 @@
...
+python3-sh_package() {
+	depends="python3"
+	noarch=yes
+	pycompile="sh.py"
+	short_desc="${short_desc/Python2/Python3}"
+	pkg_install() {
+		vmove usr/lib/python3*
+	}
+	post_install() {
+		vlicense LICENSE.txt
```
asergi: Move `vlicense` to `pkg_install()` and remove the above `post_install()`.

PR #5669, opened 2017-02-03
obs: fix the version number inside obs

Sir-Boops:
This makes obs report the proper version number instead of "0.0.1"


chneukirchen:
Tweaked revbump and commit message.

PR #5670, opened 2017-02-03
New package: gcovr-3.3

misery:


srcpkgs/gcovr/template
```diff
@@ -0,0 +1,32 @@
...
+
+post_install() {
+	vlicense LICENSE.txt
+}
+
+python3-gcovr_package() {
+	noarch=yes
+	pycompile_module="gcovr"
+	depends="python3-setuptools"
+	short_desc="${short_desc/Python2/Python3}"
```
Vaelatern: Alternatives for python3.
See `youtube-dl` for an example.

misery: done

PR #5676, opened 2017-02-04
New package: chessx-1.4.0

cipr3s:


cipr3s:
Should we use only_for_archs here since there doesn't seem to be any svg & multimedia qt modules on aarch64 and armv7 ?

cipr3s:
Cross-compilation should be fixed now.

Vaelatern:
`only_for_archs` is almost always wrong.

PR #5679, opened 2017-02-05
New package: python-picamera-1.12

cipr3s:


srcpkgs/python-picamera/template
```diff
@@ -0,0 +1,31 @@
...
+
+post_install() {
+	vlicense LICENSE.txt
+}
+
+python3-picamera_package() {
+	noarch=yes
+	depends="python3"
+	pycompile_module="picamera"
+	short_desc="${short_desc/Python2/Python3}"
```
Vaelatern: Not effective. Perhaps you'd rather change the original `short_desc` so this can be effective.

srcpkgs/python-picamera/template
```diff
@@ -0,0 +1,31 @@
...
+pkgname=python-picamera
+version=1.12
+revision=1
+noarch=yes
+wrksrc="picamera-${version}"
+build_style=python-module
+pycompile_module="picamera"
+hostmakedepends="python-setuptools python3-setuptools"
+depends="python"
+short_desc="A pure Python interface for the Raspberry Pi camera module"
```
Vaelatern: Maybe
`short_desc="Pure Python2 interface for the Raspberry Pi camera module"`

cipr3s:
Updated.

PR #5685, opened 2017-02-06
WIP New package: openjdk7-7.51.2.6.7

pullmoll:
The version is a combination of OpenJDK and the IcedTea release.

Unfortunately it fails to build with musl gcc-gcj, which most probably
needs a patch to drastically increase the heap size for gij.

You're welcome to try to *fix* gcc-gcj for x86_64-musl.

There is also some issue with solib name resolution: The template currently picks up `openjdk-8u121` libs for some `*.so` files of `openjdk7`. These need to be overriden somehow, but I don't know how to achieve this.


chneukirchen:
Why do we want this when we have JDK 8?

gcc-gcj will be gone in gcc 6.4.

pullmoll:
The idea was to become able to compile JDK 8 for musl libc.

Gottox:
In the long term I'd like to get rid of oracle-j{re,dk} completely. These are the first steps. (Thanks @pullmoll for taking over the project!)

pullmoll:
@Gottox I don't want to take over the project. This is just a WIP and proof of concept, that building JDK 7 with gcj works in principle. We will just have to keep the latest `gcc-6.3.x` which supports  `java` for that to happen for some forseeable future... and of course fix `gij` of `gcc-gcj-*-x86_64-musl`.

pullmoll:
It is too much for travis anyway, so I'll close this PR again. I'll keep the branch openjdk7 in my fork of void-packages, though.

chneukirchen:
We probably could just take the alpine binaries for bootstrapping...

chneukirchen:
Keeping old gcc and openjdk sounds like a Rube-Goldberg-contraption to me, but everyone can spend their time as they like. :)

pullmoll:
Thank you for coining *old gcc* for 6.3.x :grimacing:
There isn't too much time left while I disliked the *no java for musl* situation and thus spent some of it.

PR #5689, opened 2017-02-07
New package: srw-0.7

diogoleal:


Vaelatern:
`#include <sys/ttydefaults.h>` for `srw.c` should help with some build errors.
As for the others (`LINEFEED`, `RETURN`), I do not know. Maybe we could define them to whatever glibc does.
Opinions, @ebfe ?

Duncaen:
`LINEFEED` etc. is defined (https://bitbucket.org/emg/srw/src/295139501255976d05b7031e754d85834758e0cb/srw.c?at=default#srw.c-29),  but doesn't work because `CTRL` from `ttydefaults.h` is missing.

PR #5696, opened 2017-02-08
New package: Konsole-16.12.1 (konsole-qt5) and fix to ccache

weabot:
Ported Konsole 16.12.1 to xbps
ccache can't fetch with http:// but does with https://


xtraeme:
Now try to reduce those 5 commits to 1.

PR #5699, opened 2017-02-08
pjproject: update to 2.6.

CMB:


CMB:
I had to patch for libressl.  I've never touched that stuff, so I'm
a bit uncomfortable.  If someone could look over the patch,
that would be appreciated.

srcpkgs/pjproject/patches/libressl.patch
```diff
@@ -0,0 +1,40 @@
...
++#if !defined(LIBRESSL_VERSION_NUMBER) && !defined(OPENSSL_NO_EC) && OPENSSL_VERSION_NUMBER >= 0x1000200fL
+     int ret;
+     int curves[PJ_SSL_SOCK_MAX_CURVES];
+     unsigned cnt;
+@@ -1100,7 +1102,7 @@
+ 
+ static pj_status_t set_sigalgs(pj_ssl_sock_t *ssock)
+ {
+-#if OPENSSL_VERSION_NUMBER >= 0x1000200fL
++#if !defined(LIBRESSL_VERSION_NUMBER) && OPENSSL_VERSION_NUMBER >= 0x1000200fL
```
Vaelatern: We tend to stick the changes for LIBRESSL on the end of the line

srcpkgs/pjproject/patches/libressl.patch
```diff
@@ -0,0 +1,40 @@
+--- pjlib/src/pj/ssl_sock_ossl.c.orig	2017-02-07 09:37:34.014812834 -0800
++++ pjlib/src/pj/ssl_sock_ossl.c	2017-02-07 09:50:41.480264131 -0800
+@@ -114,8 +114,10 @@
+ 
+ #if OPENSSL_VERSION_NUMBER >= 0x10100000L
+ #  define OPENSSL_NO_SSL2	    /* seems to be removed in 1.1.0 */
++#ifndef LIBRESSL_VERSION_NUMBER
+ #  define M_ASN1_STRING_data(x)	    ASN1_STRING_get0_data(x)
+ #  define M_ASN1_STRING_length(x)   ASN1_STRING_length(x)
++#endif
```
Vaelatern: Can you use their #ifs or is this impossible? Breaking out on non-upstream lines may cause more issues.

srcpkgs/pjproject/template
```diff
@@ -42,13 +43,18 @@ pre_configure() {
...
 
 do_build() {
-	make dep LD=$CC
+	make -j1 dep LD=$CC
 	make -j1 LD=$CC
 }
 
+do_install() {
+	make -j1 DESTDIR="${DESTDIR}" install
+}
```
Vaelatern: Hm, is there no way to make these use the buildstyle?
Then you could use `disable_parallel_build`

CMB:
Toyam Cox <notifications@github.com> writes:

> In srcpkgs/pjproject/patches/libressl.patch:
> We tend to stick the changes for LIBRESSL on the end of the line


Ok, fixed.

> In srcpkgs/pjproject/patches/libressl.patch:
>
> @@ -0,0 +1,40 @@
> +--- pjlib/src/pj/ssl_sock_ossl.c.orig	2017-02-07 09:37:34.014812834 -0800
> ++++ pjlib/src/pj/ssl_sock_ossl.c	2017-02-07 09:50:41.480264131 -0800
> +@@ -114,8 +114,10 @@
> + 
> + #if OPENSSL_VERSION_NUMBER >= 0x10100000L
> + #  define OPENSSL_NO_SSL2	    /* seems to be removed in 1.1.0 */
> ++#ifndef LIBRESSL_VERSION_NUMBER
> + #  define M_ASN1_STRING_data(x)	    ASN1_STRING_get0_data(x)
> + #  define M_ASN1_STRING_length(x)   ASN1_STRING_length(x)
> ++#endif
>
> Can you use their #ifs or is this impossible? Breaking out on non-upstream lines may cause more issues.

Oh yes it seems that I can, because libressl defines OPENSSL_NO_SSL2.
I didn't think to check for that.  Fixed.

> In srcpkgs/pjproject/template:
>
>>  	make -j1 LD=$CC
>  }
>  
> +do_install() {
> +	make -j1 DESTDIR="${DESTDIR}" install
> +}
>
> Hm, is there no way to make these use the buildstyle?
> Then you could use disable_parallel_build

I forgot all about disable_parallel_build.  Well, I will add that and
remove do_install.  I assume it will call make install with
-j1.  Still not convinced make install really needs it, but I've seen it
in other build scripts, so playing it safe.
The other make lines in do_build are a problem, because they need
LD=$CC.  It choked on that when I commented out do_build completely.

Thank you for the review!

Vaelatern:
Perhaps in `pre_build` set `LD="${CC}"`

CMB:
> Perhaps in pre_build set LD="${CC}"

That didn't work.  What did work is export LD="${CC}" at the top of
pre_configure.  Also seems that the make dep line is superfluous.
Pushed with do_build eliminated.

PR #5700, opened 2017-02-08
New package: python-GitPython-2.1.1

lafrenierejm:


srcpkgs/python-GitPython/template
```diff
@@ -0,0 +1,29 @@
+# Template file for 'python-GitPython'
+pkgname=python-GitPython
+version=2.1.1
+revision=1
+noarch=yes
+wrksrc="GitPython-${version}"
+build_style=python-module
+pycompile_module="GitPython"
```
Vaelatern: `git`

PR #5711, opened 2017-02-11
New package: qdirstat-1.2

lemmi:


lemmi:
Can we add the necessary build dependencies to the qmake build-style? Maybe we should have a dedicated qt5-qmake build-style as well then. Shall I open an issue for this?

pullmoll:
The problem is that required hostmakedepends pkgs vary from package to package. You don't know what's in makedepends when the build style script is run AFAICT, so you can't know what to add to hostmakedepends.

Vaelatern:
But a `qt-qmake` and a `qt5-qmake` build style might be well suited to this.

lemmi:
So, i just had short peek over all packages that use the qmake build-style. About one fourth has to do the same dance with including extra stuff for cross compile, a large group apparently doesn't need it and then there those that just don't care and set nocross or unconditionally include what they need.

Anyway: at least for me it was kind of unexpected to find a dedicated qmake build-style, but in contrast to other build-styles doesn't make an attempt to handle cross. So I at least wanted to bring up the issue.

pullmoll:
@lemmi The fact that packages depending on qt4 or qt5 can now be cross compiled is rather new - a couple of months or weeks in either case. The qmake build style hasn't been touched since then, which is probably my fault :)

If you want to try to fix or improve the qmake build style, or split it up into qt-qmake (or qt4-qmake) and qt5-qmake, you're of course quite welcome.

lemmi:
That explains it :)

Ok. I'm quite busy until end of march and thus won't make promises. So maybe we can merge this package and I'll open  a separate issue for the changes to qmake.

pullmoll:
Thank you! Merged now.

PR #5716, opened 2017-02-11
apache-fop: update to 2.1

rkallos:


srcpkgs/apache-fop/template
```diff
@@ -10,8 +10,8 @@ distfiles="
  http://mirror.dkd.de/apache/xmlgraphics/fop/source/fop-${version}-src.tar.gz
```
TheDoctorsLife: Remove the whitespace at the end of the url (its right after tar.gz).

rkallos: Good catch! Sorry about that. Rebased.

srcpkgs/apache-fop/template
```diff
@@ -46,7 +46,7 @@ do_install() {
 	vcopy LICENSE ${_docs}
 	vcopy NOTICE ${_docs}
 	vcopy README ${_docs}
-	vcopy status.xml ${_prefix}
+	vcopy build.xml ${_prefix}
```
Vaelatern: What is the source of this rename?

rkallos: Thanks for bringing this up. I forgot to mention, but it appears as though status.xml does not exist after building anymore. That said, build.xml was there. I suspect it was renamed.

```
$ ./xbps-src install apache-fop
---snip---
=> apache-fop-2.1_1: running do_install ...
cp: cannot stat 'status.xml': No such file or directory
=> ERROR: apache-fop-2.1_1: failed to run do_install() at line 174.
```

Gottox: build.xml is used by ant, I don't think that status.xml is the replacement.

Gottox: please check what happened to status.xml :)

Gottox:
No progress for 8 month, closing

PR #5717, opened 2017-02-11
New package: texlive-20160523b

svenper:
There is currently a `texlive-bin` package that does not work with `musl`. Building binaries from source fixes this problem, though it doesn't build on ARM. (yet?)

With dependency `teckit-2.5.6` and shlibs addition.

### DIFF
--- a/common/shlibs
+++ b/common/shlibs
@@ -2845,3 +2845,4 @@ libbaccfg-7.4.7.so bacula-common-7.4.7_1
 libbacfind-7.4.7.so bacula-common-7.4.7_1
 libbacsql-7.4.7.so bacula-common-7.4.7_1
 libbaccats-7.4.7.so bacula-common-7.4.7_1
+libsuperlu.so.5 superlu-5.2.1_1
--- a/srcpkgs/superlu-devel
+++ b/srcpkgs/superlu-devel
@@ -1,0 +1,1 @@
+superlu
--- a/srcpkgs/superlu/template
+++ b/srcpkgs/superlu/template
@@ -1,0 +1,33 @@
+# Template file for 'superlu'
+pkgname=superlu
+version=5.2.1
+revision=1
+build_style=cmake
+hostmakedepends="gcc-fortran"
+makedepends="openblas-devel"
+short_desc="Solution of large, sparse, nonsymmetric systems of linear equations"
+maintainer="Julien Dehos <dehos@lisic.univ-littoral.fr>"
+license="BSD"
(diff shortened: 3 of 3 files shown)


chneukirchen:
FTR, my WIP package on texlive: http://sprunge.us/CXTd

(Part of the effort of putting whole TeXLive into XBPS, so we match up the packages...)

svenper:
Oh, that would be the proper way to do it! It seems your file does a lot more than mine does (and more elegantly), although mine works *right now* which would be the only advantage it really has. Feel free to close this, but otherwise we could use mine for now and you just replace it with yours when you like.

Piraty:
I also looked into this.
Idea for transition:
- new package is named `texlive`, and subpackages accordingly `texlive-<subpkg>`
- remove `provides texlive` from `texlive-bin`
- add `conflicts="texlive-bin>=..."` to new `texlive` package (why would one install both)

That way users of old setup won't face problems, and musl users can enjoy texlive as well.

@chneukirchen would you be so kind to add your template to the repo, or at least open a WIP-PR (along with a list of what's incomplete), so I / we can help completing.
Or maybe @svenper merges the two.

chneukirchen:
Problem is that tlmgr still wants to install binary packages, no?

svenper:
That could be worked around with the [`-custom-bin`](http://tug.org/texlive/custom-bin.html) parameter for `install-tl`, I think. Another problem is packaging the data files. Alpine [doesn't do it](http://git.alpinelinux.org/cgit/aports/tree/testing/texlive/APKBUILD) at all, but Arch seems to have [something](https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/texlive-core) though quite complex. I'm really busy right now, so it will be a while before I start looking seriously into this again.

Piraty:
I looked into this even further.
- alpine's broken, there is a related bugreport.
- slackware gets somewhere, as well as BLFS
- i didn't look to much into debian's buildsystem, but their texlive-maintainer (also texlive-maintainer) published some very specific info on that topic. debian splits texlive into big chunks and provides a "static" texlive distribution.

I adapted my template, based on both of your work and merged some of the aforementioned.

The `custom-bin`-variant sounds good somehow (that way the tlmgr could be used to keep tex-packages updated), but so far I didn't find out how to build the `install-tl`, it is not included in my custom build. 
But on the other hand, it may be more convenient for most users to follow debian's approach: install `tl-base` (bin etc.) + `tl-<what you need>`

PR #5733, opened 2017-02-13
New package: neovim-qt-0.2.6

mfulz:


TheDoctorsLife:
You need to use ``vlicense`` in the template to distribute the package (as the license requires to be distrbuted with the rest of the program) More information can be found about [global functions here](https://github.com/voidlinux/void-packages/blob/master/Manual.md#global-functions)

srcpkgs/neovim-qt/template
```diff
@@ -0,0 +1,13 @@
...
+revision=1
+build_style=cmake
+short_desc="Neovim client library and GUI, in Qt5"
+maintainer="Matthias Fulz <mfulz@olznet.de>"
+license="ISC"
+homepage="https://github.com/equalsraf/neovim-qt"
+distfiles="https://github.com/equalsraf/neovim-qt/archive/v${version}.tar.gz"
+checksum="b68c996386bade219c98d6e4446026f9285d033010f16a7a0695befa6737af3f"
+hostmakedepends="cmake"
+makedepends="qt5-devel"
```
TheDoctorsLife: You need to use ``vlicense`` as the only part of ``post_install()`` as the license needs to be packaged with the program. More information about [global functions](https://github.com/voidlinux/void-packages/blob/master/Manual.md#global-functions) & [post_install()](https://github.com/voidlinux/void-packages/blob/master/Manual.md#functions)

pullmoll:
The commit message should be `New package: neovim-qt-0.2.4` i.e. without revision appended.

srcpkgs/neovim-qt/template
```diff
@@ -0,0 +1,17 @@
...
+version="0.2.4"
+revision=1
+build_style=cmake
+short_desc="Neovim client library and GUI, in Qt5"
+maintainer="Matthias Fulz <mfulz@olznet.de>"
+license="ISC"
+homepage="https://github.com/equalsraf/neovim-qt"
+distfiles="https://github.com/equalsraf/neovim-qt/archive/v${version}.tar.gz"
+checksum="b68c996386bade219c98d6e4446026f9285d033010f16a7a0695befa6737af3f"
+hostmakedepends="qt5-devel cmake"
```
pullmoll: Here cmake is redundant because of the build style cmake. Remove it.

mfulz:
Done.

Duncaen:
Hmm `neovim-qt` was merged from another PR  #6210.

PR #5734, opened 2017-02-13
wireguard: simple package bump

zx2c4:


chneukirchen:
5ffbf1913d

Please use our commit message format in the future, so I can just merge it. :)

PR #5739, opened 2017-02-14
mumble: fix OpenGL overlay and TTS

Hoshpak:
Adds patches to link the OpenGl overlay against LibGL to fix it's
functionality and properlay include the speech-dispatcher headers
so mumble can be built with text-to-speech functionality enabled.
Also install the mumble-overlay script to make it easier to actually
use the overlay.


srcpkgs/mumble/patches/include-speechd-headers.patch
```diff
@@ -0,0 +1,10 @@
+--- src/mumble/mumble.pro	2017-01-29 23:28:11.271718136 +0100
++++ src/mumble/mumble.pro	2017-01-29 23:28:18.341626410 +0100
+@@ -293,6 +293,7 @@
+ speechd {
+ 	DEFINES *= USE_SPEECHD
+ 	LIBS *= -lspeechd
++	INCLUDEPATH *= /usr/include/speech-dispatcher
```
Duncaen: This looks wrong, or does it somehow use `${XBPS_CROSS_BASE}`?

Hoshpak: In what way does it look wrong, because it is an absolute path? The issue is that the speech-dispatcher headers are in /usr/include/speech-dispatcher and not /usr/include. Currently the qmake scripts for mumble fail to take this into account and the compilation fails.

Duncaen: Yes because its absolute, it should be `${XBPS_CROSS_BASE}/usr/include/speech-dispatcher`,
`XBPS_CROSS_BASE` is only set if you cross compile so it works in both cases.

Hoshpak: That seemed to work. However I'm not sure if and how environment variables from the shell are processed by and available within qmake. Since the package cannot be cross-compiled in it's current form I also couldn't test it. The speechd header file is only included in a single file so patching that file to include the header file in the correct path might be another way to solve this. I amended the commit to do that, the path stays relative this way.

PR #5749, opened 2017-02-16
New package: azpainter-2.0.1

o1ux:


TheDoctorsLife:
@o1ux Do you know if this is supposed to be able to cross compiled to other archs like arm?

PR #5751, opened 2017-02-17
New package: font-vlgothic-20141206

diogoleal:


the-maldridge:
There's currently some uncertainty as to what we want to name font packages, once that's cleared up this can be merged.

Issue #5708, opened 2017-02-10
bind not start with glibc2.25


arete:
Hi,
i have update my x86_64 system to last glibc 2.25
the bind daemon start, but not listen on the interface
the named process is defunct

```
$ps -aux
root     19990  0.0  0.0  45964  4944    pts/4    S    14:38                 0:00 named -d 4
root     19991  0.0  0.0      0     0 ?        Zsl  14:38   0:00 [named] <defunct>
```

chneukirchen:
You could try building bind without seccomp, but I thought I had adjusted the list of calls?

Duncaen:
This is something related to logging, someone told me with `-g` it works, or its not updated to the latest version.

arete:
Start named with -g option i have this error
`'/var/log/querylog' failed: permission denied`
this are the file permission
`Access: (0644/-rw-r--r--)  Uid: (  998/   named)   Gid: (  998/   named)`
now i have move the log file to folder /var/named/log and work fine.
Thanks

PR #5752, opened 2017-02-17
Glibc-2.25_1 revert

Gottox:
There are repeatedly coming up issues with glibc-2.25 both breaking systems and services.

This PR addresses this issue by reverting this release and going back to 2.24 which runs fine.


chneukirchen:
Reverting possibly breaks even more since there are binaries now that link against 2.25 features.

x86_64 works fine for me, the problems seem to appear on 32-bit platforms. I guess we need better bug reports to fix them...

chneukirchen:
Hm, only ruby references 2.25 symbols here on my system.

Gottox:
oh boy. 🙄

xtraeme:
IMHO reverting is simply useless. Let's find and fix the real issues instead.

chneukirchen:
I have a i686 machine image now, and just updated without issues. Please tell me what's broken. :)

Gottox:
IRC-Logs:
```
bind still won't run since glibc update without logging being set to stderr with -g
```

arete:
With last update bind work fine,is only an mistach of log permission #5708

Il 17 feb 2017 16:36, "Enno Boland" <notifications@github.com> ha scritto:

chneukirchen:
One i686 that sproadically appears (only on real HW so far) seems to result from https://github.com/bminor/glibc/commit/23d27709a423aec32821e9a5198a10267107bae2

chneukirchen:
We could try reverting that one, after all, it wasn't noticed before...

Hoshpak:
As I mentioned on IRC, reverting that commit solved the issue for me. Since it was possible to collect at least some information on the issue (thanks @chneukirchen for your help with that) and I am able to reproduce it reliably with grep, I decided to create an upstream bug report: https://sourceware.org/bugzilla/show_bug.cgi?id=21182 .

chneukirchen:
I have access to a Atom N270 now, and couldn't reproduce this yet.

I think `-dbg` packages on i686 are broken because they use `/usr/lib/debug/usr/lib` and not `lib32`...

Gottox:
superseeded by #5840

Issue #5727, opened 2017-02-12
kernel-libc-headers: linux/nvme.h is not included


pullmoll:
Since the update to `kernel-libc-headers-4.9.8` the file `/usr/include/linux/nvme.h` is gone.
The pkg `efivar` tries to include this file from `src/linux.c:25` and thus no longer builds.
I don't know what to do here. Update `efivar` and `efibootmgr`?

pullmoll:
efivar and efibootmgr were updated.

PR #5754, opened 2017-02-18
efivar, efibootmgr: update to 31, 14

pullmoll:
Upstream changed the version numbering to major only.
The current version 0.21 of efivar does no longer build with kernel-libc-headers-4.9.8 (see #5727).


pullmoll:
Successfully booted a live ISO using this efivar + efiboot versions in VirtualBox after fixing problem #5764.
Good enough to go?

PR #5758, opened 2017-02-18
New package: fbxkb-0.6

E100Beta:
fbxkb is a very lightweight and minimalist, although pretty old (last official update in 2006) X keyboard layout indicator. It's a good choice for lightweight WMs like i3.

I understand that pre-1.0 package with such a long period without maintenance is concerning, but due to its minimal functionality and rather small codebase I don't think it would do any harm. I've been using fbxkb for three month on two distributions and did not have any problems with it.

EDIT: added "only_for_arches" to the template, package didn't build on some architectures.

### DIFF
--- a/srcpkgs/fbxkb/template
+++ b/srcpkgs/fbxkb/template
@@ -1,0 +1,29 @@
+# Template for fbxkb
+
+pkgname="fbxkb"
+version="0.6"
+revision=1
+build_style=configure
+short_desc="Lightweight X keyboard layout indicator, uses GTK+2"
+maintainer="Pavel Kiselev <kispash@yandex.ru>"
+hostmakedepends="pkg-config"
+makedepends="xorg-server-devel gtk+-devel"
+license="MIT"
+homepage="http://fbxkb.sourceforge.net/"
+distfiles="${SOURCEFORGE_SITE}/${pkgname}/${pkgname}-${version}.tgz"
+checksum="fcbaf4ed9a70f58ea1316b19da74e2ca8b3fb0e2de5a73c849317589ce840ef2"
+
+pre_build() {
+# Substitute compiler,
+# Substitute CFLAGS (original options are subset of void's options),
+# Put LDFLAGS at the end of command (enables PIE)
+	sed -i Makefile* \
+		-e "s;^CC.*=.*$;CC = ${CC};" \
+		-e "s;^CFLAGS.*=.*$;CFLAGS = ${CFLAGS};" \
+		-e "s;^\(.*\).(LDFLAGS)\(.*\)$;\1\2 ${LDFLAGS};"
+}
+
+do_install() {
+	make PREFIX="${DESTDIR}/usr" install
+	vlicense COPYING
+}
--- a/srcpkgs/fbxkb/patches/fbxkb.patch
+++ b/srcpkgs/fbxkb/patches/fbxkb.patch
@@ -1,0 +1,41 @@
+Found on Arch User Repository, adapted for -p0
+
+
+--- fbxkb.c	2006-12-18 23:47:52.000000000 +0200
++++ fbxkb.c	2012-01-15 22:27:30.571427088 +0200
+@@ -378,10 +378,11 @@ read_kbd_description()
+             g_assert((no >= 0) && (no < ngroups));
+             if (group2info[no].sym != NULL) {
+                 ERR("xkb group #%d is already defined\n", no);
++            } else {
++                group2info[no].sym = g_strdup(tok);
++                group2info[no].flag = sym2flag(tok);
++                group2info[no].name = XGetAtomName(dpy, kbd_desc_ptr->names->groups[no]);           
+             }
+-            group2info[no].sym = g_strdup(tok);
+-            group2info[no].flag = sym2flag(tok);
+-            group2info[no].name = XGetAtomName(dpy, kbd_desc_ptr->names->groups[no]);           
+         }
+         XFree(sym_name);
+     }
+--- Makefile.common	2004-10-16 01:18:59.000000000 +0300
++++ Makefile.common	2012-01-15 23:20:02.081020564 +0200
+
+@@ -16,7 +16,7 @@ endif
+ endif
+ 
+ CC = gcc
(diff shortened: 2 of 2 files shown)


srcpkgs/fbxkb/template
```diff
@@ -0,0 +1,30 @@
...
+
+pkgname="fbxkb"
+version="0.6"
+revision=1
+only_for_archs="x86_64 i686 x86_64-musl"
+build_style=configure
+configure_args="--prefix=/usr"
+short_desc="Lightweight X keyboard layout indicator, uses GTK+2"
+maintainer="Pavel Kiselev <kispash@yandex.ru>"
+hostmakedepends="pkg-config"
```
TheDoctorsLife: Add ``xorg-server-devel`` and ``gtk+-devel`` to ``hostmakedepends`` so that it will cross compile.

srcpkgs/fbxkb/template
```diff
@@ -0,0 +1,30 @@
+# Template for fbxkb
+
+pkgname="fbxkb"
+version="0.6"
+revision=1
+only_for_archs="x86_64 i686 x86_64-musl"
```
TheDoctorsLife: Remove this, it will cross compile after you add the packages I suggested to ``hostmakedepends``

E100Beta:
@SirDoctors thanks for the help!
But Travis still fails for those platforms :(
Should I duplicate dependencies in both `hostmakedepends` and `makedepends`, or is it some other issue?
Travis log shows that `gtk+` cannot be found by `pkg-config` on non-x86 platforms, even though it is clearly downloaded.

P.S. I added `gtk+` to `hostmakedepends` to see if it helps. It didn't. I don't want to push too often, so I will remove it later with some other change.

E100Beta:
Ok, now it's ready. And now I finally get what is the difference between host and target dependencies.

srcpkgs/fbxkb/template
```diff
@@ -0,0 +1,29 @@
...
+makedepends="xorg-server-devel gtk+-devel"
+depends="gtk+"
+license="MIT"
+homepage="http://fbxkb.sourceforge.net/"
+distfiles="${SOURCEFORGE_SITE}/${pkgname}/${pkgname}-${version}.tgz"
+checksum="fcbaf4ed9a70f58ea1316b19da74e2ca8b3fb0e2de5a73c849317589ce840ef2"
+nostrip_files="fbxkb"
+
+do_build() {
+	make LDFLAGS=
```
Duncaen: This is just wrong and the main problem, why cross compilation did not work and why adding `-devel` packages to `hostmakedepends` fixed it.
See https://travis-ci.org/voidlinux/void-packages/jobs/203095033#L627, it uses `gcc` and not the cross compiler `armv7l-linux-musleabihf-gcc`.
You should remove the `do_build` function or supply all the variables to make it work correctly.
Those variables would be `LDFLAGS`, `CFLAGS`, `CC` and `LD` if the `Makefile` uses it.
Its necessary to pass the `LDFLAGS` that `xbps-src` defines, otherwise it doesn't use the default/hardened compiler flags that we use for all other packages.

srcpkgs/fbxkb/template
```diff
@@ -0,0 +1,29 @@
...
+short_desc="Lightweight X keyboard layout indicator, uses GTK+2"
+maintainer="Pavel Kiselev <kispash@yandex.ru>"
+hostmakedepends="pkg-config xorg-server-devel gtk+-devel"
+makedepends="xorg-server-devel gtk+-devel"
+depends="gtk+"
+license="MIT"
+homepage="http://fbxkb.sourceforge.net/"
+distfiles="${SOURCEFORGE_SITE}/${pkgname}/${pkgname}-${version}.tgz"
+checksum="fcbaf4ed9a70f58ea1316b19da74e2ca8b3fb0e2de5a73c849317589ce840ef2"
+nostrip_files="fbxkb"
```
E100Beta: The problem is that i have "non-PIE executable found in PIE build" error. I found this while searching the forums. Would `nopie=yes` be better? Or should I search for a problem in other areas?

Duncaen: No this is because it doesn't use voids `CFLAGS` and `LDFLAGS`, they are needed to enable PIE.
You should fix the `do_build` to use voids `CFLAGS` and `LDFLAGS`.

E100Beta: Sorry, read this one after the push. I understand the problem now, shouldn't be hard to fix.

E100Beta:
Thank you @Duncaen for helping me make this better!

The main problem (which prevented building package with PIE) was that `$(LDFLAGS)` were in the wrong place in the `Makefile`.
The way I do things right now involves patching out variables. I've seen with some other `template`s that editing files with `sed` is a more common practise. Should I do it, too?

Everything else was fixed as well and package is now fully functional/builds with right compilers/uses Void's build flags.

Duncaen:
If the `sed` expressions are simple I would go for it, patches are a bit large compared to it if its just some simple search and replace.
In this case I guess there is no way to get the patches upstream, thats why I would choose `sed`.

The template looks good now, can you verify if `configure_args="--prefix=/usr"` is really necessary?

E100Beta:
@Duncaen I did some final touches, `sed` is used to override `CC, CFLAGS, LDFLAGS` in `Makefile`s, and the patch is now as I first found it. Removed `configure_args`. Should now be good to go!

srcpkgs/fbxkb/template
```diff
@@ -0,0 +1,29 @@
...
+# Put LDFLAGS at the end of command (enables PIE)
+	sed -i Makefile* \
+		-e "s;^CC.*=.*$;CC = ${CC};" \
+		-e "s;^CFLAGS.*=.*$;CFLAGS = ${CFLAGS};" \
+		-e "s;^\(.*\).(LDFLAGS)\(.*\)$;\1\2 ${LDFLAGS};"
+}
+
+do_install() {
+	make PREFIX="${DESTDIR}/usr" install
+	vlicense ${wrksrc}/COPYING
```
Vaelatern: `vlicense COPYING`

E100Beta:
@Vaelatern Done. Plus, found out that `make PREFIX...` line was no longer needed, so I removed it and changed function to `post_install()`

PR #5762, opened 2017-02-18
connman: mark /etc/dbus-1/system.d/* as configuration files

ackalker:
This enables users to configure their own access policies.

As an example, in my case it allows me to comment out all of /etc/dbus-1/system.d/connman-nmcompat.conf in order to be able to use connmanctl as non-root user.

See https://forum.voidlinux.eu/t/connman-needs-root/231/7 for more details.


ackalker:
Argh. Gotta love these musl build errors. I've found [this](http://www.openwall.com/lists/musl/2012/10/11/1), what's the end verdict on how these things should be fixed?

pullmoll:
We have a new class of problems since the `kernel-libc-headers` were updated to `4.9.8`: conflicting declarations for various ethernet related structs.

In this case it seems it requires a patch for `inet.c` to `#include <linux/if_ether.h>` only if `__GLIBC__` is defined, i.e.:
```
#if defined(__GLIBC__)
#include <linux/if_ether.h>
#endif
```

Note: this is guesswork based on the Travis-CI output. I may be wrong here, because I didn't test it locally.

pullmoll:
Please amend and force push. The problem with musl libc should be solved.

ackalker:
Is it okay if I add the musl fix as a separate commit? It is addressing a different issue from my original PR.

pullmoll:
There should be no musl fix required. Just force push and see if Travis-CI succeeds.

ackalker:
Ah, misunderstood you. Will do.

ackalker:
No dice... https://travis-ci.org/voidlinux/void-packages/jobs/205580484#L684

pullmoll:
Okay `/usr/include/linux/if_ether.h` isn't currently handled by the #5811 patch.
You could try to patch `src/inet.c` and wrap the `#include <linux/if_tun.h>` into `#if defined(__GLIBC__)`/`#endif`.

I will try to extend the `srcpkgs/kernel-libc-headers/patches/1-5-kernel-libc-compat-musl-support.patch` to also handle `if_ether.h`, if that is possible.

PR #5772, opened 2017-02-19
python-Pyro4 use selectors2

pulux:


Vaelatern:
Please explain why `python-selectors34` is inadequate?

pulux:
selectors34 is not inadequate, see:

https://github.com/irmen/Pyro4/releases/tag/4.55

PR #5805, opened 2017-02-23
New Packages: compiz-plugins

CoolOhm:
Plugin packages for Compiz Reloaded.


srcpkgs/compiz-plugins-main/template
```diff
@@ -0,0 +1,29 @@
...
+license="GPL-2"
+distfiles="https://github.com/compiz-reloaded/compiz-plugins-main/archive/v${version}.tar.gz>${pkgname}-${version}.tar.gz"
+checksum=6fa2c77f37dbd13ec96065d4e6454ba1e5e098ebbc70d56e06d303d6db2098b1
+
+pre_configure() {
+	NOCONFIGURE=1 ./autogen.sh
+}
+
+compiz-plugins-main-devel_package() {
+	depends="${makedepends} compiz-plugins-main>=${version}_${revision}"
```
pullmoll: You can use `${sourcepkg}>=${version}_${revision}` here.

srcpkgs/compiz-plugins-experimental-devel
```diff
@@ -0,0 +1 @@
+compiz-plugins-experimental
```
pullmoll: The `compiz-plugins-experimental-devel_package()` function is missing from the template.

pullmoll: The commit message doesn't match the contents :)

CoolOhm: Hahaha woops! That's what I get for trying to rush some packaging in couple extra minutes I have lying around. :)

srcpkgs/compiz-plugins-experimental/template
```diff
@@ -0,0 +1,30 @@
+# Template file for 'compiz-plugins-experimental' of Compiz Reloaded
+
+pkgname=compiz-plugins-experimental
+version=0.8.12
+revision=1
+build_style=gnu-configure
+configure_args="--disable-static"
+
+hostmakedepends="automake compiz-bcop intltool libtool pkg-config glib"
```
pullmoll: Probably `glib-devel` to fix cross.

CoolOhm: Correct again my friend. Thanks. 😄

pullmoll:
Now, if everything works for you, it may perhaps make sense to create a meta package `compiz` or `compiz-reloaded` which depends on all the components?

CoolOhm:
Hi @pullmoll,

Thanks for the merge. I will be happy so make a meta package as soon as I am finished testing all of these on my fresh void install. 

On that note I have found 1 bug so far which is that `ccsm `expects its icons in `/usr/local/share/ccsm/icons` but they are actually put in `/usr/share/ccsm`. This lack of icons make ccsm fail to have most of its funcationality as the buttons that would use those icons will not 
load.


This is a new issue I did not run into before. I'm sure there is a packaging way to configure this. Looking at the makefile for `ccsm` it appears we simply failed to pass in the standard prefix for Void `/usr` so it defaulted to `/usr/local` .



Here is my plan: 

- I will submit fixes for any bugs I find in a couple of days (I dont expect to find many).
- Test those new packages again I'll

CoolOhm:
Hi @pullmoll,

Thanks for the merge. I will be happy to make a meta package as soon as I am finished testing all of these on my fresh void install. 

On that note I have found 1 bug so far which is that `ccsm` expects its icons in `/usr/local/share/ccsm/icons` but they are actually put in `/usr/share/ccsm/icons`. This lack of icons make ccsm fail to have most of its funcationality as the buttons that use icons do not 
load.

I have temporarily fixed it on my system with a symlink so I can keep testing things.

This is a new issue I did not run into before. However I'm sure there is a packaging way to configure this. Looking at the makefile for `ccsm` it appears I simply failed to pass in the standard prefix for Void  to the makefile (`/usr`) so it defaulted to `/usr/local` .

Here is my plan: 

- I will submit fixes for any bugs I find in a couple of days (I dont expect to find many others, but we'll see).
- Test those new packages again.
- If all is well THEN I'll submit a meta package and update the void wiki for how to instal/configure compiz .

Thanks again for the time, I really appreciate it. 🙂

Issue #5786, opened 2017-02-20
kernel-libc-headers: lacks of musl patches breaks NetworkManager


yopito:
Since the update to `kernel-libc-headers-4.9.8` that removed some patches for muslc (commit https://github.com/voidlinux/void-packages/commit/8c773d0d35c86693363dca906257cdd226b4c3da),  NetworkManager (1.02) does not build anymore with musl libc.

I suspect that other network stuff that rely on kernel interfaces might be affected.

I will try to provide patches against musl libc, but It might take ages (lack of spare time these days).

Output of NetworkManager build on x86_64 muslc platform (native):
```
[*] Updating `https://repo.voidlinux.eu/current/musl/x86_64-musl-repodata' ...
ERROR: [reposync] failed to fetch file `https://repo.voidlinux.eu/current/x86_64-musl-repodata'
: Not Found
ERROR: [reposync] failed to fetch file `https://repo.voidlinux.eu/current/nonfree/x86_64-musl-r
epodata': Not Found
ERROR: [reposync] failed to fetch file `https://repo.voidlinux.eu/current/aarch64/x86_64-musl-r
epodata': Not Found
[*] Updating `https://repo.voidlinux.eu/current/musl/nonfree/x86_64-musl-repodata' ...
x86_64-musl-repodata: [2429B 100%] 66MB/s ETA: 00m00s
x86_64-musl-repodata: 2429B [avg rate: 66MB/s]
[*] Updating `https://repo.voidlinux.eu/current/x86_64-musl-repodata' ...
[*] Updating `https://repo.voidlinux.eu/current/nonfree/x86_64-musl-repodata' ...
[*] Updating `https://repo.voidlinux.eu/current/aarch64/x86_64-musl-repodata' ...
=> Reconfiguring bootstrap packages...
ca-certificates: configuring ...
ca-certificates: configured successfully.
=> NetworkManager-1.0.12_8: removing autodeps, please wait...
=> NetworkManager-1.0.12_8: the following build options are set:
   gir: Enable support for gobject introspection data (ON)
=> NetworkManager-1.0.12_8: building ...
   [host] automake-1.15_4: found (https://repo.voidlinux.eu/current/musl)
   [host] libtool-2.4.6_1: found (https://repo.voidlinux.eu/current/musl)
   [host] pkg-config-0.29.1_2: found (https://repo.voidlinux.eu/current/musl)
[...]

PR #5811, opened 2017-02-23
kernel-libc-headers: musl libc support for compat with linux/*.h

yopito:
following issue #5786 
still WIP, but NetworkManager 1.0.2 builds (native muslc)

### DIFF
--- a/srcpkgs/blueman/patches/musl.patch
+++ b/srcpkgs/blueman/patches/musl.patch
@@ -1,6 +1,6 @@
 --- module/libblueman.c.orig	2016-03-27 09:13:07.203123088 +0200
 +++ module/libblueman.c	2016-03-27 09:13:19.497969382 +0200
-@@ -28,8 +28,13 @@
+@@ -28,6 +28,7 @@
  #include <string.h>
  #include <sys/ioctl.h>
  #include <sys/socket.h>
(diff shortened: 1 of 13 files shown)


pullmoll:
Thanks! The patch, once working, needs to be symlinked in all `cross-*musl*` packages to `<pkgname>/files/<patchname>` as well, and it needs to be applied there with
```
    _apply_patch -p0 ${FILESDIR}/<patchname>
```
after or around `_apply_patch -p1 ${FILESDIR}/gcc-6.3.0-musl.diff`.

yopito:
thanks for your help, I was not aware about the `cross-*musl*` packages.

I've just bump their revision since they apply all patches from `kernel-libc-headers` in their `_linux_headers()` routines (no need to link these files).

Let's see if Travis CI can deal with it.

pullmoll:
Ouch! I see the cross*musl* packages are still using musl 1.1.15. We should update that at the same time.

yopito:
dedicated commit ? seperate PR ?

yopito:
git rewrites: separated commits for `kernel-libc-headers` and `cross*musl`.
reworked `blueman` patches to take advantage of my `kernel-libc-headers` (build fine on native musl libc)

pullmoll:
I think the two `cross*musl*` commits should be squashed into one.
IMO grouping the update of 6 or 7 packages with the same change into one commit is ok, while there is usually the rule to have one package per commit. What do the other contributors think?

I also have a libreoffice pending here, which currently fails to build for musl libc, so merging this patch rather sooner than later would be good :)

yopito:
Hello, I've squashed my commits, added 2 forgotten `cross*musl` packages and review my comments within the patch of `linux/libc-compat.h`.
Ready to be merged or respond to reviewers ;)

pullmoll:
Merged. I'm building from scratch locally again to see if/what packages would still fail for *-musl.
Thank you for the patches.

PR #5813, opened 2017-02-23
xbps-src: implement a 'check' stage

### LINKED ISSUE #5808: RFC: make check for core packages
This is partially related to #5801 but IMO worth a separate issue entry and discussion.

Void does not make use of the `make check` or `make test` stages which many packages, specifically core libraries and utilities, offer.

This now turned out to lead to hard to track down problems as in a case where I was baffled that debugging `gcc-gcj-6.3.0_3-x86_64-musl` wouldn't work, because `gdb` could not find symbols in the debug pkg.

I believe we should take the [LFS](http://www.linuxfromscratch.org/lfs/view/8.0-rc1/) recommendations for packages which **definitely** require a `make check` serious and strive to fix what's currently failing.

The first package to fix is obviously `binutils` because it currently fails the `make -k checks`.

IMO further ignoring failing checks for such essential packages may lead to hell on a rather short path.

I think an additional, optional function `do_check()` which, if it exists in a package's template, could be invoked by `common/xbps-src/libexec/xbps-src-dobuild.sh`. It would probably be called as the last step, just before touching `$XBPS_BUILD_DONE`.

Pro? Contra? Other suggestions?

pullmoll:
Add another stage 'check' between 'build' and 'install'. It is be enabled using the
variable XBPS_CHECK_PKGS=yes and disabled if unset, set to "0" or "no" in your
local etc/conf.

A new xbps-src option `-q` for `quick` will disable XBPS_CHECK_PKGS by overriding it to 0.
If enabled, `common/xbps-src/shutils/xbps-src-docheck.sh` checks for an existing
`do_check()` function in the package's template and, if it exists, calls it.

A new template variable `checkdepends` may be present and list packages required
to run the `do_check()` function. Example: `checkdepends="bc unittest-cpp"`.

Closes #5808


pullmoll:
Should be working now. Travis-CI does not have XBPS_CHECK_PKGS=yes and thus skips the do_check() for binutils. If you try on a local branch, you can see how `make -k check` fails.

jcjordyn130:
what would this be used for?

pullmoll:
An example is given with the `binutils` update. Purpose: check functionality of compiled packages.

jcjordyn130:
ah

Gottox:
What do you think about putting it into its own file? similiar to the `update` file?

pullmoll:
You mean a file like `check` besides the template?
Sure, that's possible, but where do you see the advantage?

PR #5822, opened 2017-02-25
shadow: add patch for CVE-2016-6252

Hoshpak:
this commit adds CVE-2016-6252.patch and CVE-2017-2616.patch taken
from upstream commits 1d5a926cc2d6078d23a96222b1ef3e558724dad1 and
08fd4b69e84364677a10e519ccb25b71710ee686 which fix the corresponding
security issues.


dne:
Void uses su from util-linux, so see #5823 re: fix for CVE-2017-2616.

Hoshpak:
Right, thanks for pointing that out. Since getulong is used in other commands which are shipped in the shadow package in void, the fix for CVE-2016-6252 should still be relevant. I dropped the other patch from the commit.

PR #5834, opened 2017-02-26
gdb: configure with --with-build-sysroot, not --with-sysroot.

CMB:
When a cross-compiled gdb is built with --with-sysroot, it
looks for shared libraries in the wrong place when installed
to the target system.  So use --with-build-sysroot instead.

### DIFF
--- a/srcpkgs/gdb/template
+++ b/srcpkgs/gdb/template
@@ -1,7 +1,7 @@
 # Template file for 'gdb'
 pkgname=gdb
 version=7.12.1
-revision=1
+revision=2
 patch_args="-Np1"
 build_style=gnu-configure
 pycompile_dirs="/usr/share/gdb"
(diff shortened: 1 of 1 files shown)


Vaelatern:
Please don't set this in a function, but rather with all the other variables.

CMB:
All the other variables are processed before the various build phases
are ran, right?  E.G, the configure phase in the gnu-configure build
style would add --with-sysroot to the configure_args set in the variable
portion of the template and defeat the purpose.  So I think I'm stuck
with pre_configure.

Gottox:
gdb doesn't respect --with-sysroot at all?

CMB:
Yes it respects --with-sysroot, but --with-sysroot causes an additional
unwanted side-effect: gdb will search for libraries under the sysroot at
run time.  So when trying to debug on my ARMv7 system, it was failing to
load symbols for shared libraries, because it was looking in
/usr/armv7l-linux-musleabihf/usr/lib/debug instead of /usr/lib/debug.

PR #5840, opened 2017-02-27
glibc: revert upstream changes to memchr-sse2

Hoshpak:
the changes made in glibc upstream commit 23d27709a423aec32821e9a5198a10267107bae2
cause segmentation faults on at least some i686 processors. This commit
reverts these changes to resolve the issue on affected systems until
there's an upstream fix.

This PR is intended as an alternative to #5752. Since at least some important packages (gnutls, sshd) now link against glib-2.25 symbols, it is now longer feasible to stay on glibc-2.24 on a Void installation. Reverting the repository back to 2.24 would require a rebuild of these packages.


chneukirchen:
+1 from me.

ebfe:
cc @xtraeme

chneukirchen:
Proper upstream fix is now at https://sourceware.org/git/gitweb.cgi?p=glibc.git;a=commit;h=3abeeec5f46ff036bd9df60bb096e20314ccd078

PR #5842, opened 2017-02-27
KF5 dependencies fixes

dutsky:


Vaelatern:
Just a little information about why these are necessary?

dutsky:
Plasma needed some additional dependencies like this, and you must set it by hands in "depends" section(i don't know why it don't install automatically), otherwise Plasma won't work correctly(empty screen with one panel and without any widgets).

PR #5843, opened 2017-02-27
New package: plasma-workspace-5.8.7

dutsky:
This is part one of KDE Plasma with dependencies


common/shlibs
```diff
@@ -467,10 +467,16 @@ libsensors.so.4 libsensors-3.1.1_1
 libcap-ng.so.0 libcap-ng-0.6.2_1
 libphysfs.so.1 physfs-2.0.0_1
 libSDL_ttf-2.0.so.0 SDL_ttf-2.0.9_1
+libKScreenLocker.so.5 kscreenlocker-5.8.4_1
 libparted.so.2 libparted-3.1_1
 libparted-fs-resize.so.0 libparted-3.1_1
 libntfs-3g.so.87 ntfs-3g-2016.2.22_1
 libruby.so.2.4 ruby-2.4.0_1
+libruby.so.2.3 ruby-2.3.0_1
```
Vaelatern: But we don't have a `ruby-2.3`

srcpkgs/ksysguard/patches/musl.patch
```diff
@@ -0,0 +1,33 @@
+--- ksysguardd/Linux/diskstat.c.orig	2016-12-20 13:57:38.395757618 +0400
++++ ksysguardd/Linux/diskstat.c	2016-12-20 14:01:23.745156823 +0400
+@@ -18,6 +18,7 @@
+ 
+ */
+ 
++#define _BSD_SOURCE
```
Vaelatern: This define is depreciated on glibc systems. I don't know if it matters, it probably doesn't.

dutsky:
ruby-2.3 stanza removed. BSD_SOURCE stanzas removed.

PR #5845, opened 2017-02-27
i3blocks: add an INSTALL.msg

### LINKED ISSUE #5684: i3blocks: should depend on sysstat
The cpu_usage "block" needs the tool mpstat, which is provided by the sysstat package.

Grauwolf:
this fixes #5684 since those packages probably should not be hard dependencies


srcpkgs/i3blocks/INSTALL.msg
```diff
@@ -0,0 +1,6 @@
+Some of the included scripts may use external tools which must be installed:
+
```
Vaelatern: Remove this blank line

Grauwolf: Done!

PR #5856, opened 2017-02-28
New package: swarp-0.1

ibabushkin:
The packaged software is rather trivial, but quite useful at times - so it makes sense to include it in the repos IMO.

### DIFF
--- a/srcpkgs/swarp/template
+++ b/srcpkgs/swarp/template
@@ -1,0 +1,23 @@
+# Template file for 'swarp'
+pkgname=swarp
+version=0.1
+revision=1
+short_desc="Simple pointer warping tool for X"
+maintainer="Inokentiy Babushkin <inokentiy.babushkin@googlemail.com>"
+license="MIT"
+makedepends="libX11-devel"
+homepage="http://tools.suckless.org/x/swarp"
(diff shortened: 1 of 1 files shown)


srcpkgs/swarp/template
```diff
@@ -0,0 +1,23 @@
...
+version=0.1
+revision=1
+short_desc="Simple pointer warping tool for X"
+maintainer="Inokentiy Babushkin <inokentiy.babushkin@googlemail.com>"
+license="MIT"
+makedepends="libX11-devel"
+homepage="http://tools.suckless.org/x/swarp"
+distfiles="http://dl.suckless.org/tools/${pkgname}-${version}.tar.gz"
+checksum=ef5730fe8ee00879cbec1e91e22a7f0f7817a63375d790d775f5b7427886d45f
+
```
ebfe: judging from the sed's already done, it doesn't look like that would improve the template.

SirDoctors: If thats the case then which would be a the preferred way then? To remove the sed's or to set the build style?

ebfe: neither would work, so i'd just leave it as it is.

ibabushkin: This is based on other suckless software packaged for void - for instance `slock`. Those packages use the same approach.

PR #5858, opened 2017-02-28
New package: audit-2.7.3

zqqw:
Audit is a security logging package which is also useful for debugging and development work, as it can, for example, monitor and log access to specified files and directories, and record syscalls made by processes.


chneukirchen:
Why not use `make_dirs=`?

chneukirchen:
The manpage says "During startup, the rules in /etc/audit/audit.rules are read by auditctl.", so why is there a one-shot service?

zqqw:
The documentation presumably assumes using the included init files for sysvinit or systemd. This prompted me to look at audit-2.7.3/init.d/auditd.service, you can see the need to load the rules after the auditd service has started, and the extra auditctl service provides the equivalent to ExecStartPost. The auditctl script is just an adaptation of existing examples on the runit website and seems the accepted way to manage this. However, my reading of the documentation led me to believe that auditctl should be run at boot and augenrules run manually, while the service uses augenrules by default. So there could be some question over what is best here. Perhaps there should be an augenrules service as well, so the user could choose which to enable. Regarding the directories, 2 empty ones had to be created, and they were automatically deleted at the end of the build process when I used vmkdir.
If, in voidpackages/srcpkgs you do
$ ag -s install |ag -s INSTALL:
$ ag -s mkdir |ag -s INSTALL:
You will see where I found this idea! 'install' can make a directory and set the permissions in a single command.
I don't know if make_dirs would work better than vmkdirs, or would the empty dirs still get deleted? Is this why there are so many created in INSTALL scripts in other packages?

```
[Unit]
Description=Security Auditing Service
DefaultDependencies=no
## If auditd.conf has tcp_listen_port enabled, copy this file to
## /etc/systemd/system/auditd.service and add network-online.target
## to the next line so it waits for the network to start before launching.
After=local-fs.target systemd-tmpfiles-setup.service
Conflicts=shutdown.target
Before=sysinit.target shutdown.target
RefuseManualStop=yes
ConditionKernelCommandLine=!audit=0
Documentation=man:auditd(8) https://people.redhat.com/sgrubb/audit/

[Service]
Type=forking
PIDFile=/var/run/auditd.pid
ExecStart=/sbin/auditd
## To not use augenrules, copy this file to /etc/systemd/system/auditd.service
## and comment/delete the next line and uncomment the auditctl line.
## NOTE: augenrules expect any rules to be added to /etc/audit/rules.d/
ExecStartPost=-/sbin/augenrules --load
#ExecStartPost=-/sbin/auditctl -R /etc/audit/audit.rules
ExecReload=/bin/kill -HUP $MAINPID
# By default we don't clear the rules on exit. To enable this, uncomment
# the next line after copying the file to /etc/systemd/system/auditd.service
#ExecStopPost=/sbin/auditctl -R /etc/audit/audit-stop.rules

[Install]
WantedBy=multi-user.target
```

zqqw:
I also see this only builds on PC GLIBC platforms, and fails on others.
Perhaps the situation would be improved by adding extra configure options for ARM as these options are available in ./configure --help
  --with-alpha            enable Alpha processor support
  --with-arm              enable Arm eabi processor support
  --with-aarch64          enable Aarch64 processor support
I have no suitable hardware to try this myself.
Nothing available for musl though.

chneukirchen:
`make_dirs=` is exactly for the situations where you need empty dirs.

Inspecting auditd, it doesn't seem to load the rules itself (I don't see how SIGHUP is supposed to work there?). But the auditctl service should just use `sv check` on auditd, and then `sv once .`instead.

zqqw:
sv down auditctl sv up auditctl reloads the rules. Although if the instructions for secure logging are followed changes are locked and they can only be changed by rebooting.
" SIGHUP causes auditd to reconfigure. This means that auditd re-reads the configuration file."
It re - reads /etc/audit/auditd.conf, not /etc/audit/audit.rules.
I haven't found any patches to build on musl, audit isn't available in Alpine Linux. 
It will probably take at least a day or two to research some of this further and reconfigure the build and then try it over a few boots to test the features a bit incidentally. Thank you for the helpful tips so far!

zqqw:
Here is an improved version. It excludes MUSL and includes build options for ARM. I found that rules could be loaded regardless of whether auditd was running so they are loaded in the auditd run script before it execs.

zqqw:
Regarding the conflicts, all I did with common/shlibs was add 2 libs to the end of the list, libaudit and libauparse : 
```
libcompizconfig.so.0 libcompizconfig-0.8.12.1_1
libemeraldengine.so.0 emerald-0.8.12.4_1

libhangul.so.1 libhangul-0.1.0_1
libaudit.so.1 audit-0.1.7_1
libauparse.so.0 audit-0.1.7_1
```
But the list has since grown. And why is there that blank line near the end in shlibs? I don't know how to get round this without jeopardizing the single commit for a pull request rule, there are no instructions in the documentation for dealing with this, so I have left it for now. But I could take action if instructed to do so.

PR #5866, opened 2017-03-01
New packages from kde-apps 16.12

weabot:


dutsky:
I can't point at anything, because i'm not dev, but maybe better rename it to dolphin, konsole5, kate5, konversation5? It's more KDE conflict than qt, IMHO

Hoshpak:
Why would it even be necessary to rename the packages? These are just more recent versions of applications already in the repository and can probably happily coexist with existing software. If there aren't any dependencies on the KDE4 versions that would be broken by updating them, I believe it would be better to just update the templates to the current version. In the case of dolphin which isn't a separate package but (in the KDE 4 version) part of kde-baseapps, it would be sufficient to declare a conflict with that package so users could only install one of the available versions. KDE 4 is a bit broken anyway and should probably be removed as soon as KDE 5 is fully packaged and functional.

circhioz:
As you can see from #7750, #7780, #7748, #7218 and #7782, these packages have been already merged. I think it could be more useful closing this PR and opening an issue about KDE4 and KDE5 applications naming, if this is your concern.

asergi:
Obsolete.

PR #5869, opened 2017-03-02
base-files: fix vkpurge rm installed check

Duncaen:
`vkpurge rm 4.10.1_1` does not correctly check if the version is installed and would delete the files anyways, `vkpurge list` and `vkpurge all` use the the correct check.


Vaelatern:
3 thumbs up :)

PR #5873, opened 2017-03-02
gnome-online-accounts: update to 3.22.4

Sir-Boops:
also removed some unneeded flags.


SirDoctors:
Just being curious, but how were those flags unneeded?

Sir-Boops:
Those options are enabled by default so no need for a ++

SirDoctors:
Alright, that was what I assumed.

Sir-Boops:
Updated in 468648c9751d04c19343338af8207528bf72dc16

PR #5880, opened 2017-03-03
adobe-flash-plugin: update to 24.0.0.221

cr6void:
Template updated to version 24.0.0.221
Updated eula
Updated checksums


Vaelatern:
please name the commit correctly.

cr6void:
Ah ok.
adobe-flash-plugin is not the correct name?
So, what is the correct name for this commit?
I'm not used to working on github, sorry.
Thanks.

PR #5883, opened 2017-03-03
New package: spt-0.4

ibabushkin:
A simple pomodoro timer. Also quite useful at times. `libnotify` integration compiled in.

### DIFF
--- a/srcpkgs/spt/template
+++ b/srcpkgs/spt/template
@@ -1,0 +1,22 @@
+# Template file for 'spt'
+pkgname=spt
+version=0.4
+revision=1
+short_desc="Simple pomodoro timer"
+maintainer="Inokentiy Babushkin <inokentiy.babushkin@googlemail.com>"
+license="ISC"
+hostmakedepends="pkg-config"
+makedepends="libnotify-devel glib-devel"
+build_style="gnu-makefile"
+homepage="https://github.com/pickfire/spt"
+distfiles="https://github.com/pickfire/${pkgname}/archive/${version}.tar.gz"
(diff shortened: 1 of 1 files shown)


srcpkgs/spt/template
```diff
@@ -0,0 +1,24 @@
+# Template file for 'spt'
+pkgname=spt
+version=0.4
+revision=1
+short_desc="Simple pomodoro timer"
+maintainer="Inokentiy Babushkin <inokentiy.babushkin@googlemail.com>"
+license="Custom"
```
Duncaen: ISC afaik

srcpkgs/spt/template
```diff
@@ -0,0 +1,24 @@
...
+hostmakedepends="pkg-config"
+makedepends="libnotify-devel glib-devel"
+depends="libnotify"
```
Duncaen: This is not necessary if it links against `libnotify.so`, xbps-src detects it. 
See https://travis-ci.org/voidlinux/void-packages/jobs/207541866#L649.

srcpkgs/spt/template
ibabushkin: Thanks!

srcpkgs/spt/template
ibabushkin: Yes, you're right. Fixed now.

ibabushkin:
Is there any further change needed on this one? Or should we wait on #5917 to land?

Vaelatern:
The install ought to be taking care of the vman itself, can you confirm?

ibabushkin:
Yes, it seems to. I'll remove it and update the commit.

PR #5884, opened 2017-03-03
compiz-plugins-extra: add mouse-window-snapping

CoolOhm:
Add stable mouse-window-snapping (sometimes called 'aero snap') functionality from compiz-reloaded repos that was added after the last stable release. Since compiz-reloaded has no set release schedule I have added the functionality via  patch. The patch will be removed after the next stable release.


CoolOhm:
Are there any things I need to fix or change on this PR? I would like to get this merged before I submit the meta-package `compiz-reloaded` because I think that mouse-window-snapping is a basic necessity for a modern stacking window manager.

I understand that here at Void we prefer to build from purely from releases, but I am asking for an exception in this case. The patch code has already been into merged compiz-reloaded github project for months, is well tested and is ready for the next release. It is simply that compiz-reloaded does not have a release timeline that I am aware of, so I don't know when it will be available officially. If added I will then remove the patch after the next release.  I just wanted to include the functionality now since I have no way of knowing when that release will happen.

P.S.: I have been using the patched version daily without any issue on my main dev machine. So I can vouch for its stability.

pullmoll:
There you go. I hadn't seen this PR due to the mess of updates I've been working on :)

PR #5891, opened 2017-03-04
New Package: openra-20161019

Hoshpak:


srcpkgs/openra/template
```diff
@@ -0,0 +1,37 @@
...
+fi
+
+do_configure() {
+	make VERSION="release-${version}" version
+}
+
+do_build() {
+	make dependencies
+	make core
+	make tools
```
Vaelatern: Please remove this and use `make_build_target="dependencies core tools"` with setting a buildstyle.

srcpkgs/openra/template
```diff
@@ -0,0 +1,37 @@
...
+distfiles="https://github.com/OpenRA/OpenRA/archive/release-${version}.tar.gz"
+checksum="1cbc258466e1e1180a9ad95d44f3867191552a611dd45be0561372b1e6c055df"
+
+if [ -n "$CROSS_BUILD" ]; then
+	hostmakedepends+=" lua51"
+fi
+
+do_configure() {
+	make VERSION="release-${version}" version
+}
```
Vaelatern: This probably should be in a `pre_build`

srcpkgs/openra/template
```diff
@@ -0,0 +1,37 @@
...
+	make tools
+}
+
+do_install() {
+	make prefix=/usr DESTDIR="$DESTDIR" install-all
+	make prefix=/usr DESTDIR="$DESTDIR" install-linux-shortcuts
+	make prefix=/usr DESTDIR="$DESTDIR" install-linux-mime
+	make prefix=/usr DESTDIR="$DESTDIR" install-linux-appdata
+	make prefix=/usr DESTDIR="$DESTDIR" install-man-page
+}
```
Vaelatern: `make_install_target` and use a build style instead.

srcpkgs/openra/template
```diff
@@ -0,0 +1,37 @@
+# Template file for 'openra'
+pkgname=openra
+version=20161019
+revision=1
+wrksrc="OpenRA-release-${version}"
+noarch=yes
+hostmakedepends="mono unzip curl make"
```
Hoshpak: mono is needed because it provides the compiler, unzip and curl are needed because the build scripts download and unzip dependencies using these tools. make is indeed redundant, I will remove it.

srcpkgs/openra/template
Hoshpak: done.

srcpkgs/openra/template
Hoshpak: done, I used build_style=gnu-makefile since that is the best fit in my opinion.

srcpkgs/openra/template
Hoshpak: done. I had to additionally set make_install_args="prefix=/usr" since the Makefile doesn't honor PREFIX so it would install to /usr/local otherwise.

PR #5892, opened 2017-03-04
elogind: fix musl

### LINKED ISSUE #5887: FYI: working on musl elogind patch
Just to avoid duplicate efforts.

pullmoll:
Compiles but is yet untested.

Closes #5887

### DIFF
--- a/srcpkgs/elogind/files/007-fix-non_portable_percent_m.patch
+++ b/srcpkgs/elogind/files/007-fix-non_portable_percent_m.patch
@@ -1,0 +1,1744 @@
+--- src/cgroups-agent/cgroups-agent.c	2016-03-06 16:00:52.027398234 +0100
++++ src/cgroups-agent/cgroups-agent.c	2017-03-04 16:53:10.369345842 +0100
+@@ -44,7 +44,7 @@
+            system bus.  */
+         r = sd_bus_open_system(&bus);
+         if (r < 0) {
+-                log_debug_errno(r, "Failed to open system bus: %m");
++		log_debug_errno(r, "Failed to open system bus: %s", error_str());
+                 return EXIT_FAILURE;
+         }
+ 
+@@ -54,7 +54,7 @@
+                                "Released",
+                                "s", argv[1]);
+         if (r < 0) {
+-                log_debug_errno(r, "Failed to send signal message: %m");
++		log_debug_errno(r, "Failed to send signal message: %s", error_str());
+                 return EXIT_FAILURE;
+         }
+ 
+--- src/libelogind/sd-bus/bus-kernel.c	2015-04-19 13:32:24.607347468 +0200
++++ src/libelogind/sd-bus/bus-kernel.c	2017-03-04 16:53:10.410345905 +0100
+@@ -1161,7 +1161,7 @@
+ 
+                                 /* Anybody can send us invalid messages, let's just drop them. */
+                                 if (r == -EBADMSG || r == -EPROTOTYPE)
+-                                        log_debug_errno(r, "Ignoring invalid synchronous reply: %m");
++					log_debug_errno(r, "Ignoring invalid synchronous reply: %s", error_str());
+                                 else
+                                         return r;
+                         }
+@@ -1369,7 +1369,7 @@
+ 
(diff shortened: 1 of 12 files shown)


Gottox:
Nice!

pullmoll:
Please take look at if the `secure_getenv()` macro is sane enough or needs more tests.

Gottox:
Fails to start with:
```
[root@void-musl pullmoll-elogind]# /usr/libexec/elogind/elogind.wrapper 
Failed to create root cgroup hierarchy: Invalid argument
Failed to fully start up daemon: Invalid argument
[root@void-musl pullmoll-elogind]# 
```

Gottox:
Personally I used a simple -Dsecure_getenv=getenv macro whenever I thought it was safe enough, but your compatibility macro looks better.

pullmoll:
This is src/login/logind.c:711 function `manager_setup_cgroup()`.
Ah, the fscking `%m` is the culprit. Has to be replaced with `%s` a `strerror_r()` returned string to see what's wrong there.

pullmoll:
For logind.c now musl should print the real error message. There are more source files using "%m".

pullmoll:
Ok, now all `"%m"` are replace with "%s" and a new function `error_str()` defined in `src/shared/log.{c,h}`. That should make all error logging print sane things and work with glibc and musl equally well.

Gottox:
@wingo @Yamakuzure Is there any chance to get musl support for elogind back upstream?

Gottox:
According to the source. musl has support for %m: https://git.musl-libc.org/cgit/musl/tree/src/stdio/vfprintf.c#n615

Gottox:
Okay, I think I got the error tracket down:

The error happens when elogind checks for the existence of a cgroup in ```src/shared/cgroup-util.c:512```, the laccess macro calls ```faccessat(2)``` which return EINVAL. According to the manpage:
```
EINVAL Invalid flag specified in flags.
```

According to the source defined in https://git.musl-libc.org/cgit/musl/tree/src/unistd/faccessat.c#n36, musl does not support ```AT_SYMLINK_NOFOLLOW``` which is used in the laccess macro.

Gottox:
The patch:

cat 009-fix-faccessat-flags.patch 
```diff
--- src/shared/util.h.orig	2017-03-05 17:12:46.654974564 +0100
+++ src/shared/util.h	2017-03-05 17:14:11.998704156 +0100
@@ -1041,7 +1041,11 @@ union inotify_event_buffer {
         uint8_t raw[INOTIFY_EVENT_MAX];
 };
 
+#ifdef __GLIBC__
 #define laccess(path, mode) faccessat(AT_FDCWD, (path), (mode), AT_SYMLINK_NOFOLLOW)
+#else
+#define laccess(path, mode) faccessat(AT_FDCWD, (path), (mode), 0)
+#endif
 
 int ptsname_malloc(int fd, char **ret);
```

With this patch applied eloginctl seems to work fine.

Gottox:
The daemon runs, but loginctl does not show sessions. Anyway, with this patch applied it is good enough to merge.

pullmoll:
Ok, the 009-fix-faccessat-flags.patch is there now.

pullmoll:
> According to the source. musl has support for %m: https://git.musl-libc.org/cgit/musl/tree/src/stdio/vfprintf.c#n615

I wonder why the errors your reported were all just `Invalid argument` then.
While working on the patches, I also read one of those flamewar like threads regarding logind where someone told that the use of "%m" was causing problems for other libcs (possibly *BSD).

If the patch really isn't required, just drop it.

Gottox:
Imho it isn't required.

Gottox:
Okay, session tracking seems to work. The error was caused by a missconfigured pam on my end.

pullmoll:
Perhaps we can keep the patch in srcpkgs/elogind/files for the case that anyone with a libc who does definitely not support "%m" want to use it.

Gottox:
or just keep it in the git-history.

pullmoll:
Please merge. The musl builders have some things in the queue already :)

Gottox:
here we go :D

PR #5894, opened 2017-03-04
thunar-volman: add missing dependency gvfs

### LINKED ISSUE #5861: thunar-volman missing depedency: gvfs
Basically what it says on the tin. The package `thunar-volman` which is used for volume management in thunar does function correctly without `gvfs`. Specifically it lacks the ability to auto-mount volumes.

Simply installing `gvfs` with `xbps-install` fixes the problem.

Suggested solution: update template for `thunar-volman`.

CoolOhm:
This fixes #5861 

Pretty cut and dry, the volume management plugin for thunar can't manage volumes without a virtual filesystem being installed.


CoolOhm:
Actually I'm going to withdraw this request. See this [thread](https://forum.voidlinux.eu/t/solved-xfce-missing-trash-and-other-icons-on-the-desktop/2010/10).

PR #5898, opened 2017-03-04
slop: update to 5.3.37

E100Beta:
Latest version of `maim` needs `libslopy.so` for building, but `slop` utility needs `libslopy.so` as well. I looked at other packages, and they seem to provide `-devel` packages in such cases, but when I try with something like:

    slop-devel_pkg() {
        pkg_install() {
            vmove usr/include
            vmove usr/lib/*.so
        }
    }
`slop` doesn't get its `libslopy.so`. So as of now I build new version of `maim` with just `slop` in `makedepends` (I don't know if this is normal and I just overthink things, I judge only by the fact that all `makedepends` have `-devel` in them).

I couldn't solve this on my own and hope somebody here would help me with understanding this. Thanks in advance!


Vaelatern:
Possible example errors or patches to test these builds, and better understand your description? Thanks!

E100Beta:
@Vaelatern `maim` needs `slop` as both `depends` and `makedepends`. Do I need to do anything special with `slop` in this case, e.g. moving its header etc. to `slop-devel` package?

pullmoll:
If the main package does not have a version for the shlib (like libslopy.so.0.1.0) you can't move the `*.so` files in the devel package. This works only, iff libslopy.so is a symlink like `libslopy.so -> libslopy.so.0.1.0`, and possibly further symlinks like `libslopy.so.0.1 -> libslopy.so.0.1.0` and `libslopy.so.0 -> libslopy.so.0.1.0`
The common/shlibs entry then becomes `libslopy.so.0 slopy-5.3.37_1`

For a `cmake` style package, as in this case, there has to be a line to set the version number of a library:
```
set_target_properties(libslopy PROPERTIES VERSION 0.1.0)
```
somewhere in CMakeLists.txt
If it is not there, you should insert it using a patch or a `sed` in `pre_configure()`.

E100Beta:
@pullmoll Huge thanks for helping me! Everything seems to work and `maim` builds with `slop-devel` package. I'm going to make `maim` PR after this gets merged in case there are any more problems.

pullmoll:
@E100Beta just fix the xlint warning, i.e. use two tabs instead of spaces + tabs to indent the lines 23 + 24.
If you install `xtools` you can run `xlint srcpkgs/slop/template` locally to see if your template is ok.

E100Beta:
@pullmoll thanks again. Travis (entire site) doesn't open for me for some reason, so I couldn't see the error myself.

And I'm going to add (something like) a `./xbps-src $@; xlint srcpkg/${@: -1}/template` script to my local repository from now on :)

PR #5902, opened 2017-03-05
gvim: add python support

juliendehos:
some plugins need support for python (for instance youcompleteme: https://github.com/Valloric/YouCompleteMe#full-installation-guide)


ebfe:
Instead of adding another vim subpackage we could just build gvim with the same features as vim-huge. Opinions?

juliendehos:
Gvim with python support is quite heavier (RAM):
- current gvim = 10MB
- gvim-python = 20MB + 30MB (python interpreter)

Some gvim users may not need python support so maybe we can make two packages gvim and gvim-huge (in the same way as vim and vim-huge) ?

PR #5913, opened 2017-03-07
Updated hash and version number of remote files and updated revision.

ccheets:
Could do with testing on i386, but amd64 is good and working.


pullmoll:
See [Committing your changes](https://github.com/voidlinux/void-packages/blob/master/CONTRIBUTING.md#committing-your-changes) in CONTRIBUTING.md for how to format a PR's commit message.

In this case change the commit message to `spotify: update remote files versions`.

The fact that you bumped the revision is implicit with an update and does not need to be mentioned.

Vaelatern:
fcc1a2776b

PR #5916, opened 2017-03-07
python-smmap: Add missing dependency on python

lafrenierejm:
python-smmap has a dependency on the base Python install.  Explicitly
adding the python package as a dependency slipped through the cracks
initially because, as a Python library, there are very few use cases
where one would attempt to install smmap without already having the base
language.

The calling of `post_install()` is moved above the python3 subpackage in
the source to be more consistent with other Python packages.

### DIFF
--- a/srcpkgs/python-smmap/template
+++ b/srcpkgs/python-smmap/template
@@ -1,12 +1,13 @@
 # Template file for 'python-smmap'
 pkgname=python-smmap
 version=2.0.1
-revision=2
+revision=3
 noarch=yes
 wrksrc="smmap-${version}"
 build_style=python-module
 pycompile_module="smmap"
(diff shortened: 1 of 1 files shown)


srcpkgs/python-smmap/template
```diff
@@ -1,27 +1,30 @@
...
+	vlicense LICENSE
+}
+
 python3-smmap_package() {
 	noarch=yes
 	pycompile_module="smmap"
+	depends="python3"
 	short_desc="${short_desc/Python2/Python3}"
 	pkg_install() {
 		vmove usr/lib/python3*
```
Vaelatern: While here, add a vlicense?

lafrenierejm: Good call

PR #5917, opened 2017-03-07
RFC: gnu-makefile with good makefiles.

Vaelatern:
Cleaner than overwriting `do_build` partially every time we get a good makefile.
Needs a change to unset `good_makefile` as well, so it behaves well.
Tested on the PR of `spt` ( #5883 )

### DIFF
--- a/Manual.md
+++ b/Manual.md
@@ -645,7 +645,10 @@ arguments can be passed in via `configure_args`.
 - `gnu-makefile` For packages that use GNU make, build arguments can be passed in via
 `make_build_args` and install arguments via `make_install_args`. The build
 target can be overridden via `make_build_target` and the install target
-via `make_install_target`.
+via `make_install_target`. This build style tries to compensate for makefiles
+that do not respect environment variables, so well written makefiles, those
+that do such things as append (`+=`) to variables, should have `make_use_env`
+set in the body of the template.
 
 - `go` For programs written in Go that follow the standard package
   structure. The variable `go_import_path` must be set to the package's
--- a/common/build-style/gnu-makefile.sh
+++ b/common/build-style/gnu-makefile.sh
@@ -4,11 +4,15 @@
 do_build() {
 	: ${make_cmd:=make}
 
-	${make_cmd} \
-		CC="$CC" CXX="$CXX" LD="$LD" AR="$AR" RANLIB="$RANLIB" \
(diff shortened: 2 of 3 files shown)


chneukirchen:
I like the idea, but the naming is unclear... better proposals?

chneukirchen:
`make_use_env=1` perhaps?

common/build-style/gnu-makefile.sh
```diff
@@ -4,11 +4,19 @@
...
 
-	${make_cmd} \
+	if [ -z "$good_makefile" ]; then
+		${make_cmd} \
+			CC="$CC" CXX="$CXX" LD="$LD" AR="$AR" RANLIB="$RANLIB" \
+			CPP="$CPP" AS="$AS" OBJDUMP="$OBJDUMP" \
+			CFLAGS="$CFLAGS" LDFLAGS="$LDFLAGS" \
+			${makejobs} ${make_build_args} ${make_build_target}
+	else
 		CC="$CC" CXX="$CXX" LD="$LD" AR="$AR" RANLIB="$RANLIB" \
```
ebfe: These are all redundant, or?

chneukirchen: Are they exported by default?

ebfe: https://github.com/voidlinux/void-packages/blob/master/common/xbps-src/shutils/common.sh#L381

Vaelatern:
I like `make_use_env`

Vaelatern:
@pullmoll @chneukirchen variable changed, and added an `unset` in at least one correct place.

common/build-style/gnu-makefile.sh
Vaelatern: Agreed, probably redundant and extra copy pasting. Updated.

Manual.md
```diff
@@ -636,7 +636,9 @@ arguments can be passed in via `configure_args`.
 - `gnu-makefile` For packages that use GNU make, build arguments can be passed in via
 `make_build_args` and install arguments via `make_install_args`. The build
 target can be overridden via `make_build_target` and the install target
-via `make_install_target`.
+via `make_install_target`. This build style tries to compensate for poorly 
+written makefiles, but well written makefiles should have `make_use_env` set 
+in the body of the template.
```
roman-neuhauser: I loathe documentation based on innuendo and insinutation.  It's a *manual*, be clear!  What does "poorly written makefile" exactly mean?  What is the effect of {un,}setting `make_use_env`?

PR #5921, opened 2017-03-07
cross-*: enable fortran cross compilers and update glibc

pullmoll:
+ Enable the fortran language for all cross compilers.
+ Update glibc to 2.25 for the cross-*-gnu cross compilers.
+ Enable lapack cross compiling with `<triplet>-gfortran`. (separate PR)


pullmoll:
Will add the lapack PR after these are merged, because the build system doesn't recognize the need to build the cross compiler before building lapack.

PR #5922, opened 2017-03-08
New package: compiz-reloaded-0.8.12

CoolOhm:
This is the meta package for compiz window manager.


CoolOhm:
OK, I think I have made all the changes requested. If you would like any more changes, please let me know.

Sorry for the sloppy PR, I'll do better next time.

As always, thanks for all the help and patience @pullmoll, I really appreciate it. :)

PR #5926, opened 2017-03-08
qutebrowser: update to 0.10.1

NoctuaNivalis:


srcpkgs/qutebrowser/INSTALL.msg
```diff
@@ -0,0 +1,2 @@
+To use the new QtWebEngine backend, install python3-PyQt5-webchannel and
+python3-PyQt5-webengine.
```
The-Compiler: Doesn't python3-PyQt5-webengine depend on python3-PyQt5-webchannel anyways?

NoctuaNivalis: No, which is why this message is important. Everyone can figure out in a second you'll probably need python3-PyQt5-webengine for a webengine backend, but you need a few minutes and some error messages to figure out you need webchannel as well.

The-Compiler: I think this should be fixed in python3-PyQt5-webengine. I don't think QtWebEngine even builds without QtWebChannel, so the wrappers should depend on each other too.

NoctuaNivalis: Seems like it. Let's ask @asergi, the maintainer of python-PyQt5, if there's a reason not to have this dependency.

asergi: Fixed in 268f46fbf8c7e13bd7001a4a2f5776e2f0d425a3. Thank you.

PR #5935, opened 2017-03-09
New package: quilt-0.65

diogoleal:


srcpkgs/quilt/template
```diff
@@ -0,0 +1,17 @@
+# Template file for 'quilt'
+pkgname=quilt
+version=0.65
+revision=1
+build_style=gnu-configure
+configure_args="--prefix=/usr --sysconfdir=/etc --without-rpmbuild"
```
ebfe: `--prefix=/usr --sysconfdir=/etc` is already set by default: https://github.com/voidlinux/void-packages/blob/master/common/environment/configure/gnu-configure-args.sh#L7

diogoleal: Fixed!

PR #5933, opened 2017-03-08
cross*musl: use new kernel-libc-headers (missed files)

yopito:
This PR is an add to  #5811 : sorry guys, I forgot to check that `kernel-libc-headers/patches/*` were used in **each**  `cross-*musl/template`.

This should fix some stange broken builds that are fine with `cross-<arch>-musl` but `cross-aarch64-musl` regarding inclusion of various `*.h` related to network and using kernel userland interfaces (`linux/*.h`).
All broken `cross-aarch64-musl` starting from 1c977b0ba08bb50b6040d0f5af51cda07c4b4b33 (2017-02-24) might be reconsidered.

Example : (previous) NetworkManager : https://build.voidlinux.eu/builders/aarch64-musl_builder/builds/1391/steps/shell_3/logs/stdio

Again, really sorry to missed that.

PR #5942, opened 2017-03-09
various aarch64 musl unbreak

yopito:
following merge of PR  #5933, few packages are unbreak : fine with `cross-aarch64-musl` .
Do I need to bump revision of each package ?
Need to provide distinct PR for each ?


ebfe:
> Do I need to bump revision of each package ?
> Need to provide distinct PR for each ?
No, everythings fine.

Assigning @chneukirchen to merge, since this has to wait until the cross builders are rebuilt on aarch64.

PR #5943, opened 2017-03-09
New package: rustup-1.2.0

ibabushkin:
This is the rust toolchain manager most people use. I missed it in the repositories. Currently the build does some ugly tricks to avoid breakage (see comments in the `template`). I also haven't added any alternatives declarations yet, that's something to be discussed, as this package would obviously conflict with `rust` and `cargo`. Also, building this on musl might be possible, I haven't tried that yet.

### DIFF
--- a/srcpkgs/rustup/template
+++ b/srcpkgs/rustup/template
@@ -1,0 +1,47 @@
+# Template file for 'rustup'
+pkgname=rustup
+version=1.2.0
+revision=1
+only_for_archs="i686 x86_64"
+conflicts="rust cargo"
+wrksrc="${pkgname}.rs-${version}"
+hostmakedepends="cargo rust pkg-config perl"
+makedepends="libressl-devel libtls15 zlib-devel"
+short_desc="An installer for the systems programming language Rust"
+maintainer="Inokentiy Babushkin <twk@twki.de>"
+homepage="https://rustup.rs/"
+license="MIT, Apache-2.0"
+distfiles="https://github.com/rust-lang-nursery/${pkgname}.rs/archive/${version}.tar.gz"
+checksum="b79434f7c893468d122755ec7f8c4a41643fa4fdd951d2d5453ef93f55921a31"
+nocross=yes
+
+do_build() {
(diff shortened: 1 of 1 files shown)


pullmoll:
Sorry, clicked the wrong button. The "Merge brach 'master' into rustup" is crap. Seems I need some sleep :)

srcpkgs/rustup/template
```diff
@@ -0,0 +1,51 @@
+# Template file for 'rustup'
+pkgname=rustup
+version=1.0.0
+revision=1
+only_for_archs="i686 x86_64"
```
pullmoll: This is almost always the wrong approach. only_for_archs is meant for packages which by design won't build for other architectures (assembler parts, arch specific hardware), not for the "won't build" case.
Just remove that line and let's see if and how musl fails.

ibabushkin: I will try. IIRC I decided to do so because upstream didn't provide musl builds, while rust supports building for musl - which seemed odd.

ebfe: @pullmoll the package would probably build, but be useless. iirc rustup installs the upstream rust binary releases which are not available for musl systems.

pullmoll: @ebfe ok, this is good a reason to keep only_for_archs. I'll try to remember that for rust related PRs.

ibabushkin:
Alright, seems to build. I'm not sure, however, whether the correct target triple gets set on musl - that would require some checking.

srcpkgs/rustup/template
ibabushkin: Yep, see this [reddit thread](https://www.reddit.com/r/rust/comments/5iipmg/rustup_linked_statically_with_musl/). I guess having a working rustup could still be beneficial on a musl system, as it allows you to multiplex locally built toolchains - but it's quite likely impossible to build those. I guess I'll undo the change then.

ibabushkin:
As mentioned on IRC:
It possibly makes sense to move the invocation of rustup-init to `do_install` - so the user gets the latest stable at the time of installation. This would require us to ship the rustup-init binary to everyone, however.

asergi:
#10613

PR #5948, opened 2017-03-10
New package: python-pyaes-1.6.0

ananteris:


srcpkgs/python-pyaes/template
```diff
@@ -0,0 +1,16 @@
...
+pkgname=python-pyaes
+version=1.6.0
+revision=1
+noarch=yes
+wrksrc="pyaes-${version}"
+build_style=python2-module
+hostmakedepends="python-setuptools"
+depends="python"
+pycompile_module="pyaes.py"
+short_desc="Pure-Python Implementation of the AES block-cipher and common modes of operation"
```
pullmoll: I think `short_desc="Pure Python implementation of the AES block-cipher"` would suffice. See xlint warning about the length of short_desc being > 72.

srcpkgs/python-pyaes/template
```diff
@@ -0,0 +1,16 @@
...
+maintainer="ananteris <ananteris@protonmail.ch>"
+homepage="https://pypi.python.org/pypi/pyaes"
+license="MIT"
```
pullmoll: `license="MIT"` requires a `post_install()` with `vlicense LICENSE.txt`.

srcpkgs/python-pyaes/template
ananteris: Fixed

srcpkgs/python-pyaes/template
ananteris: Fixed

srcpkgs/python-pyaes/template
```diff
@@ -0,0 +1,20 @@
+# Template file for 'python-pyaes'
+pkgname=python-pyaes
+version=1.6.0
+revision=1
+noarch=yes
+wrksrc="pyaes-${version}"
+build_style=python2-module
+hostmakedepends="python-setuptools"
+depends="python"
+pycompile_module="pyaes.py"
```
asergi: `pycompile_module="pyaes"`

srcpkgs/python-pyaes/template
```diff
@@ -0,0 +1,20 @@
+# Template file for 'python-pyaes'
+pkgname=python-pyaes
```
asergi: Add python3 flavor. See `python-sh` as example.

srcpkgs/python-pyaes/template
ananteris: @asergi what is the significance of this? there are 310 packages already in the repo that have a template that uses pycompile_module="${module}.py" entry.. should all of these be updated?

ananteris: I see what the difference is now, infact python-pycrypto should not be it's own package, it's already provided by python-crypto. I've removed it.

asergi:
fcee34f61a3a0756eba93641443f190578bbe3f3

PR #5950, opened 2017-03-10
[WIP] New package: amdgpu-pro-16.60.379184

pullmoll:
+ Bought a Radeon RX 460 which doesn't work with linux4.10 (corrupt display, system dead)
+ Template based on AUR package amdgpu-pro (x86_64)
+ Untested. I didn't dare to ruin my Void before a peer review.
+ Unsure if identical compatibility symlinks in multiple subgpkgs are ok.
+ Need help with conflicts, shlib_provides, ...
+ alternatives for libs?


masato9000:
I'm giving this a shot (I have a VM I can sacrifice with an R9 passed through).  I'll post any results I get.

However, right off the bat I can say that building on 4.10 fails due to the change from struct fence to struct dma_fence, so some patchwork will be needed.  I don't know anything about the dkms system, so I don't know how that would be done (you only want to apply to 4.10 kernels or you'll break LTS; Can this be done?)  I'm going to go ahead and try it on 4.9 at any rate...

pullmoll:
Forgot to say that 4.10 is working now. It was not the kernel, but the firmware which needed an update. The previous linux-firmware-amd was broken for the polaris10 and polaris11 GPUs.

I don't know if and which advantages you can expect from amdgpu-pro and if it's worth trying.
If there isn't strong demand for this proposed pkg, I'd rather close the PR.

masato9000:
I was just trying it because the PR was here and I can use VMs for this without endangering my real system.  I don't normally use the r9 in Linux (it's for an os-that-shall-not-be-named)

Building just the module seems to work fine in 4.9.  I've confirmed it's loading.  X works with the regular mesa stack (including the regular amdgpu module).  I didn't get around to testing the AMD xorg module stack.  I was going to try it with regular mesa first, though I really expect there would be problems just like with nvidia, and those probably will be requirements.

At any rate, if there _does_ end up being a demand for this, then the struct fence issue will have to be addressed for the newest kernels.  I may toy around with it some, and if I find anything interesting, I'll note it, but if you'd rather close this, it won't make me cry...

ibrokemypie:
+1 demand for amdgpu pro, not on amd at this current moment, but probably getting a 480 or vega rx in the coming months

PR #5951, opened 2017-03-10
Update libtirpc and rpcbind

Hoshpak:


common/shlibs
```diff
@@ -2799,3 +2798,5 @@ libomxvideosched.so.0 libomxil-bellagio-0.9.3_1
 libnss_libvirt_guest.so.2 libvirt-3.1.0_1
 libvirt-admin.so.0 libvirt-3.1.0_1
 libnss_libvirt.so.2 libvirt-3.1.0_1
+
```
pullmoll: You should remove that empty line. I think it was my fault to not have removed it when merging that PR?

pullmoll:
Looks good to me but I'd suggest to wait for @xtraeme 's approval for the patches.

common/shlibs
Hoshpak: Removed it. I'm not sure where exactly the empty line comes from. I didn't insert it but noticed on several occasions that one was inserted when doing a rebase on top of master. As you can see, I do not always notice and remove it, especially if I rebase the same branch several times.

ebfe:
https://www.cvedetails.com/cve/CVE-2017-8779/

PR #5952, opened 2017-03-10
New package: sc-controller-0.3.8

Vintodrimmer:


srcpkgs/sc-controller/template
```diff
@@ -0,0 +1,14 @@
+# Template file for 'sc-controller'
+
+pkgname=sc-controller
+version=0.3.8
+revision=1
+build_style=python2-module
+makedepends="python-setuptools"
+depends="gtk+3 python-gobject python-cairo"
```
Vaelatern: `$ xmindep gtk+3 python-gobject python-cairo
gtk+3 python-gobject
`
What is the purpose of `gtk+3` here?

Vintodrimmer: If I understand it correctly, it uses gtk+ 3 for the interface.

Vaelatern: Then I wonder why it does not link against gtk+3

Vintodrimmer: Hmm, let's summon @kozec and ask him/her.

kozec: Him :)

As @Vintodrimmer said, sc-controller uses gtk3 for GUI, but there is no linking as GUI is written in python.

Vintodrimmer:
Anything else I should do, besides waiting?

Vintodrimmer:
Fixed the checksum and added dependency, pointed out by @KeepBotting in https://github.com/voidlinux/void-packages/pull/6293.

Since I have no Idea, what to do in such situations, I'll just keep the package up to date.

asergi:
> Dependencies
>
>  python 2.7, GTK 3.10 or newer and PyGObject
>  python-gi-cairo and gir1.2-rsvg-2.0 on debian based distros (included in PyGObject elsewhere)
>  python-pylibacl
>  setuptools

`python-pylibacl` is missing or is that a optional dependency?

KeepBotting:
Program works fine without it, from my end it looks like it's optional. Logic dictates that it might be required for the program to work on systems making use of ACLs but I don't have a system configured like that to test with.

kozec:
ACL is used to check if user has access to `/dev/uinput` and display _warning_ if not. If ACL is not available, it just checks for file mode, so that warning may be displayed unnecessarily under some circumstances. But everything will work even in such case.

Vintodrimmer:
Can confirm that I don't have `python-pylibacl` installed and `sc-controller` works as intended, without any additional wornings.

What would be the correct way to implement such a dependency, if it's a python module?

srcpkgs/sc-controller/template
```diff
@@ -0,0 +1,14 @@
+# Template file for 'sc-controller'
+
+pkgname=sc-controller
+version=0.3.11
+revision=1
+build_style=python2-module
```
asergi: `pycompile_module="scc"` is missing.

KeepBotting:
@Vintodrimmer you're a real trooper, keeping this package updated so well :package: 

@asergi Any chance for another review on this? Been needing this package in the repo recently and it's still sitting as a PR :(

Vintodrimmer:
@KeepBotting 

That's strange. Line 7 in the template is `pycompile_module="scc"`

KeepBotting:
@Vintodrimmer So it is! Can't see any reason not to get this merged then! It's been sitting here for long enough.

codingHahn:
@Vaelatern Something holding this back?

PR #5959, opened 2017-03-11
blender: unbreak musl (Alpine patch)

ebfe:
Patches from Alpine and upstream. jemalloc seems broken on musl (seen this in other packages too), so added a buildoption to disable it on musl.

### DIFF
--- a/srcpkgs/blender/template
+++ b/srcpkgs/blender/template
@@ -1,14 +1,14 @@
 # Template file for 'blender'
 pkgname=blender
 version=2.78c
-revision=1
+revision=2
 build_style="cmake"
 makedepends="
  libgomp-devel libpng-devel tiff-devel python3-devel glu-devel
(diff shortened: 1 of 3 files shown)


srcpkgs/blender/template
```diff
@@ -42,6 +43,11 @@ configure_args="
...
 -DPYTHON_INCLUDE_DIRS=$py3_inc"
 
+build_options="jemalloc"
+build_options_default=""
 case "$XBPS_TARGET_MACHINE" in
-	*-musl) broken="http://build.voidlinux.eu/builders/x86_64-musl_builder/builds/3998/steps/shell_3/logs/stdio";;
+	*-musl) # crashes at startup in jemallocs free()
+		;;
+	*) build_options_default+=" jemalloc"
+                ;;
```
Gottox: Tabs vs Spaces

ebfe: xlints fixed

PR #5961, opened 2017-03-11
New package: rix-0.3.0

diogoleal:


pullmoll:
No manpage? No `/usr/share/doc/rix`?
It doesn't build for musl libc because `rix.y` is probably missing an include?

Sorry, I have to give :-1:

PR #5962, opened 2017-03-11
New package: fuse-emulator

skrzyp:
Back to pushing packages after some months of inactivity.

**FUSE** is a **F**ree **U**nix **S**pectrum **E**mulator.

I had packaged it locally for some days, works well, I'm using it together with Z80 assembler.

About ROM files inside: Amstrad (current holder of ZX Spectrum copyrights) allowed to use them freely with any emulator, redistribute, disassemble or patch, with copyright strings retained.

**NOTE**: `fuse(1)` is different than `fuse(8)`, I don't see a problem here.

### DIFF
--- a/srcpkgs/fuse-emulator/template
+++ b/srcpkgs/fuse-emulator/template
@@ -1,0 +1,15 @@
+# Template file for 'fuse-emulator'
+pkgname=fuse-emulator
+version=1.3.4
+revision=1
+wrksrc="fuse-${version}"
+build_style=gnu-configure
+configure_args="--verbose --with-gtk --with-libspectrum-prefix=/usr --enable-desktop-integration"
+hostmakedepends="pkg-config perl"
+makedepends="gtk+-devel alsa-lib-devel glib-devel libao-devel libgcrypt-devel libspectrum-devel zlib-devel libpng-devel"
+short_desc="Free Unix Spectrum Emulator"
+maintainer="Jakub Skrzypnik <j.skrzypnik@openmailbox.org>"
+license="GPL-2"
+homepage="http://fuse-emulator.sourceforge.net"
+distfiles="${SOURCEFORGE_SITE}/${pkgname}/fuse-${version}.tar.gz"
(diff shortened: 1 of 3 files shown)


pullmoll:
IMHO you should rename the binary and manpage to the pkgname to avoid confusion, though (pun? not intended). Why would we force people to prepend a number just to get to the fuse (not fuse-emulator) manpage?
```
post_install() {
        mv ${DESTDIR}/usr/bin/fuse{,-emulator}
        mv ${DESTDIR}/usr/share/man/man1/fuse{,-emulator}.1
        mv ${DESTDIR}/usr/share/applications/fuse{,-emulator}.desktop
        sed -i ${DESTDIR}/usr/share/applications/fuse-emulator.desktop -e 's|fuse|fuse-emulator|g'
}
```
or something similar.

skrzyp:
But we can't easily change other paths, like `/usr./share/fuse` (explicitly hardcoded in source, despite being configurable by autoconf script)...
I think it's just a name for userland software, while fuse is a name of kernel<->user space filesystem support. This is why `man` has groups 
```
                   1         General commands (tools and utilities).
                   2         System calls and error numbers.
                   3         Library functions.
                   3p        perl(1) programmer's reference guide.
                   4         Device drivers.
                   5         File formats.
                   6         Games.
                   7         Miscellaneous information.
                   8         System maintenance and operation commands.
```

Vaelatern:
I disagree with @pullmoll on this. Upstream decided it should be this way, and there is nothing technically wrong here, and it doesn't conflict with any other packages either.

pullmoll:
I retract from commenting on this. Do whatever you think is good.

skrzyp:
IMO We should just keep it as is :)

skrzyp:
Bump for visibility, as I recently got to install this package on more than 2 machines.

skrzyp:
What happened with that?

Gottox:
I'm not sure about the binary name, as it tends to conflict with `libfuse`. Anyway, get this merged. In doubt we can rename the binary later.

please rebase and ping me afterwards.

PR #5963, opened 2017-03-11
mono: update to 4.8.0.495

Hoshpak:


srcpkgs/mono/template
```diff
@@ -28,3 +28,12 @@ pre_configure() {
 post_install() {
 	vlicense LICENSE
 }
+
+mono-devel_package() {
+	short_desc+=" - development files"
+	depends="${sourcepkg}>=${version}_${revision}"
+	pkg_install() {
+		vmove usr/include
+		vmove "usr/lib/*.a"
```
pullmoll: vmove usr/lib/pkgconfig

Hoshpak: Right, thanks. I added it.

PR #5965, opened 2017-03-11
New package: kakoune-20170311

skrzyp:
Kakoune is a vim-like text editor, but using a different approach to text editing. It's mostly selection based, uses multiple cursors and has improved (more efficient) key grammar.

Despite being "unreleased" it's pretty usable, I'm using it since late 2015, exclusively since past few months. Did some serious work with it, nothing bad happened.

I've asked the author about versioning scheme, he just said that's only the marketing matter and still not decided about versioning scheme. Kinda weird, but software itself works very well.

### DIFF
--- a/srcpkgs/kakoune/template
+++ b/srcpkgs/kakoune/template
@@ -1,0 +1,18 @@
+# Template file for 'kakoune'
+pkgname=kakoune
+version=0.0.20170311
+revision=1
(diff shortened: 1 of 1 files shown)


chneukirchen:
Now we have gcc 6, this should build fine.

I'd change the version to `0.0.YYMMDD` in case the author starts to do proper relases some day.

Else +1

skrzyp:
Good idea, let me fix this

PR #5966, opened 2017-03-11
rpi-kernel: update to 4.9.13

pbui:
- Move to latest LTS kernel (4.9)

- I'm not 100% about the header situation, but the `arch/arm/mach-bcm2709` directory no longer exists in 4.9, so I removed those lines from the template

- I've tested both the armv6l and armv7l kernels on my Raspberry Pis.


E100Beta:
@pbui Travis wants shorter descriptions, I suggest removing the part about git revision:

    srcpkgs/rpi-kernel/template:13: short_desc should be less than 72 chars
    srcpkgs/rpi-kernel/template:180: short_desc should be less than 72 chars

I don't know if it's the right place to ask, but I was curious - how do you choose which git commit to base the package on for raspberry kernels? Just the last commit, or is there some other criterion?

pbui:
@E100Beta I didn't change the description (that was in put in place by the original package author @xtraeme).  I'm fine with shortening it, but I don't know if it is my place to update it.

In regards to which git revision, I usually pick the one after merging/rebasing with the mainline (ie. 4.9.13).  In this case, however, I picked one that include some bugfixes after that rebase.

CMB:
Tested with 4.9.25 on an rpi3 running the rpi2 ARMV7 musl image.
Seems good.

pbui:
I've updated to 4.9.29 and tested it on both armv6l and armv7l.

PR #5973, opened 2017-03-12
qemu: add smartcard support

Sir-Boops:
First build will fail.

### DIFF
--- a/srcpkgs/libcacard/template
+++ b/srcpkgs/libcacard/template
@@ -1,0 +1,24 @@
+# Template file for 'libcacard'
+pkgname=libcacard
+version=2.5.2
+revision=1
+build_style=gnu-configure
+hostmakedepends="pkg-config"
+makedepends="nss-devel glib-devel pcsclite-devel"
+short_desc="A library to provide emulation of smart cards to a virtual card reader"
+maintainer="Sir_Boops <admin@boops.me>"
+homepage="https://www.spice-space.org/page/Libcacard"
+license="LGPL-2.1"
+distfiles="https://www.spice-space.org/download/libcacard/libcacard-${version}.tar.xz"
+checksum=2e39d52346a64af3bd7ee6bff658c4dcfee05a5ab8c0cf70d225fa58d3df4ecb
+
+libcacard-devel_package() {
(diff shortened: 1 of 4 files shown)


Sir-Boops:
The i686 build timed, so everything seems good

srcpkgs/qemu/template
```diff
@@ -19,7 +19,7 @@ makedepends="libpng-devel libjpeg-turbo-devel pixman-devel snappy-devel
  libbluetooth-devel libssh2-devel libusb-devel pulseaudio-devel libnfs-devel
  $(vopt_if sdl SDL-devel) $(vopt_if sdl2 SDL2-devel) $(vopt_if gtk gtk+-devel)
  $(vopt_if spice spice-devel) $(vopt_if virgl virglrenderer-devel)
- $(vopt_if opengl 'libepoxy-devel libdrm-devel MesaLib-devel')"
+ $(vopt_if opengl 'libepoxy-devel libdrm-devel MesaLib-devel') libcacard-devel"
```
Gottox: Can you add it as a build_option?

Sir-Boops: Added as a default for i686 & x86_64

Duncaen: Do we need this by default, I really dont want it.

Sir-Boops: No, we really don't need it as a default option, i only added it as one as i wasn't aware some don't want it. I'll remove it from the default build options.

Sir-Boops:
Put adding libcacard into it's own commit

pullmoll:
For consistency please use `New package: libcacard-2.5.2` as the first commit message.

srcpkgs/libcacard/template
```diff
@@ -0,0 +1,24 @@
+# Template file for 'libcacard'
+pkgname=libcacard
+version=2.5.2
+revision=1
+build_style=gnu-configure
+hostmakedepends="pkg-config"
+makedepends="nss-devel glib-devel pcsclite-devel"
+short_desc="A library to privide emulation of smart cards to a virtual card reader"
```
ebfe: typo

PR #5975, opened 2017-03-12
collectd: Update to 5.7.1

coredumb:
- Added a bunch more modules
  - Split as much as it made sense
  See discussion on issue 5463


ebfe:
Merging this as is will break installations/configs that use any of the plugins i guess.

coredumb:
Probably yes

Vaelatern:
Maybe we can make a metapackage, like the xorg situation?

scadu:
So, any decision? I wanted to bump collectd to 5.8.0 and saw this pr.
I'm not sure if many people uses Void on turbo production.
We can always add a nice info about package split in INSTALL.msg.

PR #5991, opened 2017-03-16
Sigrok

lemmi:
http://sigrok.org/wiki/Main_Page

### DIFF
--- a/common/shlibs
+++ b/common/shlibs
@@ -2818,3 +2818,4 @@ libmutter-cogl-pango-0.so mutter-3.24.0_1
 libmozjs-.so mozjs38-38.8.0_1
 libostree-1.so.1 libostree-2017.3_1
 librhash.so.0 rhash-1.3.4_1
+libserialport.so.0 libserialport-0.1.1_1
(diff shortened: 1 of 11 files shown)


pienjo:
pulseview and sigrok-cli list a runtime dependency for sigrok-firmware-fx2lafw, but is this really appropriate? This dependency is required for a specific group of logic analyzers, but is not needed for other hardware supported by pulseview (or sigrok-cli)

pienjo:
Libsigrok comes with a useful set of default udev-rules (in contrib/z60_libsigrok.rules), but these don't get installed by default. I would suggest installing this file somewhere along with a gentle reminder.

pienjo:
pulseview comes with a set of icons. For some reason these don't get installed (which makes setting trigger modes rather tricky) Sorry to be a nuisance :/

lemmi:
Hey thanks for looking over that. I'm having another go at the packages this week, once i caught up on some other things.
I know about the udev rules, but we (void) don't use the plugdev group anywhere and I haven't figured out the right approach to that. As for the firmwares: it's not easy enough to mark optional dependencies, which is why I prefer a working environment with a few more packages.

PR #5994, opened 2017-03-16
New package: glyr-1.0.10

autrimpo:
A cli tool and a library for searching for music metadata.


srcpkgs/glyr/template
```diff
@@ -0,0 +1,30 @@
...
+homepage="https://github.com/sahib/glyr"
+distfiles="https://github.com/sahib/glyr/archive/${version}.tar.gz"
+checksum=77e8da60221c8d27612e4a36482069f26f8ed74a1b2768ebc373c8144ca806e8
+
+libglyr_package() {
+	short_desc="A music metadata searchengine library"
+	pkg_install() {
+		vmove "usr/lib/*.so.*"
+	}
+}
```
Vaelatern: What is the purpose of this subpackage?

srcpkgs/glyr/template
```diff
@@ -0,0 +1,30 @@
+# Template file for 'glyr'
+pkgname=glyr
+version="1.0.10"
+revision=1
+build_style=cmake
+hostmakedepends="pkg-config"
+makedepends="libcurl-devel glib-devel sqlite-devel"
+depends="libcurl glib sqlite"
```
Vaelatern: Not necessary.
https://travis-ci.org/voidlinux/void-packages/jobs/211925782#L870-L873
```
=> libglyr-1.0.10_1: running pre-pkg hook: 04-generate-runtime-deps ...
   SONAME: libcurl.so.4 <-> libcurl>=7.19_1
   SONAME: libglib-2.0.so.0 <-> glib>=2.18.0_1
   SONAME: libsqlite3.so.0 <-> sqlite>=3.8.11.1_3
   SONAME: libc.so.6 <-> glibc>=2.8_1
```

srcpkgs/glyr/template
Vaelatern: Likely should be removed. The main package can not function without it, and it doesn't appear to serve a constructive purpose.

srcpkgs/glyr/template
```diff
@@ -0,0 +1,30 @@
...
+		vmove "usr/lib/*.so.*"
+	}
+}
+
+libglyr-devel_package() {
+	depends="libglyr>=${version}_${revision}"
+	short_desc="A music metadata searchengine library - development files"
+	pkg_install() {
+		vmove usr/include
+		vmove usr/lib/pkgconfig
```
Vaelatern: And `vmove usr/lib/*.so`

common/shlibs
```diff
@@ -2799,3 +2799,4 @@ libomxvideosched.so.0 libomxil-bellagio-0.9.3_1
 libnss_libvirt_guest.so.2 libvirt-3.1.0_1
 libvirt-admin.so.0 libvirt-3.1.0_1
 libnss_libvirt.so.2 libvirt-3.1.0_1
+libglyr.so.1 libglyr-1.0.8_1
```
Vaelatern: When you remove the subpackage (unless you can argue for why it should stay), this should be changed as well.

srcpkgs/glyr/template
autrimpo: Some software (like lyvi which I intend to package later) requires the library but not the cli tool. When packaging glyr I used curl as a reference, which splits the packages in the same way. I find this split reasonable, since the tool and the library are two separate entities and users (me as well actually) might be interested in just the library, not the tool . One could argue this is needless clutter just to save a few KB of disk space, but then again, so are the -devel subpackages in a lot of cases. If you still think it is unreasonable I'll remove it though.

srcpkgs/glyr/template
autrimpo: I assume this was meant in the case of the lib subpackage being removed. May I ask why put it into the -devel subpackage instead of the main package?

autrimpo:
In the end I've removed the lib subpackage since I didn't get any response and fixed all the mentioned issues.

Duncaen:
Merged in 965b905ebc2e3e87088833f3befa735fb04266a1.

PR #5944, opened 2017-03-09
icu-58.2, libevent-2.1 revbumps and firefox{,-esr}, thunderbird and chromium update

Duncaen:

PR #5995, opened 2017-03-16
libwebp: update to 0.6.0. (soname bump)

ebfe:


ebfe:
Test built everything on x86_64.

pullmoll:
Another job that will be taking several days to finish. We really ought to be sponsored by someone who gives us (free) access to powerful hardware and sufficient disk space :)

Perhaps we should contact hetzner.de and ask them for a favor.
I think not only my voidlinux.de is hosted there (now with 200GiB disk space) but also the official repo(s) and some of the builders?

ebfe:
Ideally this is merged/pushed with #5944. I'm not sure the slow builds are purely a lack of hardware issue. The current qt5 build took 12h (failed with out of space...), which seems unreasonable. Maybe we should look in how to utilize existing hardware better. Of course free hardware would be good too ;)

pullmoll:
Yeah, reconciling this revbump with #5944 would certainly help.

For the builders I think one bottleneck is memory and thus also the available tmpfs space. Having the masterdirs below /tmp would help *tremendously*.

But memory on the aarch64* builders e.g. is just 8GiB and that doesn't even suffice to link the Qt5WebEngine library. Recently @chneukirchen added another 8GiB or so of swap space, and of course swapping is slow. More physical memory would be desirable in this case.

Then some builders always seem to be short on disk space, or at least run out of disk space every couple of days. This could perhaps be solved by removing outdated distfiles with `./xbps-src purge-distfiles` once a day. Handling packages is another story and difficult or impossible to handle in a cron job.

I could try and contact the hetzner.de team and ask them, if they would be willing to support our joint effort. I think we would then of course mention them and put their logo in prominent places on voidlinux.eu, the forum, the wiki etc.

biopsin:
Think ImageMagic template needs a patchver update too, as 0 is no longer availible
```
pkgname=ImageMagick
_majorver=6.9.8
_patchver=2
version=${_majorver}.${_patchver}
revision=2
checksum=be809993975dbc84b69987a8b914b656849351396c519b54199083967fceb143
```

ebfe:
This was merged into #5944.

PR #6000, opened 2017-03-18
encfs: update to 1.9.1.

lemmi:


lemmi:
1.9 switches to cmake, which makes cross compilation feasible. Maybe someone can test this, don't know whether there were breaking changes.

Issue #6006, opened 2017-03-19
jack source location has changed


biopsin:
new location..
```
# git
distfiles="https://github.com/jackaudio/jack2/archive/v${version}.tar.gz"
checksum=5bc6336e6ac9799e3cb241915e2ba5d01b030589bbb2afae39579a59ef0f2f56

# wrksrc location
wrksrc=jack2-${version}
```

PR #6007, opened 2017-03-19
jack: change distfiles location

Vaelatern:
#6006 
Letting travis test.
This does not do a rebuild so we can merge this without worrying about the builders.


Vaelatern:
So the sum changed. Will compare the results to see if any actual change.

Gottox:
```diff
# diff -upr jack-1.9.10 jack2-1.9.10
Only in jack2-1.9.10: .gitignore
```

PR #6009, opened 2017-03-19
RFC: Apparmor implementation

coredumb:
**Purpose**: Provide a fully functional AppArmor system for Void
**Requirements**: 
  - https://github.com/voidlinux/void-runit/pull/66
  - https://github.com/coredumb/void-apparmor-profiles
  I've made the choice to put AppArmor profiles in a dedicated repo as it makes it easier to track modifications.
The goal is to provide profiles tailored for Void default services/applications.
Right now only 3 profiles are provided:
    - usr.bin.iputils-ping
    - usr.bin.dhcpcd
    - usr.bin.ssh

**Usage**: `ln -s /etc/sv/apparmor/ /var/service/`
A limitation of AppArmor is that profiles **must** be loaded before the process starts. As a result services that need to be confined need to depend on 'apparmor' service in their run script.
`[ ! -d /var/service/apparmor/ ] || grep sshd /sys/kernel/security/apparmor/profiles || exit 1`

**TODO**: 
  - Write more profiles
  - More test on different configurations


thypon:
Could you provide system benchmark analysis?

We should document these three configurations
- Apparmor Enabled / Kernel Compiled with Apparmor
- Apparmor Disabled / Kernel Compiled with Apparmor
- Kernel compiled without apparmor

You could use `unixbenchmark` with https://github.com/amluto/virtme to test

thypon:
BTW i think it's a good idea to provide a pluggable apparmor/selinux/tomoyo system instead of hardwiring apparmor to default void-runit.

coredumb:
"hardwiring" AppArmor in void-runit as a core-service is important as it is required for AppArmor profiles to be loaded before the confined process start.
I don't see any other way to do this nicely except than putting it into void-runit core-services.

For the support of other LSM, if it is required, I don't think there would be any problem to have them plugged in core-services the same way with an option in rc.conf.

Now the problem with any of these MAC systems is that they're useless without any rules/profiles and supporting 2 main ones - AFAIU Tomoyo is using binary format for its rules that may not be suitable for packaging (I may be wrong about that though) - would require a substantial amount of work.
Enabling them without providing any default sane rules, along with the userlands, to be used out of the box pretty much ensures that they won't ever be used.

Although it's a small start here, I'm trying to provide a useful *and* fully functional AppArmor system.
Useful in the sense that it comes with a default set of profiles covering basic/defaults services for Void.

For the benchmarks, let me find some time to run all that :)

thypon:
I'm not arguing that apparmor should not be loaded during startup phase before every service. I'm only saying that it should stay outside the main core services maybe in a -apparmor subpackage of void-runit.
Maybe parsing the commandline or using /{proc,sys} is enough to know which LSM is currently loaded.

coredumb:
Ok I get you!
Let me think this through

coredumb:
Ok so I added a subpkg runit-void-apparmor.
We still need 00-pseudofs.sh to mount securityfs. I've remade the PR as https://github.com/voidlinux/void-runit/pull/67

The core-service script deployed will in turn silently exit if apparmor is not the default LSM
In case we enable multiple LSM one would have to set "security=<LSM>" in grub

srcpkgs/runit-void/files/09-apparmor.sh
```diff
@@ -0,0 +1,30 @@
...
+if [ "$APPARMOR" != "disable" -a -n "$APPARMOR" ]; then
+    if [ "$APPARMOR" != "complain" -a "$APPARMOR" != "enforce" ]; then
+        printf '! AppArmor incorrect option %s - ignoring profiles\n' "$APPARMOR"
+        exit 0
+    fi
+
+    [ "$APPARMOR" = "complain" ] && AACOMPLAIN="-C"
+
+    if [ -d /etc/apparmor.d -a -x /usr/bin/apparmor_parser ]; then
+        for profile in $(find /etc/apparmor.d -maxdepth 1 -type f); do
```
Duncaen: maybe just use `for profile in /etc/apparmor.d/*; do`? or maybe with .conf, not sure if its necessary to check if there are files or folders, i don't think this will fail.

srcpkgs/runit-void/files/09-apparmor.sh
```diff
@@ -0,0 +1,30 @@
...
+# AppArmor is not enabled in kernel, silently exit
+[ ! -d /sys/kernel/security/apparmor ] && exit
+
+# Load config
+[ -r /etc/rc.apparmor ] && . /etc/rc.apparmor
+
+# Proceed to load profiles depending on user settings
+msg "Loading AppArmor profiles..."
+
+if [ "$APPARMOR" != "disable" -a -n "$APPARMOR" ]; then
```
Duncaen: I guess it would be nicer to interpret unset as disabled, this is the standard in other tools afaik.
just check for `-n "$APPARMOR"`.

srcpkgs/runit-void/files/09-apparmor.sh
```diff
@@ -0,0 +1,30 @@
+# vim: set ts=4 sw=4 et:
+
+# AppArmor is not enabled in kernel, silently exit
+[ ! -d /sys/kernel/security/apparmor ] && exit
```
Duncaen: I think you cant just exit here, the files are sourced afaik, this would break booting.

srcpkgs/runit-void/files/09-apparmor.sh
coredumb: apparmor_parser will then try to load everythin recursively as well as files supposed to be included and not loaded directly which ends with a lot of error.

Indeed you end up with your profiles loaded but I think it's nicer to only load profiles

srcpkgs/runit-void/files/09-apparmor.sh
coredumb: You're right I overlooked that, I should use return instead of exit
Will fix

srcpkgs/runit-void/files/09-apparmor.sh
coredumb: I wanted to match explicit "disable" as well but you're right check on unset would be sufficient 
will fix

coredumb:
So I made some unixbenchmark runs on linux4.10-4.10.4_1. 
The result is an average on 5 runs on 2CPUs VM:
  - No LSM enabled (Void repo)
    - Index score: **1650**
  - AppArmor enabled Kernel with securityfs not mounted
    - Index score: **1412**
  - AppArmor enabled and the 3 profiles loaded
    - Index score: **1661**

I'm pretty sure the result of the run with AppArmor disabled is not relevant. I'll see if I have 2.5h ahead of me to rerun it :)

thypon:
The results are totally unexpected. I would expect a slightly slower performance on Apparmor kernel enabled (it needed profiling enabled) and worsen with Apparmom userspace enabled.

I think i have to run the tests myself :S

coredumb:
When you have ~3% drop with a full grsec kernel I wouldn't expect anything noticeable
with just a LSM module enabled.
Now benchmarking a userland app confined would certainly give some more relevant numbers.

thypon:
Are the Linux patches needed?

coredumb:
Yes the patches are very much needed ... Mainline AppArmor is running very old code and is full of bugs.

srcpkgs/apparmor/files/apparmor/run
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+[ -r conf ] && . ./conf
+
+# Find all profiles in $APPARMORD and load them
+for profile in $(find ${APPARMORD} -maxdepth 1 -type f); do
+	apparmor_parser -a $profile
+done
+exec chpst -b apparmor pause
```
Gottox: Why do we need a sv-service if there's already a core-service? Can you explain?

srcpkgs/apparmor/files/apparmor/run
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+[ -r conf ] && . ./conf
+
+# Find all profiles in $APPARMORD and load them
+for profile in $(find ${APPARMORD} -maxdepth 1 -type f); do
```
Gottox: `APPARMORD` should default to `/etc/apparmor.d`:
```bash
${APPARMORD:-/etc/apparmor.d}
```
conf file can be removed then.

olivier-mauras: https://github.com/voidlinux/void-packages/pull/13309 let's get the details there ^^

PR #6011, opened 2017-03-20
poppler: update to 0.53.0. (soname bump)

ebfe:
Maybe also add to #5944.


ebfe:
55a0bc62954a3f4dfb2487a030c9237992cf7565

PR #6015, opened 2017-03-21
xbps: bump to depend on glibc>=2.25_1

Duncaen:
We have versioned symbols for glibc, if you update an old system with
the previous glibc release installed, only xbps is updated and breaks
because of those versioned symbols.
Increasing the required glibc version in common/shlibs and bumping xbps
fixes this and would update glibc too if it upates itself.

https://forum.voidlinux.eu/t/update-of-fresh-install-breaking-xbps/2143

### DIFF
--- a/common/shlibs
+++ b/common/shlibs
@@ -17,36 +17,36 @@
 # one (order top->bottom) is preferred over the next ones.
 #
 libc.so musl-0.9.9_1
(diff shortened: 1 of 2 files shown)


chneukirchen:
FWIW, let's do this on every major glibc release.

PR #6030, opened 2017-03-25
kodi: update to 17.1

Bubble-be:


xtraeme:
Disabling musl is not an option. If you want to get this merged please fix this for musl as well.

Bubble-be:
Thing is, the current package circumvents the jdk requirement by using a package from https://repo.voidlinux.eu/distfiles/kodi-16.1-generated-addons.tar.xz. There is no package there for v17.1 and I'm not sure how it was created.

Forcing it to v17.0 seems cheating.

xtraeme:
Those files are generated via bootstrap and it's generated manually to contain only these files required to avoid java.

Bubble-be:
Any hint on how I can do that for this new version?

Gottox:
The generated files got updated at repo, so this particular issue should be fixed.

Nevertheless, kodi on musl needs some fixes at `xbmc/cores/DllLoader/exports/util/EmuFileWrapper.{cpp,h}` which missuses the FILE struct.

You may also rebase your work on my template at https://github.com/Gottox/void-packages/tree/kodi-17.0 which has everything except the FILE issue fixed.

Hoshpak:
Alpine has updated their kodi package in the meantime. I believe the previous patch for EmuFileWrapper.cpp also originated there so it might be possible to use the updated patch for our kodi package: https://git.alpinelinux.org/cgit/aports/tree/community/kodi?h=master

Hoshpak:
The alpine patch replaced "if !defined(__GLIBC__) " with "#if !defined(GLIBC)". I'm not sure if that's strictly wrong but it at least doesn't seem to work on i686, hence the compilation failure. Apart from that, please squash all the commits into one. It's generally not necessary to add additional commits to a PR, you can just amend the existing one and force push after that.

Bubble-be:
Thanks for the guidance, you can probably tell these are my first steps in git...

When looking up git squash, I usually encounter something along the lines of: "Don’t include any commit you’ve already pushed to a central server – doing so will confuse other developers by providing an alternate version of the same change."

This will not be a problem in this case, I guess?

Hoshpak:
Right. This advice is generally valid if you are working together with other people on a branch. In this case however, it's your private branch which you want to get merged into the main branch in void packages and you can rebase and amend as you like. You just have to force-push afterwards.

Btw, I looked through the glibc header files on my system and couldn't find any definition for ```GLIBC``` while ```__GLIBC__``` is defined as the major version number so reverting the conditions back to ```__GLIBC__``` in the alpine patch seems the right way to handle it.

Gottox:
@Bubble-be we will take care of that.

Bubble-be:
@xtraeme  Kodi pushed an update to 17.2 in part to combat malicious subtitle files https://kodi.tv/article/kodi-v172-minor-bug-fix-and-security-release

Can you please update the distfiles? After that it's just changing the version number and verification hash to update.

PR #6035, opened 2017-03-26
New package: python-packaging

Hoshpak:
Introduces the new package python-packaging which is a dependency of python-cryptography and also declares that dependency in the package.


asergi:
`cryptography` is patc the `packaging` module provided by `setuptools` at the moment (which is already a dependency). The new `python-packaging` package will be introduced with `setuptools>=34`.

asergi:
`cryptography` is modified to use the `packaging` module provided by `setuptools` at the moment (which is already a dependency). The new `python-packaging` package will be introduced with `setuptools>=34`.

asergi:
As of 1.9, `cryptography` no longer depends on `setuptools` and/or `packaging`.

PR #6038, opened 2017-03-26
New package: tmux-mem-cpu-load-3.4.0

diogoleal:


srcpkgs/tmux-mem-cpu-load/INSTALL.msg
```diff
@@ -0,0 +1,7 @@
+
+Edit $HOME/.tmux.conf to display the program's output in status-left or status-right. For example:
+
+  set -g status-interval 2
+  set -g status-left "#S #[fg=green,bg=black]#(tmux-mem-cpu-load --colors --interval 2)#[default]"
+  set -g status-left-length 60
```
Vaelatern: I think this should be removed.

diogoleal: removed!

PR #6045, opened 2017-03-27
tcsh: fix memory issue on musl

weabot:
tcsh on musl would fail to start, complaining it's out of memory. That's because tcsh uses its own implementation of malloc by default which doesn't seem to be compatible with musl. This enables the flag to use the system's malloc instead, when on musl.


chneukirchen:
Good catch.

PR #6047, opened 2017-03-28
inspircd: Fix permissions on /var/run/inspircd

the-maldridge:
Without the chown, inspircd just sits there trying to write the PID to a directory that it can't write to.


Vaelatern:
Revbump

PR #6063, opened 2017-03-31
gvim: huge version

juliendehos:
gvim with python, perl and lua support (see #5902)


ebfe:
1b5c491446cc2498b205bbed1867b73682154c96

PR #6070, opened 2017-04-01
surf: updated to 2.0

phikal:


cr6void:
Hi @phikal !

In the template, since this is the first build for surf 2.0, you must set "revision=1" (not 3).

In "makedepends=" remove webkitgtk2-devel and replace it with webkit2gtk-devel.

Also, "nopie=yes" seems to be necessary. Without it, I get this error:

`=> surf-2.0_1: running post-install hook: 06-strip-and-debug-pkgs ...
   Stripped executable: /usr/bin/surf
=> ERROR: surf-2.0_1: non-PIE executable found in PIE build: /usr/bin/surf
=> ERROR: surf-2.0_1: failed to run post-install_06-strip-and-debug-pkgs() at line 58.`

srcpkgs/surf/template
```diff
@@ -1,7 +1,7 @@
 # Template file for 'surf'
 pkgname=surf
-version=0.7
-revision=2
+version=2.0
+revision=3
```
Gottox: Revision is set to 1 when you increase the version.

Use `revision=1` instead.

srcpkgs/surf/template
```diff
@@ -1,7 +1,7 @@
...
 build_style=gnu-makefile
 hostmakedepends="pkg-config"
 makedepends="webkitgtk2-devel"
```
Gottox: surf-2.0 uses webkit 2, please change.

Gottox:
Don't add nopie, fix the LDFLAGS instead.

phikal:
>Don't add nopie, fix the LDFLAGS instead.
nopie seemed to work, is it still necessary to fix the LDFLAGS?

Duncaen:
> nopie seemed to work - is it still necessary to fix the LDFLAGS?

Yes, `nopie` is a workaround to ignore the problem.
The package has to use the LDFLAGS, otherwise the hardening flags we use by default don't apply, which is a bad idea especially for a browser.

phikal:
I hope it's fixed now, I haven't had had too much experience until now with void packaging, so I'm sorry if this was a bit of a fuss.

cr6void:
@Gottox & @Duncaen I'm really sorry for my improper use of `nopie`

Like @phikal , I am not very experienced with void packaging, but I want to help! :wink:

Gottox:
Not a prob 👍

cr6git:
@Duncaen What did you say about `nopie` ? It's a bad idea for a browser ? 

Take a look at these templates: [chromium](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/chromium/template#L16), [firefox](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/firefox/template#L13), and [firefox-esr](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/firefox-esr/template#L14). 
All three have `nopie=yes` !

PR #6084, opened 2017-04-03
gnome-terminal: update to 3.24.0

Sir-Boops:


Sir-Boops:
It currently fails to compile is musl as the package file for "python-atspi" seems to be missing from the musl repo

PR #6088, opened 2017-04-04
New package: termbox-1.1.0

diogoleal:


common/shlibs
```diff
@@ -2803,3 +2803,4 @@ libdouble-conversion.so.1 double-conversion-2.0.1_1
 libspectrum.so.8 libspectrum-1.2.2_1
 libbearssl.so.0 bearssl-0.3_1
 libXfont2.so.2 libXfont2-2.0.1_1
+termbox-1.1.0_1 termbox-1.1.0_1
```
pullmoll: This is definitely wrong. If you have a `-devel` subpackage, there will be some *.so.* file to add here.

diogoleal:
this it okay?

Vaelatern:
No. In the devel package, do not move `.so*` but instead the `.so` files.
And a `common/shlibs` entry is needed, in the form of the others, with the correct versioning information.

diogoleal:
Okay, adjusted!

ebfe:
Where did the the -devel package go?

the-maldridge:
@diogoleal this needs the -devel package back to be mergable.

diogoleal:
@ebfe 

-devel added and branch updated.

PR #6090, opened 2017-04-04
nvidia340: bump xserver-abi version

cr6void:


Vaelatern:
Does it not also conflict with `>21`?

cr6void:
I think this update was necessary.

Before my commit: nvidia340 was broken by xorg-server-1.19.3_1(abi conflict).

I did this commit in the same way as [#6089](https://github.com/voidlinux/void-packages/pull/6089).

Tested on my computer, no problem for me.

PR #6093, opened 2017-04-05
shlibs: add missing libjawt.so

juliendehos:


Gottox:
Can you explain why that's needed?

juliendehos:
It is needed to compile libreoffice with java support.

Issue #5706, opened 2017-02-10
Package Request: hledger


silvernode:
**Hledger** is a plain text double entry accounting suite written in [Haskell](https://www.haskell.org/). Hledger features a cli, Ncurses UI, and web frontend.

Project page: https://github.com/simonmichael/hledger
Dependencies: zlib, ncurses (pkg-config, stack, ghc, gcc, make)

* Notes:
Each component (hledger, hledger-ui, hledger-web) can be downloaded and built independently via the releases page, or built together using 'stack install' on master. Least build trouble seems to be using master instead of the current releases. Maybe you can get the releases to build?. 

I use this software as my primary finance application. It makes the accounting part fun and motivates me to keep up on my savings. Thanks in advance to anyone who decides to take on the task.

Vaelatern:
Does not build easily with resolver=lts-7.19, needs other packages not mentioned at all.

silvernode:
I only listed what was required for me to build it without a template. Strange that it has issues. I'll look around for clues to what else it needs but the docs aren't exactly clear on that.

Vaelatern:
A first stab attempt. Does not succeed but could use some finessing.
```
# Template file for 'hledger'
pkgname=hledger
version=1.1
revision=1
wrksrc="${pkgname}-${pkgname}-${version}"
build_style=haskell-stack
nocross="Can't cross haskell yet"
stackage="lts-7.19"
hostmakedepends=""
makedepends=""
depends=""
short_desc="Simple, precise, plain text accounting"
maintainer="Toyam Cox <Vaelatern@gmail.com>"
license="GPL-3"
homepage="http://hledger.org"
distfiles="https://github.com/simonmichael/hledger/archive/hledger-1.1.tar.gz"
checksum=b47055fb218c7e8b254bc663c5a63398504d2dfa8daf80646b04983f4d488edb

post_extract() {
        # Unable to parse cabal file /builddir/hledger-hledger-1.1/extra/hledger-chart/hledger-chart.cabal: NoParse "version" 2
        # README says: Warning, this package has been unmaintained since 2011 and will not install.
        rm -rf ${wrksrc}/extra/hledger-chart
}
```

Vaelatern:
5e94074

PR #6098, opened 2017-04-05
New package: hledger-1.2

ibabushkin:
`hledger` was requested in #5706. This currently splits stuff in subpackages, I still need assurance from travis to see whether some fixes are needed.

### DIFF
--- a/srcpkgs/hledger/template
+++ b/srcpkgs/hledger/template
@@ -1,0 +1,49 @@
+# Template file for 'hledger'
+pkgname=hledger
+version=1.2
+revision=1
+build_style=haskell-stack
+nocross=yes
+short_desc="Hledger is an accounting program for money, time, or other commodities"
+maintainer="Inokentiy Babushkin <twk@twki.de>"
+license="GPL-3"
+homepage="http://hledger.org/"
+hostmakedepends="ghc stack"
+makedepends="zlib-devel ncurses-devel"
+wrksrc="${pkgname}-${pkgname}-${version}"
+stackage="lts-8.6"
+distfiles="https://github.com/simonmichael/${pkgname}/archive/${pkgname}-${version}.tar.gz"
+checksum="bd00115272feada9e388b1aafe6f105369f0d0f0286f65b4ed35fb4ffcd3ad54"
+nopie=yes  # network-2.6.3.1 build fails with a relocation
+
+subpackages="hledger-ui hledger-web hledger-api"
+
+post_install() {
+	vman hledger/doc/hledger.1
+	vman hledger/doc/other/*.5
+}
+
+
+hledger-ui_package() {
+	depends="${sourcepkg}>=${version}_${revision}"
+	pkg_install() {
+		vmove usr/bin/hledger-ui
+		vman hledger/doc/other/hledger-ui.1
+	}
+}
+
+hledger-web_package() {
+	depends="${sourcepkg}>=${version}_${revision}"
+	pkg_install() {
+		vmove usr/bin/hledger-web
+		vman hledger/doc/other/hledger-web.1
+	}
+}
+
+hledger-api_package() {
+	depends="${sourcepkg}>=${version}_${revision}"
+	pkg_install() {
+		vmove usr/bin/hledger-api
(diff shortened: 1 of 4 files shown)


ibabushkin:
To kick off the discussion. `stack` doesn't build on musl. Any solutions besides building only for... well, non-musl (which is essentially only `i686` and `x86_64`).

srcpkgs/hledger/template
```diff
@@ -0,0 +1,49 @@
+# Template file for 'hledger'
+pkgname=hledger
+version=1.2
+revision=1
+build_style=haskell-stack
+nocross=yes
```
Vaelatern: Is there any real reason to keep this?

srcpkgs/hledger/template
```diff
@@ -0,0 +1,49 @@
...
+homepage="http://hledger.org/"
+hostmakedepends="ghc stack"
+makedepends="zlib-devel ncurses-devel"
+wrksrc="${pkgname}-${pkgname}-${version}"
+stackage="lts-8.6"
+distfiles="https://github.com/simonmichael/${pkgname}/archive/${pkgname}-${version}.tar.gz"
+checksum="bd00115272feada9e388b1aafe6f105369f0d0f0286f65b4ed35fb4ffcd3ad54"
+nopie=yes  # network-2.6.3.1 build fails with a relocation
+
+subpackages="hledger-ui hledger-web hledger-api"
```
Vaelatern: This variable is redundant.

srcpkgs/hledger/template
```diff
@@ -0,0 +1,49 @@
+# Template file for 'hledger'
+pkgname=hledger
+version=1.2
+revision=1
+build_style=haskell-stack
+nocross=yes
+short_desc="Hledger is an accounting program for money, time, or other commodities"
```
Vaelatern: `short_desc="Simple, precise, plain text accounting"`

srcpkgs/hledger/template
Vaelatern: Nvm, yes.

srcpkgs/hledger/template
ibabushkin: Yeah, the order of subpackages shouldn't be important.

srcpkgs/hledger/template
ibabushkin: what bit? the build style or the nocross setting?

Vaelatern:
I'm currently testing my modifications and will push soon

ibabushkin:
Yeah, I'm waiting for travis to finish.

ibabushkin:
Turns out, waiting was useless.

chneukirchen:
We don't have ghc on musl, so...

ibabushkin:
That was to be expected :P I guess I'll add the arch restriction then, so we only have non-musl `i686` and `x86_64`. What should we do about the travis timeout?

Vaelatern:
```
WARNING: Installation path /destdir/hledger-1.2/usr/bin not found on the PATH environment variable
=> hledger-1.2_1: running post_install ...
=> ERROR: hledger-1.2_1: vinstall: cannot find 'hledger/doc/other/*.5'...
=> ERROR: hledger-1.2_1: failed to run post_install() at line 12.
```
on x86_64

Vaelatern:
Don't add an arch restriction.

Vaelatern:
I think I fixed that problem I just shared. Re-running.

ibabushkin:
Nice! How exactly? I'm a bit baffled about why the paths don't match the directory layout of the tarball.

Vaelatern:
vman does not take wildcards, so instead I'm looping over the relevant files and vmaning each one

ibabushkin:
ah, stupid me, that makes sense. I suppose the warning regarding paths is irrelevant due to the chroot environment?

Vaelatern:
Build time (MM:SS): 39:59 with 24 cores at it.
But it works for me after simple testing, and the packages seem to have what they need, so for now, merging!

ibabushkin:
Alright, let's see how this works out! I already thought about how we could reduce the perverse build time, so stay tuned :)

PR #6108, opened 2017-04-06
New package: hoogle-5.0.9

ibabushkin:
Hoogle is a search engine for Haskell libraries. I think it's useful to have as a distribution package as it takes some time to compile and updating it by hand from a local stack or cabal installation is not that good of an idea.


ibabushkin:
As expected, the musl build fails. No arch restriction added though, as do `pandoc` and `hledger`, for example.

PR #6127, opened 2017-04-08
perl-List-MoreUtils: update to 0.419

Hoshpak:


chneukirchen:
I think this needs a new package for the XS now.

Hoshpak:
I didn't think it was strictly necessary since no software currently in the repository depends on the XS part. It certainly doesn't hurt having the package around to provide the same functionality perl-List-MoreUtils provided previously for people wanting to use it. I added the perl-List-MoreUtils-XS package to the PR.

chneukirchen:
I thought the XS was used automatically, but I may be wrong. Looks good now.

PR #6136, opened 2017-04-08
font-iosevka: Add 01-08 variants

SevereOverfl0w:
Uses the pack distribution to include most of the variants of iosevka.


Gottox:
@diogoleal does this make sense?

diogoleal:
@Gottox yeah! it's correct!

@SevereOverfl0w update this PR to font-iosevka version 1.12.5.

Issue #6099, opened 2017-04-05
gajim is started with py3


emdete:
gajim is python2 code but is started with "python" (from /usr/bin/gajim) which defaults to python3. this leads to an error:

```
File "gajim.py", line 58
    except getopt.error, msg1:
                       ^
SyntaxError: invalid syntax
```

because py3 does not know about that exception syntax.

PR #6143, opened 2017-04-09
gajim: add patch to prefer python2 in configure script

Hoshpak:
This is to address #6099 . Gajim uses shell scripts as wrappers to start the actual application. It detects the python interpreter during the configure step and writes the absolute path to the interpreter in the scripts. Since these scripts are shell scripts and the python interpreter is not in the shebang line, it is not caught by the shebang rewrite of xbps-src. This can lead to situations where /usr/bin/python is in the wrapper scripts and is linked to python 3 so the start of gajim fails. This PR proposes a patch that puts python2 first in the configure script so, if it is available, the script will pick it up and write /usr/bin/python2 into the wrapper scripts which seems to me to be the cleanest way to deal with the issue.


asergi:
There is no need to patch the `configure`: `configure_args="--enable-site-packages PYTHON=python2"`.

Hoshpak:
```PYTHON=python2``` or ```PYTHON=/usr/bin/python2```? The path set in the variable is written into the scripts without changing it. For the automatic detection, it writes the absolute path. Just setting python2 should also work in most cases unless PATH is empty or messed up.

The idea behind patching the configure script was to still rely on the automatic detection without directly hardcoding the path to the python interpreter.

asergi:
`PYTHON=python2` is fine.

Hoshpak:
PR amended

PR #6159, opened 2017-04-11
dovecot: update to 2.2.29

Hoshpak:
fixes CVE-2017-2669


srcpkgs/dovecot/template
```diff
@@ -1,7 +1,7 @@
 # Template file for 'dovecot'
 # revbump dovecot-plugin-pigeonhole when updating dovecot!
```
ebfe: ^

Hoshpak: sorry, added revbump.

PR #6164, opened 2017-04-11
opera: update to 44.0.2510.1218

cr6void:


cr6void:
I know, the commit name is wrong. 
I'm really sorry, I must be stupid because I still can't find how to rename my own commits! 
So I will close this PR to make the new one. 
For me github is not easy to use, I need to learn... :(

PR #6170, opened 2017-04-13
gnome-settings-daemon: update to 3.24.1

cr6void:


cr6void:
The build failed on musl but I don't know what these errors mean... Did I make a mistake?

PR #6191, opened 2017-04-14
libreoffice: enable java

juliendehos:


pullmoll:
We don't have openjdk for *-musl, so this option can only be an option which is enabled for non-musl targets.

srcpkgs/libreoffice/template
```diff
@@ -86,12 +86,14 @@ fi
 case "$XBPS_TARGET_MACHINE" in
 	i686*)	# Broken unit tests
 		CXXFLAGS="-DDISABLE_CVE_TESTS=1"
+		makedepends+=" apache-ant openjdk "
```
pullmoll: We also have i686-musl, even if there is no official builder for this combination.
I'm not yet in a position to tell if libreoffice builds at all for i686-musl, but you could change the case to i686) for now.

juliendehos:
Ok. I've updated the template for detecting *-musl.

PR #6192, opened 2017-04-14
New package: nnn-1.0

d-kraus:


srcpkgs/nnn/template
```diff
@@ -0,0 +1,17 @@
+# Template file for 'nnn'
+pkgname=nnn
+version=1.0
+revision=1
+build_style=gnu-makefile
+makedepends="ncurses-devel"
+depends="ncurses"
```
Vaelatern: Pretty sure this `depends=` is redundant and should be removed.
https://travis-ci.org/voidlinux/void-packages/jobs/222032934#L684-L685

PR #6198, opened 2017-04-16
Update Rakudo/MoarVM/nqp to 2017.03

the-eater:
`probe_cflags_fix.patch` has been pushed upstream most likely it will be in the next release: https://github.com/MoarVM/MoarVM/pull/582

Currently we need to fix the packages `rakudo-star` and `panda`

And we need to add `zef`.


the-eater:
After some discussion we decided to kill `panda` and `rakudo-star` this also means `zef` will not be packaged.

yopito:
This PR is outdated since version 2017.10 of these are present in master. Someone to close this PR ?

PR #6207, opened 2017-04-18
New package: klystrack-1.7.0

allan:


srcpkgs/klystrack/template
```diff
@@ -0,0 +1,48 @@
...
+ 8f957d1250fa36d343d3eecd6585c5890342b0a1b3d8ecaec22217e95d1dbf96"
+
+do_build() {
+	test -d klystron ||
+		mv klystron-${_klystronsha} klystron
+	cd klystrack-${version}-final
+
+	sed -i \
+		-e 's/
$//' \
```
Gottox: I would prefere using the dos2unix tool or an escape sequence here. Wild control characters may do strange things when using cat for example.

allan: changed it to `\r`.  looks less wild now.

PR #6213, opened 2017-04-19
dotherside: shlibs update

gangsterakato:


Vaelatern:
Rebuild of dotherside needed too (revbump)

Gottox:
updating shlibs is only needed when a the so version has changed.

gangsterakato:
@Gottox, since the update of the package I got the error `could not load: libDOtherSide.so` when trying to open a program compiled with `dotherside`.

When renaming the library from from libDOtherSide.so.0.0.0 to libDOtherSide.so it works.
What can be done in this case?

Gottox:
looks like an ldopen issue. this may be fixed by patchng the depending program.

gangsterakato:
@Gottox, the fact is that the program worked well with the last version, without any configuration.

Gottox:
which one was it?

PR #6215, opened 2017-04-19
New Package: git-lfs

kulinacs:


srcpkgs/git-lfs/INSTALL.msg
```diff
@@ -0,0 +1 @@
+Run 'git-lfs install' to configure git to use git-lfs
```
Gottox: missing EOL.

kulinacs: Fixed.

PR #6218, opened 2017-04-19
weston: add egl patches

Sir-Boops:
These patches are needed for nvidia support, i haven't tested them on a non nvidia system


the-maldridge:
Are these patches still relevant?

Sir-Boops:
I've not seen any movement upstream so yes they are. And I've now also tested these patches on a non nvidia system and they don't effect it.

Duncaen:
I dont think we should import patches that large, it bloats the repository.

asergi:
What @Duncaen said. Also, it's not sure these patches will be accepted upstream.
So, @the-maldridge revert it.

the-maldridge:
I very much dispute the bloat comment, but I've reverted the patch anyway.

PR #6221, opened 2017-04-20
dotherside: fixing shared library availability in standard package

gangsterakato:


gangsterakato:
@filcuc seems to have fixed DOtherSide upstream, the patch isn't necessary anymore and the standard package now works as it is.

PR #6239, opened 2017-04-22
New package: sgfutils-0.25

xelxebar:


srcpkgs/sgfutils/template
```diff
@@ -0,0 +1,14 @@
+# Template file for 'sgfutils'
+pkgname=sgfutils
+version=0.25
+revision=1
+build_style=gnu-makefile
+nopie=yes
```
Duncaen: `nopie=yes` is most likely wrong and indicates that the build doesn't use Voids default `CFLAGS` and `LDFLAGS`.

Duncaen:
You can force push instead fo opening the PR over and over again.

xelxebar:
@Duncaen 
Thank you for the review. I patched upstream's Makefile, so the `nopie` should be fixed now.
This is my first pull request ever, so yeah, sorry for the spam. Think I'm getting the hang of it now.

PR #6243, opened 2017-04-22
octoxbps: don't use --force

Piraty:
As it has already been discussed a few times, this is a dangerous flag to have in a package managing too, especially if user's may not be aware of it.
Whover wants to --force on his xbps-system, should do that on the terminal.

### DIFF
--- a/srcpkgs/octoxbps/patches/no-force.patch
+++ b/srcpkgs/octoxbps/patches/no-force.patch
@@ -1,0 +1,75 @@
+This patch is also offered to upstream: https://github.com/aarnt/octoxbps/pull/3
+
+--- src/xbpsexec.cpp
++++ src/xbpsexec.cpp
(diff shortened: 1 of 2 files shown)


Duncaen:
What about fixing it upstream?

Piraty:
I have a PR with the same patch, until that makes it into the project eventually, void could use this patch.

Vaelatern:
Link the patch file to the upstream pull request (as in, link to the PR for upstream) and I think this is mergable.

Piraty:
done

the-maldridge:
@Vaelatern is this merge-able even with the broken cross issues?

Vaelatern:
Yes. It never cross built.

ebfe:
just ftr: ujpstream disagrees with this patch.

PR #6245, opened 2017-04-23
icu: update to 59.1. (soname bump)

ebfe:
nodejs/v8  broken, waiting for a fix


ebfe:
lots of other packages broken too: http://site.icu-project.org/download/59#TOC-ICU4C-char16_t

ebfe:
Everything seems to build now, had to switch nodejs to use the bundled icu for now (cc @Gottox).

Gottox:
🙄  oh boy. nodejs.

@ebfe can you use a new build_option instead?

ebfe:
If it makes you happy i can add a build_option, but it seems pretty useless to have a build_option you can't enable because the build is broken with it.

Gottox:
@ebfe thanks! :)

ebfe:
@Gottox done

ebfe:
58880f4aa0e4c4f1abe94534d474473dcfbef83f

PR #6246, opened 2017-04-23
ffmpeg: update to 3.3.

ebfe:


Gottox:
armv7 needs to be fixed.

ebfe:
More like our travis environment has to be fixed, i don't have time to investigate what's different on travis.

Gottox:
armv7 builds locally. #yolo

PR #6257, opened 2017-04-24
New package: qgo-2.1.1

xelxebar:
A Go game client to complement the gnugo and sgfutils packages.

For some reason qt5-multimedia-devel doesn't playing well with the cross-compilation and will fail without the final conditional in the template file. I have no clue why we need to munge `makedepends` into `hostmakedepends` like that, but texstudio is doing it, so I copied them.


Vaelatern:
Please try this with `hostmakedepends+=' qt5-host-tools`

xelxebar:
@Vaelatern 
Thanks for taking a look at this. The configure step ends up calling `/usr/lib/qt5/bin/lrelease` which, unfortunately, isn't in `qt5-host-tools`.

Vaelatern:
It is in `qt5-tools`

xelxebar:
Could you please provide a little more context on what problem you're trying to solve? I have no idea what's going on. :P

The package compiles with or without `qt5-host-tools` and the template already contains `qt5-tools`. Are you asking that I replace line 16 with explicit packages?

PR #6261, opened 2017-04-25
new package espeakup 0.80

equalsraf:
kernel module is loaded on service startup.


CMB:
A few quick comments.  
srcpkgs/espeakup/files/log/ needs to be
srcpkgs/espeakup/files/espeakup/log.

You should add an
exec 2>&1
to the top of srcpkgs/espeakup/files/espeakup/run,
(E.G. just after shebang)
since espeakup can write messages to stderr that belong in the log.

I'd also add an
[ -r ./conf ] && . ./conf
after that, and execute espeakup as:
exec espeakup -d ${OPTS}
so that the user can pass options in the config file.
E.G., they might want to specify OPTS='-V whatever' to change
the default voice.

Otherwise, LGTM.

equalsraf:
Thanks @CMB I've added all the changes.

CMB:
Wonderful, thanks.  I don't see any reason why this couldn't be merged.

srcpkgs/espeakup/template
```diff
@@ -0,0 +1,15 @@
+pkgname="espeakup"
+version="0.80"
+revision=1
+short_desc="A light weight connector for espeak and speakup"
```
Vaelatern: `Lightweight connector for espeak and speakup`

Vaelatern:
- Every template must begin with a comment, as per most other templates.

CMB:
> * Every template must begin with a comment, as per most other templates.

To be fair, this is not mentioned in the manual.

grep -ci comment Manual.md

The xlint tool doesn't complain, either.
It's unwritten convention, and most of us get it for free by using xnew
to start new package templates.

Also in this template pkgname and version are quoted, but by convention,
they are usually unquoted.

equalsraf:
Added comment, and unquoted the name/version.

Vaelatern:
Last thing is the commit name, as per the Manual.
But github lets me fix that for you.
@CMB thank you for noticing. Fixed that. 81bcc70bcd

PR #6267, opened 2017-04-26
New package: poco-1.7.8

juliendehos:


srcpkgs/poco-devel/template
```diff
@@ -0,0 +1 @@
+../poco/template
```
ebfe: poco-devel should be a symlink to poco

juliendehos: ok, fixed

srcpkgs/poco/template
```diff
@@ -0,0 +1,34 @@
...
+post_install() {
+	vlicense LICENSE
+}
+
+poco_package() {
+	short_desc+=" - runtime libraries"
+	pkg_install() {
+		vmove usr/lib/*.so.*
+	}
+}
```
Vaelatern: this entire `poco_package` is useless and xbps-src doesn't even try to use it. Please remove it.

srcpkgs/poco/template
```diff
@@ -0,0 +1,34 @@
...
+
+poco-devel_package() {
+	depends="${sourcepkg}>=${version}_${revision}"
+	short_desc+=" - development files"
+	pkg_install() {
+		vmove usr/include
+		vmove usr/lib/*.so*
```
Vaelatern: `vmove usr/lib/*.so`

juliendehos: ok, fixed (vmove and poco_package)

PR #6276, opened 2017-04-27
pgadmin3: fix startup and distfile

arete:


Vaelatern:
Can you show an upstream source for that patch?

arete:
hi, 
the patch is from arch linux

https://git.archlinux.org/svntogit/community.git/tree/trunk/pgadmin3-fix-segfault.patch?h=packages/pgadmin3

PR #6279, opened 2017-04-28
Add net-tools hostname to alternative system

SevereOverfl0w:
Seems to be working for me. But there may be a reason it was not included in the alternatives originally, but I'm not certain. It may just have been needs driven, much like now.


chneukirchen:
I guess this was just older than the alternatives mechanism.

Please use the `package: description` commit format, then +1.

SevereOverfl0w:
Reworded the commit message.

PR #6284, opened 2017-04-28
New package: streamripper

Piraty:
It's a bit old, but works as intended (I use it for a few weeks now). This version is the latest one that got a release number.

There are nonetheless two repos, obviously formerly maintained by the original authors (https://github.com/streamripper/streamripper/ and https://sourceforge.net/p/streamripper/code/commit_browser) which both have about two handful of patches for minor bugs i think. They diverged at some point...

Maybe i merge them together into one repo one day and use that instead ...


Vaelatern:
remember the commit formatting :)

PR #6289, opened 2017-04-29
New package: hlint-2.0.5

ibabushkin:
`hlint` is a (or even *the*?) linter for haskell code.

Note that I wasn't able to get stack/ghc to generate a PIE. adding `make_build_args="--ghc-options -fPIC"` didn't yield the expected result, so that might be worth taking a look at.

### DIFF
--- a/srcpkgs/hlint/template
+++ b/srcpkgs/hlint/template
@@ -1,0 +1,19 @@
+# Template file for 'hoogle'
+pkgname=hlint
+version=2.0.5
+revision=1
+build_style=haskell-stack
+nocross=yes # Can't yet cross compile Haskell
+# make_build_args="--ghc-options -fPIC"
+stackage="lts-8.12"
+short_desc="Haskell source code suggestions"
+maintainer="Inokentiy Babushkin <twk@twki.de>"
+license="BSD-3"
+homepage="https://github.com/ndmitchell/hlint"
+distfiles="https://github.com/ndmitchell/${pkgname}/archive/v${version}.tar.gz"
+checksum="7547bc716e69367bb808f878c528bd9f014fa84de202b89c0f70fe96fb3c9421"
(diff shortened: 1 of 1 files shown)


Vaelatern:
My issue is the `nopie=yes` part. You sure it doesn't build when PIE is enabled?

ibabushkin:
On 2017-04-30, Toyam Cox wrote:
> My issue is the `nopie=yes` part. You sure it doesn't build when PIE is enabled?

As I said, `make_build_args="--ghc-options -fPIC"` doesn't seem to give
us a PIE, since `xbps-src` complains, that a non-PIE binary is present
in a normal (i.e. non-`nopie`) build. I don't know why this happens
though, GHC should be capable of generating PIC code.

ebfe:
FWIW, every haskell-stack package we have set nopie=yes currently.

ibabushkin:
On 2017-05-06, Michael Gehring wrote:
> FWIW, every haskell-stack package we have set nopie=yes currently.

Correct, though some explain it with a comment along thelines of
`network-x.y.z` won't *build* otherwise. This is a different issue
though, and I personally think it *can* be resolved, I'll look into it
when I get.

the-maldridge:
Merged with minor changes.  Feel free to open a new PR for PIE consistency if and when you sort out that issue.

ibabushkin:
On 2017-07-05, Michael Aldridge wrote:
> Merged with minor changes.  Feel free to open a new PR for PIE consistency if and when you sort out that issue.
> 
> -- 
> https://github.com/voidlinux/void-packages/pull/6289#issuecomment-312985311

I just now realized some breakage of the package (basically, data files
being handled incorrectly). I'll fix that and look at the PIE situation
then. Next time, better local testing in a properly isolated environment
:P.

PR #6293, opened 2017-04-30
New package: sc-controller-0.3.9

KeepBotting:
Needed this today and decided to make a template. I believe I have adhered to all relevant guidelines, please provide feedback if not.


ebfe:
#5952

KeepBotting:
@ebfe That template is outdated, missing a dep & wrong checksum. I have fixed those issues.

Vintodrimmer:
Good day, 

Seems that you are correct.

Though I don't really know, how the old checksum got into the template, since I pushed a commit only after the successfull build on my machine.

Shouldn't `kernel-libc-headers` be part of any installation, though?

KeepBotting:
@Vintodrimmer `kernel-libc-headers` was not present on my system. I had to install it manually.
```
[branon@schwifty ~]$ xbps-query -RX kernel-libc-headers
base-chroot-0.65_2
glibc-devel-2.25_3
```
The revdeps seem to suggest that no, it will not be part of every installation. As I do not have either of them installed. I am probably an edge case because I run a very minimal installation. Others (including yourself) have probably pulled `kernel-libc-headers` in as a dependency somewhere down the line.

Vintodrimmer:
Just checked and you are right again.

It was pulled by `musl-devel`, which was pulled by `gcc`. Fixed my template too.

ebfe:
Headers are usually only required at compile time.

KeepBotting:
@ebfe they are required for runtime in this case, see [this issue](https://github.com/kozec/sc-controller/issues/199)

Vintodrimmer:
@ebfe I tried forcefuly removing them and got the error, while running ` sc-controller`.

KeepBotting:
It's been a week. @ebfe any update on this or #5952?

KeepBotting:
Closing this in deference to #5952. Would like to see sc-controllers merged soon!

PR #6319, opened 2017-05-03
New package: R-cran-data.table-1.10.4 and build-style R-cran

wagnerflo:


wagnerflo:
This is my second try to get this package accepted. In contrast to the first time, all CI checks pass now. Is there anything holding you back merging this? Please provide feedback if there are issues...

ebfe:
First we would have to decide if we actually want to have xbps packages for R packages.

On the formalities:
The R build-style should probably go into it's own commit and the _ be dropped from the `_cran*` variables. Build-style and the variables should be documented in Manual.md.

wagnerflo:
Thanks for the feedback.

I'm very much willing to work on all the formalities as well as maintaining a bunch of CRAN packages, but would prefer the decision to be made first. Where's the right place to discuss this?

ebfe:
cc @chneukirchen

chneukirchen:
I'm not really into R, what's the benefit over installing these things locally?

wagnerflo:
My primary motivation is being able to install/update using a single system: The distribution package manager.

This is IMHO valid for any and all programming language that come with it's own package manager: For a rather basic system having installed Python and Perl isn't uncommon. In that case I wouldn't want to have to remember a system update "chain" of xbps, CPAN and whatever Python package manager is currently en vogue, just to be sure to not miss anything.

Same argument applies to R/CRAN.

wagnerflo:
Bump! Any decision on this would be appreciated...

wagnerflo:
@ebfe, @chneukirchen: This is the deciding factor for switching a bunch of machines from Arch to Void. Please decide how to go along with this! No hard feelings if that turns out to be a no...

chneukirchen:
From my POV, i'd say a slight negative. There are many CRAN packages people want to use, and keeping them up to date is a lot of work (which I won't do). If someone maintains them properly, I guess it would be ok.

That said, Arch only has CRAN packages in AUR afaics (54 packages).

ebfe:
I'd probably continue to use the R internal package manager, but don't mind having these pacakges if there are people who use and **maintain** them.

wagnerflo:
Any package I make templates for and/or use I'm also willing to maintain!

chneukirchen:
Also, I don't really use R, so if someone wants to take over maintainership, it would probably make sense to do so and coordinate releases with other CRAN rebuilds then. Just adopt.

wagnerflo:
That package template doesn't seem particularly complicated so I could do that. I'll start getting to work next week: Adopt R, separate pull with the build-style and associated changes, then a bunch of CRAN packages.

Thanks everyone!

PR #6323, opened 2017-05-04
New package: udftools-1.3

lyzech:


srcpkgs/udftools/template
```diff
@@ -0,0 +1,13 @@
+# Template file for 'udftools'
+pkgname=udftools
+version=1.3
+revision=1
+only_for_archs="x86 x86_64"
```
ebfe: why? cross builds seem to work fine, musl needs some fixes.

ebfe: also x86 is spelled i686 in void ;)

lyzech:
Thank you !

PR #6325, opened 2017-05-04
New package: brother-brscan4-0.4.4

pienjo:
I know there has been an earlier attempt to create a template for this package before. I tried to address the issues that were raised back then, as far as possible.


the-maldridge:
I'm not happy that this creates an /etc/opt directory.  Apart from that I'm happy with the way you're doing most things.  If another maintainer will 'shipit' for this PR then I'm good to merge.

pienjo:
There's little I can do about /etc/opt; there's hard-coded paths in the binary. The only workaround I can imagine is patching the binary and truncate those strings, but I'm not sure that's a viable solution.

the-maldridge:
Yeah I know, but I can still be unhappy about it.

srcpkgs/brother-brscan4/template
```diff
@@ -0,0 +1,75 @@
+# Template file for 'brother-brscan4'
+pkgname=brother-brscan4
+version=0.4.4
+revision=2
```
Vaelatern: `revision=1`

srcpkgs/brother-brscan4/template
```diff
@@ -0,0 +1,75 @@
...
+conf_files="/etc/opt/brother/scanner/brscan4/Brsane4.ini /etc/opt/brother/scanner/brscan4/brsanenetdevice4.cfg /etc/opt/brother/scanner/brscan4/models4/*.ini"
+
+if [ "$XBPS_TARGET_MACHINE" = "x86_64" ]; then
+	_debpkgid="3.amd64"
+	distfiles="http://download.brother.com/welcome/dlf006645/brscan4-${version}-${_debpkgid}.deb"
+	checksum=5dd9edcc670a19b66882932290bf867de1669855f05b0d340c5817b69f104659
+	_deblibdir="usr/lib64"
+elif [ "$XBPS_TARGET_MACHINE" = "i686" ]; then
+	_debpkgid="3.i386"
+	distfiles="http://download.brother.com/welcome/dlf006646/brscan4-${version}-${_debpkgid}.deb"
```
Vaelatern: This line is the same in both. Please deduplicate.

srcpkgs/brother-brscan4/INSTALL.msg
```diff
@@ -0,0 +1,6 @@
+In order to access networked scanners and multifunctionals, the SANE backend 
+must be configured using
+
```
Vaelatern: Spurious newline. Remove.

srcpkgs/brother-brscan4/INSTALL.msg
```diff
@@ -0,0 +1,6 @@
...
+brsaneconfig4
+
```
Vaelatern: And here.

srcpkgs/brother-brscan4/template
pienjo: I assume you mean the "distfiles" line? They are not the same; the URL is different ( ".../dlf00664**5**/.."  versus ".../dlf00664**6**/..").

pienjo:
@Vaelatern 
Thank you for your review - I'll address them. I fully understand the problem with this package - this package doesn't make me happy either. In fact, having bought this printer doesn't make me happy in retrospect. 

I won't hold any grudges if this PR gets rejected because the package is just too dirty and doesn't meet the expected quality level. I wasn't even sure whether I should PR it or not because of this, but on the other hand I was hoping to save someone else some time.

Maybe packages like these shouldn't be in the official repository. I've encountered other packages currently in the void-packages repo that are just as bad, if not worse (including brother-brscan3).

pienjo:
Brother updated their package, so I updated the template..

PR #6327, opened 2017-05-05
gstreamer1: update to 1.12.0.

ebfe:


ebfe:
armv7 builds fine locally, same error as in #6246. Not sure what's wrong on travis.

PR #6330, opened 2017-05-05
New package: superlu-5.2.1

juliendehos:


srcpkgs/superlu/template
```diff
@@ -0,0 +1,36 @@
...
+distfiles="http://crd-legacy.lbl.gov/~xiaoye/SuperLU/${pkgname}_${version}.tar.gz"
+checksum=28fb66d6107ee66248d5cf508c79de03d0621852a0ddeba7301801d3d859f463
+configure_args="-DUSE_XSDK_DEFAULTS=TRUE -DCMAKE_BUILD_TYPE=RELEASE"
+
+pre_configure() {
+	export LDFLAGS="-fPIC"
+}
+
+post_extract() {
+	mv ${XBPS_BUILDDIR}/SuperLU_${version} ${XBPS_BUILDDIR}/superlu-${version}
```
ebfe: set wrksrc instead

srcpkgs/superlu/template
```diff
@@ -0,0 +1,36 @@
...
+post_install() {
+	vlicense License.txt
+}
+
+superlu-devel_package() {
+	depends="${sourcepkg}-${version}_${revision}"
+	short_desc+=" - development files"
+	pkg_install() {
+		vmove "usr/include"
+		vmove "usr/lib/libsuperlu.so*"
```
ebfe: Only the unversioned .so goes into the -devel package. Also needs adjustment in common/shlibs

srcpkgs/superlu/template
```diff
@@ -0,0 +1,36 @@
...
+build_style=cmake
+hostmakedepends="gcc-fortran"
+makedepends="openblas-devel"
+short_desc="Solution of large, sparse, nonsymmetric systems of linear equations"
+maintainer="Julien Dehos <dehos@lisic.univ-littoral.fr>"
+license="BSD"
+homepage="http://crd-legacy.lbl.gov/~xiaoye/SuperLU/"
+distfiles="http://crd-legacy.lbl.gov/~xiaoye/SuperLU/${pkgname}_${version}.tar.gz"
+checksum=28fb66d6107ee66248d5cf508c79de03d0621852a0ddeba7301801d3d859f463
+configure_args="-DUSE_XSDK_DEFAULTS=TRUE -DCMAKE_BUILD_TYPE=RELEASE"
```
ebfe: CMAKE_BUILD_TYPE should already be set by the cmake build-style

juliendehos:
ok, fixed (buildtype, wrksrc and .so).

PR #6342, opened 2017-05-07
New packages: libyui

Gottox:


Gottox:
* [ ] Fix crossbuild
* [ ] Fix musl

ebfe:
musl fixes: https://github.com/ebfe/void-packages/commits/libyui

Gottox:
Thanks! lubyui is really bitchy about x-compiling

PR #6222, opened 2017-04-20
New package: openblas-0.2.19

juliendehos:
Oops, I deleted a branch by mistake : #4799

PR #6343, opened 2017-05-07
openblas: fix cpu targets

juliendehos:
Adds runtime arch selection for i686 and x86_64 (otherwise only the arch used during compilation is provided).
Removes useless includes (already provided by dependencies).
#6222


srcpkgs/openblas/template
```diff
@@ -29,7 +30,6 @@ openblas-devel_package() {
 	depends="${sourcepkg}>=${version}_${revision}"
 	short_desc+=" - development files"
 	pkg_install() {
-		vmove usr/include
```
Gottox: Why is this line remove?

juliendehos: Oups, I removed the line to avoid header conflicts (blas and lapack headers are already provided by original packages) but I forgot to include openblas specific headers... Thanks

ebfe: You are sure these are compatible? Using the lapacke-devel headers and the .so from openblas sounds like asking for trouble.

juliendehos: I'm quite sure about that. Openblas is "just" an optimized implementation and it is intended to be 100% compatible. For example, Armadillo FAQ says: "You can link with high-speed OpenBLAS instead of standard BLAS by changing -lblas -llapack to -lopenblas -llapack" (http://arma.sourceforge.net/faq.html). I made some diffs one day and I'm currently rechecking that but compilation is rather long...

ebfe: lapack but the conflict is with lapacke.

juliendehos: Ok, I've updated the template. Now it compiles openblas with dynamic cpu detection, puts headers in /usr/include/OpenBLAS and updates cmake config. I've tested with a small cmake project; seems ok.

ebfe: Debian installs the headers to /usr/include/openblas but also doesn't ship the lapacke.h headers. tbh, i'm not sure what the right approach is.

juliendehos: Indeed, the classical include path seems to be /usr/include/openblas not /usr/include/OpenBLAS (https://github.com/opencv/opencv/blob/master/cmake/OpenCVFindOpenBLAS.cmake). I've fixed the template.

srcpkgs/openblas/template
```diff
@@ -15,12 +15,32 @@ distfiles="https://github.com/xianyi/${pkgname}/archive/v${version}.tar.gz"
...
+	armv7*) make_build_args+=" TARGET=ARMV7" ;;
+	aarch64*) make_build_args+=" TARGET=ARMV8" ;;
+	i686*) make_build_args+=" BINARY=32 DYNAMIC_ARCH=1" ;;
+	x86_64*) make_build_args+=" BINARY=64 DYNAMIC_ARCH=1" ;;
 esac
 
+do_install() {
+
+	# install in /opt
+	make PREFIX=${XBPS_BUILDDIR}/opt install
```
Vaelatern: I'm pretty sure we don't want to install to `/opt`

Vaelatern:
Why the step with installation to `/opt`?

srcpkgs/openblas/template
```diff
@@ -1,12 +1,12 @@
...
 version=0.2.19
-revision=1
+revision=2
 wrksrc="OpenBLAS-${version}"
 build_style=gnu-makefile
-make_build_args="HOSTCC=gcc USE_OPENMP=1"
+make_build_args=" HOSTCC=gcc USE_OPENMP=1"
 hostmakedepends="perl gcc-fortran"
-makedepends="cblas-devel lapack-devel lapacke-devel libgomp-devel"
+makedepends="libgomp-devel"
```
Vaelatern: `xmindep` doesn't agree here. 
```
$ xmindep cblas-devel lapack-devel lapacke-devel libgomp-devel
cblas-devel lapacke-devel libgomp-devel
```

juliendehos: Indeed, the /opt step is not necessary; fixed.
I think the cblas/lapack/lapacke dependencies can be removed since openblas provides its own optimized implementation of these libraries.

Vaelatern: Are they statically linked?

juliendehos: they are in openblas libs:

$ nm -D /usr/lib/liblapacke.so | grep cggsvd
00000000001b3630 T LAPACKE_cggsvd
...
$ nm -D /usr/lib/libopenblasp-r0.2.19.so | grep cggsvd
00000000020e5790 T LAPACKE_cggsvd
...
$ nm /usr/lib/libopenblasp-r0.2.19.a | grep cggsvd
lapacke_cggsvd.o:
0000000000000000 T LAPACKE_cggsvd
...

srcpkgs/openblas/template
```diff
@@ -15,12 +15,30 @@ distfiles="https://github.com/xianyi/${pkgname}/archive/v${version}.tar.gz"
...
+	echo "SET(OpenBLAS_VERSION \"${version}\")" \
+		> ${PKGDESTDIR}/usr/lib/cmake/openblas/OpenBLASConfig.cmake
+	echo "SET(OpenBLAS_INCLUDE_DIRS /usr/include/openblas)" \
+		>> ${PKGDESTDIR}/usr/lib/cmake/openblas/OpenBLASConfig.cmake
+	echo "SET(OpenBLAS_LIBRARIES /usr/lib/libopenblas.so)" \
+		>> ${PKGDESTDIR}/usr/lib/cmake/openblas/OpenBLASConfig.cmake
+
+	# move headers to a subdirectory to avoid conflicts with cblas and lapacke
+	mkdir -p ${PKGDESTDIR}/usr/include/openblas
+	mv ${PKGDESTDIR}/usr/include/*.h ${PKGDESTDIR}/usr/include/openblas/
```
Vaelatern: This stuff seems like `post_install` material. The call to `make` s handled by the build style.

juliendehos: ok, fixed.

Gottox:
@Vaelatern have your complaints been met?

ebfe:
ping @Vaelatern

ebfe:
#7208

PR #6347, opened 2017-05-08
[nomerge] New package: dblatex-0.3.10

the-maldridge:
dblatex converts from asciidoc to latex so that you can ultimately convert to pdf.  Why it does this I'm not entirely sure since you can just go from asciidoc to PDF...

This package builds, but the python script that should wind up in /usr/bin doesn't ever get there.  It seems there's something wrong with out python-module build style and it doesn't ever run the build-scripts phase.  Since this module seems to date from the 2.4 days, I look for suggestions.


newbluemoon:
I made a template once and found that the following worked:

```
do_install() {
        python ./setup.py install --nodeps --root "/destdir/${pkgname}-${version}" --prefix=/usr --catalogs=/usr/share/xml/docbook
}

```

Using the `--nodeps` option `xbps-src` doesn’t have to pull the whole of the texlive dependencies. But I’m not sure how to handle it when actually installing the package: should dblatex install them via post install script and `tlmgr` or should there be just a message with a list of needed texlive packages for manual installation or something else?

the-maldridge:
Given that post-install scripts are highly discouraged, I'm a fan of a post install message that tells the user to go install them.

newbluemoon:
I just realised that I did not set the `build_style` variable. The package built fine on `x86_64`, `i686`, and `armv7l-musl`.

the-maldridge:
Using the build_style python-module will not work as that does not install /usr/bin/dblatex.

newbluemoon:
I’m trying to package lilypon (a music engraving program) and it depends on dblatex to build the documentation. So I also packaged dblatex and it’s working for me. As mentioned above, I don’t use the `build_style` variable and also not the `find ...` command in `do_install()`, the rest is the same.

So my question is if I shall include dblatex when I make a PR for lilypond (despite this open PR) or if you would like to fix it first. :)

the-maldridge:
Depend on this Pr. It's marked nomerge because if some proposed Tex
changes, but it can be merged without those.

On Sep 15, 2017 01:31, "newbluemoon" <notifications@github.com> wrote:

I’m trying to package lilypon (a music engraving program) and it depends on
dblatex to build the documentation. So I also packaged dblatex and it’s
working for me. As mentioned above, I don’t use the build_style variable
and also not the find ... command in do_install(), the rest is the same.

So my question is if I shall include dblatex when I make a PR for lilypond
(despite this open PR) or if you would like to fix it first. :)

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/voidlinux/void-packages/pull/6347#issuecomment-329717604>,
or mute the thread

newbluemoon:
Ok, thanks for the quick reply! In that case I will build lilypond without documentation and hence no dblatex dependency. I can add it later when it’s in the repo.

the-maldridge:
@newbluemoon Now that I'm not acking an email right after I woke up I can give more explanation.  Void does not have native TeX packages, instead we rely on TexLive to install everything and tlmgr to manage it.  This works okay but it means that building packages that involve LaTeX in any way is painfully slow and often just fails outright.

This PR is here because Proxmox needs to build the documentation to be able to build anything (perl programs can be really really weird).  I'd rather not merge anything that needs dblatex until the TeX situation can be sorted out.  Until such time as that happens, please build your packages with no docs if possible, its not ideal but its a loss I'm willing to defend until the TeX ecosystem is native on Void.

newbluemoon:
@the-maldridge Thank you for your detailed explanation. Lilypond also depends on texlive so I know what you’re talking of ;-) I will disable the documentation for the time being. However, upstream also offers a documentation tarball. Would it be a good idea to create a ‘lilypond-doc’ package based on it?

the-maldridge:
Yes, I will reply on the lilypond PR.

newbluemoon:
Ok, thanks. Will do shortly.

PR #6350, opened 2017-05-08
openttd: use --with-xdg-basedir instead of hardcoded --personal-dir

### LINKED ISSUE #6349: openttd: remove hardcoded data directory, move to XDG instead
The template for `openttd` contains `--personal-dir` hardcoded, while upstream started to support XDG-style directories. It's more comfortable for me to store data like that in some reasonable place, like `~/data/local` (as bound to `XDG_DATA_HOME` environment variable) instead of polluting the root of home directory.

Solution: rebuild with `--with-xdg-basedir`autoconf parameter, remove hardcoded `--personal-dir=.openttd`.

CC: @pullmoll

Gottox:
fixes #6349


asergi:
Fixed in 43900e347b567eeeffbd293c39f2ebddd7224aa7.

maxice8:
please close @asergi

PR #6365, opened 2017-05-10
New package: brisk-menu-0.3.5

Jazqa:


srcpkgs/brisk-menu/template
```diff
@@ -2,6 +2,7 @@
 pkgname=brisk-menu
 version=0.3.5
 revision=1
+only_for_archs="i686 x86_64"
```
ebfe: That's not a fix. 

> ...
> configure: error: glib-compile-schemas not found.

Try adding glib to `hostmakedepends`.

Jazqa: Thanks.

PR #5651, opened 2017-01-31
New package: vidcutter-3.0.1

bra1nwave:
#5638

PR #6382, opened 2017-05-12
vidcutter: distfile checksum changed

bra1nwave:
The Author updated the distfile without version bump (See [AUR](https://aur.archlinux.org/cgit/aur.git/commit/?h=vidcutter&id=22f8f034cb7eff0c03cf3c6e45d471c4e2abacf7))
Something like this has happened in the [past](https://github.com/voidlinux/void-packages/pull/5651).

### DIFF
--- a/srcpkgs/vidcutter/template
+++ b/srcpkgs/vidcutter/template
@@ -1,7 +1,7 @@
 # Template file for 'vidcutter'
 pkgname=vidcutter
 version=3.2.0
-revision=1
+revision=2
 noarch=yes
 build_style=python3-module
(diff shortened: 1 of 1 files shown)


Duncaen:
Can you diff old and the new archive? Would be nice to find out what happened, and maybe suggest the author to avoid this.
This blocked a few builds yesterday and I had to retrigger the jobs manually after marking the packages as broken.

bra1nwave:
> Can you diff old and the new archive?

Sure, here you go: https://pastebin.com/UJNzxe7U

PR #6383, opened 2017-05-12
New package: pdnsd-1.2.9a

E100Beta:
`pdnsd` is a caching DNS server with (nearly unique) ability to preserve cache after restart.

By default it installs `pdnsd.conf.example` in `/etc`. It is not usefult by default and has some pitfalls. Therefore, to shorten the time from installation to getting it running I decided to provide my own, based on `pdnsd.conf` used in Alpine, with some redactions based on documentation and ArchWiki.

I tested the package locally for half a day on `x86_64-musl`, didn't encounter any problems so far.

**Things to consider:** Should I write a short INSTALL.msg to explain how to set everything up (put 127.0.0.1 in `/etc/resolv.conf` and write protect it, etc.)?


E100Beta:
Oh, and while poking around git of the project (which stops in April 2012), there are a couple of changes that never got released. There are quite a bit of potentially neat improvements there. I can provide it as a patch, if it's okay.

srcpkgs/pdnsd/files/pdnsd.conf
```diff
@@ -0,0 +1,45 @@
+# Minimal pdnsd.conf focused on performance.
+# See pdnsd.conf(5) and pdnsd.conf.sample for more settings.
```
the-maldridge: When a sample is provided we don't package a default, please remove this file.

srcpkgs/pdnsd/INSTALL.msg
```diff
@@ -0,0 +1,15 @@
+To start using pdnsd:
```
the-maldridge: This is excellent content for the wiki, but does not belong in the INSTALL.msg file.  This file is reserved for information that a system operator MUST be aware of when installing a package such as breaking changes or issues that would cause problems down the road.

srcpkgs/pdnsd/files/pdnsd/run
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+exec pdnsd 2>&1
```
the-maldridge: Must pdnsd run as root?  you created a system account below but didn't use it; I suspect this run script is incomplete.

srcpkgs/pdnsd/template
```diff
@@ -0,0 +1,25 @@
+# Template file for 'pdnsd'
+pkgname=pdnsd
+version=1.2.9a
+revision=1
+build_style=gnu-configure
+configure_args="--prefix=/usr --sysconfdir=/etc --sbindir=/usr/bin --enable-ipv6 --with-random-device=/dev/urandom"
+conf_files="/etc/pdnsd.conf"
+system_groups="pdnsd"
```
the-maldridge: please prefix this with an _

srcpkgs/pdnsd/template
```diff
@@ -0,0 +1,25 @@
...
+system_accounts="pdnsd"
```
the-maldridge: prefix with _

srcpkgs/pdnsd/template
```diff
@@ -0,0 +1,25 @@
...
+short_desc="Proxy DNS server with permanent caching"
+maintainer="Pavel Kiselev <kispash@yandex.ru>"
+license="GPL-3"
+homepage="http://members.home.nl/p.a.rombouts/pdnsd"
+distfiles="${homepage}/releases/pdnsd-${version}-par.tar.gz"
+checksum=bb5835d0caa8c4b31679d6fd6a1a090b71bdf70950db3b1d0cea9cf9cb7e2a7b
+
+do_install() {
```
the-maldridge: Don't override do_install, instead put your custom calls (vlicense and vsv) in post_install.

srcpkgs/pdnsd/INSTALL
```diff
@@ -0,0 +1,5 @@
+case "$ACTION" in
+post)
+	chown -R pdnsd:pdnsd /var/cache/pdnsd
```
the-maldridge: Perhaps this should be done in the run script to enforce ownership of this directory on startup?  The run script will initially have root authority and then should drop to the authority of the daemon, you could make this chown happen there instead of here to always have the correct permissions.

E100Beta: It's enforced in config file in `run_as` variable. 
Should I prefer program's internal privelege dropping mechanisms or method you suggest?

the-maldridge: It doesn't matter much to me, I provided my suggestion with reason above.

srcpkgs/pdnsd/INSTALL.msg
E100Beta: Okay, I'll remove this and add a wiki article after this merges.

srcpkgs/pdnsd/files/pdnsd/run
E100Beta: Same answer as here:  https://github.com/voidlinux/void-packages/pull/6383#discussion_r145331617

srcpkgs/pdnsd/INSTALL
ebfe: If you keep this: don't use absolute paths in INSTALL scripts.

srcpkgs/pdnsd/files/pdnsd.conf
E100Beta: The default was not working out of the box, but I agree, providing the entire other config is an overkill, I probably should add this, and a couple of "gotcha" moments I had with this file, to the wiki afterwards instead.

srcpkgs/pdnsd/INSTALL
E100Beta: @ebfe I'm a bit confused as to why. And how to _not_ use absolute path in that case? Is there something I missed in the manual?

I'll see about how I can enforce this at runtime, but `cache_dir` is a variable in the config and can't be set using command line. Does that mean I need to grep a config before I start a service?

ebfe: https://github.com/voidlinux/void-packages/blame/master/Manual.md#L939-L940

E100Beta: Thank you, sorry I missed it :(

E100Beta:
Okay, I adressed the requests above and tested it on my system, seems to be working fine.

Sorry I delayed this for such a long time!

Issue #6375, opened 2017-05-11
Libreoffice: CPU usage 100%


alasic:
As soon as launched, libreoffice uses 100% of one core.
I think the problem is recent. It appeared after the java support addition.

E100Beta:
@alasic same, and I don't think it's related to java. `musl` systems have java disabled, but when I launch Libreoffice i hear my laptop cry in pain with all its fan's noise capability :(

alasic:
I’m agree with you. My message wasn’t clear enough. I am quite sure that the issue was posterior to the java support addition, but not immediately posterior: last time I used libreoffice base, it was with the java support, and my fan wasn’t trying to elevate my laptop… ☹ So that’s would be more recent.

De : Pavel Kiselev<mailto:notifications@github.com>
Envoyé le :jeudi 11 mai 2017 20:52
À : voidlinux/void-packages<mailto:void-packages@noreply.github.com>
Cc : alasic<mailto:alasic@hotmail.fr>; Mention<mailto:mention@noreply.github.com>
Objet :Re: [voidlinux/void-packages] Libreoffice: CPU usage 100% (#6375)


@alasic<https://github.com/alasic> same, and I don't think it's related to java. musl systems have java disabled, but when I launch Libreoffice i hear my laptop cry in pain with all its fan's noise capability :(

Piraty:
same here

E100Beta:
A marginally funny finding - the CPU usage drops to 0 when I open Preferences. 
Also I noticed frequent CPU spikes when using Firefox. Both it and Libreoffice didn't do that before. I suspect one of the libs they both depend on.

Aaaaand massive breakthrough!
========================
After stumbling around not knowing what I'm doing for a bit, I fired up `gdb`, it complained on `glib`, then  I downgraded to `glib-2.52.1` and the issue disappeared.

PR #6388, opened 2017-05-13
glib: downgrade to 2.52.1

E100Beta:
Fixes https://github.com/voidlinux/void-packages/issues/6375


ebfe:
https://github.com/GNOME/glib/commits/glib-2-52. There are very few commits between 2.52.1 and 2.52.2. Only https://github.com/GNOME/glib/commit/2b47c8360c70a8455d61fa4bb94ec57b93bf228e looks like it cause this. Care to try with this patch reverted?

ebfe:
https://github.com/GNOME/glib/commit/5d74233476d61771b4d67d50bec5420f551ef922 might be the fix.

E100Beta:
@ebfe Huge thanks! Tested `glib-2.52.2_2` just now, Libreoffice works fine!

ebfe:
Thanks for identifying the source :)

xtraeme:
Michael you are the MAN

Issue #6206, opened 2017-04-18
Newsbeuter - STFL Parser Error near


neton-bricio:
I have a problem running newsbeuter using **Musl**. See what appears on terminal.

> XDG: configuration directory '/home/$USER/.config/newsbeuter' not accessible, using '/home/$USER/.newsbeuter' instead.
> Starting newsbeuter 2.9...
> Loading configuration...done.
> Opening cache...done.
> Loading URLs from /home/$USER/.newsbeuter/urls...done.
> Loading articles from cache...done.
> STFL Parser Error near ''.
> Aborted

neton-bricio:
any sugestion for this bug?

ebfe:
https://github.com/akrennmair/newsbeuter/issues/364

PR #6392, opened 2017-05-14
Newsbeuter: fix STFL error on musl

E100Beta:
fixes https://github.com/voidlinux/void-packages/issues/6206

Also `xlint` complained about not using package's license.
Is there a better way to conditionally patch the package than what I came up with?


Duncaen:
Don't patch conditionally.

E100Beta:
@Duncaen I know it's bad, just didn't want to rob `glibc` users of functionality... Well, okay, transliteration isn't that critical of a function. I reworked the commit.

E100Beta:
Okay, I got around git's segfaults and **now** it's reworked.

E100Beta:
Er... Now there're two commits. Okay, this is getting ridiculous...
*After some git tutorials*
Okay, it should be 100% fine.
I am REALLY sorry for all this. (BTW git crashes were my fault too, I was experimenting with localization on musl and it seems git didn't like it)

neton-bricio:
@E100Beta Thanks for this.

PR #6399, opened 2017-05-15
xfce4-terminal: update to 0.8.5.1

cr6git:


cr6git:
I see the problem. 

for version=0.8.5.1 , ${version%.*} in the path returns "0.8.5", see [line620](https://travis-ci.org/voidlinux/void-packages/jobs/232345892#L620). 

But it should be "0.8" instead, see the Xfce [repository](http://archive.xfce.org/src/apps/xfce4-terminal/0.8/). 

I'm so clumsy, stupid me! 
I apologize.

Gottox:
Not a prob, thanks!

PR #6400, opened 2017-05-15
linux4.4: update to 4.4.68

cr6git:


cr6git:
I use this kernel since this morning and it seems to work just fine. 
_(I guess the build will fail because it's too large and take too much time, as usual.)_

cr6git:
Hmm.. I just noticed the recent changes made by @Duncaen in the [dotconfigs](https://github.com/voidlinux/void-packages/commit/a260aeee134e516991f0131d5dcd280b9660bcb2) of linux4.9 

Do we need to make the corresponding changes in linux4.4 ? _(Side issue: is it relevant for linux3.18 also?)_ 

Regards.

cr6git:
I close this PR because kernel 4.4.69 has been released today, see [kernel.org](https://www.kernel.org/) ...

PR #6404, opened 2017-05-15
Update: mu-0.9.18_2

emacsomancer:
bump to trigger musl build (used not to build due to dependency on emacs)

[I contacted the old maintainer and he asked me to take over maintainership as he is no longer using Void]


ebfe:
Fixed commit message.

PR #6407, opened 2017-05-16
New package: isync-git-20170513_1

emacsomancer:


chneukirchen:
We generally don't add git packages, and especially not ones without stable tarballs. Can you explain why this package is needed?

emacsomancer:
@chneukirchen the last tarball release from 2015-11-08 doesn't seem to contain the ability to specify the "SubFolders" options (Verbatim &c.), which is a fairly major feature. And one I've been using for a few years.

I followed the basic approach of Void's `compton-git` package.

Vaelatern:
Please don't. The isync project needs to tag a release before we can update our package.

There are essential fixes in master that I use as well.

PR #6409, opened 2017-05-16
[rfc] drop gstreamer 0.x

ebfe:


chneukirchen:
It probably would be a good idea to rename the other packages then.

ebfe:
> It probably would be a good idea to rename the other packages then.

let's wait for the next soname bump

asergi:
- `FeedReader`, `farstream` and `telepathy-farstream` fixed (they are false-positive, already depend on Gstreamer 1.x)
- `gmusicbrowser` supports Gstreamer 1.x (I'll push the needed changes later aka free time)
- you need to remove `xfce4-mixer` from the `xfce4` meta-package as well
- you can already push the changes to `gnome-video-effects`, `snappy-player`, `pocketsphinx` and `xfburn` if you want

ebfe:
835b40ca9258b1f7e465b104cde3b16b4171a1fa

PR #6415, opened 2017-05-16
New package: font-ephifonts-20170517

aashiks:
Fonts from www.ephifonts.com including Helvetica, Gotham and Futura.


the-maldridge:
I'm not entirely confident that those are distributable, can you point to something concrete that says  they are?  I'm also not entirely happy with this as one monolithic package, perhaps you could split it up into the various font families?

aashiks:
http://www.ephifonts.com/disclaimer says they are distributable.

I packaged it for my convinience. I don't have time to deal with split ups and so on. It is up to you to decide whether or not to merge.

the-maldridge:
> I don't have time to deal with split ups and so on.
Then will you have time to maintain this package?  We have an expectation that all packages are actively maintained and updated as needed.

aashiks:
I will have time to deal with things that are actually relevant - like I said, I can deal with a packagethat I use. I can't deal with splitting it up since I don't see sense in doing so - and no time for that. I find it okay to deal with it as is. Again, it is up to you to merge.

the-maldridge:
You are correct, and I have chosen not to merge it as it does not meet our standards for packages and you have made clear you do not wish it to.

PR #6419, opened 2017-05-17
base-files: update to 0.140

isnogudus:
Updated base-files. New group uucp with gid 26 added to etc/groups. Same in arch linux. I need this for my uucp package (comming soon).


Gottox:
Please use system_accounts in your template instead.

See [Manual.md](https://github.com/voidlinux/void-packages/blob/master/Manual.md#creating-system-accountsgroups-at-runtime)

isnogudus:
Hi,

I discussed it with duncan on irc. I need the group at compile time, because the source package will set the right permissions. uucp needs suid/sgid flags on some executables. The other way is to change the owner after the installation.
Most BSD/Linux distributions have an uucp user or group.
Therefore system_accounts does not help here. What should I do to build a uucp package?

Best regards

Markus

PR #6427, opened 2017-05-18
New package: font-vollkorn-ttf-4.015

ivierlin:


srcpkgs/font-vollkorn-ttf/template
```diff
@@ -0,0 +1,28 @@
+# Template file for 'font-vollkorn-ttf'
+pkgname=font-vollkorn-ttf
```
the-maldridge: this should probably be just 'vollkorn' to match the pure upstream name, but as we don't have an official policy on fonts this can be corrected later if needed.

PR #6433, opened 2017-05-18
New package: Abricotine

praticamentetilde:
[Here](http://abricotine.brrd.fr) the website. I just made the template, not the application itself.


Anachron:
It's electron based so very unlikely to get merged.

praticamentetilde:
Atom is based on electron, and it's in the official repository. Do you mean that the package should depend on an `electron` package?

the-maldridge:
@praticamentetilde we're actively trying to get rid of the electron based apps since they aren't designed for packaging.  Atom comes from a time where we thought the quality of electron derived codebases would improve with time, this assertion turned out to be very much false.

praticamentetilde:
In this case, I guess I should close this.

the-maldridge:
feel free to leave it for review, another maintainer may be willing to merge it.

the-maldridge:
Other maintainers have had plenty of time to claim this PR, seems that (as expected) no one wants to shoulder another electron app.

PR #6450, opened 2017-05-19
wps-office: add update file and template cleanup

the-maldridge:


the-maldridge:
Travis fails this due to an unset patchLevel variable.  I've been keeping the variable though so that I don't have to add it back every other update when a patchLevel is present.

Vaelatern:
I don't like the installation into `/opt` but the cleanup looks good.

PR #6458, opened 2017-05-19
New package: hunspell-ru_RU-ieyo-0.3.9_1

kainonergon:


srcpkgs/hunspell-ru_RU-ieyo/template
```diff
@@ -0,0 +1,22 @@
...
+create_wrksrc=yes
+noarch=yes
+provides="hunspell-ru_RU-${version}_${revision}"
+replaces="hunspell-ru_RU>0"
+
+do_install() {
+	vinstall russian-aot-ieyo.aff 644 usr/share/hunspell ru_RU.aff
+	vinstall russian-aot-ieyo.dic 644 usr/share/hunspell ru_RU.dic
+	vlicense copyright.txt
+	vlicense lgpl.txt
```
Vaelatern: The `vlicense` lines are not necessary for `GPL` licenses.

srcpkgs/hunspell-ru_RU-ieyo/template
```diff
@@ -0,0 +1,22 @@
...
+short_desc="Russian dictionary for hunspell, variant with ie and yo"
+maintainer="Andrey Raugas <kainonergon@gmail.com>"
+license="LGPL-2.1"
+homepage="http://extensions.openoffice.org/en/projectrelease/russkiy-orfograficheskiy-slovar-aot-eyo-russian-spellcheck-dict-ieyo-based-works"
+distfiles="${SOURCEFORGE_SITE}/project/aoo-extensions/5149/3/dict_ru_ru-aot-${version}-ieyo.oxt>dict-ru_RU-ieyo.zip"
+checksum=3b374cf18f00fd0a9b6514eb8413438ce5d425a2e04435ffd81d8acc7ee1e578
+create_wrksrc=yes
+noarch=yes
+provides="hunspell-ru_RU-${version}_${revision}"
+replaces="hunspell-ru_RU>0"
```
Vaelatern: It would be preferred to have `alternatives` for these. This does not appear to be a "next" package for users of hunspell-ru_RU, but rather a variant.

If this is a variant, consider naming the files something other than `ru_RU.dic` and `.aff`, so they can both be installed at the same time.

kainonergon: If I set `alternatives` for `hunspell-ru_RU-ieyo`, should `alternatives` for `hunspell-ru_RU` also be set? If so, should I include changes for `hunspell-ru_RU` in the same PR?

Vaelatern: Yes and yes, but I don't think this should be a case of alternatives. hunspell's man page states:
```
       en_US and de_DE are base dictionaries, they consist of aff and dic file
       pairs: en_US.aff, en_US.dic and de_DE.aff, de_DE.dic.  En_geo, en_med,
       de_med are special dictionaries: dictionaries without affix file.
       Special dictionaries are optional extension of the base dictionaries
       usually with special (medical, law etc.)  terms. There is no naming
       convention for special dictionaries, only the ".dic" extension:
       dictionaries without affix file will be an extension of the preceding
       base dictionary (right order of the parameter list needs for good
       suggestions). First item of -d parameter list must be a base
       dictionary.
```
I read this as saying that there isn't an enforced naming convention, so you don't need to conflict or alternatives with `hunspell-ru-RU`

kainonergon: In my opinion there is no naming convention for special dictionaries, but base dictionary should be named ru_RU, or else spell checking settings will miss it. For example hunspell-fr_FR package uses the same name for different variants of dictionary.

the-maldridge:
Merged with minor commit change.

PR #6463, opened 2017-05-20
New package: xxkb-1.11.1_1

kainonergon:


E100Beta:
@kainonergon `vinstall ${FILESDIR}/XXkb ...` means you want to install file from <directory where your `template` is>/files/XXkb. Either it's not what you want or you didn't add this file.

Also, I dont think you need 3 `vmkdir` commands, `vinstall` automatically creates needed path.

Also-also, do you think that instead of doing most things in `do_install()` manually it would be easier to just `make DESTDIR=${DESTDIR} install` and then adjust things as needed?

kainonergon:
Yes, I mean to use /files/XXkb as default config. Just added it. I'll try to use `make DESTDIR=${DESTDIR} install`.

srcpkgs/xxkb/template
```diff
@@ -0,0 +1,32 @@
...
+makedepends="libX11-devel libXt-devel libXext-devel libXpm-devel"
+short_desc="Keyboard layout switcher and indicator"
+maintainer="Andrey Raugas <kainonergon@gmail.com>"
+license="PerlArtistic"
+homepage="https://sourceforge.net/projects/xxkb/"
+distfiles="${SOURCEFORGE_SITE}/${pkgname}/${pkgname}-${version}-src.tar.gz"
+checksum=ceaa33bed81d8ae5bf77e89787d0736f7564c3de047c35a57bcbf6f4d1888142
+
+do_build() {
+	xmkmf
```
E100Beta: Try changing to `CC=cc xmkmf`

srcpkgs/xxkb/template
```diff
@@ -0,0 +1,32 @@
+# Template file for 'xxkb'
+pkgname=xxkb
+version=1.11.1
+revision=1
+make_dirs="/usr/share 0755 root root
+	/usr/share/${pkgname} 0755 root root
+	/usr/share/X11/app-defaults 0755 root root"
+hostmakedepends="imake"
```
E100Beta: Try adding  `xorg-cf-files`.
[xautolock](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/xautolock/template) seems to use `imake`, maybe you'll find its contents useful.

kainonergon:
@E100Beta , I followed your suggestions and it seems it builds OK now. Thanks.

E100Beta:
@kainonergon glad I could help.
One last thing - maintainers prefer having one commit per package/change (read [this](https://github.com/voidlinux/void-packages/blob/master/CONTRIBUTING.md#review) for more detail, as well as how to handle this in the future).

Since you already have multiple commits, I propose you squash them. If you never did this before, I think [this](http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git#5201642) might help you.
Good luck with getting this merged :+1:

Vaelatern:
This in general looks good. I am okay with the results of the package. However, the template can do with some cleaning.
Using a `build_style="gnu-makefile"` and making use of `pre_build()` and `post_install()` would make me happier. This would call the `make` commands for you.
I am not thrilled about the custom configuration file in the repo. Is there a source for this file in the tarball?
@E100Beta Thank you for your help so far.

kainonergon:
@Vaelatern, there is a configuration file in the source tarball, but I am not sure it is a good one. So I borrowed default config from arch linux xxkb package. If that is not acceptable, I will remove it. The user can set needed config by himself.

kainonergon:
@Vaelatern, I've checked config file from the source tarball and have found that it is not colmplete and does not work.

Also I tried to use `build_style="gnu_makefile"`. Its `do_install` works OK, but its `do_build` does not work. It sets `CFLAGS="$CFLAGS"` argument for `make` and it would not build. So I had to use my own `do_build`.

Vaelatern:
Can you try with `make_use_env=yes` and not a custom `do_build`?

Vaelatern:
Oh nevermind, I see what you are saying.

srcpkgs/xxkb/template
```diff
@@ -0,0 +1,28 @@
...
+maintainer="Andrey Raugas <kainonergon@gmail.com>"
+license="PerlArtistic"
+homepage="https://sourceforge.net/projects/xxkb/"
+distfiles="${SOURCEFORGE_SITE}/${pkgname}/${pkgname}-${version}-src.tar.gz"
+checksum=ceaa33bed81d8ae5bf77e89787d0736f7564c3de047c35a57bcbf6f4d1888142
+build_style="gnu-makefile"
+make_install_target="install install.man"
+
+pre_build() {
+	CC=cc xmkmf
```
Vaelatern: What is this about?

srcpkgs/xxkb/template
```diff
@@ -0,0 +1,28 @@
...
+}
+
+do_build() {
+	make CC=$CC CDEBUGFLAGS="${CFLAGS}" EXTRA_LDOPTIONS="${LDFLAGS}"
```
Vaelatern: You could just set these variables with `export` in the `pre_build`, probably. That in combination with `make_use_env=yes` might fix the problem.

srcpkgs/xxkb/template
kainonergon: `xmkmf - create a Makefile from an Imakefile`
CC=cc was suggested by @E100Beta

E100Beta: I suggested that because it was in only other template I found which used `imake`: [xautolock](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/xautolock/template). I didn't actually test if it builds fine without this.

kainonergon: I'll try it without `CC=cc`.

kainonergon: It failed https://travis-ci.org/voidlinux/void-packages/builds/235588909
I will keep `CC=cc` then

the-maldridge:
@kainonergon can you drop the _1 from the commit name and reply to @Vaelatern's comments?  This looks otherwise good to merge.

PR #6435, opened 2017-05-18
Add debug-me

SevereOverfl0w:
I likely need feedback on the haskell-stack changes.

PR #6474, opened 2017-05-22
git-annex: update to 6.20170519

freshprince:
~upstream doesn't provide tarballs so I fetch the stable release from git repo~


chneukirchen:
There are tarballs at https://git.joeyh.name/index.cgi/git-annex.git/snapshot/git-annex-6.20170519.tar.gz

freshprince:
thanks @chneukirchen. i've updated the template.

freshprince:
btw.. this package would benefit if https://github.com/voidlinux/void-packages/pull/6435/commits/3a0b6b7caede4beea5d96cbca314202e5f91cf9a was already merged

PR #6475, opened 2017-05-22
efibootmgr: make one boot option per installed kernel

freshprince:
This might be more a RFC than a pull request.

I have `MODIFY_EFI_ENTRIES=1` in `/etc/default/efibootmgr-kernel-hook` and I'd like to have one boot option per installed kernel in my efi table.

Right now the post-install script installs one boot option per major version and tries to maintain the original boot order.  My patches change that behaviour so that you get one boot option per installed kernel and the newest installed kernel will get the first position in the boot order.

With this change it's easier to boot an older kernel if something breaks.

I don't know if this breaks workflow for some people so I'm happy to adjust the behaviour if I get feedback.


jannisch:
I really like the idea! The current behaviour is just inspired by gummiboot.
xbps doesn't delete old kernels from minor version updates in /boot so they are pretty useless and your proposed script is also much simpler. On the other hand, it's hard to remove the old kernels because this can only be done manually afaik (which does not update the efi loader)

freshprince:
Thanks for the feedback.  I was thinking of adding an option to always keep the original boot order but I assumed that people with `MODIFY_EFI_ENTRIES=1` usually want to boot the newest kernel automatically.

> it's hard to remove the old kernels because this can only be done manually afaik (which does not update the efi loader)

I use `vkpurge` to remove old kernels which also runs kernel.d scripts.

jannisch:
this whole script is problematic as https://github.com/voidlinux/void-packages/issues/5165 shows, you can't really expect the users' reactions. Part of the problem is that _many_ firmware implementations are broken.
Although most distributions offer prebuilt efistub kernels I haven't seen a comparable script as well.

Maybe moving the code of these hooks into one small cli tool would be an idea.     
  

> I use vkpurge to remove old kernels which also runs kernel.d scripts.

Oh, I forgot about vkpurge (I don't delete old kernels at all).

freshprince:
If there's no more input and void members agree then I humbly request this to be merged.

the-maldridge:
I see nothing technically wrong with this commit, but it does seem to be at odds with our policy of not screwing up people's machines (see previous issues with EFI and varying qualities of implementation).  Perhaps you can better explain how this works without screwing up a working system?

freshprince:
> Perhaps you can better explain how this works without screwing up a working system?

It doesn't screw up a working system because it doesn't do anything when `MODIFY_EFI_ENTRIES` isn't set (which is the default).

freshprince:
I also update the documentation in `INSTALL.msg` since some of it wasn't correct or won't be correct anymore when this (eventually) gets pulled.

freshprince:
I have another patch in the pipeline that adds an option to keep the current EFI boot order untouched which should be useful if the user has a different os he wants to boot to by default: https://github.com/freshprince/void-packages/commit/a10d46a7ba580329607eede28577ae9a3791d526
What is the right way to go here? Wait for this to get merged and then create another pull request or add it to this one or stop working on this stuff alltogether since people don't want it in void?

freshprince:
> Perhaps you can better explain how this works without screwing up a working system?

When this change gets applied and the user has `MODIFY_EFI_ENTRIES=1`.  A new entry in the EFI boot manager will be added for every newly installed kernel (as with the old setup). But with this patch the entry for the newly installed kernel will also be added in the bootorder as first place. So after a reboot you will automatically boot the latest installed kernel.

If a kernel gets removed with `xbps-remove` or `vkpurge` the corresponding entry in the EFI boot manager will be removed.

So I guess one of the biggest changes here is that the bootorder will change when installing a new kernel and I already have a patch to make this configurable. (Maybe make it off by default?)

The idea was to get similar behaviour as grub1 had (one boot entry per kernel).

Only entries with the naming scheme "void-${kernel_version}" will get created/removed and all other entries will be left untouched.

If we want to maintain the original bootorder, I could add this patch https://github.com/freshprince/void-packages/commit/ace51b1ce38a0e35922f164270a5e043f4816d11 and maybe inverse the `KEEP_EFI_BOOT_ORDER` variable (keep boot order by default).

freshprince:
i'll set 

```
preserve=/etc/files/kernel.d/*/50-efibootmgr
```

in `xbps.d`.

PR #6476, opened 2017-05-22
Fcitx m17n support

aashiks:
This commit introduces three new packages that are required to use m17n via fcitx


aashiks:
Have to split into multiple packages

PR #6478, opened 2017-05-22
New package: m17n-db-1.7.0

aashiks:
Bringing in support for m17n-db first, to be followed by libm17n and fcitx integration.


srcpkgs/m17n-db/template
```diff
@@ -0,0 +1,12 @@
+# Template file for 'm17n-db'
+pkgname=m17n-db
+version=1.7.0
+revision=1
+only_for_archs="i686 x86_64"
```
the-maldridge: I highly doubt this is architecture dependent since its `build_style=gnu-configure`.  If upstream does not support cross compiling you can use `nocross=yes` to disable non-native builds.  If its only broken on specific sub-architectures, use the case structure paradigm to selectively mark this broken with the broken variable set to the travis build URL of the failing build.

asergi:
#5484

srcpkgs/m17n-db/template
aashiks: I am closing off this pull request because I have moved on from void-linux and is no longer interested in doing anything with it. It was nice to use and so on, but I have had better experiences with other communities. Good luck to you guys !

PR #6483, opened 2017-05-22
poppler: update to 0.55.0.

ebfe:


asergi:
> build system:
>          * Fail by default if libjpeg is not available
>          * Fail by default if libopenjpeg2/1 is not available

- you can remove `--enable-libjpeg`
- I think it's time to build `poppler` with `libopenjpeg2` too

What do you think? Any objections?

ebfe:
dropped useless configure switch, enabled openjpeg2.

PR #6487, opened 2017-05-23
New packages: Malayalam Fonts

aashiks:
A set of free and open source fonts released by https://smc.org.in, a developer collective from India


srcpkgs/font-meera-ttf/template
```diff
@@ -0,0 +1,24 @@
...
+checksum=4b03f9625c1693b38136d009e290fb9ae79a28da8b8754ad3de30f7aec04ffc7
+
+do_install() {
+	vmkdir usr/share/fonts/TTF
+	install -m644 *.ttf ${DESTDIR}/usr/share/fonts/TTF
+
+	# install fontconfig files
+	vinstall 65-0-smc-meera.conf 644 \
+		etc/fonts/conf.avail 65-0-smc-meera.conf
+}
```
the-maldridge: This file needs a trailing newline.

the-maldridge:
Why do all the versions have a datecode appended to them?

srcpkgs/font-meera-ttf/template
aashiks: Will fix this.

aashiks:
The datecodes are as per the font versions in the font files that upstream produces. I just host them since they don't do it in a manner that makes it easy to package.

the-maldridge:
Given that the artifacts for this package aren't actually being pulled from the upstream source, I'll need a second opinion on this before merging.

aashiks:
Where would you like the opinion from ? The upstream devs ?

the-maldridge:
A second Void Linux Member.

aashiks:
Alright. I am part of the collective that makes this fonts.  Do you want me to get the person who maintains Debian packages for these fonts involved ?

the-maldridge:
Unless that person is also a member of the Void Linux packages team, then they do not satisfy the constraint above.  A far easier solution would just be to pull the fonts from the upstream distribution URL, as I hinted at above.

aashiks:
The upstream URLs does not included a version and they don't package it in compressed formats. Me and the other distro packagers are working with them on this. Do consider merging, if not I will come back to this pull request once I get the upstream to host it in a proper manner and update the files as you specified.

the-maldridge:
Our requirements and expectations stand.  Please feel free to revisit and reopen this PR after your organization has made the changes you mention.

aashiks:
Thanks. Will do so.

chneukirchen:
We have fontforge, can't we build them ourselves?

PR #6488, opened 2017-05-23
nvidia: update to 381.22

freshprince:
dkms builds fine on x86_64 for 4.10 and 4.11


common/shlibs
```diff
@@ -139,7 +139,7 @@ libnvidia-gtk2.so.346.47 nvidia-gtklibs-346.47_1 ignore
 libnvidia-gtk3.so.346.47 nvidia-gtklibs-346.47_1 ignore
 libnvidia-glcore.so.346.47 nvidia340-libs-340.46_1 ignore
 libnvidia-glsi.so.346.72 nvidia-libs-346.72_1 ignore
-libnvidia-fatbinaryloader.so.378.13 nvidia-libs-378.13_1 ignore
+libnvidia-fatbinaryloader.so.381.22 nvidia-libs-381.22_1
```
Hoshpak: I think you should leave the ignore in place here.

srcpkgs/nvidia/template
```diff
@@ -197,7 +192,7 @@ do_install() {
 	vman nvidia-smi.1
 
 	# nvidia-modprobe
-	vinstall nvidia-modprobe 4755 usr/bin
+	vbin nvidia-modprobe
```
Hoshpak: Although xlint will complain, I believe vinstall to be correct here. The nvidia installer explicitly installs nvidia-modprobe as a suid binary.

freshprince:
thanks for the feedback @Hoshpak!

I've updated the pull request

freshprince:
https://github.com/voidlinux/void-packages/commit/716683264715fb95213226312f53a82b1bb37c90

PR #6491, opened 2017-05-23
phpMyAdmin: update to 4.7.0

the-maldridge:
This should also fix the issue of there not being a package for this since the noarch package exists.


Duncaen:
#6490 miscommunication all day.

the-maldridge:
Would it be void if it wasn't?

PR #6492, opened 2017-05-24
New package: gradio-5.0

logenkain:


silvernode:
Logan was nice enough to package Gradio for me pon request. I have his package installed and everything checks out over here. Radio has been playing for hours. all stations I have tested worked as expected. Great little program

srcpkgs/gradio/template
```diff
@@ -0,0 +1,16 @@
+# Template file for 'gradio'
+pkgname=gradio
+version=5.0
+revision=1
+build_style=gnu-configure
+configure_script="./autogen.sh"
+only_for_archs="x86_64 i686 x86_64-musl i686-musl"
```
Duncaen: this seems wrong, I can't imagine that it is architecture dependent.

logenkain: I suspect you may be right. I'll change it to nocross. Not sure how to actually test it though.

logenkain:
In case it's relevant: https://travis-ci.org/logenkain/void-packages/jobs/235458932
^ arm error

the-maldridge:
I have a commit on top of this that fixes cross compiling.  It just needed a few extra packages to be available during build and then it was fine.  The build came out fine with no errors, but short of pulling out a pi I can't confirm that it actually works.  Given that our standard of "working" for new packages has been green builds on Travis, I'm merging this as it sits.

Quite an enjoyable program, I have it running in the background at the moment.

silvernode:
Awesome, thanks a lot, glad this got merged. Also I am glad you like it @the-maldridge I agree that it is an enjoyable program.

PR #6494, opened 2017-05-24
go: update to 1.8.3.

ebfe:
As always manual rebuild of host go on the aarch64 builders required before pushing the revbumps.


ebfe:
updated to go 1.8.3

srcpkgs/flannel/template
```diff
@@ -10,7 +10,7 @@ license="Apache-2.0"
 go_import_path="github.com/coreos/flannel"
 homepage="https://$go_import_path"
 distfiles="$homepage/archive/v${version}.tar.gz"
-checksum=f8b837bc042c917cd37cfe131132ba1fde5268fadb7632da9c16f57d0905161b
+checksum=3c3148eac3537fba6c5377922366ed36b1415d20111434172c5c42fec74db366
```
Gottox: Any idea what has changed here?

thypon: They probably force pushed a tag to a different revision.

ebfe: the only diff is in a version string of a dependency.

Gottox: 🙄 when will they learn.

PR #6505, opened 2017-05-25
orca: fix build-time detection of liblouis.

CMB:
Contracted braille breaks if we can't find the .pc file for
liblouis at build time, so liblouis-devel belongs in makedepends.

### DIFF
--- a/srcpkgs/orca/template
+++ b/srcpkgs/orca/template
@@ -1,11 +1,11 @@
 # Template file for 'orca'
 pkgname=orca
 version=3.24.0
-revision=1
+revision=2
 noarch=yes
 build_style=gnu-configure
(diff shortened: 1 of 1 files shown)


Vaelatern:
Amazing for a noarch package.

CMB:
> Amazing for a noarch package.

I think so too.  But Orca needs to know which braille tables are
available.  There's no way to get that info at runtime from the library,
so the only recourse is to go poking around in the tables directory --
which is /usr/share/liblouis/tables on a Void install.  The library
also doesn't provide a way to get that path at runtime, but it can be
gotten through pkg-config at build-time.  Very unusual.

Anyway, thank you for the merge.

PR #6507, opened 2017-05-26
gajim: fix plugin installer

lonwillett:
The plugin-installer plugin in gajim has a few minor build problems that prevent it from working. I believe that this should fix it.

### DIFF
--- a/srcpkgs/gajim/patches/include-root-cert.patch
+++ b/srcpkgs/gajim/patches/include-root-cert.patch
@@ -1,0 +1,22 @@
+--- plugins/Makefile.am.orig	2017-02-01 14:16:03.000000000 +0100
++++ plugins/Makefile.am	2017-05-22 17:25:16.121616663 +0200
+@@ -7,7 +7,7 @@
+ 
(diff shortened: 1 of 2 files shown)


Vaelatern:
Can you open a PR with upstream?
I can't wrap my head around this patch right now. Not sure exactly what you are changing (my problem, not yours).

lonwillett:
The problem is that the gajim feature that allows installing plugins on the fly does not work.

When I set out to solve this, I found that the problem is that they added certificate pinning for the downloads, as a security measure. This proved to be broken on void for two reasons.

The first is that they forgot to install the certificate file. This is already fixed upstream, but not released.

The second is that python-pyasn1 needs to be added as a dependency. This needs to be done in the template file.

I broke the one commit per pull request here intentionally, because I wasn't clear how important it was to you maintainers (I'm new to void), and it seemed inappropriate, since the fix is really two parts.

So what should I do? The options are:

 - Squash the two commits. This will fix the problem for now. The patch file will need to be thrown out when there is a new upstream release.

 - Throw out the patch file, but include the changes to the template file. This will not fix the problem, but at least it should then work when there is a new upstream release.

- Break the two commits into two PRs.

 - Leave things as they are.

Let me know what I should do.

PR #6508, opened 2017-05-26
gajim-otr: fixed broken xbps-src fetch

lonwillett:
The gajim web server seems to regenerate its plugin's zip files
frequently (hourly?), and the file order and/or other properties seem
to be indeterminate.

The bottom line is that the contents constantly change, so the
xbps-src digest check fails, even though the actual contents of the
archive haven't changed.

This adds a custom fetch function, which checksums the files in a
reproducible fashion. It's rather a kludge, but seems to work.


srcpkgs/gajim-otr/template
```diff
@@ -1,18 +1,45 @@
 # Template file for 'gajim-otr'
 pkgname=gajim-otr
 version=1.9.8
-revision=2
+revision=3
 noarch=yes
 wrksrc="gotr"
 pycompile_dirs="usr/share/gajim/plugins/gotr"
-hostmakedepends="unzip"
+hostmakedepends="wget unzip"
```
Vaelatern: No need, you can use `xbps-uhelper fetch`

srcpkgs/gajim-otr/template
```diff
@@ -1,18 +1,45 @@
 # Template file for 'gajim-otr'
 pkgname=gajim-otr
 version=1.9.8
-revision=2
+revision=3
```
Vaelatern: No reason to revbump since the generated package shouldn't change.

Vaelatern:
Is there no better solution? Perhaps a different source?

PR #6509, opened 2017-05-26
New package: python-axolotl-curve25519-0.1

lonwillett:
curve25519 with ed25519 signatures, used by libaxolotl.


lonwillett:
Since this package is a dependency of python-axolotl, and therefore needs to be included by it to pass the build checks, I am removing it as a seperate PR.

PR #6510, opened 2017-05-26
New package: python-axolotl-0.1.39

lonwillett:
Port of libaxolotl-android written by Moxie Marlinspike.


lonwillett:
Since this package is a dependency of gajim-omemo, and therefore needs to be included by it to pass the build check, I am closing it a separate PR.

PR #6511, opened 2017-05-26
New package: gajim-omemo-1.2.1

lonwillett:
OMEMO plugin for gajim

### DIFF
--- a/srcpkgs/gajim-omemo/template
+++ b/srcpkgs/gajim-omemo/template
@@ -1,0 +1,48 @@
+# Template file for 'gajim-omemo'
+pkgname=gajim-omemo
+version=1.2.1
+revision=1
+noarch=yes
+wrksrc="omemo"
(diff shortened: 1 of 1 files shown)


Vaelatern:
(I am not adopting this PR right now).
At first brush, the packages seem reasonable. Nothing obviously wrong with them except that fetch function in `gajim-omemo`, but there is a separate discussion on that in #6508.
The rule of 1 commit per template is broken here, please iron out the merge commits and squash c497857 with d8d50d2
I like your commit naming otherwise.
The `gajim-omemo` package could probably stand to have a build style, and depend on `gajim`.
Your commented out `depends` in `python-axolotl-curve25519` should be uncommented or deleted. Probably uncommented.

lonwillett:
All comments taken on board. Except that I am not sure what to do about the dodgy fetch function.

PR #6513, opened 2017-05-26
spotify: update to 1.0.55

teh-jazzman:


CMB:
Same PR as 6504, but the submitter of 6504 typoed the version
number in the commit message.  Maybe 6504 should be closed
and this one should be used instead, since you're the maintainer?

teh-jazzman:
Oh, didn't see the previous pull request. The version number could be a problem, otherwise I don't really care whos version gets to be pulled. :)

PR #6522, opened 2017-05-26
gajim-otr: fixed broken xbps-src fetch

lonwillett:
The gajim web server seems to regenerate its plugin's zip files
frequently (hourly?), and the file order and/or other properties seem
to be indeterminate.

The bottom line is that the contents constantly change, so the
xbps-src digest check fails, even though the actual contents of the
archive haven't changed.

This adds a custom fetch function, which checksums the files in a
reproducible fashion. It's rather a kludge, but seems to work.

### DIFF
--- a/srcpkgs/gajim-otr/template
+++ b/srcpkgs/gajim-otr/template
@@ -11,8 +11,35 @@ short_desc="OTR Plugin for Gajim"
 homepage="https://dev.gajim.org/gajim/gajim-plugins/wikis/OffTheRecordPlugin"
 maintainer="Helmut Pozimski <helmut@pozimski.eu>"
 license="GPL-3"
-distfiles="https://ftp.gajim.org/plugins_0.16_zip/gotr.zip"
-checksum="00ba5bf027b82d5f6a9867af5295b6518835b138c5ac22a891102bfb26b32a1d"
+#distfiles="https://ftp.gajim.org/plugins_0.16_zip/gotr.zip"
+checksum="c80f7dca53cc9e1c7ba0cfa955721217a7dd27220f0f07b99a79f6ad478dfc05"
+
+do_fetch() {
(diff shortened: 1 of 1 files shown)


lonwillett:
Oops. Sorry. Accidentally closed this PR, and have to resubmit it now.

As for the small things from the comments:

 - Right you are about the version number bump. I wasn't thinking.

 - I changed the "wget" to "$XBPS_FETCH_CMD" and removed the dependency on it.

As for the bigger issue, as to whether or not this kludge is the way to go, I'll admit that it's not entirely clear.

I couldn't find another obvious place to obtain the source archive from. I did just notice that Helmut Pozimski (the package maintainer) opened an upstream request a couple of months ago, asking for a stable source package. I don't know what magic he used to get this built in the meantime.

I settled on this kludge as a reasonable, hopefully temporary, fix to the problem. It allows the package to build, while still checking a digest. The less kludgey solutions that I can think of seem like overkill for a small problem that will hopefully go away. The simpler solution of just not checking a digest seems like a bad idea. I'm open suggestions, but it seems like something should be done.

Hoshpak:
There wasn't any magic involved when I submitted the package, it was just a simple manual workaround ;). I wasn't aware of the fact that they generate the plugin archives as often and submitted the package with the hash sum of the one I got when I downloaded the archive. After merging the build failed and someone with write access to the repository corrected the hash sum to the current one. So it was built successfully once and is on the mirrors now. After that, I submitted the upstream request https://dev.gajim.org/gajim/gajim-plugins/issues/197 which unfortunately didn't receive any replies yet.

Vaelatern:
I appreciate opening the upstream issue. I'd rather wait on this to determine if upstream is willing to do consistent downloads. If they are not, then we can try to work out something.

Issue #6514, opened 2017-05-26
KDE frameworks package versions removed from KDE mirror


Hoshpak:
The KDE frameworks packages in Void are currently at version 5.29. I just discovered that these versions were removed from the KDE server. https://download.kde.org/stable/frameworks/ only lists versions from 5.30 on. This will lead to issues if rebuilds are necessary.

The packages should either be updated to the current version or the distfile URLs be changed to another server still hosting the files (if such a server exists).

Vaelatern:
Finding such a server with older software versions would be preferred.

amak79:
Older releases can be found here: https://download.kde.org/Attic/frameworks/

Hoshpak:
Thanks, I will create a pull request.

ebfe:
#6052 merged

PR #6052, opened 2017-03-30
Update KDE Framework to 5.36

yopito:

PR #6526, opened 2017-05-27
Change distfile location for kdeframeworks packages

Hoshpak:
This is to address #6514 . Change the distfiles URLs for the kdeframeworks packages from stable to Attic on the KDE mirror. No rebuild/revbump needed.


Vaelatern:
I'll have to check if the attic also has the most recent versions, or if it's just the older versions

Hoshpak:
The directory contains only the versions that were removed from the stable directory, currently these are frameworks versions up to 5.29. If these packages would be updated in the near future, one would have to revert the URLs back to the stable directory. If you think it's better to leave it as it currently is, I don't mind closing this pull request and leaving it unchanged. My only concern was that possibly necessary rebuilds of the packages will inevitably fail if the URLs are not changed and since KDE 5 seems pretty much to be stuck, I wouldn't expect updates to these packages in the near future.

ebfe:
@Vaelatern ping

yopito:
will be unecessary if #6052 is merged.

Hoshpak:
Lukcily made obsolete by #6052 having been merged.

PR #6527, opened 2017-05-27
linux4.11: fix iommu error on AMD processors

Piraty:
see:
 https://bugs.launchpad.net/ubuntu/+source/xserver-xorg-video-amdgpu/+bug/1683184
patch:
 https://patchwork.freedesktop.org/series/21684/

I used this patch on 4.11.2 a few days and on 4.11.3 since the template was updated to .3
Without this patch, my machine doesn't boot stable unless i use iommu=soft cmdline option.


Vaelatern:
I am in favor of merging. Is there a timeline for when this might hit mainstream linux?

Piraty:
I can't tell, i did not have contact with the amd developers.
But i will check regularly and report back in this PR / irc.

Thanks for merging

PR #6530, opened 2017-05-27
New package: pithos-1.3.1

OSN64:


Vaelatern:
First review:
Template doesn't look awful, commit message is totally wrong (but depending on your comfort with git this may just be something I should fix myself) and your INSTALL.msg is at least wrong, but probably should be removed.
Pending actual review later.

OSN64:
Ah, ill fix the commit message, missed it in the CONTRIBUTING.md. How should i specify optional dependencies?

Vaelatern:
It's in Manual.md
Currently we don't specify optional deps, but how optional are they?

OSN64:
They enable these plugins. So they are not needed to run the application.
Mpris: Allows player control via dbus. - python3-dbus 
Lastfm: Scrobble tracks listened to on Last.fm - python-pylast

the-maldridge:
merged with minor changes

PR #6538, opened 2017-05-28
skalibs: allow cross compilation.

CMB:


srcpkgs/skalibs/template
```diff
@@ -1,20 +1,27 @@
 # Template file for 'skalibs'
 pkgname=skalibs
 version=2.5.1.1
-revision=1
-nocross=yes
+revision=2
```
Vaelatern: Don't bother bumping revision if just enabling cross building.

srcpkgs/skalibs/template
```diff
@@ -1,20 +1,27 @@
...
 configure_args="--libdir=/usr/lib --enable-static --enable-shared --enable-clock
  --enable-monotonic --enable-force-devr --datadir=/usr/share/$pkgname"
 short_desc="General purpose libraries for building software from skarnet.org"
 maintainer="bougyman <bougyman@voidlinux.eu>"
 license="ISC"
 homepage="http://skarnet.org/software/skalibs/"
-distfiles="http://skarnet.org/software/${pkgname}/${pkgname}-${version}.tar.gz"
-checksum=aa387f11a01751b37fd32603fdf9328a979f74f97f0172def1b0ad73b7e8d51d
+distfiles="http://skarnet.org/software/${pkgname}/${pkgname}-${version}.tar.gz
+ https://github.com/CMB/${_sysdepspkg}/archive/${version}${_sysdepsrev}.tar.gz"
```
Vaelatern: You fundamentally don't understand that the `revision` variable is for internal (Void) use only.

CMB:
> You fundamentally don't understand that the revision variable is for internal (Void) use only.

The sysdeps repo I created is also for Void use only.  So pinning the
sysdeps to the version+revision of the Void package seemed like a
reasonable thing to do.  But on the other hand, I can't imagine sysdeps
changing across a package revision, so maybe I'm just overthinking
things.

CMB:
> Don't bother bumping revision if just enabling cross building.

Fixed, thanks.  And the sysdeps collection is no longer tied to
revision.

Vaelatern:
This does make the package un-bumpable except through you.

CMB:
> This does make the package un-bumpable except through you.

True.  So I fixed the versioning on my sysdeps collection, as well as
the template.  Now it just has the same version as skalibs itself.
I couldn't think of a better way to handle these then putting them in a
repo.  Dumping a pile of files into srcpkgs/skalibs/files/sysdeps
seemed wrong.
It took some work to create this; I'd really love to get it right.

Thanks!

Gottox:
@Vaelatern state of this PR? approve or wait?

Vaelatern:
Okay, I'll approve this, but I don't want your repo to be necessary for all builds. Please `+=` the checksum and the repo location within the conditional for cross building.

CMB:
Done; the location and checksum are conditional.
I get the impression that you're really not comfortable with this PR.
And that's cool.  If it doesn't get merged, it's fine.  It can sit in
the queue -- or I can close it -- until someone comes up with a better way
to do this that doesn't involve tossing the sysdeps into an external
repo.

ebfe:
> until someone comes up with a better way to do this that doesn't involve tossing the sysdeps into an external repo.

Can't you just put these (along with instructions on ow to generate them) into `files/`?

CMB:
Right now it's 81 files plus the readme.  I could cut that down to 54,
since i686, x86_64, and x86_64-musl are just included for completeness.
That seems like a lot to put in files/, but if that's really not a
problem, it would be the way to go.

PR #6540, opened 2017-05-28
docker: fetch binaries from get.docker.com for arm support

praticamentetilde:
I've edited the template in order to dowload the binaries from https://get.docker.com for easy arm support. Please note that the binaries for `aarch64` remain 32 bit. If packaging from binaries like this is not allowed, I will close this pull request.


Vaelatern:
Crossbuilding is strongly preferred.

PR #6552, opened 2017-05-30
New package: font-3270-1.2.21

diogoleal:


chneukirchen:
Sure it needs fontforge-devel?

diogoleal:
@chneukirchen you're right. removed fontforge-devel.

PR #6558, opened 2017-05-31
New package: skype-5.3.0.1

lonwillett:
Repackaging of the Debian version of Skype for Void.

This is the newest ("beta") version for x86_64.

Nothing interesting here, but I just thought it might be nice to share the template with others that have to use skype.

### DIFF
--- a/srcpkgs/skype/template
+++ b/srcpkgs/skype/template
@@ -1,46 +1,39 @@
 # Template file for 'skype'
+# This just repackages the debian package.
 pkgname=skype
-version=4.3.0.37
-revision=5
-short_desc="P2P software for high-quality voice communication"
-maintainer="Juan RP <xtraeme@voidlinux.eu>"
-homepage="http://www.skype.com"
-license="skype"
-distfiles="http://download.skype.com/linux/${pkgname}-${version}.tar.bz2"
-checksum=8c99dc3978a588fb13303df7c8134379fb55b8dd460efefbc79ae594269b892d
-
-only_for_archs="i686"
+version=5.3.0.1
+revision=1
+only_for_archs="x86_64"
 repository="nonfree"
+build_style="fetch"
 nostrip="yes"
-depends="pulseaudio qt hicolor-icon-theme ca-certificates alsa-plugins-pulseaudio pavucontrol qt-webkit"
-lib32mode="full"
+short_desc="Skype for Linux Beta"
+maintainer="Lon Willett <xgit@lonw.net>"
+license="skype"
+homepage="https://www.skype.com"
+distfiles="https://repo.skype.com/deb/pool/main/s/skypeforlinux/skypeforlinux_${version}_amd64.deb"
+checksum=3fdb61292cfdeef42ddb473d019884a4a8c5798f871793086f5edb6fb14d4e23
+
+do_extract() {
+	ar p ${XBPS_SRCDISTDIR}/${pkgname}-${version}/skypeforlinux_${version}_amd64.deb data.tar.xz | tar xJpf - ./usr
+}
 
 do_install() {
-	# Executable
-	vbin ${FILESDIR}/skype.sh skype
-	vinstall skype 755 usr/lib/skype
-	
-	# Data
-	vmkdir usr/share/skype
-	vcopy avatars usr/share/skype
-	vcopy lang usr/share/skype
-	vcopy sounds usr/share/skype
-	
-	# DBus Service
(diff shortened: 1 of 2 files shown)


lonwillett:
@cr6git 
So I see. Thanks.

Vaelatern:
And pray tell why you don't edit the `skype` template that we already have?

lonwillett:
@Vaelatern 

I certainly borrowed from the existing skype template. The reasons for making this a new package are:

 - The "new" skype is a radically different package, not a small update to the existing one. Also, it is designated as "beta", although I shan't be surprised if we never see a non-beta version of skype for linux.
 - Their are no conflicts between them, since all the location names have changed from "skype" to "skypeforlinux".
 - The "new" skype is only for 64 bit. The "old" skype is only for 32 bit.
 - There are compatibility issues which mean that one may want to have both versions installed (I do). At the least, I have discovered that the "old" and "new" versions don't seem to do video calls between each other. So I have to use the "old" version to do a video call with some people that aren't using the "new" version. And there may well be more issues that I'm not aware of.

Bottom line: neither of the two versions is ideal, so it seems easier to have them as separate packages, so that one can see which one works better. I don't feel confident in recommending that the "old" skype be replaced by the "new" one.

Perhaps the name of the package could be chosen a bit better, although the advantages of using the same name as upstream are manifest.

m-wynn:
Skype for Linux clients version 4.3 or older are set to be retired on July 1, 2017.

https://www.skype.com/en/download-skype/skype-for-linux/

srcpkgs/skypeforlinux/template
```diff
@@ -0,0 +1,40 @@
...
+	vbin usr/bin/skypeforlinux
+
+	vmkdir usr/share
+	vcopy usr/share/skypeforlinux usr/share
+
+	# These files must go in expected location, but
+	# perhaps some symbolic links are called for?
+	vmkdir usr/share/doc
+	vcopy usr/share/doc/skypeforlinux usr/share/doc
+	#vlicense usr/share/doc/skypeforlinux/copyright
```
the-maldridge: This probably shouldn't be commented out.

srcpkgs/skypeforlinux/template
```diff
@@ -0,0 +1,40 @@
...
+	#vlicense usr/share/doc/skypeforlinux/third-party_attributions.html
```
the-maldridge: Neither should this...

the-maldridge:
So come July 1st we will need to remove the old skype template.  I'd much prefer if this template was called 'skype' and just seamlessly updated from the old version.  @lonwillett can you make those changes so that this becomes just a seamless upgrade?

lonwillett:
@the-maldridge @m-wynn 

I'm not sure the upgrade can be considered seamless, but with the skype-32bit package disappearing, I suppose this may as well be renamed to just skype.  This is now done. I also installed the license files in the correct location, and put symbolic links to them from /usr/share/doc/skypeforlinux.

srcpkgs/skype/template
```diff
@@ -1,46 +1,39 @@
-# Template file for 'skype'
+# Template file for 'skypeforlinux'
```
the-maldridge: This should say 'skype'

srcpkgs/skype/template
```diff
@@ -1,46 +1,39 @@
...
+	vbin usr/bin/skypeforlinux
+
+	vmkdir usr/share
+	vcopy usr/share/skypeforlinux usr/share
+
+	vlicense usr/share/doc/skypeforlinux/copyright
+	vlicense usr/share/doc/skypeforlinux/third-party_attributions.html
+	# The binary needs to be able to find these under /usr/share/doc
+	vmkdir usr/share/doc/skypeforlinux
+	ln -s ../../licenses/skype/copyright ../../licenses/skype/third-party_attributions.html ${DESTDIR}/usr/share/doc/skypeforlinux
```
the-maldridge: I'm not sure this does what you intended it to do.

lonwillett: I am new to Void, so perhaps there is some canonical way to accomplish this that I don't know about. But it seems to work for me: the "copyright" and "third-party-attributions.html" files get installed in /usr/share/licenses/skype, which is where Void policy says they should be, and symbolic links to them are installed in /usr/share/doc/skypeforlinux, which is where the binary expects to find them.

the-maldridge: Ah, I did not realize that skype expected to be able to read those files.  That's good enough for me.

PR #6563, opened 2017-05-31
New package: nanomsg-1.0.0

The-Fred:
This package is required by another that I have also created a template for. I will create a PR for that one next.


The-Fred:
Actually I got 6 out of 8 passed. No idea what to do about aarch64/arm64 and armv7hf.

PR #6567, opened 2017-06-01
glew: update to 2.0.0

ibrokemypie:


Gottox:
We'll take #6596 as it correctly bumps all dependencies. Thanks for your effort nonetheless!

PR #6570, opened 2017-06-01
New package: nanomsg-1.0.0

The-Fred:
This package is required by another that I have also created a template for. I will create a PR for that one next.

Also after reading other pull request dialogue I added nocross="yes".  It seems like the correct thing to do. And also made a new branch as that also seems like the correct thing to do.

### DIFF
--- a/common/shlibs
+++ b/common/shlibs
@@ -2768,4 +2768,4 @@ libgcab-1.0.so.0 gcab-0.7_1
 libappstream-glib.so.8 appstream-glib-0.6.13_1
 libappstream-builder.so.8 appstream-glib-0.6.13_1
 libflatpak.so.0 flatpak-0.9.3_2
-libnanomsg.so.5.0.0 nanomsg-1.0.0_1
+libnanomsg.so.5 nanomsg-1.0.0_1
--- a/common/shlibs
+++ b/common/shlibs
@@ -2768,3 +2768,4 @@ libgcab-1.0.so.0 gcab-0.7_1
 libappstream-glib.so.8 appstream-glib-0.6.13_1
 libappstream-builder.so.8 appstream-glib-0.6.13_1
 libflatpak.so.0 flatpak-0.9.3_2
+libnanomsg.so.5.0.0 nanomsg-1.0.0_1
--- a/srcpkgs/nanomsg/template
+++ b/srcpkgs/nanomsg/template
@@ -1,0 +1,36 @@
+# Template file for 'nanomsg'
+pkgname=nanomsg
+version=1.0.0
+revision=1
+short_desc='Simple, high-performance implementation of "scalability protocols"'
+homepage='http://nanomsg.org/'
+license="MIT"
+depends="glibc"
+makedepends="cmake"
+build_style="cmake"
+shlib_provides="lib${pkgname}.so.5.0.0"
+nocross="yes"
+distfiles="https://github.com/nanomsg/nanomsg/archive/$version.tar.gz>nanomsg-${version}.tar.gz"
+checksum="24afdeb71b2e362e8a003a7ecc906e1b84fd9f56ce15ec567481d1bb33132cc7"
+
+pre_configure() {
+	configure_args="-DCMAKE_INSTALL_PREFIX=/usr"
+}
+
+do_build() {
+	cd $wrksrc/build
+	cmake --build .
+}
+post_build() {
+	cd $wrksrc/build
+	ctest
+}
+
+do_install() {
+	cd $wrksrc/build
+	DESTDIR=${DESTDIR} cmake --build . --target install
+}
+
+post_install() {
+	vlicense COPYING
+}
--- a/srcpkgs/nanomsg/template
+++ b/srcpkgs/nanomsg/template
@@ -9,10 +9,6 @@ build_style=cmake
 distfiles="https://github.com/nanomsg/nanomsg/archive/$version.tar.gz>nanomsg-${version}.tar.gz"
 checksum="24afdeb71b2e362e8a003a7ecc906e1b84fd9f56ce15ec567481d1bb33132cc7"
 
-post_build() {
-	ctest
-}
-
 post_install() {
 	vlicense COPYING
 }
--- a/srcpkgs/nanomsg/template
+++ b/srcpkgs/nanomsg/template
@@ -5,30 +5,17 @@ revision=1
 short_desc='Simple, high-performance implementation of "scalability protocols"'
 homepage='http://nanomsg.org/'
 license="MIT"
-depends="glibc"
-makedepends="cmake"
-build_style="cmake"
+hostmakedepends="cmake"
+build_style=cmake
 shlib_provides="lib${pkgname}.so.5.0.0"
-nocross="yes"
+nocross="https://travis-ci.org/voidlinux/void-packages/jobs/238104905"
 distfiles="https://github.com/nanomsg/nanomsg/archive/$version.tar.gz>nanomsg-${version}.tar.gz"
 checksum="24afdeb71b2e362e8a003a7ecc906e1b84fd9f56ce15ec567481d1bb33132cc7"
 
-pre_configure() {
-	configure_args="-DCMAKE_INSTALL_PREFIX=/usr"
-}
+configure_args="-DCMAKE_INSTALL_PREFIX=/usr"
 
-do_build() {
-	cd $wrksrc/build
-	cmake --build .
-}
 post_build() {
-	cd $wrksrc/build
 	ctest
-}
-
-do_install() {
-	cd $wrksrc/build
-	DESTDIR=${DESTDIR} cmake --build . --target install
 }
 
 post_install() {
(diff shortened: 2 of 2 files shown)


srcpkgs/nanomsg/template
```diff
@@ -0,0 +1,36 @@
+# Template file for 'nanomsg'
+pkgname=nanomsg
+version=1.0.0
+revision=1
+short_desc='Simple, high-performance implementation of "scalability protocols"'
+homepage='http://nanomsg.org/'
+license="MIT"
+depends="glibc"
+makedepends="cmake"
```
Vaelatern: ew no no no no no no no

srcpkgs/nanomsg/template
```diff
@@ -0,0 +1,36 @@
+# Template file for 'nanomsg'
+pkgname=nanomsg
+version=1.0.0
+revision=1
+short_desc='Simple, high-performance implementation of "scalability protocols"'
+homepage='http://nanomsg.org/'
+license="MIT"
+depends="glibc"
```
Vaelatern: also ew no no.

srcpkgs/nanomsg/template
```diff
@@ -0,0 +1,36 @@
...
+makedepends="cmake"
+build_style="cmake"
+shlib_provides="lib${pkgname}.so.5.0.0"
```
Vaelatern: and probably not necessary.

srcpkgs/nanomsg/template
```diff
@@ -0,0 +1,36 @@
...
+pre_configure() {
+	configure_args="-DCMAKE_INSTALL_PREFIX=/usr"
+}
+
+do_build() {
+	cd $wrksrc/build
+	cmake --build .
+}
+post_build() {
+	cd $wrksrc/build
```
Vaelatern: no no

srcpkgs/nanomsg/template
```diff
@@ -0,0 +1,36 @@
...
+checksum="24afdeb71b2e362e8a003a7ecc906e1b84fd9f56ce15ec567481d1bb33132cc7"
+
+pre_configure() {
+	configure_args="-DCMAKE_INSTALL_PREFIX=/usr"
+}
+
+do_build() {
+	cd $wrksrc/build
+	cmake --build .
+}
```
Vaelatern: please remove.

srcpkgs/nanomsg/template
```diff
@@ -0,0 +1,36 @@
...
+makedepends="cmake"
+build_style="cmake"
+shlib_provides="lib${pkgname}.so.5.0.0"
+nocross="yes"
+distfiles="https://github.com/nanomsg/nanomsg/archive/$version.tar.gz>nanomsg-${version}.tar.gz"
+checksum="24afdeb71b2e362e8a003a7ecc906e1b84fd9f56ce15ec567481d1bb33132cc7"
+
+pre_configure() {
+	configure_args="-DCMAKE_INSTALL_PREFIX=/usr"
+}
```
Vaelatern: can be set in the main body, not in `pre_configure`

srcpkgs/nanomsg/template
```diff
@@ -0,0 +1,36 @@
...
+	cd $wrksrc/build
+	cmake --build .
+}
+post_build() {
+	cd $wrksrc/build
+	ctest
+}
+
+do_install() {
+	cd $wrksrc/build
```
Vaelatern: ew no.

srcpkgs/nanomsg/template
```diff
@@ -0,0 +1,36 @@
...
+	DESTDIR=${DESTDIR} cmake --build . --target install
+}
```
Vaelatern: Heck remove the whole function.

srcpkgs/nanomsg/template
```diff
@@ -0,0 +1,36 @@
...
+version=1.0.0
+revision=1
+short_desc='Simple, high-performance implementation of "scalability protocols"'
+homepage='http://nanomsg.org/'
+license="MIT"
+depends="glibc"
+makedepends="cmake"
+build_style="cmake"
+shlib_provides="lib${pkgname}.so.5.0.0"
+nocross="yes"
```
Vaelatern: Please offer a link to a failing cross build.

srcpkgs/nanomsg/template
Vaelatern: `build_style=cmake`

srcpkgs/nanomsg/template
```diff
@@ -0,0 +1,36 @@
...
+	ctest
+}
+
+do_install() {
+	cd $wrksrc/build
+	DESTDIR=${DESTDIR} cmake --build . --target install
+}
+
+post_install() {
+	vlicense COPYING
```
Vaelatern: good

srcpkgs/nanomsg/template
```diff
@@ -0,0 +1,23 @@
...
+pkgname=nanomsg
+version=1.0.0
+revision=1
+short_desc='Simple, high-performance implementation of "scalability protocols"'
+homepage='http://nanomsg.org/'
+license="MIT"
+hostmakedepends="cmake"
+build_style=cmake
+shlib_provides="lib${pkgname}.so.5.0.0"
+nocross="https://travis-ci.org/voidlinux/void-packages/jobs/238104905"
```
Vaelatern: Seems like the build is fine, and the binary probably is too. You just can't run the tests in a cross environment. This is not enough of a reason to be `nocross`.

srcpkgs/nanomsg/template
```diff
@@ -0,0 +1,23 @@
...
+distfiles="https://github.com/nanomsg/nanomsg/archive/$version.tar.gz>nanomsg-${version}.tar.gz"
+checksum="24afdeb71b2e362e8a003a7ecc906e1b84fd9f56ce15ec567481d1bb33132cc7"
+
+configure_args="-DCMAKE_INSTALL_PREFIX=/usr"
```
Vaelatern: Not necessary at all.
Line 26 of common/build-style/cmake.sh
`cmake_args+=" -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_BUILD_TYPE=Release"`

srcpkgs/nanomsg/template
```diff
@@ -0,0 +1,23 @@
+# Template file for 'nanomsg'
+pkgname=nanomsg
+version=1.0.0
+revision=1
+short_desc='Simple, high-performance implementation of "scalability protocols"'
+homepage='http://nanomsg.org/'
+license="MIT"
+hostmakedepends="cmake"
```
Vaelatern: Not necessary at all.

Vaelatern: since the build_style is now `cmake`

srcpkgs/nanomsg/template
Vaelatern: Hm, still think it's not necessary, but forgot about `common/shlibs` changes, which are necessary for this package.

srcpkgs/nanomsg/template
The-Fred: You sure? I ran into a problem when I tried to build this in a new environment. At some point there was a failure because there was no cmake and adding the hostmakedepends fixed it.

srcpkgs/nanomsg/template
The-Fred: Good to know and good to learn the right way to do things. I see a lot of pkgs using that and of course I look at how other packages do things when I'm not sure.  I have to remind myself to start the template off simple and only add things as needed.

srcpkgs/nanomsg/template
The-Fred: Conclusion is?

- Remove nocross and don't run tests?
- Remove nocross and run tests but ignore failures?
- Something else?

common/shlibs
```diff
@@ -2768,3 +2768,4 @@ libgcab-1.0.so.0 gcab-0.7_1
 libappstream-glib.so.8 appstream-glib-0.6.13_1
 libappstream-builder.so.8 appstream-glib-0.6.13_1
 libflatpak.so.0 flatpak-0.9.3_2
+libnanomsg.so.5.0.0 nanomsg-1.0.0_1
```
Vaelatern: `libnanomsg.so.5` not `libnanomsg.so.5.0.0`

srcpkgs/nanomsg/template
```diff
@@ -0,0 +1,14 @@
+# Template file for 'nanomsg'
+pkgname=nanomsg
+version=1.0.0
+revision=1
+short_desc='Simple, high-performance implementation of "scalability protocols"'
```
Vaelatern: Why do you double-quote "scalability protocols"? Are they only so-called and are not actually?

The-Fred: Due to proper attribution and respect to the original author's intent.

Vaelatern:
I squashed commits and added a `-devel` package, but did not bother to fix your quotes to match the general style.

common/shlibs
ebfe: Why? The soname actually is `libnanomsg.so.5.0.0`. That's what other binaries link against and that's what's in the `shlib_provides`.

PR #6576, opened 2017-06-02
neomutt: update to 20170602

freshprince:


Vaelatern:
Hm. I don't like the changes in dependencies when the changelog doesn't support a change in how documentation is built.

freshprince:
They just fixed the build process so that it doesn't fail anymore when xsltproc and w3m isn't installed: https://github.com/neomutt/neomutt/commit/0d8ca5bd48cdfeb0d8e67bd3aa9d48aac9c4cc64

Here they silently added those dependencies: https://github.com/neomutt/neomutt/commit/dd499e12066bcacf110b9fa2aaf1b13fb90294e4

We could skip this release.

Gottox:
@freshprince do you expect neomutt will release a fix soon? otherwise we can cherrypick those patches too.

freshprince:
Cherrypicking sounds like the way to go. I have no idea what the release scedule of neomutt is.

I've updated my PR.

PR #6579, opened 2017-06-03
New package: xrick-021212

the-maldridge:
The source is stated to be released under GPL, but the data assets aren't technically released under anything and its not entirely clear who owned them in the first place.  This package builds on all architectures.

Its worth noting this is restricted, so I'm not really sure why Travis built it...


Vaelatern:
Because travis builds everything, makes testing things easier and we never use the product of Travis builds anyway.

PR #6581, opened 2017-06-03
radicale: update to 1.1.3.

lemmi:


lemmi:
So.. radicale 2 is out. It is not compatible with the 1.x.x series.

https://github.com/Kozea/Radicale/releases
http://radicale.org/1to2/

I would like to support version 2, but keep version 1 alive for transition. 
So either 

1. I move the current radicale package into radicale1 and name the 2.x.x series radicale, or
2. Keep the current package and package radicale2.

Thoughts?

Gottox:
I favor 2.

Vaesper:
Either is fine, I just want an option to install radicale2.

PR #6595, opened 2017-06-04
mergerfs: update to 2.22.0

cr6git:


cr6git:
As always, aarch64 & armv7hf are annoying.

PR #6601, opened 2017-06-04
Samba4 with cross compiling support

Gottox:


srcpkgs/samba4/template
```diff
@@ -0,0 +1,98 @@
+# Template file for 'samba'
+pkgname=samba4
+version=4.6.7
+revision=1
+build_style=waf
```
yopito: what is this build style ?

Gottox: [A world of pain](https://waf.io/)

yopito: my bad: I was not aware this build style was already defined.

PR #6604, opened 2017-06-05
mpd: update to 0.20.9

cr6git:


cr6git:
Thanks @Gottox 

I just noticed _a detail_ in the template: 

`homepage="http://www.musicpd.org/"`
`distfiles="http://www.musicpd.org/download/mpd/${version%.*}/mpd-${version}.tar.xz"`

Could be _reduced_ to: 

`homepage="http://www.musicpd.org"`
`distfiles="${homepage}/download/${pkgname}/${version%.*}/${pkgname}-${version}.tar.xz"`

PR #6616, opened 2017-06-06
R: adopt package; add R-cran build style

wagnerflo:
As discussed with @chneukirchen: Adopting the R package. Also adding a new build style for CRAN packages to use for a bunch of templates I will be commiting and maintaining.


chneukirchen:
Maintainer change will require a rebuild to show up in binary packages (you can skip that one tho).

I don't think we should start using arrays in templates, and I'd rather not start adding magic `cran*` variables...

wagnerflo:
I can drop `cranname` and instead just do `${pkgname#R-cran-}` everywhere and `crandepends` is just convenience, of course.

But `cranversion` serves a purpose I'm not sure how to handle otherwise: CRAN package names can contain a dash-and-number component at the end. So they conflict with xbps revisions.

We can, of course, put the burden on the package maintainer to replace dash with dot for the `version` variable and also to remember to not replace in the distfiles URL.

I'd prefer to automated that in some way, though...

chneukirchen:
You can use other chars in the version variable, e.g. `2.2.2r2` and then translate r to . in the build_style.

wagnerflo:
I've amended the second commit. No more special variables. Please have a look!

chneukirchen:
I think `wrksrc` can be set relative. Else :+1:

(They don't put version numbers into the tarball dirs?)

wagnerflo:
In a template, maybe. But in an build-style environment file relative paths for `wrksrc` don't work. That is because [prepending `$XBPS_BUILDDIR`](https://github.com/voidlinux/void-packages/blob/272114cd05e19542b0d88448d31cb19206fd1a67/common/xbps-src/shutils/common.sh#L491) is done before [sourcing the environment file](https://github.com/voidlinux/void-packages/blob/272114cd05e19542b0d88448d31cb19206fd1a67/common/xbps-src/shutils/common.sh#L516).

(Yes. No version numbers in the directories within the CRAN tarballs.)

wagnerflo:
For clarification: Should I bump the revision on R to make the maintainer change show up in the binary package or will you merge the request like this?

PR #6617, opened 2017-06-06
New package: minetest-game-0.4.16

jcjordyn130:


Vaelatern:
Please look before you package. That said, your template should have looked more like the pre-existing template.

Vaelatern:
I should read more carefully. But is there any reason for this to be a package?

jcjordyn130:
@Vaelatern The only thing included with minetest by default is the minimum development test, and iirc minetest only reads games from it's prefix.

Vaelatern:
Then this PR should actually be a minetest package that also downloads minetest_game in the distfiles, and installs minetest_game in a `post_install`

jcjordyn130:
Oh yeah, I thought about that.

jcjordyn130:
Closing issue as this commit should be in minetest and not a new package.

PR #6618, opened 2017-06-07
linux4.1: update to 4.1.40

cr6git:
Seems to be a _significant_ update. I use it since yesterday.

### DIFF
--- a/srcpkgs/linux4.1/template
+++ b/srcpkgs/linux4.1/template
@@ -1,6 +1,6 @@
 # Template file for 'linux4.1'
 pkgname=linux4.1
-version=4.1.39
+version=4.1.40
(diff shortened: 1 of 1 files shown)


cr6git:
Just noticed it, sorry.. 
This [lines](https://github.com/voidlinux/void-packages/pull/6618/files#diff-6f7975289c7413aa104469cc2b9e8e70R49) should be changed. 
As you can see in the [repository](http://rcn-ee.com/deb/sid-armhf/), there is no 4.1.40-armv7-x2 folder! 
The latest is [4.1.38-armv7-x8](http://rcn-ee.com/deb/sid-armhf/v4.1.38-armv7-x8/). :worried:

PR #6619, opened 2017-06-07
qemu: disable sdl by default, use gtk instead

Gottox:
This pull request disables sdl support in qemu and replaces it with gtk3.

### DIFF
--- a/srcpkgs/qemu/template
+++ b/srcpkgs/qemu/template
@@ -1,7 +1,7 @@
 # Template file for 'qemu'
 pkgname=qemu
 version=2.9.0
-revision=2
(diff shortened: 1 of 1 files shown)


xtraeme:
Why?

Gottox:
Primarily because window scaling is broken with sdl{,2} which has become pretty annoying since I started using a 4k.

Secondary because of easier accessibility of common commands in the task bar.

Gottox:
see this for example:
screenshot from 2017-06-07 10-21-12

xtraeme:
Ok, makes sense. Thanks.

PR #6631, opened 2017-06-08
vidcutter: update to 4.0.0

cr6git:
Added `nocross=yes` because `python3-PyQt5` cannot be cross compiled 

https://travis-ci.org/voidlinux/void-packages/jobs/237947223#L620
https://travis-ci.org/voidlinux/void-packages/jobs/237947226#L482
https://travis-ci.org/voidlinux/void-packages/jobs/237947227#L482


asergi:
- distfile checksum changed (again)
- lastest stable release still seems to be [3.5.0](https://github.com/ozmartian/vidcutter/releases), not `4.0.0`

As a side note, there is no need to add `nocross` here, it's duplicate.

cr6git:
@asergi ok I understand and will remove `nocross` ... 

The checksum **changed** (again) because the disfile has changed.

You can see in [releases/tag/4.0.0](https://github.com/ozmartian/vidcutter/releases/tag/4.0.0): "ozmartian tagged this **3 days** ago" 

3 days ago? I made this Pull Request 7 days ago, and the checksum was correct at the time. 
So the tarball has been replaced 3 days ago, this is the only explanation.

PR #6634, opened 2017-06-09
New package: base-container-0.001

Vaelatern:
Based on the `base-system` metapackage, but just for creating little containers on already running Linux systems.


liwakura:
Whats the difference to `base-voidstrap` ?

Vaelatern:
Oh, duh.

Issue #2422, opened 2015-08-31
Template cleanup TODO


pullmoll:
This a list of issues reported by `xlint` when running it over all templates in the repository. These are usually minor and not too important things.

If you are the Mr. Proper type and have time to spare, you could look into this or that and make `xlint` happy.

There are some real issues in there as well.

#### Variable name issues
- [x] srcpkgs/glew19/template:25: conflicts needs depname with version
- [x] srcpkgs/catalyst/template:20: custom variables should use _ prefix: XBPS_FETCH_CMD="wget --referer='http://support.amd.com/en-us/download/desktop?os=Linux+x86' "
- [x] srcpkgs/h2o/template:6: custom variables should use _ prefix: configure_flags="-DUSE_BUNDLED_SSL=OFF"
- [x] srcpkgs/connman/template:5: custom variables should use _ prefix: lib32disaled=yes
- [x] ~~srcpkgs/bind/template:26: custom variables should use _ prefix: named_desc="BIND DNS server"~~

#### Indentation issues
- [x] srcpkgs/AlsaMixer.app/template:14: indent with tabs
- [x] srcpkgs/GraphicsMagick/template:11: indent with tabs
- [ ] srcpkgs/JAI/template:12: indent with tabs
- [ ] srcpkgs/aubio/template:31: indent with tabs
- [ ] srcpkgs/awoken-icons/template:19: indent with tabs
- [ ] srcpkgs/bananapi-uboot/template:24: indent with tabs
- [x] srcpkgs/base-devel/template:10: indent with tabs
- [ ] srcpkgs/bgs/template:15: indent with tabs
- [ ] srcpkgs/brother-dcp197c-cupswrapper/template:22: indent with tabs
- [ ] srcpkgs/catalyst/template:49: indent with tabs
- [ ] srcpkgs/chicken/template:35: indent with tabs
- [ ] srcpkgs/chromaprint/template:17: indent with tabs
- [ ] srcpkgs/cinnamon-menus/template:21: indent with tabs
- [ ] srcpkgs/cppcheck/template:14: indent with tabs
- [ ] srcpkgs/doom3/template:45: indent with tabs
- [ ] srcpkgs/dovecot-plugin-pigeonhole/template:10: indent with tabs
- [ ] srcpkgs/dssi/template:21: indent with tabs
- [x] srcpkgs/fcgi/template:25: indent with tabs
- [ ] srcpkgs/fdupes/template:13: indent with tabs
[...]

PR #6640, opened 2017-06-09
Quotes not needed.

The-Fred:


Vaelatern:
Eh, but not a bug.

Vaelatern:
You made several mistakes with these contributions.
1) It's not something we care about
2) It isn't substantial in any way
3) None of your commit messages were well enough formed for inclusion.
4) Would have just been a commit to confuse the package history.
5) Again, we just don't care. We don't even mention it in package reviews.

The-Fred:
I think any contribution large or small should be welcomed. Having all existing templates conform may not seem substantial but for someone writing new template(s) it is quite useful. For example if I look at another template I take that as an example for the one I'm writing. Having all templates be as close to some consistency as is possible is good not bad.

Duncaen:
We only apply style fixes if it touches the template for something else, e.g. an update.
The repository is already really large and such changes just add to it, they don't improve anything.
Those quotes are really irrelevant, we had a list way more important fixes in https://github.com/voidlinux/void-packages/issues/2422 but even there we came to the conclusion that its not worth it without real changes.

The-Fred:
@Duncaen this helps to understand Void philosophy.  Good to read that there's flexibility.

PR #6641, opened 2017-06-09
Quotes not needed.

The-Fred:


The-Fred:
https://github.com/voidlinux/void-packages/pull/6640#issuecomment-307495414

PR #6646, opened 2017-06-09
sddm: fix display of user avatars

m-wynn:
sddm 0.14.0 introduced an issue with displaying user avatars.  The
simple fix was committed to master, but wasn't released.
https://github.com/sddm/sddm/issues/807

It looks like at least Arch and Gentoo include this patch, as it's rather
simple but provides a major advantage.  It's not clear when sddm
plans to have a new release.


m-wynn:
Should these be one commit?

PR #6649, opened 2017-06-09
electrum: Update to 2.8.3

Sir-Boops:
This also adds two new python packages as electrum now depends on them

### DIFF
--- a/srcpkgs/python-pysocks/template
+++ b/srcpkgs/python-pysocks/template
@@ -1,0 +1,29 @@
+# Template file for 'python-pysocks'
+pkgname=python-pysocks
+version=1.6.7
+revision=1
+wrksrc=PySocks-${version}
+noarch=yes
+build_style=python-module
+depends="python"
+hostmakedepends="python-setuptools python3-setuptools"
+short_desc="A Python2 SOCKS and HTTP proxy wrapper"
+maintainer="Sir_Boops <admin@boops.me>"
+license="Custom"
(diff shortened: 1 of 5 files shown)


Vaelatern:
Python modules look wrong, lacking a version 2 and ver. 3 package. See other python packages to be a general guide.

Sir-Boops:
Added the python3 version of the two new python packages

Vaelatern:
All transitional dummy packages are no longer relevant, since new packages don't need to migrate from old number schemes

Vaelatern:
Descriptions also need to follow the model of other python packages

srcpkgs/python-pysocks/template
```diff
@@ -0,0 +1,36 @@
+# Template file for 'python-pysocks'
+pkgname=python-pysocks
+version=1.6.7
+revision=1
+wrksrc=PySocks-${version}
+noarch=yes
+build_style=python-module
+depends="python"
+hostmakedepends="python-setuptools python3-setuptools"
+short_desc="PySocks lets you send traffic through SOCKS and HTTP proxy servers"
```
Vaelatern: How this would look:
`[-] python-pysocks-1.6.7_1       Pysocks ....`
don't repeat the package name!

Sir-Boops:
Updated the descriptions and removed the python3.4 version of the packages

Gottox:
@Vaelatern okay from you?

PR #6663, opened 2017-06-10
New package: vapoursynth-editor-r16

DirectorX:


lemmi:
Please use a email where we can reach you. 
https://github.com/voidlinux/void-packages/blob/master/Manual.md#mandatory_vars

PR #6669, opened 2017-06-11
python-smmap: update to 2.0.3

lafrenierejm:


srcpkgs/python-smmap/template
```diff
@@ -1,7 +1,7 @@
 # Template file for 'python-smmap'
 pkgname=python-smmap
-version=2.0.1
-revision=3
+version=2.0.3
+revision=4
```
Vaelatern: `revision=1`

PR #6671, opened 2017-06-11
python-GitPython: update to 2.1.5

lafrenierejm:


srcpkgs/python-GitPython/template
```diff
@@ -1,7 +1,7 @@
 # Template file for 'python-GitPython'
 pkgname=python-GitPython
-version=2.1.3
-revision=1
+version=2.1.5
+revision=2
```
Vaelatern: No revision bump if there is a version bump.

PR #6672, opened 2017-06-11
python-gitdb: update to 2.0.2

lafrenierejm:


srcpkgs/python-gitdb/template
```diff
@@ -1,7 +1,7 @@
 # Template file for 'python-gitdb'
 pkgname=python-gitdb
-version=2.0.0
-revision=1
+version=2.0.2
+revision=2
```
Vaelatern: No revbump if bump.

PR #6684, opened 2017-06-11
mednafen: update to 0.9.45.1

cr6git:


cr6git:
One job [failed](https://travis-ci.org/voidlinux/void-packages/jobs/241818078#L1520) on armv7hf-musl, and I don't know why.

PR #6687, opened 2017-06-12
libogre: build against freeimage

masato9000:
Our libogre build doesn't support any image formats other than dds (no png, jpg, etc). For better support it needs to link to devil or freeimage.

I've gone with freeimage here, as that's what the ogre folks choose as default.


masato9000:
probably can't go breaking the package on arm64...
going to have to look into the use of function png_init_filter_functions_neon in freeimage...

PR #6695, opened 2017-06-12
ModemManager: remove duplicate libqmi-devel

The-Fred:


Vaelatern:
This may fall under no real change, again. @Duncaen ?

PR #6696, opened 2017-06-12
New package: sc3-plugins-3.8.0

jarbowski:


jarbowski:
do I have to add nocross=yes ?

PR #6697, opened 2017-06-12
New package: sc3-plugins-3.8.0

jarbowski:
Extension plugins for the SuperCollider3 audio synthesis server.

Sorry for the double pr


Vaelatern:
Stop closing PRs without need.
There was no need.

jarbowski:
I don't see anything obvious, would it be enough to add x86_64-musl=broken?

jarbowski:
@Vaelatern 
since supercollider didn't build neither on musl, what do you recommend?

jarbowski:
@thypon 
since supercollider didn't build neither on musl, what do you recommend?

thypon:
`only_for_archs` should be enough

jarbowski:
@thypon 
ok, thank you

the-maldridge:
@Vaelatern @thypon Are you satisfied with this package?

thypon:
Not until it contains an update file. The custom `do_fetch` breaks the `update-check`.

I already started a review.

jarbowski:
@thypon 
searching a solution to make an update file, I found that the libpurple-telegram package which is similar to this one, doesn't contain any update file.
so, is it necessary and how can I resolve this?

thypon:
https://github.com/voidlinux/void-packages/blob/master/srcpkgs/urbanterror/update something like this should work.

jarbowski:
@thypon 
Is it okay now, can I rebase and propose it?

thypon:
It is ok for me.

thypon:
rebase and I merge

PR #6701, opened 2017-06-12
python-ipython_widgetsnbextension: update to 2.0.0.

### LINKED ISSUE #6700: python{,3}-ipython_widgetsnbextension: version mismatch, please update
When I create a Jupyter notebook (using either Python 2 or Python 3 kernel) and add a cell with the following contents:

```python
from ipywidgets import IntSlider
from IPython.display import display
slider = IntSlider(min=1, max=10)
display(slider)
```
and run the cell, I get a slider but with following message added:
```
The installed widget Javascript is the wrong version. It must satisfy the semver range ~2.1.4.
```
When I create a clean virtualenv and install Jupyter in it:
```sh
$ virtualenv3 venv
$ source venv/bin/activate
$ pip install jupyter
```
and open and run the same example notebook, the slider appears without any extra message.

Looling at the output of `pip freeze` in the virtualenv and comparing it with the output when not in a virtualenv, I see:

| system | virtualenv |
| - | - |
| ipywidgets==6.0.0 | ipywidgets==6.0.0 |
| | jupyter==1.0.0 |
| jupyter-client==5.0.1 | jupyter-client==5.0.1 |
| jupyter-console==5.1.0 | jupyter-console==5.1.0 |
| jupyter-core==4.3.0 | jupyter-core==4.3.0 |
| ... | ... |
| widgetsnbextension==1.2.6 | widgetsnbextension==**2.0.0** |

This seems to indicate that the version of packages `python{,3}-ipython_widgetsnbextension` in Void Linux are too old to use with `python{,3}-ipython_ipywidgets`. Please consider updating them, unless there is a compelling reason to keep them at their present versions.

ackalker:
This fixes #6700.


ackalker:
Sorry for hammering CI with a forced push, I didn't realize that https://travis-ci.org/voidlinux/void-packages/jobs/242453499 was a genuine build check failure and not a builder crash.
(Unlike the text boxes for the other subjobs which are all clickable, for the `XLINT=1` subjob only the subjob number appears to be clickable.)

Anyway, unlike the other Jupyter and IPython packages, although the source code for this package states that it is 3-clause-BSD licensed, it doesn't contain any license files at all for me to use with`vlicense`.  This could be considered an upstream 'bug' of sorts.

Of course I could work around this by retrieving the file `COPYING.md` from from the upstream repository from which the ipywidgets package gets built, but I'm unsure of the legal implications,

ackalker:
Thanks for merging!

PR #6707, opened 2017-06-13
New package: onedrive-d-0.0.0.20170617

ibrokemypie:
Currently there are no tagged releases upstream, though I have requested the developer to make them available. This however seems to work, so leaving this here until the developer gets back about releases.


srcpkgs/onedrive-d/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'onedrive-d'
+pkgname=onedrive-d
+version=4511242e99c1dbf978265bf839cdd6f7a065277a
```
Gottox: As you said, this is WIP, but a general note: Never ever use commithashes for versions. They will break updates. In Void we're using yyyymmdd or 0.0.0.yyyymmdd for versions instead.

srcpkgs/onedrive-d/template
```diff
@@ -0,0 +1,19 @@
...
+revision=1
+build_style=gnu-makefile
+makedepends="git dmd libcurl-devel sqlite-devel"
+nopie=true
```
Gottox: dmd can't build pie binaries?

ibrokemypie: nope, seems dmd itself is also nopie

srcpkgs/onedrive-d/template
```diff
@@ -0,0 +1,20 @@
+# Template file for 'onedrive-d'
+pkgname=onedrive-d
+version=20170528
+_commit="4511242e99c1dbf978265bf839cdd6f7a065277a"
+revision=1
+build_style=gnu-makefile
+makedepends="git dmd libcurl-devel sqlite-devel"
```
cr6git: Move git & dmd to hostmakedepends: 

`hostmakedepends="git dmd"`
`makedepends="libcurl-devel sqlite-devel"`

And because of dmd [restriction](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/dmd/template#L19), add this line: 

`only_for_archs="i686 x86_64"`

Vaelatern:
https://github.com/skilion/onedrive/issues/194

srcpkgs/onedrive-d/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'onedrive-d'
+pkgname=onedrive-d
+version=4511242e99c1dbf978265bf839cdd6f7a065277a
+revision=1
+build_style=gnu-makefile
+makedepends="git dmd libcurl-devel sqlite-devel"
```
Vaelatern: Because of the dmd restriction, @cr6git , you do not need to add that line. Please remove the `only_for_archs` line, @ibrokemypie

srcpkgs/onedrive-d/template
```diff
@@ -0,0 +1,22 @@
...
+short_desc="Free Client for OneDrive on Linux"
+maintainer="ibrokemypie <ibrokemypie@bastardi.net>"
+license="GPL-3"
+homepage="https://github.com/skilion/onedrive"
+
+do_fetch() {
+	git clone ${homepage} ${wrksrc}
+	cd ${wrksrc}
+	git reset --hard ${_commit}
+}
```
Vaelatern: Alternately set the distfile to `https://github.com/skilion/onedrive/archive/${commit}.tar.gz`

srcpkgs/onedrive-d/template
```diff
@@ -0,0 +1,22 @@
+# Template file for 'onedrive-d'
+pkgname=onedrive-d
+version=20170529
```
Vaelatern: Ideally use `0.0.0.20170529` instead, so if the first version is `0.0.1` then we still don't have a problem for updates.

srcpkgs/onedrive-d/template
ibrokemypie: this isn't working for me for some reason...

srcpkgs/onedrive-d/template
ibrokemypie: have done

srcpkgs/onedrive-d/template
ibrokemypie: i think its all in a subdirectory within the archive, so the archive name has no impact on the extracted folder. Have done so anyway, but ad to set wrksrc

srcpkgs/onedrive-d/template
ibrokemypie: I have removed the line

Vaelatern:
I will be making small changes and pushing this, unless there are objections.

PR #6709, opened 2017-06-13
inxi: update to 2.3.20

cr6git:


Vaelatern:
https://github.com/smxi/inxi/commit/b42c4a7f54e99aa006b5103a41d1e98f9a747be2

cr6git:
@Vaelatern Here it's the same problem as with tint2: if I update the inxi template, it will automatically generates a new commit in this PR, and you don't want that!! 

I'm stuck... :disappointed:

cr6git:
Closed by [#6736](https://github.com/voidlinux/void-packages/pull/6736).

PR #6718, opened 2017-06-13
xterm: update to 329

cr6git:


cr6git:
Sorry for the _conflicts_ !

PR #6725, opened 2017-06-14
New package: rsu-client-4.3.5

ibrokemypie:


Vaelatern:
Please see the end of the Manual where it tells you how to format commit msgs, including capitalization and version numbers.

the-maldridge:
This is installing a bunch of stuff into /opt which is in general not permitted by the package manger per both the FHS and our internal package standards.  If possible this needs to be reworked to install within the standard system tree.  If not possible and explanation needs to be provided here and in the template for why non-standard locations must be used for install, and a fragment should be provided for /etc/profile.d to put this package's custom /bin directory into the path.

asergi:
Abandoned by author.

PR #6735, opened 2017-06-15
[donotmerge] New package: linux-steam-integration-0.3

ibrokemypie:
This builds and installs fine, however it doesn't actually seem to run correctly, something which I think is upstream.
Have submitted an issue https://github.com/solus-project/linux-steam-integration/issues/8


asergi:
Abandoned by author.

PR #6748, opened 2017-06-16
Update coursera-dl to 0.9.0

Hoshpak:


the-maldridge:
Besides the xlint error this looks good. Please fix that and this can be merged.

Hoshpak:
What would be the proper way to fix it? This error is about the python-keyring package which I didn't create but just fixed the dependencies on. The source tarball does not contain a license file so there is no file to call vlicense on.

the-maldridge:
Good point, I should have looked closer.  That error is quite tricky as the original author doesn't seem to understand how the MIT license must be used to be binding.  I am not a lawyer but I have read the long-winded arguments of numerous people who condemn the bad example being set by that rather prolific developer and as more of his code catches on, we may need to patch xlint to deal with these issues.  That, or send enough people to annoy him that he'll put the license in the tarball like a normal developer...

srcpkgs/python-SecretStorage/template
```diff
@@ -0,0 +1,30 @@
+# Template file for 'python-SecretStorage'
+pkgname=python-SecretStorage
+version=2.3.1
+revision=1
+wrksrc="SecretStorage-${version}"
+build_style=python-module
+noarch=yes
+pycompile_module="secretstorage"
+hostmakedepends="python-setuptools python3-setuptools"
+depends="python-cryptography python3-cryptography"
```
asergi: `depends` fixed in 0b20605dd42b9fb808b968d9422eee85036e8e24.

PR #6754, opened 2017-06-16
alacritty: update to 0.1.0.20170614

cr6git:


cr6git:
:worried: Travis _does not agree_ ?

PR #6755, opened 2017-06-16
linux-lts: use the newest LTS branch

scadu:


chneukirchen:
4.4 is not EOL yet. `linux-lts` updates conservatively, if you want to jump between LTS, just install the appropriate kernel yourself.

PR #6760, opened 2017-06-17
firefox-esr: update to 52.2.0

asergi:


cr6git:
````
	x86_64*) build_options_default+=" rust";;
esac
````

@asergi Please could you explain why you enabled rust only for x86_64* ? Not available for 32-bit ? Regards.

asergi:
d08658e9105ad476925f29b0d769862e71bfa7cd

cr6git:
ok. and what about @jnbr 's comment in [#13016](https://github.com/voidlinux/void-packages/pull/13016) ?

cr6git:
I found the solution.

PR #6764, opened 2017-06-17
hg-git: update to 0.8.6

cr6gtx:
Modified "short_desc=" to be less than 72 chars.


cr6gtx:
Ah ok I just spotted the trailing whitespace at line 12. 
Sorry, I'm going to fix this.

PR #6769, opened 2017-06-17
New package: yadm-1.10.0

NeoChontrix:


srcpkgs/yadm/template
```diff
@@ -0,0 +1,21 @@
...
+short_desc="Yet Another Dotfiles Manager"
+maintainer="John <johnz@posteo.net>"
+license="GPL-3"
+homepage="https://thelocehiliosan.github.io/yadm/"
+distfiles="https://github.com/TheLocehiliosan/${pkgname}/archive/${version}.tar.gz"
+checksum=3f77b51197fe14d24f670dbc3e5edc4ae329fdd91c03db297ab929e7bc1168bb
+
+do_build() {
+	make yadm.md
+}
```
Vaelatern: Please use a build style.

srcpkgs/yadm/template
```diff
@@ -0,0 +1,21 @@
+# Template file for 'yadm'
+pkgname=yadm
+version=1.10.0
+revision=1
+hostmakedepends="make"
```
Vaelatern: Not necessary.

Vaelatern:
Not yet ready, revisit after a build style is used.

Vaelatern:
@NeoChontrix  please do not close this PR.
Please do not add new commits.
What you want instead is to "squash" commits, or "fixup" your commit.
Then `push -f` to force an update to this PR.

NeoChontrix:
Fixed it and removed the unnecessary make stuff that I only had in for testing.

NeoChontrix:
Or not, what would be the build type for just extracting?

Vaelatern:
OK except...
should this be a package? It's a manpage and shell script. I'm not comfortable saying absolutely no, but I'd like someone else's feedback before merging.

the-maldridge:
We've packaged less useful things of equivalent complexity in the past, though I personally like to not rely on root access to bootstrap my dotfiles (no system packages needed).

PR #6773, opened 2017-06-18
Manual.md: clarify the maintainer field

the-maldridge:
We have enough packages that people have added and abandoned, IMHO we shouldn't accept anymore where people are outright refusing contact information on something they're supposedly maintaining.

@voidlinux/pkg-committers What do you think?


lemmi:
It won't solve all the problems with people not being reachable, but it's a start.

lemmi:
If we do this, we maybe should open an issue mentioning everybody with packages that use such an address and ask them to update their packages. That way they at least get a notification over their registered github email.

Vaelatern:
We already basically assume that pkg commiters who are not in IRC are not reachable. How many times would we benefit from always being able to email contributors?

the-maldridge:
I think the statement that's being made is to let the people know that they are not using a valid address and this is no officially in violation of a package management policy.  Perhaps a deadline would be set so that if the individuals in question did not respond in a timely manner their packages would be changed over to "Orphaned <nobody@voidlinux.eu>".

chneukirchen:
I could live with github noreply addresses, but there doesn't seem a way to send any messages to a github user (except by at-mentioning them).

the-maldridge:
I'm coming at this from a technical as well as philisophical angle.  I look at it that if you aren't willing to put a real email onto the package, are you actually maintaining it?  If the github emails actually went somewhere, like an aliased forwarder, then that would be fine.  As is those aren't even valid addresses, its just a format of a URI that github has seeded out (there isn't an MX record backing any of those subdomains, nor an A record).

chneukirchen:
They are CNAMES to an A record. (They really should use a null MX there...)

Gottox:
ok with me.

Gottox:
And maybe we should contact the maintainers once a package isn't updated for weeks/months.

PR #6777, opened 2017-06-18
New package: masscan-1.0.4.

scadu:


scadu:
Need to clean up this a little.

srcpkgs/masscan/template
```diff
@@ -0,0 +1,13 @@
+# Template file for 'masscan'
+pkgname=masscan
+version=1.0.4
+revision=1
+build_style=gnu-makefile
+patch_args='-p1'
```
Vaelatern: Since this is your patch, please do not set `patch_args` but instead patch as follows:
```
--- src/pixie-backtrace.c
+++ src/pixie-backtrace.c 
```

scadu: The patch is actually backported from https://github.com/robertdavidgraham/masscan/commit/8e58f7caf78c1d3e8e042dae72d4017cf5a1af86 using git format-patch. Unfortunately the fix has been merged few days after the release.

Vaelatern: Great! This is good to know.
I will be merging this soon.

PR #6783, opened 2017-06-18
ceph: update to 11.2.0

the-maldridge:


srcpkgs/ceph/template
```diff
@@ -1,49 +1,37 @@
 # Template file for 'ceph'
 pkgname=ceph
-version=10.2.3
+version=11.2.0
 revision=3
```
Vaelatern: `revision=1`

Vaelatern:
Waiting on change to `revision=1`, appropiate shlibs changes, and ideally
```
    *-musl) broken="Needs a lot of include patching at minimum";;
    i686*) broken="doesn't properly detect i686 build";;
esac
nocross="This will take much more engineering"
```

PR #6784, opened 2017-06-19
mcabber: update to 1.1.0

damadmai:
distfiles URL changed to http for now because they use a StartCom certificate which is no longer trusted


damadmai:
Did it like #6753 
Verified checksum with https://mcabber.com/files/mcabber-1.1.0.tar.bz2.sha256 manually

PR #6786, opened 2017-06-19
xbps-src: Make update-check return a meaningful exit status.

wagnerflo:
I was thinking that having the `./xbps-src update-check` command return a non-zero exit status when if finds new packages, would be useful for scripting.


Duncaen:
This would break current scripts like https://github.com/voidlinux/void-updates, add the awk into your script if you need a check, the exit status is for errors.

Issue #6789, opened 2017-06-19
kde distfiles broken


ebfe:
frameworks/5.29.0 no longer available from upstream

Hoshpak:
also see #6514 and #6526

yopito:
KDE Framewok update to 5.35 (PR #6052) will fix it for few months once get merged.

ebfe:
#6052 merged

PR #6790, opened 2017-06-19
libgit2: update to 0.26.0. (soname bump)

ebfe:
ktexteditor (#6789), libgit2-glib broken


Vaelatern:
`libgit2-glib` broke last time too, iirc

ebfe:
It always brakes when there are API changes.

chneukirchen:
IIRC we decided nothing uses libgit2-glib and it can be removed.

Else, good to merge.

ebfe:
> IIRC we decided nothing uses libgit2-glib and it can be removed.

Unfortunately no longer true:
```
$ xrevshlib libgit2-glib
gitg
gnome-builder
```

ebfe:
ktexteditor, libgit2-glib fixed. cargo broken...

chneukirchen:
Merged manually.

PR #6793, opened 2017-06-20
New package: trickle-0.0.0.20150113

ibrokemypie:
ugly.
no idea how, but every other distro has this. There is a stable, non-git tarball, but it is too outdated to build afaict. Any suggestions appreciated, but so far this builds and runs.
Also, there dont seem to be any alternatives... (limiting bandwidth of individual processes)

### DIFF
--- a/srcpkgs/trickle/template
+++ b/srcpkgs/trickle/template
@@ -1,0 +1,30 @@
+# Template file for 'trickle'
+pkgname=trickle
+version=0.0.0.20150113
+revision=1
+_commit="66551ad94ad3d8af83e1e4422804676ac8762f47"
+# https://travis-ci.org/voidlinux/void-packages/builds/244963468
+only_for_archs="i686 x86_64"
+build_style=gnu-configure
+makedepends="autoconf automake libtool libevent-devel"
+short_desc="Userland bandwidth shaper for Unix-like systems"
+maintainer="ibrokemypie <ibrokemypie@bastardi.net>"
+license="BSD-3"
+homepage="https://github.com/mariusae/trickle"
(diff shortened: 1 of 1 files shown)


srcpkgs/trickle/template
```diff
@@ -0,0 +1,30 @@
+# Template file for 'trickle'
+pkgname=trickle
+version=0.0.0.20150113
+revision=1
+_commit="66551ad94ad3d8af83e1e4422804676ac8762f47"
+# https://travis-ci.org/voidlinux/void-packages/builds/244963468
+only_for_archs="i686 x86_64"
```
Vaelatern: Why?

ibrokemypie: why?

Gottox: only_for_archs is meant for packages that are not able to run on certian platforms (like closed source apps). It is not meant for packages that are able to run but haven't been ported yet.

Either mark those platforms as broken or - preferably - fix them.

srcpkgs/trickle/template
```diff
@@ -0,0 +1,30 @@
...
+build_style=gnu-configure
+makedepends="autoconf automake libtool libevent-devel"
```
Vaelatern: Read the section on `hostmakedepends` vs. `makedepends`.

srcpkgs/trickle/template
Vaelatern: This line is bogus. Fix your dependencies.

the-maldridge:
Rather than screwing up the versioning on this package, why not ask upstream to make a new release.

asergi:
Abandoned by author.

PR #6804, opened 2017-06-22
New package: goaccess-1.2

BKer:


the-maldridge:
When I first looked at this I would have sworn it was written in Go.

srcpkgs/goaccess/template
```diff
@@ -0,0 +1,18 @@
+# Template file for 'goaccess'
+pkgname=goaccess
+version=1.2
+revision=1
+build_style=gnu-configure
+nocross="Uses built artifacts to continue build https://travis-ci.org/voidlinux/void-packages/jobs/245691122"
+configure_args="--enable-utf8 --enable-geoip=legacy"
```
the-maldridge: try also specifying --HOST and --BUILD, this should fix your nocross issue as well.

PR #6806, opened 2017-06-22
New package: NXEngine-evo-2.4.1

ibrokemypie:
Need some input on this one.
Since the thing being packaged is technically a standalone project, I dont think it's worth even considering packaging the game's assets along with it, especially  with the legal fuzziness, however alone this package is useless, so I am unsure whether it would be suitable to have an INSTALL script download and set up the game's files (which are freeware) for the user on install, or simply leave instructions in the INSTALL.msg.
Any feedback would be great.
This is not yet ready to merge.

### DIFF
--- a/srcpkgs/NXEngine-evo/INSTALL.msg
+++ b/srcpkgs/NXEngine-evo/INSTALL.msg
@@ -1,0 +1,4 @@
+NXEngine-evo is simply a re-implementation of the Cave Story engine, as such,
+none of the game's asets or data is included. To utilise NXEngine-evo you
+must download the game's data and extract it all into /usr/share/NXEngine-evo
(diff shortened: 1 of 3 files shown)


the-maldridge:
So if you can't package the assets into a -data package, where exactly can they be legally downloaded as "freeware" from?

ibrokemypie:
https://github.com/isage/nxengine-evo/wiki/Running
Download from http://www.cavestory.org/downloads/cavestoryen.zip

the-maldridge:
I'm not really happy with this as a package that requires content that doesn't seem to be clearly stated as distributable.  In the absence of a license, there is no right given of any kind for any purpose.  If such a time comes as a license if provided with the cavestoryen.zip, please feel free to revisit and reopen this PR.

PR #6807, opened 2017-06-22
python-xlib: update to 0.19

cr6git:
Python Xlib is now maintained at [Github](https://github.com/python-xlib/python-xlib), where Python 3 support is being added and the code has been relicensed to LGPL, see the [old homepage](http://python-xlib.sourceforge.net/).


the-maldridge:
If support is being added, can this be built as a dual mode module?

cr6git:
You mean it would be better to have `python-xlib` and `python3-xlib` in one template? 
As it is the case for [python-xdg](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/python-xdg/template), for example?

the-maldridge:
Go ahead and drop the 3.4 subpackage.  Any time you see 3.4 stuff on a python package, that's a leftover from when we did the migration from versioned python to python that keys on the major version of the series.

the-maldridge:
And this can be just one commit with the update message, there's no need to have two commits for this.

cr6git:
My attempt is a complete failure, sorry. 😞 
And I don't how to have only one commit on github.

the-maldridge:
I can make the update locally, I don't think GitHub has advanced enough tools to do this online.

the-maldridge:
resolved in 6c3a0b674546b2cbb22077c8d358ea4970095098

asergi:
> ** Compatibility
> 
> Support for Python versions older than 2.7 has been dropped. Support for
> Python 3 (3.3, 3.4 and 3.5) has been added. Note that Python-Xlib now
> depends on the six package (>=1.10) for combined Python 2 / 3 support.

Fixed in 2114d19b3865d8024a9f2b4363713c7662a579da.

PR #6809, opened 2017-06-22
ncurses: add more compat symlinks

ackalker:
Add `libncurses.so.5`, `libtinfo.so.5` as symlinks to `libncurses.so.${version}`.

These are needed to use e.g. the prebuilt Clang toolchain included in Android AOSP repo.

Proof:
```sh
$ readelf -d  prebuilts/clang/host/linux-x86/clang-2690385/bin/clang++ | grep NEEDED
 0x0000000000000001 (NEEDED)             Shared library: [libc++.so]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libpthread.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [libdl.so.2]
 0x0000000000000001 (NEEDED)             Shared library: [libncurses.so.5]
 0x0000000000000001 (NEEDED)             Shared library: [libtinfo.so.5]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
```
After updating `ncurses` packages with this fix:
```sh
$ ldd prebuilts/clang/host/linux-x86/clang-2690385/bin/clang++
	linux-vdso.so.1 (0x00007ffeebb30000)
	libc++.so => /home/miki/src/android-x86/prebuilts/clang/host/linux-x86/clang-2690385/bin/../lib64/libc++.so (0x00007fe4c2543000)
	libm.so.6 => /usr/lib/libm.so.6 (0x00007fe4c2113000)
	libpthread.so.0 => /usr/lib/libpthread.so.0 (0x00007fe4c1ef5000)
	libdl.so.2 => /usr/lib/libdl.so.2 (0x00007fe4c1cf1000)
	libncurses.so.5 => /usr/lib/libncurses.so.5 (0x00007fe4c1a97000)
	libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x00007fe4c1880000)
	libc.so.6 => /usr/lib/libc.so.6 (0x00007fe4c14dc000)
	librt.so.1 => /usr/lib/librt.so.1 (0x00007fe4c12d4000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fe4c2426000)
```


the-maldridge:
Since the soname was bumped that signifies a breaking change, we do not link known breaking builds together.  This seems like an excellent patch for you to make locally, but not one that will be accepted into void-packages.

PR #6835, opened 2017-06-24
pmount f2fs support added.

dziq:
Patch for mounting f2fs formated drives.


dziq:
Fix in another pull request

PR #6837, opened 2017-06-24
New package: qt5ct-0.33

xYuusha:


ibrokemypie:
The file in the install message seems to get overwritten upon starting x...

xYuusha:
Are you talking about the ~/.profile or the /etc/X11/Xsession.d/100-qt5ct?

I have never heard of a display manager/starting x from terminal that deletes any of those, and i have been using the same .profile for years.

Could you tell me what dm are you using or how are you startig X so i can test it?

ibrokemypie:
/etc/X11/Xsession.d/100-qt5ct and running startx.
I was creating that while x was already running, and then restarting x l, if that is of significance.

xYuusha:
I'm sorry but i cannot reproduce that behaviour.
When creating the Xsession file and running startx, the file is not overwritten but it's not loaded. I think is because when loading startx, it loads the standard xinitrc or a custom .xinitrc in the home directory, and none of both usually runs Xsession scripts. In that use case, i think the file that should be modified is the .profile or even .xprofile or .bash_profile, depending on how are you launching qt programs.
I'm using sddm and the Xsession scripts are automatically loaded, so maybe that directory is loaded by some desktop managers.

the-maldridge:
@xYuusha does this need to be added to the X profile or just the profile in general?  You could add a file to /etc/profile.d/ to deal with certain files not being read.

xYuusha:
The official documentation advices to add the environment variable to .profile or use the /etc/X11/xsession.d directory but the only important thing is to have the environment variable to exist in the context that is launching the QT programs.

The export can be in the Xprofile or the .pam_environment (which i think most DM source) or even in bash_profile or it can also be exported just before launching a Qt program if you are launching them from the terminal. This is also useful if you just want a single program to be affected by qt5ct instead of every program.

Using /etc/profile.d/ works completely fine using a DM or using startx so i think is a good global solution.

the-maldridge:
Sounds good, mention me once you've made that change.

xYuusha:
@the-maldridge Just updated the INSTALL.msg. I hope is easier to understand now and gives a better solution than before.

the-maldridge:
@xYuusha I see what you changed, but I was more expecting you to add a file to this package that would be installed in /etc/profile.d/ and your install message should be similar to that of openjdk which explains you need to re-source the profile.

xYuusha:
@the-maldridge Sorry, i misunderstood the statement.
In the install process the file is added to /etc/profile.d and it's working correctly.

PeridotGemTech:
How come this hasn't been merged?

the-maldridge:
@PeridotGemTech because life is a hectic and complicated thing.

Issue #6798, opened 2017-06-20
Package request: goldendict


AcarBurak:
The best dictionary application (by a wide margin) and indispensable for translators and non-native speakers.

The stable release may be dated, but it is actively developed and I'd been using the git qt5 version from the official Arch repo for quite some time without any issue; it is stable.

https://github.com/goldendict/goldendict

asergi:
cba8baa31881336ec548c72086c80d2327ecdb09

AcarBurak:
Thank you.

PR #6839, opened 2017-06-24
New Packages: Goldendict-1.5.0-RC2 and ed

NeoChontrix:
https://github.com/voidlinux/void-packages/issues/6798


the-maldridge:
In general we don't permit RC releases as packages.  Is there a specific reason you have used RC2 instead of the point release?

NeoChontrix:
Mainly due to the fact that the last release is quite old (nearly 7years) and rc2 is 1y old.


_Sent from my ZUK Z2 Plus using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

srcpkgs/goldendict/template
```diff
@@ -0,0 +1,26 @@
...
+hostmakedepends="qt5-qmake pkg-config qt5-tools"
+makedepends="tiff-devel qt5-devel libvorbis-devel zlib-devel\
+	libXtst-devel hunspell-devel qt5-tools-devel qt5-declarative-devel\
+	qt5-webkit-devel lzo-devel bzip2-devel libao-devel qt5-svg-devel\
+	qt5-x11extras-devel ffmpeg-devel"
+short_desc="A feature-rich dictionary lookup program"
+maintainer="John <johnz@posteo.net>"
+license="GPL-3"
+homepage="http://goldendict.org/"
+distfiles="https://github.com/goldendict/goldendict/archive/1.5.0-RC2.tar.gz"
```
faerietree: Is there a reason for not using ${version} here? As reminder for removing RC2 later?
Does the RC2 not belong to the version? (following semantic [versioning](semver.org))
```
-version=1.5.0
+version=1.5.0-RC2
```

faerietree:
Long release cycle. Is there a special reason for it? Especially as the author writes [here](http://www.goldendict.org/download.php)

> If you experience problems with the [in 2010] released version, use the latest GIT snapshot

I personally prefer command line like [wordnet + dict, thes wrappers](https://github.com/Herd-Base/wordnet-dict) but `goldendict` certainly fulfills the requirements to get into `void-packages`. :)

srcpkgs/goldendict/template
NeoChontrix: You can't write version=1.5.0-RC2 as version ^^ it has a - in it which isn't allowed.
And the not used $version in distfiles is simply a mistake, I normally wanted to but it in there but I forgot it.

I can set the version to 1.5.0RC2 if you would prefer that.

NeoChontrix:
I can't say why the release cycle is that long, there were around 1.5k commits between the last stable and RC2 and most projects would have done a lot releases out of that.

NeoChontrix:
I just checked the old release and tried to compile it and if I interpret the error messages correctly then it depends on an older version of hunspell and I don't think it would be wise to also package an older version of that or patching the stable version of goldendict for supporting hunspell.

srcpkgs/eb/template
```diff
@@ -0,0 +1,36 @@
...
+pkgname=eb
+version=4.4.3
+revision=1
+build_style=gnu-configure
+hostmakedepends="perl"
+makedepends="zlib-devel"
+short_desc="C library and tools for accessing CD-ROM books"
+maintainer="John <johnz@posteo.net>"
+license="BSD-3"
+homepage="https://github.com/sgqy/libeb-fork"
```
the-maldridge: libeb-fork?

srcpkgs/eb/template
```diff
@@ -0,0 +1,36 @@
...
+distfiles="https://github.com/2ion/eb/archive/${version}.tar.gz"
```
the-maldridge: Why do the distfiles come from a separate github user than the home page?

srcpkgs/eb/template
```diff
@@ -0,0 +1,36 @@
...
+checksum=16274f4d5ec34cc37ed308927390a1d0d09e0e6a2b35fbff7d17aacca0524a37
+nocross=yes
```
the-maldridge: This needs a comment either linking to an upstream issue or a travis build log for why this can't be cross compiled.  If its broken on specific architectures use a case structure and the `broken=` mechanism instead.

srcpkgs/goldendict/template
```diff
@@ -0,0 +1,27 @@
+# Template file for 'goldendict'
+pkgname=goldendict
+version=1.5.0RC2
```
the-maldridge: You've satisfied my concerns with the version.

srcpkgs/goldendict/template
```diff
@@ -0,0 +1,27 @@
...
+revision=1
+_version_upstream="1.5.0-RC2"
```
the-maldridge: It should be possible to generate this from the version by using some bash string manipulation.  Then you wouldn't need this variable and you could just expand this where needed.

srcpkgs/eb/template
NeoChontrix: The official homepage is nonexistent and the fork has some configure stuff fixed. (It worked and I wasn't in the mood to test an 'original' package)

_Sent from my ZUK Z2 Plus using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

srcpkgs/eb/template
NeoChontrix: OK then I may have chosen the wrong homepage, I will check that later

_Sent from my ZUK Z2 Plus using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

srcpkgs/goldendict/template
NeoChontrix: Mkay, so then someone should explain me how to do such stuff, my bash knowledge is rather limited.

_Sent from my ZUK Z2 Plus using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

the-maldridge: You can either google how or leave that as it sits.  Its not a blocking issue.

srcpkgs/eb/template
NeoChontrix: I will recheck cross compilation, I think I may can fix it ( didn't had the time when I created the package to look more closely into it)

The [log](https://travis-ci.org/NeoChontrix/void-packages/jobs/246555733) is just saying that tests can't be run when cross compiling

NeoChontrix:
@the-maldridge Fixed Cross Compilation and the wrong "homepage"

srcpkgs/goldendict/template
```diff
@@ -0,0 +1,21 @@
+# Template file for 'goldendict'
+pkgname=goldendict
+version=1.5.0RC2
+revision=1
+wrksrc="${pkgname}-${version/RC/-RC}"
+build_style=qmake
+hostmakedepends="qt5-qmake pkg-config qt5-tools"
+makedepends="tiff-devel qt5-devel libvorbis-devel zlib-devel\
+	libXtst-devel hunspell-devel qt5-tools-devel qt5-declarative-devel\
```
the-maldridge: One last change: these should be indented by a single space, and you don't need to escape the newlines here since its a quoted string.

NeoChontrix: And done

PR #6841, opened 2017-06-25
New package: Gifsicle-1.88

ColdPhoenix00:


srcpkgs/gifsicle/template
```diff
@@ -0,0 +1,12 @@
+# Template file for 'gifsicle'
+pkgname=gifsicle
+version=1.88
+revision=1
+build_style=gnu-configure
+makedepends="libX11-devel"
+short_desc="Command-line GIF manipulation tool"
+maintainer="ColdPhoenix <coldphoenix@interia.pl>"
+license="GPL-2"
+homepage="https://www.lcdf.org/${pkgname}"
```
the-maldridge: Don't put ${pkgname} into the homepage variable, it doesn't get resolved when the update checker writes out the update notifications and so makes the URL useless.

Other than that this looks good.

ColdPhoenix00:
@the-maldridge Fixed hostname variable, as you requested.

srcpkgs/gifsicle/template
ColdPhoenix00: Fixed hostname variable, as you requested.

the-maldridge:
On further inspection, I think this package should be called 'Gifsicle' but I'm not sure.  @Vaelatern second opinion please.

ColdPhoenix00:
Since @Vaelatern isn't responding, I've changed the package name to Gifsicle.

PR #6842, opened 2017-06-25
Perl-Wx

ibrokemypie:
still working on getting the final package to build. All deps are fine though


pullmoll:
The error output for perl-Wx-0.9932 looks as if it is trying to build C++ code using the C compiler. Perhaps a matter of setting CC=$CXX in the template somewhere?

asergi:
Abandoned by author.

Issue #6489, opened 2017-05-23
phpMyAdmin package is in the repo but not findable with xbps-query


praticamentetilde:
Hi all. I've just built from source `phpMyAdmin` because https://repo.voidlinux.eu does not host a prebuilt package (at least i wasn't able to find it with xbps-install nor xbps-query).

liwakura:
Cant reproduce:
```
$ xbps-query -Rs phpmyadmin
[-] phpMyAdmin-4.6.5.2_1 Web-based MySQL database administration tool
```

In addition, phpMyAdmin is not built, its written in an interpreted language (PHP) and therefore a noarch package.

liwakura:
I just noted that im using musl (where phpMyAdmin is available), whereas the glibc repo does indeed have no phpMyAdmin package.

the-maldridge:
This is a noarch build problem.  I can review this in more detail later this evening.  Its possible that we have more noarch packages that have not been built.  Personally I'm in favor of dropping this package, but we'll see, @stpx do you wish to weigh in on this?

faerietree:
hey praticamentetilde, do you know the command `mysql -u <user> -p`? It's quite handy. Just use like

    mysql -u <user> -p
    # enter  your secret password (stdin)
    USE <database_name>;
    SHOW TABLES;
    DESCRIBE <table_name>;

(No need for uppercase letters. Quote table names et alia if e.g. a reserved keyword or to be more compatible to other database systems than MySQL.)

Apart from that, are you really running a rather minimalistic Void OS on a server and yet require PHPMyAdmin? No offense meant.

praticamentetilde:
@faerietree No, I'm just running Void on a ThinkPad. And even if I've used the shell before, it's just handier to have a web interface to see what's going on instead of a bunch of text based tables that flow over the next line if you have more than 6 or 7 columns.

praticamentetilde:
@liwakura I know that PHP is an interpreted language: I was just referring to the "build" process involved in the creation of the package.

faerietree:
Respected. Note though this wrap around problem can be worked around:
e.g.
* append `\G` to the end of your MySQL command to display a row's columns vertically.
* consider capturing the buffer in a file and using `tail -f /path/to/file | less -S`
* to prevent line wrap in general in `xterm` provide `+aw` option.

The benefit of Void is to be bare bones, simplistic, swift, repairable.

Anachron:
[...]

PR #6844, opened 2017-06-25
remove phpMyAdmin

liwakura:
Currently broken in repo (#6489)
Update PR's were not merged (#6490,#6491)

Since there are several security issues in the current version, remove it.

Edit: Also, the maintainer @stpx seems to be inactive


the-maldridge:
Since an hour later you opened a PR to update it, can you pick one to move forward with and close the other?

faerietree:
I second this decision because (see also #6489)
* One can deploy lightweight alternatives like [`adminer`](https://duckduckgo.com/html?q=adminer) easily as it is a single file.
* One can use e.g. `mysql -u root -p`
* But foremost this package does not fulfill the [requirements of void-packages](https://github.com/voidlinux/void-packages/blob/master/Manual.md#quality-requirements).

liwakura:
@faerietree We removed it, but it seems like you argue in favor of removing it?

PR #6845, opened 2017-06-25
phpMyAdmin: update to 4.7.1

liwakura:
Alternative to #6844


liwakura:
As discussed in irc, phpMyAdmin wont be continued.

PR #6846, opened 2017-06-25
python-mechanize: update to 0.3.5

Sir-Boops:
Also updated the destfiles and homepage links

Version 0.3.X has breaking changes over 0.2.X [Changelog](https://github.com/python-mechanize/mechanize/blob/master/ChangeLog#L26)


asergi:
@the-maldridge @Sir-Boops broken

> * Support HTML 5 (all html is now parsed using html5lib)

asergi:
Also, `calibre` needs to be checked after this.

Sir-Boops:
#6869 fixes the dependency issue as for calibre i tried a few recipes that import mechanize and they seem to work as intended using 0.3.5

the-maldridge:
@asergi good catch.

PR #6849, opened 2017-06-25
avr-gcc: update to 7.1.0

liwakura:


the-maldridge:
before this gets merged, arduino needs to be checked to make sure this doesn't break that package.

liwakura:
broken, https://github.com/voidlinux/void-packages/pull/6861#issuecomment-311193399

PR #6855, opened 2017-06-26
New package: ttf-tlwg-0.6.3

ColdPhoenix00:


the-maldridge:
Your template looks good, can you add the undocumented parameter `font_dirs="/usr/share/fonts/TTF"` so that XBPS will rebuild the font cache when this gets installed?  Once its got that mention me in a comment and I'll merge it.

ColdPhoenix00:
@the-maldridge Just added it.

PR #6859, opened 2017-06-26
tty-clock-git: update to 20170115

cr6git:
Changed `short_desc` to Digital clock (issue [#12](https://github.com/xorg62/tty-clock/issues/12)). 

Changed `license` to GPL-3, as the Debian [package](https://packages.debian.org/sid/tty-clock) ([Copyright](http://metadata.ftp-master.debian.org/changelogs/main/t/tty-clock/tty-clock_2.3-1_copyright)), 
because of: 
`$ xlint srcpkgs/tty-clock-git/template`
`srcpkgs/tty-clock-git/template:9: license GPL without version`


the-maldridge:
It would be very nice for this to not be a -git package anymore but instead to move over to using the tarballs, it looks like this package predates upstream using sane numbers.  This is not a review comment, but is a suggestion I'd like to open for discussion.

cr6git:
`# Template file for 'tty-clock'`
`pkgname=tty-clock`
`version=2.3`
`revision=1`
`makedepends="ncurses-devel"`
`short_desc="Digital clock in ncurses"`
`maintainer='Juan RP <xtraeme@voidlinux.eu>'`
`license="GPL-3"`
`homepage="https://github.com/xorg62/tty-clock"`
`distfiles="${homepage}/archive/v${version}.tar.gz"`
`checksum=343e119858db7d5622a545e15a3bbfde65c107440700b62f9df0926db8f57984`
`do_build() {`
`make CC="$CC" CFLAGS="$CFLAGS" LDFLAGS="$LDFLAGS -lncursesw"`
`}`
`do_install() {`
`vbin tty-clock`
`vman tty-clock.1`
`}`

the-maldridge:
My primary concern is that looking at the release history for tty-clock it isn't strictly increasing.  They went backwards in version numbers for a while.

cr6git:
@the-maldridge I have a question about another package: icewm ([template](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/icewm/template)) is version 1.3.8 dated 2013-11-17. 
On the [homepage](http://www.icewm.org/) they say there is an "Unofficial but maintained branch on [github](https://github.com/bbidulock/icewm)." 
The AUR [package](https://aur.archlinux.org/packages/icewm-git) uses this branch. 
Should we do the same ? Thank you.

the-maldridge:
@cr6git please open a separate issue on that, I think it deserves some discussion.

PR #6867, opened 2017-06-27
intel-ucode: -force- default load -w/o dracut dependency-

thypon:


chneukirchen:
Won't this break the existing dracut hook?
(esp. since hostonly dracut will only add the updates actually needed, resulting in a smaller initrd.)

chneukirchen:
also if we do it in grub already, we should do it for amd too.

chneukirchen:
(I prefer if dracut remains working, since we use it for the live cd and other boot managers too.)

thypon:
ucode is optional as nobody even notices it doesn't work without human intervention.

chneukirchen:
It's not really optional when whole series of CPU are broken when it's not loaded.

thypon:
```
<Gottox> hehe, funny how intel-ucode does nothing on void and nobody noticed :D
<delirein> it does it's job
<delirein> but you have to add a dracut toggle
<Gottox> ah, that piece was missing
<delirein> early_microcode=yes in /etc/dracut.conf.d/intel.conf
<Gottox> delirein: are you sure that works?
<delirein> intel-ucode.conf
<delirein> and then xbps-reconfigure -f linux4.11
<delirein> for me it works
```

chneukirchen:
Your argument does not support breaking it nevertheless.  At least install the single files too, so dracut works as well.

thypon:
Dracut is not working right now. Nonetheless, the fix works for dracut/mkinitcpio even starting bash as init since it loads before even dracut/mkinitcpio kicks in.

An alternative could be patching both mkinitcpio and dracut to force enable intel-ucode when installed.

cr6git:
@thypon this has been explained on the [forum](https://forum.voidlinux.eu/t/solved-intel-ucode/763?u=cr6) for a long time.

thypon:
@cr6git It's not an excuse to have not working as expected.

chneukirchen:
Dracut is working fine when you enable it with `early_microcode=yes`.

chneukirchen:
I guess we could ship that file with intel-ucode, but we can't easily guarantee dracut is rerun i guess.

thypon:
What I'm questioning is wheather or not it should be enabled by default, when installed.

The proposed solution enables it by default on any initrd generator on voidlinux.

thypon:
Yep, bundling the dracut config could be a solution for dracut.

chneukirchen:
:+1: for installing it by default.

cr6git:
I would like to draw your attention to the fact that this microcode update IS NOT automatically loaded when rebooting the machine. 
It is necessary to generate a new initramfs image with dracut to enable the new microcode. 
This point should be indicated in a post-install message, at least.

chneukirchen:
Didn't we have multiple kernel updates since then?

cr6git:
Yes, new initramfs are automatically generated with kernel updates, no problem ;) 
But: when you install the package `intel-ucode` for the first time, you must generate manually a new initramfs, this is not automatic. 
People are not necessarily aware of that...

Issue #6698, opened 2017-06-12
Package request: webkit-ng


SevereOverfl0w:
Links: https://github.com/annulen/webkit https://github.com/annulen/webkit/wiki/Building-QtWebKit-on-Linux

I believe there are some security concerns around qt5-webkit & it is generally abandoned. Using this fork makes Qutebrowser far more usable, on a wider range on applications.

https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/qt5-webkit-ng might we worth referencing.

SevereOverfl0w:
It appears that arch linux has removed the qt5-webkit-ng package, and replaced qt5-webkit with it.

annulen:
Fedora is preparing to do it as well: http://lupinix.blogspot.ru/2017/06/improving-qtwebkit-security-for-fedora.html

SevereOverfl0w:
I have a working build locally, I've been testing it locally for some time with qutebrowser. It seems to work okay, there is a warning due to the duplicated shlibs.

Unfortunately the build takes a while locally, so I've not had much time to fix it.

I just opened the PR to give this some steam :smile:

The-Compiler:
Security wise, the old QtWebKit is worse than WebKitGTK 2.4 - we're talking about ~3 years of delta from upstream WebKit, including all security fixes in that timespan, which are missing from the current QtWebKit package.

There's also a lot of other bugfixes; a lot of websites break or segfault with the legacy QtWebKit package.

PR #6876, opened 2017-06-27
New package: qt5-webkit-ng-5.212.0.alpha2

SevereOverfl0w:
I'm pretty sure this will need review. Key points:

* Manipulation of -alpha2 to .alpha2 & passing xlint
* How I'm working with shlibs is probably wrong.


SevereOverfl0w:
Oh, and this resolves #6698

The-Compiler:
Should this be a separate `-ng` package rather than replacing `qt5-webkit` like Arch/Fedora now do?

ebfe:
Whatever happens: please don't increase the total number of webkit variants in the repo.

SevereOverfl0w:
Seems to have timed out the build. It does take ~2 hours locally. Do I need to speed it up somehow, or bump up the allowed time?

mrnhmath:
This will be shipped with Qt in the next version, so it's better to replace `qt5-webkit`.

mrnhmath:
This will be the next version of `qt5-webkit`, so it's better replacing it.

SevereOverfl0w:
@ebfe Looking to you as authority on this (feel free to defer to someone else). Should I replace  qt5-webkit with this package? I'm in #voidlinux so feel free to catch me there if you want to chat rather than long-form.

This comment seems relevant to the discussion on replacement also: https://github.com/voidlinux/void-packages/issues/6698#issuecomment-313652398

ebfe:
I don't have an opinion on this, except i'd like to avoid having even more webkit variants in the repo.

Since this is a qt5-* package maybe @pullmoll has a more informed opinion.

pullmoll:
`qt5-webkit` is a separate package, not contained in the `qt5-everywhere-opensource…` tarball, so in case qt5-webkit-ng is a *drop-in replacement* I'm all for replacing `qt5-webkit` with this.

SevereOverfl0w:
There are some minor changes: https://github.com/project-kermit/kermit/issues/115 but largely is fully compatible. As seen by arch adopting qt5-webkit-ng as a replacement for qt5-webkit. I will make the change.

SevereOverfl0w:
I've replaced qt5-webkit with this package.

srcpkgs/qt5-webkit/template
```diff
@@ -1,55 +1,30 @@
...
+build_style=cmake
+configure_args="-DPORT=Qt"
+hostmakedepends="bison flex gperf perl python ruby pkg-config cmake"
 makedepends="libwebp-devel libxslt-devel libXcomposite-devel sqlite-devel libjpeg-turbo-devel
  gst-plugins-base1-devel qt5-declarative-devel qt5-location-devel
- qt5-sensors-devel qt5-webchannel-devel"
-short_desc="A cross-platform application and UI framework (Qt5) - WebKit2 component"
-maintainer="Juan RP <xtraeme@voidlinux.eu>"
+ qt5-sensors-devel qt5-webchannel-devel hyphen-devel"
+short_desc="Classes for a WebKit2 based implementation and a new QML API"
```
annulen: That's not an accurate description actually. It should be something like "WebKit component for Qt 5" or "Qt port of WebKit engine"

srcpkgs/qt5-webkit/patches/qt-musl-execinfo.patch
```diff
@@ -1,20 +0,0 @@
---- Source/WTF/wtf/Assertions.cpp	2016-02-04 13:58:47.000000000 +0100
-+++ Source/WTF/wtf/Assertions.cpp	2016-03-17 17:42:38.194682870 +0100
-@@ -61,7 +61,7 @@
- #include <windows.h>
- #endif
- 
--#if (OS(DARWIN) || (OS(LINUX) && !defined(__UCLIBC__))) && !OS(ANDROID)
```
annulen: This patch may help with musl

srcpkgs/qt5-webkit/template
```diff
@@ -1,55 +1,30 @@
...
+maintainer="Dominic Monroe <monroef4@googlemail.com>"
 license="GPL-3, LGPL-2.1"
-homepage="https://www.qt.io/"
```
annulen: That was correct homepage actually

srcpkgs/qt5-webkit/patches/qt-musl-execinfo.patch
SevereOverfl0w: I added this patch. Good call, musl wasn't compiling previously.

pullmoll:
I see there are problems with cross compiling for aarch64 and armv7 (cmake error `Unknown CPU ''`).
AFAIK we currently have qt5-webkit for those cross architectures, so switching should continue to provide the package if possible. Otherwise some packages relying on qt5-webkit-devel won't update for these architectures.

annulen:
Looks like your toolchan file for cmake does not contain CMAKE_SYSTEM_PROCESSOR

SevereOverfl0w:
I've added a commit which attempts to set it. The CI doesn't seem to be engaged right now. How can I test this @pullmoll

pullmoll:
Hmm.. instead of changing the build-style/cmake.sh file you could just add these lines:
```
        configure_args+=" -DCMAKE_SYSTEM_PROCESSOR=${XBPS_TARGET_MACHINE}"
fi
```
which will then be appended to the cmake call in the build style.

But I may be wrong here. cmake is not my strong side.

annulen:
What about other packages that can be missing CMAKE_SYSTEM_PROCESSOR? E.g. webkitgtk is affected in the same way, as we share build system

pullmoll:
AFAICT webkitgtk uses `build_style=gnu-configure` and there the processor is derived from `--target`; or was it `--host`?

I'm not against modifying the build-style, it's just that such a change needs at least some tests to make sure it does not break other packages. I don't know how it could do that, but then I am no cmake expert.

SevereOverfl0w:
My thought was also to ensure the variable is set for all build using
cmake. It may bite them in the future, this seems standard.

annulen:
@pullmoll  https://github.com/voidlinux/void-packages/blob/f6cb0c4741378084b23730a6936f5afc99e6cc1a/srcpkgs/webkit2gtk/template

pullmoll:
Ah, webkit2gtk, not webkitgtk (or webkitgtk2).. we have far too many of these :)

pullmoll:
Ok, so in webkit2gtk there is `configure_args+=" -DCMAKE_SYSTEM_PROCESSOR=${XBPS_TARGET_MACHINE%-musl}"` which means that e.g. `aarch64-musl` becomes `aarch64` in CMAKE_SYSTEM_PROCESSOR.

If the build-style is modified, it should thus also strip the optional -musl part from the target CPU.

ebfe:
```
srcpkgs/librsync/template:6:configure_args="-DCMAKE_SYSTEM_PROCESSOR=$XBPS_TARGET_MACHINE"
srcpkgs/openimageio/template:8: -DCMAKE_SYSTEM_PROCESSOR=${XBPS_TARGET_MACHINE}"
srcpkgs/rr/template:6:configure_args="-DCMAKE_SYSTEM_PROCESSOR=${XBPS_TARGET_MACHINE} -DBUILD_TESTS=OFF"
srcpkgs/webkit2gtk/template:41: configure_args+=" -DCMAKE_SYSTEM_PROCESSOR=${XBPS_TARGET_MACHINE%-musl}"
srcpkgs/x265/template:18:       configure_args+=" -DCMAKE_SYSTEM_PROCESSOR=${CROSS_BUILD%-musl}"
```

Half of these are wrong (missing %-musl) so it might make sense to move this
into the cmake build_style.

annulen:
>it should thus also strip the optional -musl part from the target CPU

At least in case of WebKit, "-musl" suffix won't hurt, and may even get used in future. Though this variable is supposed to have a value of `uname -p` on the target system.

SevereOverfl0w:
Handled the -musl suffix.

SevereOverfl0w:
Should I remove the unrequired uses of CMAKE_SYSTEM_PROCESSOR too?

SevereOverfl0w:
All looks good except for aarch64, which seems to have mysteriously dropped off mid-build https://travis-ci.org/voidlinux/void-packages/jobs/256469467

I'm not sure what to make of it. Someone else seems to have hit it here: https://forum.qt.io/topic/75145/cross-compile-qt-5-7-for-raspberry-pi-3-problem-font-text-words-in-application-not-showing-up-during-running/32 but I don't understand most of the conversation.

SevereOverfl0w:
I've disabled gold for aarch64, and that has worked (great!). Hopefully that line can be removed when aarch64 is supported by the gold linker for whatever it is triggering that bug.

I think this is looking good now.

annulen:
Note that webkit2gtk already disable gold on aarch64. There is no need to reinvent the wheel

https://github.com/voidlinux/void-packages/blob/f6cb0c4741378084b23730a6936f5afc99e6cc1a/srcpkgs/webkit2gtk/template#L28

SevereOverfl0w:
It seems to build okay on aarch64-musl without tweaking the gold linker. But you're right. I'm wondering if I need the arm options too. @asergi contributed them, so I'm guessing you would know how to detect whether it is required for that platform?

I will use the configure args strategy inside of an `if` until then.

pullmoll:
FYI: Testing the PR locally with non-musl architectures. This will take some time.
Our muslrepo.voidlinux.eu server is currently down and building all dependencies will take even more time.

Other than that the PR looks good and can IMO be merged. Perhaps lets wait for the [musl builders](https://build.voidlinux.eu/waterfall) to reappear.

pullmoll:
The aarch64 build failed: [Build log](http://misc.voidlinux.de/qt5-webkit_aarch64_do_build.log)
The aarch64-musl build failed (the same?): [Build log](http://misc.voidlinux.de/qt5-webkit_aarch64-musl_do_build.log)
Now testing armv7hf and armv7hf-musl

pullmoll:
Note: The `common/build-style/cmake.sh` change breaks cross compiling `x265` for `armv7hf{,-musl}`.

pullmoll:
I can't seem to find a fix for the cross build problems. Tried to add `${makedepends}` to `hostmakedepends` for cross building, tried to explicitly add `MesaLib-devel` to makedepends and/or hostmakedepends, tried to use `CXXFLAGS="-std=c++11"` as in the previous qt5-webkit template... all to no avail.

More ideas are welcome.

SevereOverfl0w:
@annulen Does the build error above seem like a break on our side, or your side?

SevereOverfl0w:
Regarding the x265 break, it seems that this is due to the different values of the variables:

CROSS_BUILD | XBPS_TARGET_MACHINE
---|---
`armv7hf` | `armv7l`

A quick test suggests that only armv7 is different on this. So I'm going to switch to using CROSS_BUILD in the cmake build style. I tried to build x265, but failed at the strip stage on both master & this branch, but the build seems to have passed.

annulen:
@SevereOverfl0w Following patch may help:

```diff
diff --git a/Source/WebCore/platform/graphics/opengl/TemporaryOpenGLSetting.cpp b/Source/WebCore/platform/graphics/opengl/TemporaryOpenGLSetting.cpp
index c502a0c3b95..8ae96b94ce8 100644
--- a/Source/WebCore/platform/graphics/opengl/TemporaryOpenGLSetting.cpp
+++ b/Source/WebCore/platform/graphics/opengl/TemporaryOpenGLSetting.cpp
@@ -29,7 +29,11 @@
 #if ENABLE(GRAPHICS_CONTEXT_3D)
 #include "TemporaryOpenGLSetting.h"
 
-#if USE(OPENGL_ES_2)
+#if PLATFORM(QT)
+#define FUNCTIONS m_functions
+#include "OpenGLShimsQt.h"
+#define glIsEnabled(...) m_functions->glIsEnabled(__VA_ARGS__)
+#elif USE(OPENGL_ES_2)
 #include <GLES2/gl2.h>
 #include "OpenGLESShims.h"
 #elif PLATFORM(IOS)
@@ -38,10 +42,6 @@
 #include <OpenGL/gl.h>
 #elif PLATFORM(GTK) || PLATFORM(EFL) || PLATFORM(WIN)
 #include "OpenGLShims.h"
-#elif PLATFORM(QT)
-#define FUNCTIONS m_functions
-#include "OpenGLShimsQt.h"
-#define glIsEnabled(...) m_functions->glIsEnabled(__VA_ARGS__)
 #endif
 
 namespace WebCore {
diff --git a/Source/WTF/wtf/Platform.h b/Source/WTF/wtf/Platform.h
index f3d72521721..0666ac710fa 100644
--- a/Source/WTF/wtf/Platform.h
+++ b/Source/WTF/wtf/Platform.h
@@ -1052,6 +1052,9 @@
 #if PLATFORM(QT)
 #ifdef __cplusplus
 #include <qglobal.h>
+#if QT_VERSION >= QT_VERSION_CHECK(5,8,0)
+#include <QtGui/qtguiglobal.h>
+#endif
 #endif
 #if defined(QT_OPENGL_ES_2) && !defined(USE_OPENGL_ES_2)
 #define USE_OPENGL_ES_2 1
```

However, it builds fine for aarch64 and mips here even without patch

SevereOverfl0w:
I got this build error for aarch64.

```
In file included from /builddir/qtwebkit-5.212.0-alpha2/Source/WebCore/platform/graphics/opengl/GraphicsContext3DOpenGLCommon.cpp:72:0:
/builddir/qtwebkit-5.212.0-alpha2/Source/WebCore/platform/graphics/opengl/GraphicsContext3DOpenGLCommon.cpp: In member function 'bool WebCore::GraphicsContext3D::checkVaryingsPacking(Platform3DObject, Platform3DObject) const':
/builddir/qtwebkit-5.212.0-alpha2/Source/WebCore/platform/graphics/opengl/GraphicsContext3DOpenGLCommon.cpp:417:21: error: 'GL_MAX_VARYING_FLOATS' was not declared in this scope
     ::glGetIntegerv(GL_MAX_VARYING_FLOATS, &maxVaryingFloats);
                     ^
/builddir/qtwebkit-5.212.0-alpha2/Source/WebCore/platform/graphics/qt/OpenGLShimsQt.h:38:77: note: in definition of macro 'LOOKUP_GL_FUNCTION'
 #define LOOKUP_GL_FUNCTION(f, ...) OpenGLShimsQt::doNothing(), FUNCTIONS->f(__VA_ARGS__)
                                                                             ^~~~~~~~~~~
/builddir/qtwebkit-5.212.0-alpha2/Source/WebCore/platform/graphics/opengl/GraphicsContext3DOpenGLCommon.cpp:417:7: note: in expansion of macro 'glGetIntegerv'
     ::glGetIntegerv(GL_MAX_VARYING_FLOATS, &maxVaryingFloats);
       ^
/builddir/qtwebkit-5.212.0-alpha2/Source/WebCore/platform/graphics/opengl/GraphicsContext3DOpenGLCommon.cpp: In member function 'void WebCore::GraphicsContext3D::texSubImage2D(GC3Denum, GC3Dint, GC3Dint, GC3Dint, GC3Dsizei, GC3Dsizei, GC3Denum, GC3Denum, const void*)':
/builddir/qtwebkit-5.212.0-alpha2/Source/WebCore/platform/graphics/opengl/GraphicsContext3DOpenGLCommon.cpp:1727:16: error: 'GL_HALF_FLOAT_ARB' was not declared in this scope
         type = GL_HALF_FLOAT_ARB;
                ^~~~~~~~~~~~~~~~~
At global scope:
cc1plus: warning: unrecognized command line option '-Wno-expansion-to-defined'
make[2]: *** [Source/WebCore/CMakeFiles/WebCore.dir/build.make:40695: Source/WebCore/CMakeFiles/WebCore.dir/platform/graphics/opengl/GraphicsContext3DOpenGLCommon.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:772: Source/WebCore/CMakeFiles/WebCore.dir/all] Error 2
make: *** [Makefile:163: all] Error 2
=> ERROR: qt5-webkit-5.212.0.alpha2_1: failed to run do_build() at line 48.
```

With the above patch it failed at ~99% about missing references to a bunch of gl related stuff, here's a snippet of the failure, in case that can help before I'm able to get the full log.

```
../../lib/../Source/WebCore/CMakeFiles/WebCore.dir/platform/graphics/opengl/GraphicsContext3DOpenGLES.cpp.o: In function `WebCore::GraphicsContext3D::renderbufferStorage(unsigned int, unsigned int, int, int)':
GraphicsContext3DOpenGLES.cpp:(.text._ZN7WebCore17GraphicsContext3D19renderbufferStorageEjjii+0x40): undefined reference to `glRenderbufferStorage'
../../lib/../Source/WebCore/CMakeFiles/WebCore.dir/platform/graphics/opengl/GraphicsContext3DOpenGLES.cpp.o: In function `WebCore::GraphicsContext3D::getIntegerv(unsigned int, int*)':
GraphicsContext3DOpenGLES.cpp:(.text._ZN7WebCore17GraphicsContext3D11getIntegervEjPi+0x28): undefined reference to `glGetIntegerv'
../../lib/../Source/WebCore/CMakeFiles/WebCore.dir/platform/graphics/opengl/GraphicsContext3DOpenGLES.cpp.o: In function `WebCore::GraphicsContext3D::getShaderPrecisionFormat(unsigned int, unsigned int, int*, int*)':
GraphicsContext3DOpenGLES.cpp:(.text._ZN7WebCore17GraphicsContext3D24getShaderPrecisionFormatEjjPiS1_+0x40): undefined reference to `glGetShaderPrecisionFormat'
../../lib/../Source/WebCore/CMakeFiles/WebCore.dir/platform/graphics/opengl/GraphicsContext3DOpenGLES.cpp.o: In function `WebCore::GraphicsContext3D::depthRange(float, float)':
GraphicsContext3DOpenGLES.cpp:(.text._ZN7WebCore17GraphicsContext3D10depthRangeEff+0x18): undefined reference to `glDepthRangef'
../../lib/../Source/WebCore/CMakeFiles/WebCore.dir/platform/graphics/opengl/GraphicsContext3DOpenGLES.cpp.o: In function `WebCore::GraphicsContext3D::clearDepth(float)':
GraphicsContext3DOpenGLES.cpp:(.text._ZN7WebCore17GraphicsContext3D10clearDepthEf+0x18): undefined reference to `glClearDepthf'
../../lib/../Source/WebCore/CMakeFiles/WebCore.dir/platform/graphics/qt/QFramebufferPaintDevice.cpp.o: In function `QFramebufferPaintDevice::QFramebufferPaintDevice(QSize const&, QOpenGLFramebufferObject::Attachment, bool)':
QFramebufferPaintDevice.cpp:(.text._ZN23QFramebufferPaintDeviceC2ERK5QSizeN24QOpenGLFramebufferObject10AttachmentEb+0xa4): undefined reference to `glClearColor'
QFramebufferPaintDevice.cpp:(.text._ZN23QFramebufferPaintDeviceC2ERK5QSizeN24QOpenGLFramebufferObject10AttachmentEb+0xc8): undefined reference to `glClear'
../../lib/../Source/WebCore/CMakeFiles/WebCore.dir/platform/graphics/opengl/Extensions3DOpenGLES.cpp.o: In function `WebCore::Extensions3DOpenGLES::getExtensions()':
Extensions3DOpenGLES.cpp:(.text._ZN7WebCore20Extensions3DOpenGLES13getExtensionsEv+0x14): undefined reference to `glGetString'
../../lib/../Source/WebCore/CMakeFiles/WebCore.dir/platform/graphics/opengl/Extensions3DOpenGLES.cpp.o: In function `WebCore::Extensions3DOpenGLES::supportsExtension(WTF::String const&)':
Extensions3DOpenGLES.cpp:(.text._ZN7WebCore20Extensions3DOpenGLES17supportsExtensionERKN3WTF6StringE+0x148): undefined reference to `eglGetProcAddress'
Extensions3DOpenGLES.cpp:(.text._ZN7WebCore20Extensions3DOpenGLES17supportsExtensionERKN3WTF6StringE+0x158): undefined reference to `eglGetProcAddress'
Extensions3DOpenGLES.cpp:(.text._ZN7WebCore20Extensions3DOpenGLES17supportsExtensionERKN3WTF6StringE+0x168): undefined reference to `eglGetProcAddress'
Extensions3DOpenGLES.cpp:(.text._ZN7WebCore20Extensions3DOpenGLES17supportsExtensionERKN3WTF6StringE+0x178): undefined reference to `eglGetProcAddress'
Extensions3DOpenGLES.cpp:(.text._ZN7WebCore20Extensions3DOpenGLES17supportsExtensionERKN3WTF6StringE+0x1c8): undefined reference to `eglGetProcAddress'
../../lib/../Source/WebCore/CMakeFiles/WebCore.dir/platform/graphics/opengl/Extensions3DOpenGLES.cpp.o:Extensions3DOpenGLES.cpp:(.text._ZN7WebCore20Extensions3DOpenGLES17supportsExtensionERKN3WTF6StringE+0x1d8): more undefined references to `eglGetProcAddress' follow
../../lib/../Source/WebCore/CMakeFiles/WebCore.dir/platform/graphics/opengl/Extensions3DOpenGLCommon.cpp.o: In function `WebCore::Extensions3DOpenGLCommon::Extensions3DOpenGLCommon(WebCore::GraphicsContext3D*)':
Extensions3DOpenGLCommon.cpp:(.text._ZN7WebCore24Extensions3DOpenGLCommonC2EPNS_17GraphicsContext3DE+0x74): undefined reference to `glGetString'
Extensions3DOpenGLCommon.cpp:(.text._ZN7WebCore24Extensions3DOpenGLCommonC2EPNS_17GraphicsContext3DE+0xc8): undefined reference to `glGetString'
```

Given that it seems your patch got us osme of the way, is there anything we've obviously done wrong?

pullmoll:
It looks like `/usr/include/qt5/QtGui/qopenglext.h` defines the missing symbol `GL_MAX_VARYING_FLOATS`. Perhaps the patch should include this file as well after the line `#include <QtGui/qtguiglobal.h>`.

SevereOverfl0w:
I had to supplement the qtguiglobal.h also with GL/gl.h as there was a missing define. That then got to 99% and then this was missing: https://www.khronos.org/registry/EGL/sdk/docs/man/html/eglGetProcAddress.xhtml

Which i haven't tried rebuilding with. I'm getting a sense of something being fundamentally wrong here. Too many hacks are being required.

Unfortunately I don't have the skill at all to diagnose this.

SevereOverfl0w:
I see a reference to https://github.com/annulen/webkit/issues/601 I'll check it out to see if I can get the crossbuild working.

Looks like next qutebrowser won't support the current qt5-webkit in the repos.

jcjordyn130:
@SevereOverfl0w any updates?

SevereOverfl0w:
My interest has waned a little. Getting this working on multiple architectures proved too much.

There may be a new release, which could be worth trying to see if it helps.

On 19 April 2018 05:14:36 BST, jcjordyn130 <notifications@github.com> wrote:
>@SevereOverfl0w any updates?
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/voidlinux/void-packages/pull/6876#issuecomment-382605063

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #6882, opened 2017-06-28
Version bump: libpurple-facebook

dziq:
Facebook changed it's API and old libpurple-facebook plugin stoped working. This is the new version of this plugin which works fine.


the-maldridge:
@dziq please name your PR and commit according to the [Manual](https://github.com/voidlinux/void-packages/blob/master/Manual.md)

ebfe:
af448bd0273c58229acdc01303ae27bbb335d241

Issue #4147, opened 2016-05-07
RFC: increase musl libc's DEFAULT_STACK_SIZE


pullmoll:
As it turns out even an increasing number of patches for several core libraries which try to reduce the stack frame footprint of some of their functions doesn't help to entirely avoid crashes of some applications with musl libc.

The default stack size is defined in `musl-1.1.14/src/internal/pthread_impl.h` as

```
#define DEFAULT_STACK_SIZE 81920
```

We could try with a substantially increased size to see if this helps failing libraries like `pixman` and `cairo` (#4133) causing `SpaceFM` to crash, and `libfaad2` (#2461) causing `firefox` and `icecat` to crash.

The required size would probably be at least 256K, i.e.

```
#define DEFAULT_STACK_SIZE 262144
```

judging by the size of the structures the mentioned libraries expect to reside in the stack frame.

If we try, and at the same time remove the patches for `libfaad2`, `pixman` and `cairo` (pending), we would need some users to try out applications using these libraries on a musl Void Linux installation.

richfelker:
The current default was chosen empirically such that it solved real-world problems, and such that further incremental increase (without going much larger) did not solve any additional real-world problems. It also allows a fairly reasonable/typical buffer size, 64k, plus sufficient additional room for a few call frames and ability to use libc functions with on-stack buffers.

Really programs should not be using the stack for large buffers to begin with, but if they are, they need to declare their need for a large stack.

xtraeme:
@richfelker I suppose that you are saying that applications should use `pthread_attr_setstacksize()` instead?

pullmoll:
Making all applications which use e.g. `libfaad2`, `pixman` or `cairo` call `pthread_attr_setstacksize()` in their startup code is not really feasible IMO, or less feasible than patching the libraries.

richfelker:
That's why I think the right fix is probably getting rid of the huge on-stack buffers and making them use malloc.

xtraeme:
Hmm true, makes sense.

pullmoll:
As far as I understood the intention of the large on-stack buffers in e.g. pixmap, the intention was to avoid the overhead of malloc()/free() up to some limit. pixman for example allocates an even bigger buffer if the one in the stack is not sufficient, and then free()s the pointer, if its address is different from the on-stack thing.

So far there were no problems with turning such buffers into `static __thread` allocated, which is a little less invasive than changing the code to use malloc()/free().

ebfe:
[...]

PR #6889, opened 2017-06-28
musl: add bigstack build option

ebfe:
This enables building musl with a bigger default pthread stack size
(currently 8MB to match glibc default stack size), to make it easier to
diagnose these issues.

#4147


the-maldridge:
LGTM

chneukirchen:
musl patch missing?

ebfe:
> musl patch missing?

Argh, yes. Fixed.

PR #6895, opened 2017-06-29
ufw: update to 17.10.0

cr6git:


cr6git:
Wrong `pkgname` , stupid me!!

PR #6898, opened 2017-06-29
docker-machine: update to 0.12.0

cr6git:


srcpkgs/docker-machine/template
```diff
@@ -1,18 +1,17 @@
...
 wrksrc="machine-${version}"
 build_style=go
 go_import_path="github.com/docker/machine"
 go_package="${go_import_path}/cmd"
-hostmakedepends="git"
 short_desc="Docker Machine management for a container-centric world"
 maintainer="Sergi Alvarez <pancake@nopcode.org>"
 license="Apache-2.0"
-homepage="https://github.com/docker/machine/"
-distfiles="https://github.com/docker/machine/archive/v${version}.tar.gz"
```
ebfe: Please stop adding unnecessary variables to `distfiles`.

cr6git: @ebfe I will stop. 
I merely thought that it was better to reduce the written text, in favour of variables. 
I apologize. 
I will learn my lesson. 😉

PR #6910, opened 2017-06-30
sigrok-cli: Add build dependency on python3

pienjo:
One of the most important features of the sigrok suite are the protocol decoders. The template intends to build sigrok-cli with these decoders enabled, but this silently fails because of the missing dependency.


the-maldridge:
Shouldn't this also depend on python3?  If the python artifacts are separate from the main program, please create a subpkg that will contain the python components.

pienjo:
Sigrok is a bit weird in this respect. I'll try to explain.

The runtime dependency isn't really with the commandline tools, but rather with the decoding library - which uses python3 to perform the actual decoding, but that's almost an implementation detail, and is not exposed. The libsigrokdecode library ships with these python scripts, and has a shlib-requires on python3 - so sigrok-cli already indirectly depends on python3.

The problem is that (before my changes) sigrok-cli has a build dependency on libsigrokdecode (which links against libpython), but libsigrokdecode-devel doesn't automatically pull in python3-devel, so the configure script fails and makes a wrong decision. Since it is marked as an optional dependency it'll go ahead and compile without the protocol decoders. 

Maybe I fixed the wrong package - maybe I should've fixed libsigrokdecode-devel instead, but I'm not entirely sure how. 

PS: There's another package that uses libsigrokdecode : pulseview, and it currently has a makedepends on python3-devel as well, for the very same reason. I think it's probably best to treat them the same.

asergi:
@pienjo the depends of `libsigrokdecode-devel` is wrong:
```diff
 libsigrokdecode-devel_package() {
-       depends="${sourcepkg}>=${version}_${revision}"
+       depends="${makedepends} ${sourcepkg}>=${version}_${revision}"
```

pienjo:
@asergi Thank you, I shall change that.

Issue #6892, opened 2017-06-28
screen: update to 4.6.0 fails


ghost:
I can't update [screen](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/screen/template) to version 4.6.0, see my [build log](https://paste2.org/2WfxvWXZ).

the-maldridge:
closed by #6919

PR #6919, opened 2017-07-01
screen: update to 4.6.0

Sir-Boops:


Sir-Boops:
I'm unsure about #6892 as this upgrade worked flawlessly for me

PR #6921, opened 2017-07-01
Sauerbraten: add musl and i686

NeoChontrix:
Just for the sacke of having also other platforms in the template and I just go an assume that the comment is correct and it wouldn't even run on other platforms even if compiled on them.


Vaelatern:
@the-maldridge is `-musl` a viable candidate here?

NeoChontrix:
@Vaelatern the game is compiling and running on musl.. therefore I would say yes (but I'm not the-maldrige)

the-maldridge:
The official statement from upstream is that the enet state synchronization library does not cross compile.  Since the architectures you suggest are not cross compiled, I see no reason not to add the lines you provide.  You are aware though that we do not run an i686-musl builder though, right?

Honestly it might be more correct to drop `only_for_archs` and instead add `nocross` since we may someday operate native builders for other architectures.

Issue #4550, opened 2016-08-03
Package request: caddy


jannisch:
[homepage](https://caddyserver.com/) / [github](https://github.com/mholt/caddy)

Caddy is a modern and easy to use HTTP server written in Go with built-in Let's Encrypt.
I made a few attempts but since I don't know anything about the language and `go get` I failed and thought it might be a quick job for a Go dev here.

PR #6925, opened 2017-07-02
New package: caddy-0.10.4

SevereOverfl0w:
#4550 is related.


SevereOverfl0w:
This might qualify for a system account to be required as part of the package.

the-maldridge:
Generally system accounts are only needed if there's a long running service.

PR #6927, opened 2017-07-02
compton-git: Add a proper default compton.conf

ColdPhoenix00:


Vaelatern:
I vote to close. Upstream has not issued a `compton.conf` and this would be customization in the distro that I do not think we desire.

biopsin:
also this is present in /etc/skel if one needs hint (at least I have one there)

Vaelatern:
@biopsin that's just you.

PR #6935, opened 2017-07-03
chromium-widevine: update to 59.0.3071.115

diogoleal:


the-maldridge:
It was my understating that this had long since been obsoleted and that modern versions of Firefox shipped widevine internally (which is why we had this package).

Duncaen:
This does not work, chrome doesnt ship widevine as a shared object anymore, test your updates.

PR #6943, opened 2017-07-03
New package: Carla-1.9.7

ibrokemypie:
not building because
```
=> Carla-1.9.7_1: running pre-pkg hook: 03-rewrite-python-shebang ...
=> ERROR: Carla-1.9.7_1: failed to run pre-pkg_03-rewrite-python-shebang() at line 10.
=> ERROR: Carla-1.9.7_1: failed to run pre-pkg_03-rewrite-python-shebang() at line 10.

```
which I cant find any documentation about...

### DIFF
--- a/srcpkgs/Carla/template
+++ b/srcpkgs/Carla/template
@@ -1,0 +1,19 @@
+# Template file for 'Carla'
+pkgname=Carla
+version=1.9.7
+revision=1
+build_style=gnu-makefile
+python_version=3
+hostmakedepends="pkg-config"
+makedepends="python3-PyQt5-devel-tools python3-PyQt5 libmagic liblo-devel alsa-lib-devel pulseaudio-devel libX11-devel qt5-devel gtk+3-devel fluidsynth-devel fftw-devel mxml-devel zlib-devel python3-rdflib"
+depends="python3 python3-PyQt5"
+short_desc="A fully-featured audio plugin host"
+maintainer="ibrokemypie <ibrokemypie@bastardi.net>"
+license="GPL-3"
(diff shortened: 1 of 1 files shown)


srcpkgs/Carla/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'Carla'
+pkgname=Carla
+version=1.9.7
+revision=1
+build_style=gnu-makefile
+python_version=3
```
Vaelatern: 3.5

srcpkgs/Carla/template
```diff
@@ -0,0 +1,19 @@
...
+hostmakedepends="pkg-config"
+makedepends="python3-PyQt5-devel-tools python3-PyQt5 libmagic liblo-devel alsa-lib-devel pulseaudio-devel libX11-devel qt5-devel gtk+3-devel fluidsynth-devel fftw-devel mxml-devel zlib-devel python3-rdflib"
```
Vaelatern: PLEASE break this up into multiple lines.
Standard is to break at a space, no \ at the end of line needed or wanted, and any lines continued like this should begin with a space. See other packages with long `depends` for an example. Like `base-system`.

srcpkgs/Carla/template
```diff
@@ -0,0 +1,19 @@
...
+depends="python3 python3-PyQt5"
+short_desc="A fully-featured audio plugin host"
+maintainer="ibrokemypie <ibrokemypie@bastardi.net>"
+license="GPL-3"
```
Vaelatern: Where do you get that?

asergi:
Abandoned by author.

PR #6949, opened 2017-07-03
New packages: yarn and yarn-bin

kulinacs:
Unfortunately, as yarn requires itself to build, the yarn-bin package is required.


srcpkgs/yarn/template
```diff
@@ -0,0 +1,34 @@
+# Template file for 'yarn'
+pkgname=yarn
+version=0.27.5
+revision=1
+hostmakedepends="yarn-bin"
+depends="nodejs"
+short_desc="Fast, reliable, and secure dependency management"
+maintainer="Nicklaus McClendon <nicklaus@kulinacs.com>"
+license="2-cause-BSD"
+homepage="https://github.com/yarnpkg/${pkgname}"
```
the-maldridge: homepage should not have any variables in it.

srcpkgs/yarn/template
```diff
@@ -0,0 +1,34 @@
...
+
+do_build() {
+	yarn run build-dist
+}
+
+do_install() {
+	vmkdir /usr/share/yarn/
+	vmkdir /usr/bin
+	vcopy dist/lib/ /usr/share/yarn/lib/
+	vcopy dist/bin/ /usr/share/yarn/bin/
```
the-maldridge: is there a reason this can't be vbin'd ?

srcpkgs/yarn-bin/template
```diff
@@ -0,0 +1,29 @@
...
+do_extract() {
+	tar xfz ${XBPS_SRCDISTDIR}/${pkgname}-${version}/yarn-v${version}.tar.gz --strip-components=1 -C ${wrksrc}
+}
+
+do_install() {
+	vmkdir /usr/share/yarn/
+	vmkdir /usr/bin
+	vcopy bin /usr/share/yarn/bin
+	vcopy lib /usr/share/yarn/lib
+	ln -rs ${DESTDIR}/usr/share/yarn/bin/yarn ${DESTDIR}/usr/bin/yarn
```
the-maldridge: Is there a reason this can't be vbin'd?

the-maldridge:
I'm also inclined to believe that the homepage for both of these should be https://yarnpkg.com/

srcpkgs/yarn/template
kulinacs: Developer recommends installing to /usr/share/yarn/ 
https://github.com/yarnpkg/yarn/blob/master/scripts/build-deb.sh#L42

srcpkgs/yarn-bin/template
kulinacs: Developer recommends installing to /usr/share/yarn/ 
https://github.com/yarnpkg/yarn/blob/master/scripts/build-deb.sh#L42

srcpkgs/yarn/template
kulinacs: Updated

kulinacs:
Homepage changed. Developer recommends putting everything in /usr/share/yarn, as the executables expect ../lib to contain the libraries for yarn, and these libraries are architecture independent.

kulinacs:
Reworked to install to /usr/bin and /usr/lib/yarn

the-maldridge:
It's a nodejs package so I fully expect this to be unmaintainable, but it looks decent.

Issue #4106, opened 2016-04-30
mongodb: crashes on start


jsumners:
I have a fresh install of Void (installed on 2016-04-14). I have not been able to get mongodb to launch on it. I just tried again after the 3.2.5_1 package was released. It is still crashing on launch:

```
2016-04-15T20:19:46.708-0400 I CONTROL  [main] ***** SERVER RESTARTED *****
2016-04-15T20:19:46.719-0400 I CONTROL  [initandlisten] MongoDB starting : pid=10846 port=27017 dbpath=/var/lib/mongodb 64-bit host=ice-guild.info
2016-04-15T20:19:46.719-0400 I CONTROL  [initandlisten] db version v3.2.4
2016-04-15T20:19:46.719-0400 I CONTROL  [initandlisten] git version: e2ee9ffcf9f5a94fad76802e28cc978718bb7a30
2016-04-15T20:19:46.719-0400 I CONTROL  [initandlisten] OpenSSL version: LibreSSL 2.3.3
2016-04-15T20:19:46.719-0400 I CONTROL  [initandlisten] allocator: tcmalloc
2016-04-15T20:19:46.719-0400 I CONTROL  [initandlisten] modules: none
2016-04-15T20:19:46.719-0400 I CONTROL  [initandlisten] build environment:
2016-04-15T20:19:46.719-0400 I CONTROL  [initandlisten]     distarch: x86_64
2016-04-15T20:19:46.719-0400 I CONTROL  [initandlisten]     target_arch: x86_64
2016-04-15T20:19:46.719-0400 I CONTROL  [initandlisten] options: { config: "/etc/mongodb/mongodb.conf", net: { bindIp: "127.0.0.1" }, storage: { dbPath: "/var/lib/mongodb" }, systemLog: { destination: "file", logAppend: true, path: "/var/log/mongodb/mongod.log", quiet: true } }
2016-04-15T20:19:46.770-0400 I STORAGE  [initandlisten] wiredtiger_open config: create,cache_size=1G,session_max=20000,eviction=(threads_max=4),config_base=false,statistics=(fast),log=(enabled=true,archive=true,path=journal,compressor=snappy),file_manager=(close_idle_time=100000),checkpoint=(wait=60,log_size=2GB),statistics_log=(wait=0),
[...]

PR #6964, opened 2017-07-05
mongodb: unbreak

ebfe:


ebfe:
Still segfaults with the default config when initializing wiredtiger #4106 
```
#0  raise (sig=5) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00005555569481e5 in mongo::breakpoint() ()
#2  0x0000555555c356b8 in mongo::fassertFailedNoTraceWithLocation(int, char const*, unsigned int) ()
#3  0x00005555566bb0e5 in mongo::WiredTigerKVEngine::WiredTigerKVEngine(std::string const&, std::string const&, mongo::ClockSource*, std::string const&, unsigned long, bool, bool, bool, bool) ()
#4  0x00005555566b1f06 in ?? ()
#5  0x0000555556598b77 in mongo::ServiceContextMongoD::initializeGlobalStorageEngine() ()
#6  0x0000555555c1bd04 in ?? ()
#7  0x0000555555c3d514 in main ()
```

PR #6975, opened 2017-07-06
New package: libudev0-shim-1

autrimpo:
A shim package for libudev.so.0 compatibility, required for running Steam with native libraries.


Gottox:
Is this any useful outside of steam?

autrimpo:
Probably not, even the Arch official package is a dependency of only `steam-native-runtime`. But the package is tiny (less than 3kB when built on my system), builds extremely fast and most probably won't be updated often (if at all), so the maintenance costs are almost zero.

Gottox:
I would prefer to ship with steam directly.

autrimpo:
As in ship the shim in the steam package? Sounds like a good idea.

Gottox:
👍 

Closing

PR #6980, opened 2017-07-07
R: update to 3.4.1.

wagnerflo:


ebfe:
Do the R-cran-* packages need a rebuild?

wagnerflo:
I don't think so. I've googled around about how ABI changes to libR.so are supposed to be handled, announced or whatever but came up with nothing. So I've simply built this package and tested it with a script using `R-cran-data.table` (which is partially implemented in a shared library) and came across no problems.

PR #6994, opened 2017-07-07
corebird: fix cross compiling

Sir-Boops:
* Tried on armv7l-musl and it seems to work fine ( Only launched it, i didn't go any deeper )

### DIFF
--- a/srcpkgs/corebird/template
+++ b/srcpkgs/corebird/template
@@ -1,7 +1,7 @@
 # Template build for 'corebird'.
 pkgname=corebird
 version=1.5.1
-revision=1
+revision=2
 short_desc="Native Gtk+ Twitter Client"
 maintainer="Enno Boland <gottox@voidlinux.eu>"
 license="GPL-3"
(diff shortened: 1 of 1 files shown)


srcpkgs/corebird/template
```diff
@@ -14,10 +14,13 @@ hostmakedepends="pkg-config intltool vala glib-devel gdk-pixbuf libgee08-devel
...
-	arm*) broken="https://build.voidlinux.eu/builders/armv7l-musl_builder/builds/12500" ;;
-esac
+if [ -n "${CROSS_BUILD}" ]; then
+	hostmakedepends+=" gspell-devel"
+fi
 
 pre_configure() {
 	autoreconf -fi
+	if [ -n "${CROSS_BUILD}" ]; then
+		sed -i '13080,13083d' configure
```
ebfe: Please use a pattern that has a chance to survive an update.

ebfe: Even better: drop this and `hostmakedepends+=gstreamer1-devel"

Sir-Boops: Wow, ya that's alot cleaner 😄

srcpkgs/corebird/template
```diff
@@ -14,9 +14,9 @@ hostmakedepends="pkg-config intltool vala glib-devel gdk-pixbuf libgee08-devel
 makedepends="gtk+3-devel rest-devel libgee08-devel sqlite-devel gspell-devel
  libsoup-devel json-glib-devel gettext-devel gst-plugins-base1-devel"
 
-case "$XBPS_TARGET_MACHINE" in
-	arm*) broken="https://build.voidlinux.eu/builders/armv7l-musl_builder/builds/12500" ;;
-esac
+if [ -n "${CROSS_BUILD}" ]; then
+	hostmakedepends+=" gspell-devel gstreamer1-devel"
```
asergi: Just add both to hostmakedepends directly and then you can remove  the if statement at all.

Sir-Boops: Ok, I was under the impression that this was the way you should add arch specific dependencies

PR #6996, opened 2017-07-08
New package: ifupdown-0.8.19

the-maldridge:
This package provides the popular /etc/network/interfaces mechanism to configure the network at boot.


cr6git:
@the-maldridge I used a lot `ifupdown` a long time ago in Debian because it was installed by default. 
Now we have it in Void but I don't know how to make it work with `dhcpcd` / `wpa_supplicant`. 
Please could you explain what is the **correct way** to use `ifupdown` in Void ? 
Thank you. 😃

the-maldridge:
I can't speak for correctness, but its build directly from the debian source tarball.  I'd suggest consulting with the upstream documentation to configure to your liking.  I mostly needed it for a handful of places that I absolutely needed the network to be up before runlevel 2 launched, and because its a lot more convenient to bring up vxlans and bridges with it.

Feel free to draft a wiki page if you have the time, I'm in a bit of a busy spot where I can't contribute to void as much as I'd like.

PR #7008, opened 2017-07-10
orca, brltty: fix various issues and deps.

CMB:


CMB:
As far as I can tell, the build failures for this PR are caused by
the shebang line in /usr/bin/itstool:
#!/usr/bin/python -s
itstool is python2, but python3 was installed first.

asergi:
`itstool` "fixed".

PR #7016, opened 2017-07-11
srcpkgs/*: fix (L)GPL without version lints

ebfe:
There are still quite a few left (no license file, multiple licenses, ...), but this fixes the ones that where easy to identify and correct.


ebfe:
done

PR #7021, opened 2017-07-11
New package: godm-1.4

kulinacs:
Package name changed from upstream's gdm to avoid the name conflict.


kulinacs:
Closed, as included in #7070

ebfe:
godm is not a very good name. Coupled with the `short_desc` it won't even show up on a search for `gdm`.

kulinacs:
On July 18, 2017 6:11:06 AM CDT, Michael Gehring <notifications@github.com> wrote:
>godm is not a very good name. Coupled with the `short_desc` it won't
>even show up on a search for `gdm`.
>
>-- 
>You are receiving this because you modified the open/close state.
>https://github.com/voidlinux/void-packages/pull/7021#issuecomment-316032267

Good point. I can change the `short_desc` to something like "Go Dependency Manager - gdm"

I'm not sure what else to name it. Maybe go-gdm or gdm-go?

PR #7030, opened 2017-07-12
libextractor: update to 1.4

### LINKED ISSUE #4291: csound conflicts with libextractor
```
 /usr/bin/extract
      csound-6.07.0_1
      libextractor-1.3_4
```

Sir-Boops:
* Also fixes #4291


srcpkgs/libextractor/template
```diff
@@ -1,20 +1,21 @@
...
 maintainer="Martin Riese <grauehaare@gmx.de>"
 license="GPL-3"
 homepage="http://www.gnu.org/software/libextractor/"
 distfiles="${GNU_SITE}/$pkgname/$pkgname-$version.tar.gz"
-checksum=868ad64c9a056d6b923d451d746935bffb1ddf5d89c3eb4f67d786001a3f7b7f
+checksum=84128170a4a9aa3a19942dd53fdf30ed17b56d7fae79b5f6e7e17a0d65d1f66c
 build_style=gnu-configure
 configure_args="--disable-static"
-hostmakedepends="automake gettext-devel libtool pkg-config"
+hostmakedepends="automake gettext-devel libtool pkg-config iso-codes"
```
ebfe: What is iso-codes for? There is an optional check for it, but it doesn't seem to be used anywhere. If it is used somewhere it likely is also needed at runtime.

srcpkgs/libextractor/patches/gif.patch
```diff
@@ -0,0 +1,20 @@
...
+       EGifCloseFile (gif_file, NULL);
+@@ -137,7 +137,7 @@
+ 	       DGifGetExtensionNext(gif_file, &ext)) &&
+ 	      (NULL != ext) ) ; /* keep going */
+     }
+-#if defined (GIF_LIB_VERSION) || GIFLIB_MAJOR < 5 || GIFLIB_MINOR < 1
++#if defined (GIF_LIB_VERSION) || GIFLIB_MAJOR <= 5 || GIFLIB_MINOR < 1
+   DGifCloseFile (gif_file);
+ #else
+   DGifCloseFile (gif_file, NULL);
```
ebfe: remove the sed line in do_extract instead.

srcpkgs/libextractor/template
Sir-Boops: That was put there to satisfy the configure check for it thinking it had some use. I also don't see any use for it digging around so I'll remove it.

srcpkgs/libextractor/patches/gif.patch
Sir-Boops: Wow, I'm blind sometimes 😆

PR #7033, opened 2017-07-13
New package: SASM-3.8.0

ChipOManiac:


srcpkgs/SASM/template
```diff
@@ -0,0 +1,15 @@
+# Template file for 'SASM'
+pkgname=SASM
+version=3.8.0
+revision=1
+only_for_archs="i686 x86_64"
```
the-maldridge: This is wrong.  fix the crossbuilds and try again.  If this really can't be built for other architectures than mark it broken.

srcpkgs/SASM/template
```diff
@@ -0,0 +1,15 @@
...
+build_style=qmake
+hostmakedepends="qt-qmake make"
+makedepends="qt-devel"
+depends="qt nasm gdb gcc"
```
the-maldridge: this shouldn't need to depend on qt, the others look okay though.

srcpkgs/SASM/template
ChipOManiac: Changed.

srcpkgs/SASM/template
ChipOManiac: Changed.

PR #7035, opened 2017-07-13
New package: VeraCrypt-1.21

GloryS:


srcpkgs/VeraCrypt/template
```diff
@@ -0,0 +1,40 @@
+# Template file for 'VeraCrypt'
+pkgname=VeraCrypt
+_appname=veracrypt
+version=1.21
+revision=1
+create_wrksrc=yes
+only_for_archs="i686 x86_64"
```
NeoChontrix: Works fine on Musl and nocross is probably more appropriate.

`nocross="wxWidget"`

srcpkgs/VeraCrypt/template
```diff
@@ -0,0 +1,40 @@
...
+build_style=gnu-makefile
+hostmakedepends="wxWidgets wxWidgets-devel yasm fuse fuse-devel pkg-config makeself"
+makedepends="wxWidgets wxWidgets-devel yasm fuse fuse-devel pkg-config makeself"
+depends="wxWidgets fuse"
```
NeoChontrix: `hostmakedepends="yasm pkg-config"
makedepends="wxWidgets-devel fuse-devel"`
This should be enough.

GloryS: You right, this enough.

srcpkgs/VeraCrypt/template
GloryS: Hmm...What about arm? Developer mentions only x86.
https://www.veracrypt.fr/en/Supported%20Operating%20Systems.html

NeoChontrix: OK, haven't checked the site. Then you may only have to add x86-64-musl.

GloryS: Add musl archs...

srcpkgs/VeraCrypt/template
```diff
@@ -0,0 +1,40 @@
+# Template file for 'VeraCrypt'
+pkgname=VeraCrypt
+_appname=veracrypt
+version=1.21
+revision=1
+create_wrksrc=yes
+only_for_archs="i686 x86_64 i686-musl x86_64-musl"
```
the-maldridge: This is almost certainly wrong.  Fix the cross builds or if it truly can't be cross compiled mark it broken for the applicable architectures.

srcpkgs/VeraCrypt/template
```diff
@@ -0,0 +1,40 @@
...
+depends="wxWidgets fuse"
+short_desc="VeraCrypt is a free open source disk encryption software"
+maintainer="Viatcheslav E. Kouznetsov <glorysmith@gmail.com>"
+license="Apache 2.0"
+homepage="https://www.veracrypt.fr/en"
+_filename="${pkgname}_${version}_Source.tar.bz2"
+distfiles="https://launchpadlibrarian.net/328011925/${_filename}"
+checksum=44717d24b8deff2db9dd0365572063e38bbe4a59438a8c3c07675b5eed370358
+
+do_extract() {
```
the-maldridge: This looks fishy, you should probably be setting wrksrc to what this extracts to instead.

srcpkgs/VeraCrypt/template
```diff
@@ -0,0 +1,40 @@
...
+	tar xf ${XBPS_SRCDISTDIR}/${pkgname}-${version}/${_filename} \
+	  -C ${wrksrc}/
+}
+
+do_build() {
```
the-maldridge: No, you should set build_wrksrc instead.  This breaks all the flags that should be passed here.

srcpkgs/VeraCrypt/template
```diff
@@ -0,0 +1,40 @@
...
+	cd ${wrksrc}/src
+	make
+}
+
+do_install() {
+	cd ${wrksrc}/src
+	vbin Main/${_appname}
+	vbin Setup/Linux/${_appname}-uninstall.sh
+	vdoc License.txt
```
the-maldridge: No, this should be vlicense not vdoc so that it winds up in the right place.

maxice8:
@the-maldridge VeraCrypt has already been updated to 1.22

PR #7043, opened 2017-07-14
New package: mysql++-3.2.3

DirectorX:


srcpkgs/mysql++/template
```diff
@@ -0,0 +1,21 @@
...
+makedepends="libmysqlclient-devel"
+depends="libmysqlclient-devel"
+short_desc="A C++ wrapper for MySQLs C API"
+maintainer="DirectorX <void.directorx@protonmail.com>"
+license="LGPL-2.1 custom"
+homepage="https://tangentsoft.net/mysqlpp"
+distfiles="${homepage}/releases/${pkgname}-${version}.tar.gz"
+checksum=c804c38fe229caab62a48a6d0a5cb279460da319562f41a16ad2f0a0f55b6941
+
+if [ -n "$CROSS_BUILD" ]; then
```
Gottox: needless if-statement. remove it around the hostmakedepends.

Also use hostmakedepends=, as it is only set once in the template.

DirectorX: But `hostmakedepends` is absolutely not necessary when building natively

srcpkgs/mysql++/template
```diff
@@ -0,0 +1,21 @@
...
+	hostmakedepends+=" libmysqlclient-devel"
+fi
+
+post_install() {
+	vlicense COPYING.txt
+}
```
Gottox: A -devel subpackage is missing.

srcpkgs/mysql++/template
Gottox: Yea, it's in makedepends; xbps will handle this case and only install it once on non-crossbuilds.

srcpkgs/mysql++/template
```diff
@@ -0,0 +1,21 @@
+# Template file for 'mysql++'
+pkgname=mysql++
+version=3.2.3
+revision=1
+build_style=gnu-configure
+makedepends="libmysqlclient-devel"
+depends="libmysqlclient-devel"
```
Gottox: This is only needed in the `-devel` subpackage

srcpkgs/mysql++/template
DirectorX: Nope `mysql++` is `-devel` package. mysql++ provides C++ headers and `*.so` & `*.a` files.

Gottox: Is there a reason against a `-devel` subpkg?

DirectorX: Adding `-devel` subpkg is same as renaming `pkgname` to `mysql++-devel`

ebfe:
> Adding `-devel` subpkg is same as renaming `pkgname` to `mysql++-devel`

No
```
/usr/lib/libmysqlpp.so.3
/usr/lib/libmysqlpp.so.3.2.3
```

Don't belong into `-devel`

DirectorX:
Ok. So I need to patch the Makefile as well.

ebfe:
> Ok. So I need to patch the Makefile as well.
Huh?

Gottox:
Why do you need to patch the makefile?

xbps-src just move certain files from the main package into a subpackage. See zlib or any other library as an example.

DirectorX:
The Makefile `install` target just installs the `*.so` files. I need to prevent this from happening while packing the non  `-devel` aka the  main package.

Gottox:
See this example:
```
zlib-devel_package() {
	depends="zlib>=${version}_${revision}"
	pkg_install() {
		vmove usr/include
		vmove usr/lib/pkgconfig
		vmove "usr/lib/*.a"
		vmove "usr/lib/*.so"
		vmove usr/share
	}
}
```
this moves certain files and directories to the subpkg. you can do the same for mysql++.

DirectorX:
I think you don't get what mysql++ is.

This package is nothing but development files... 

https://travis-ci.org/voidlinux/void-packages/jobs/253752373#L2321

Gottox:
Yes, if that would be the case, a devel subpkg would be needless.

But that is built on travis contains shared objects: https://travis-ci.org/voidlinux/void-packages/jobs/253752373#L2376-L2378

DirectorX:
OK now everything clear, I will take zlib as example.

DirectorX:
@Gottox ping

srcpkgs/mysql++/patches/01-fix-cross.patch
```diff
@@ -0,0 +1,30 @@
...
+ #include <iomanip>
++#include <cstring>
+ 
+ #include <stdlib.h>
+ #include <time.h>
+ 
+ using namespace std;
+@@ -51,11 +52,11 @@
+ 	localtime_r(&t, ptm);
+ #else
```
ebfe: This patch looks wrong. localtime_r should be available on all archs/libcs so this branch should never matter. Maybe fix the configure script instead.

srcpkgs/mysql++/template
```diff
@@ -0,0 +1,29 @@
...
+post_install() {
+	vlicense COPYING.txt
+}
+
+mysql++-devel_package() {
+	depends="mysql++>=${version}_${revision} libmysqlclient-devel"
+	short_desc+=" - development files"
+	pkg_install() {
+		vmkdir usr/include/mysql++
+		vcopy lib/*.h usr/include/mysql++
```
ebfe: The unversioned /usr/lib/libmysqlpp.so goes into the -devel too.

DirectorX: seriously?
If so, the main package will be empty...

let me show you the tree:
```
usr/
├── include
│   └── mysql++
│       ├── autoflag.h
│       ├── beemutex.h
│       ├── cmdline.h
│       ├── common.h
│       ├── comparable.h
│       ├── connection.h
│       ├── cpool.h
│       ├── datetime.h
│       ├── dbdriver.h
│       ├── exceptions.h
│       ├── field.h
│       ├── field_names.h
│       ├── field_types.h
│       ├── insertpolicy.h
│       ├── manip.h
│       ├── myset.h
│       ├── mysql++.h
│       ├── mystring.h
│       ├── noexceptions.h
│       ├── null.h
│       ├── options.h
│       ├── qparms.h
│       ├── query.h
│       ├── querydef.h
│       ├── refcounted.h
│       ├── result.h
│       ├── row.h
│       ├── scopedconnection.h
│       ├── sql_buffer.h
│       ├── sql_types.h
│       ├── sqlstream.h
│       ├── ssqls.h
│       ├── ssqls2.h
│       ├── stadapter.h
│       ├── stream2string.h
│       ├── tcp_connection.h
│       ├── tiny_int.h
│       ├── transaction.h
│       ├── type_info.h
│       ├── uds_connection.h
│       ├── utility.h
│       ├── vallist.h
│       └── wnp_connection.h
└── lib
    ├── libmysqlpp.so -> libmysqlpp.so.3
    ├── libmysqlpp.so.3 -> libmysqlpp.so.3.2.3
    └── libmysqlpp.so.3.2.3
```

srcpkgs/mysql++/template
```diff
@@ -0,0 +1,30 @@
...
+		vmove "usr/lib/*.so"
```
DirectorX: ping @ebfe @Gottox

srcpkgs/mysql++/template
ebfe: Yes, seriously. And no, the main package wouldn't be empty if you move `/usr/lib/libmysqlpp.so` into -devel.

DirectorX:
ping @Gottox

DirectorX:
ping @Gottox

PR #7048, opened 2017-07-15
lollypop: update to 0.9.244

cr6git:


pullmoll:
See [38b55b9](https://github.com/voidlinux/void-packages/commit/38b55b905a5c8aa3afb3e846ce279772690b792e)

PR #7052, opened 2017-07-15
xcalc: fix wrong template

cr6git:


Duncaen:
Increasing the revision is useless.

PR #7069, opened 2017-07-17
New package: urlview-0.9

LiamHowley:
Urlview is a complement to mutt, which can be found in the repositories. Without it mutt is unwieldy to use.
I hope this is useful.

Thanks,
Liam


xYuusha:
It gives me segfault using x86_64 glibc.
It's fixed using this patch.
https://bugs.gentoo.org/show_bug.cgi?id=70254#c1

LiamHowley:
That's patched and should work now.

xYuusha:
The package works till i manually copy the url_handler.sh to the PATH.
I think is a good idea to add it to the package alongside
sample.urlview.

And by the way, does the build strips the manpages? They are referenced
in the config but they are not created at buildtime or installed.

LiamHowley:
Hi xYuusha,
Thanks for the patience. I'd had an existing config on my machine from a src build, so I'd missed a few things. Man pages and url_handler are added now. If there's anything else I've overlooked, let me know.  

One thing: Xlint fails because of "indent with tabs". I saw it referenced in #2422 . Don't know if there's a solution other than non-indented code. Better that it's readable. Let me know if you want that changed.

xYuusha:
Everything work fine for me now.

On the xlint side, i think the error is weird mostly because i changed the spaces to tabs and it stopped failing. Maybe your editor is changing the tabs to spaces or something like that.

LiamHowley:
Ahh! Sleep!!!! I was reading "indent with tabs" as the error instead of the solution. Fixed that.

PR #7070, opened 2017-07-18
TICK Monitoring Stack

kulinacs:


the-maldridge:
@kulinacs please contact me OOB to discuss this PR.

srcpkgs/influxdb/template
```diff
@@ -1,46 +1,41 @@
...
-build_style=go
-hostmakedepends="git-perl mercurial"
-go_import_path="github.com/influxdata/influxdb"
-go_package="${go_import_path}/cmd/influx
- ${go_import_path}/cmd/influxd
- ${go_import_path}/cmd/influx_inspect
- ${go_import_path}/cmd/influx_stress
- ${go_import_path}/cmd/influx_tsm"
+version=1.3.1
+revision=6
```
Vaelatern: no

srcpkgs/influxdb/template
```diff
@@ -1,46 +1,41 @@
...
+_branch=1.3
+_commit=5887e92
+build_style="go"
```
Vaelatern: You added quotes where not needed.

srcpkgs/influxdb/template
```diff
@@ -1,46 +1,41 @@
...
+nocross=1
```
Vaelatern: Reason?

srcpkgs/telegraf/files/telegraf.conf
```diff
@@ -0,0 +1,2477 @@
+# Telegraf Configuration
```
Vaelatern: Please lose this file. Perhaps there can be a good configuration file found elsewhere, but I don't want this file in our repo.

srcpkgs/influxdb/template
```diff
@@ -1,46 +1,41 @@
...
+go_import_path="github.com/influxdata/${pkgname}"
```
Vaelatern: no need to add a variable here, especially when the variable is longer than the string it replaces.

srcpkgs/influxdb/template
kulinacs: https://travis-ci.org/kulinacs/void-packages/jobs/258683433

the-maldridge: 1) That doesn't look like the cross build is actually failing.
2) If the cross build is actually failing, set nocross to be the link of the failing build log (ideally plaintext) as xbps-src will dump the contents of that variable into the logs for why this package wasn't cross built.

kulinacs:
Requested changes have been made.

PR #7073, opened 2017-07-18
New package: wimlib-1.11.0

KenjiTakahashi:


srcpkgs/wimlib/INSTALL.msg
```diff
@@ -0,0 +1,5 @@
+Some optional packages must be installed for additional functionality:
+
+ - cdrtools: Making ISO images
+ - mtools, syslinux: Making disk images
+ - cabextract: Extracting WinPE from WAIK
```
Vaelatern: I'd rather not have this `INSTALL.msg`.

srcpkgs/wimlib/template
```diff
@@ -0,0 +1,15 @@
+# Template file for 'wimlib'
+pkgname=wimlib
+version=1.11.0
+revision=1
+build_style=gnu-configure
+configure_args="--with-libcrypto"
+hostmakedepends="pkg-config"
+makedepends="libxml2-devel ntfs-3g-devel libressl-devel"
+depends="libxml2 ntfs-3g libressl fuse"
```
Vaelatern: This line is fully redundant. Please remove it.
https://travis-ci.org/voidlinux/void-packages/jobs/254893440#L1199-L1206
all packages listed there ^ are automatically pulled in. libcrypto is libressl, so that's libressl pulled in as well.

srcpkgs/wimlib/INSTALL.msg
KenjiTakahashi: What would you have instead? I've looked around and this seemed like the only possible "solution/workaround", that a bunch of other packages are using already (e.g. `syslinux`, `empathy`, `corebird`).

srcpkgs/wimlib/template
KenjiTakahashi: Oh, I didn't know the system is that clever, nice :-). Will do.

KenjiTakahashi:
Got around to it again. Removed `depends` line, split into `lib` and `-devel` packages and added `common/shlibs` entry. Also updated to `1.12`, which got released in the meantime.

PR #7078, opened 2017-07-18
compiz: update to 0.8.14

CoolOhm:
Update compiz-reloaded from 0.8.12.x to 0.8.14. .

This PR will at first attempted to update all compiz packages, that failed.

EDIT: the all-in-one PR has failed. Thus emerald will be in another PR and the comiz-plugins-* will be in yet another PR.


Gottox:
bling it on

PR #7080, opened 2017-07-18
mpv: update to 0.26.0

cr6git:


srcpkgs/mpv/template
```diff
@@ -18,11 +18,11 @@ makedepends="
  libXv-devel libxkbcommon-devel libva-glx-devel libvdpau-devel alsa-lib-devel
  pulseaudio-devel libbluray-devel libcdio-paranoia-devel libdvdread-devel
  MesaLib-devel harfbuzz-devel libXScrnSaver-devel jack-devel libdvdnav-devel
- wayland-devel libwayland-egl libuuid-devel libguess-devel libXrandr-devel samba-devel
```
Gottox: Why was samba removed?

chneukirchen:
In any case, samba makes mpv GPL-3.

Is there a single user that needs the samba dependency?

srcpkgs/mpv/template
cr6git: Samba was already disabled by default... 

`Checking for Samba support (makes mpv GPLv3)                      : disabled`

Gottox:
Don't think so.

Our samba is still gplv2 iirc.

cr6git:
@Gottox GPL-2 ? The [samba template](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/samba/template#L18) is GPL-3.

Gottox:
hmm... true. anyway, the change is fine.

PR #7085, opened 2017-07-19
emerald: update to 0.8.14

CoolOhm:
This is part 2 of 3 for the compiz upgrade.

compiz core packages. [done]
emerald packages []
compiz-plugin-* packages []


CoolOhm:
I am now getting a error:
"configure: error: cannot run test program while cross compiling"

I am closing the PR while  I look into it.

PR #7095, opened 2017-07-20
intellij-idea-community-edition: move to nonfree

### LINKED ISSUE #7094: IntelliJ IDEA might have to be moved to nonfree
While IntelliJ IDEA itself is under Apache 2.0, while working on a package for PyCharm I have discovered that most libraries used by IDEA have different licenses, some of which might be considered "nonfree": in particular, OROMatcher seems to be the primary culprit, but I'm also suspicious of the licenses for SQLJet and SVNKit (though these two are only used by the "svn4idea" plugin, which could be packaged separately).

[License in question](https://github.com/JetBrains/intellij-community/blob/master/license/oromatcher_license.txt)

This might, however, be fixable on the IDEA side - a successor to OROMatcher exists [under Apache 2.0](https://jakarta.apache.org/oro/).

Gottox:
fixes #7094


asiekierka:
Mind if I work on this one? I'd also like to update to the latest version along the way.

asiekierka:
https://github.com/voidlinux/void-packages/pull/7100

Gottox:
fixed by #7100

PR #7101, opened 2017-07-21
vimb: update to 3.0alpha

cr6git:


cr6git:
I understand you do not accept alpha release, sorry, forget about it.

PR #7114, opened 2017-07-22
base-files: Make vkpurge check for root privileges when using rm option

ColdPhoenix00:


Vaelatern:
Looks okay to me. Anybody else?

PR #7118, opened 2017-07-23
rclone: update to 1.37

cr6git:


cr6git:
See [#7119](https://github.com/voidlinux/void-packages/pull/7119), @diogoleal himself updated his package. 
I prefer to close this PR then.

Issue #6258, opened 2017-04-24
Package Request: anbox


ivierlin:
https://github.com/anbox/anbox

Duncaen:
0 releases.

PR #7120, opened 2017-07-23
New package: anbox-0.0.0.20170714

thypon:
This package provides the possibility to run android application natively on Linux.

The applications runs isolated using container technology (ala docker).

This PR also includes the config needed by the kernel to enable these apps to work inside VoidLinux.


thypon:
solves #6258

chneukirchen:
I'd strongly prefer to build ashmem and binder using dkms (as in anbox upstream), instead of polluting our distro kernels with this stuff.

chneukirchen:
I'd strongly prefer to build ashmem and binder using dkms (as in anbox upstream), instead of polluting our distro kernels with this stuff.

thypon:
Polluting: I LOLled. I don't think we can consider 4000 self contained sloc pollution in a 20 million sloc Linux kernel codebase. But you should ask the kernel guys that thought android binder should be merged in mainline kernel, maybe.

Btw it is not possible to pursue what you proposed since anbox binder/ashmem are only synced with latest ubuntu supported kernel and they don't even compile on recent kernels. Otherwise, one possibility is a Linux mainline patch to made these modules loadable, or maybe you could simple `rm /dev/binder /dev/ashmem` if you consider them harmful.

Gottox:
How big are those modules?
dkms is only a workaround for licencing issues. imho it's not the right tool to split up 'bloatness'.

thypon:
They are not modules; they need to be compiled into the kernel image. The anbox modules are modified to compile outside the main kernel image.

thypon:
Updates?

Issue #6911, opened 2017-07-01
Can't delete first character on tty login


cr6git:
I think it's directly related to void-mklive [#114](https://github.com/voidlinux/void-mklive/issues/114). 
Steps to reproduce the bug: 
1. Boot Void linux and wait for the tty login prompt. 
2. Type some characters 
3. Hit backspace 

I can't delete the **first** character. 
The problem disappears on second login attempt after that.

cr6git:
This issue will be fixed [soon](https://github.com/karelzak/util-linux/issues/454#issuecomment-310658081). 😃

cr6git:
Fixed now in util-linux-2.30.1 !

PR #7122, opened 2017-07-23
util-linux: update to 2.30.1

cr6git:
I can confirm this new version solves the tty bug [#6911](https://github.com/voidlinux/void-packages/issues/6911).


Gottox:
fixed by 8244076b10e1a2b77263e63023369a385465ba70

PR #7124, opened 2017-07-23
Add dbus as qutebrowser dependency

arcadius3d:
Apparently qutebrowser fails to load and gives an error saying that D-bus library appears to be incorrectly set up. Installing dbus fixed the issue for me.


Gottox:
Looks okay, please increase the revision and ping me afterwards. I'll merge

PR #7125, opened 2017-07-23
yarn: update to 0.28.4

kulinacs:


srcpkgs/yarn-bin/template
```diff
@@ -1,15 +1,15 @@
 # Template file for 'yarn-bin'
 pkgname=yarn-bin
-version=0.27.5
-revision=1
+version=0.28.4
+revision=2
```
Gottox: set `revision=1` when you update the version.

kulinacs: fixed.

PR #7126, opened 2017-07-24
thc-hydra: update to 8.6.

kulinacs:


cr6git:
````
version=8.6
revision=3
````

Why `revision=3` ? It should be `revision=1`.

Vaelatern:
Doesn't seem to respect `$(CC)`. Probably can actually be cross compiled once this is fixed.
Doesn't seem to know where to find `int32_t` on musl, probably due to missing header file and a glibc-ism.
I'd like to see the above fixed, including a fixed revision number. Revision resets on version update.

ebfe:
6864a65edb96e98f2af9f7540bef4ecd214554d3

PR #7132, opened 2017-07-24
emerald: update to 0.8.14

CoolOhm:
This is part 3 of 3 for the compiz upgrade.

compiz core packages. [done]
compiz-plugin-* packages [done]
emerald packages []

Note: The lovely folks over at Compiz-Reloaded  have made it so emerald will not cross compile because of ... interesting.... decisions about their configure.ac file. I have applied a patch that fixes the issue and allows for cross compiling.


Gottox:
> Note: The lovely folks over at Compiz-Reloaded have made it so emerald will not cross compile because of ... interesting.... decisions about their configure.ac file. I have applied a patch that fixes the issue and allows for cross compiling.

Business as usual 😄

PR #7136, opened 2017-07-25
LMMS: Added musl-compatibility

skurtix:
Additional patches that makes this compile under musl.


srcpkgs/lmms/template
```diff
@@ -1,10 +1,10 @@
 # Template file for 'lmms'
 pkgname=lmms
 version=1.1.90
-revision=1
+revision=2
 
 build_style=cmake
-only_for_archs="x86_64 i686"
+only_for_archs="x86_64 x86_64-musl i686 i686-musl"
```
Vaelatern: Actually you can just remove this line completely.

Vaelatern:
I don't have the time necessary to vett the musl patch, sorry.

PR #7139, opened 2017-07-25
linux*: disable intel_iommu by default     

ebfe:
Various machines with integrated intel graphics hang randomly or lock up on
mode switching unless this is disabled with intel_iommu=off or
intel_iommu=igfx_off.


cr6git:
What happens to your PR @ebfe ? Your changes have not been validated ?

ebfe:
#7150

cr6git:
Interesting, different settings for different distros: 

Void kernel: `CONFIG_INTEL_IOMMU_DEFAULT_ON=y`

Arch kernel: `CONFIG_INTEL_IOMMU_DEFAULT_ON is not set` [#n6848](https://git.archlinux.org/svntogit/packages.git/tree/trunk/config.x86_64?h=packages/linux#n6848) 

Alpine kernel: `CONFIG_INTEL_IOMMU_DEFAULT_ON is not set` [#n5920](https://git.alpinelinux.org/cgit/aports/tree/main/linux-hardened/config-hardened.x86_64#n5920)

ebfe:
Yes, that's why void doesn't boot or hangs on some machine and everything else doesn't.

chneukirchen:
Also off on Debian.

I'm fine with disabling it.

thypon:
DEFAULT_ON disabled doesn't mean OFF, but IGFX_OFF

PR #7144, opened 2017-07-25
New package: hunspell-pl_PL-20170717

m-cz:


srcpkgs/hunspell-pl_PL/template
```diff
@@ -0,0 +1,23 @@
...
+noarch=yes
+create_wrksrc=yes
+
+post_extract() {
+	unzip pl_PL.zip
+}
+
+do_install() {
+	vinstall pl_PL.aff 644 /usr/share/hunspell/
+	vinstall pl_PL.dic 644 /usr/share/hunspell/
```
chneukirchen: vinstall doesnt use initial `/` usually.

Else :+1:

PR #7147, opened 2017-07-26
cmake: update to 3.9.2.

ebfe:
Someone wants to run a bulk build for all build_style=cmake packages?


ebfe:
3.9.2

ebfe:
build failure is travis only

PR #4875, opened 2016-10-02
New package: budgie-desktop-10.2.7

mar1n3r0:
The Budgie Desktop is the flagship desktop of the Solus Operating System.

PR #7148, opened 2017-07-26
nnn: update to 1.3

cr6git:


cr6git:
Once again I've made a mistake, sorry.

cr6git:
Thanks to @pullmoll and his advice [here](https://github.com/voidlinux/void-packages/pull/4875#issuecomment-252489270) , I've made some progress with git. :)

PR #7150, opened 2017-07-26
grub: disable iommu on integrated graphical card in order to support problematic CPUs/mobo combo

thypon:


asergi:
To recap:
- our kernels have intel_iommu enabled by default
- void doesn't boot at all on some machines due to above
- we disable it by default via grub cmdline

This solution doesn't make sense, imo. What about implement #7139 instead?

thypon:
Intel Iommu enabled by default is ok, for the machines that already has in place. It is better to opt-out by grub than disable the protection for all the other machines (already running) that work like a charm with intel_iommu=on.

asergi:
Adding things to grub is not the solution. It's not our job, is a user job. Stop adding things to the grub cmdline by default.

-1 to this, +1 to @ebfe's PR.

thypon:
@asergi `slub_debug=P page_poison=1` have to be passed by command line.

Setting `GRUB_CMDLINE_LINUX_DEFAULT` is distribution job. Setting `GRUB_CMDLINE_LINUX` is user job.

asergi:
> Setting `GRUB_CMDLINE_LINUX_DEFAULT` is distribution job. Setting `GRUB_CMDLINE_LINUX` is user job.

Not for just a workaround.

Again, this is not the solution. -1.

thypon:
Closing this as we didn't reach an arrangement.

PR #7153, opened 2017-07-26
Update Bullet to 2.86.1

Johnnynator:


Johnnynator:
It probably should be fine now, Travs fails because the mirror that gets used by Travis doesn't have  gstreamer1-1.12.3 and Travis doesn't like to build gstreamer.

PR #7133, opened 2017-07-24
linux: update to 4.12.

Gottox:
Updates the default kernel to 4.12.

Before merging check:
* [x] nvidia
* [x] nvidia340
* [x] acpi_call-dkms-1.2.0_2
* [ ] batman-adv-2016.4_1
* [ ] batman-adv14-2013.4.0_8
* [ ] bbswitch-0.8_2
* [x] exfat-dkms-1.2.8_2
* [ ] spl-0.6.5.10_1 See #7158 
* [x] sysdig-0.17.0_1
* [x] tp_smapi-dkms-0.42_2
* [x] virtualbox-ose-dkms-5.1.24_1
* [x] virtualbox-ose-guest-dkms-5.1.24_1
* [x] wireguard-0.0.20170706_1
* [ ] zfs-0.6.5.10_1 See #7158

PR #7155, opened 2017-07-27
exfat-dkms: bump commit

Sir-Boops:
Else compiling will fail for linux4.12 ( Tested on x86_64{-musl} )


Sir-Boops:
Should also say that this is the fix for #7133

PR #7157, opened 2017-07-27
sssd: update to 1.15.3.

xYuusha:


srcpkgs/tevent/files/cross-32bit.answers
```diff
@@ -0,0 +1,46 @@
+Checking uname sysname type: "Linux"
```
Gottox: Is this file used anywhere?

xYuusha: Sorry, I was testing some cross compilation and i accidentaly staged it. Is removed now

PR #7158, opened 2017-07-27
zfs: update to 0.7.0.

Gottox:
fixes zfs for #7133


ColdPhoenix00:
@Gottox I just fixed build errors on arm by adding these lines to zfs template:

```bash
CFLAGS='-fPIC -I=/include/tirpc'
LDFLAGS='-fPIC -fPIE'
```

emacsomancer:
Is there still an issue with this package?  I see the last attempt at a Travis build was over a month ago. Does @ColdPhoenix00's patch fix the musl issue?

Vaelatern:
I'm testing this possible fix.

emacsomancer:
Shouldn't there be a `depends` or `conflicts` which stops this from being installed concurrently with >linux4.12? ( https://github.com/zfsonlinux/zfs/releases/tag/zfs-0.7.1 ) [I know Void hasn't added 4.13 yet, but given that 4.12 must be nearing EOL, this presumably will come soon.]

Vaelatern:
The problem is that you may want a different linux than the linux metapackage.
There is no clean way to do this.

emacsomancer:
I see. So will linux4.13 be held until zfs is updated, do you know?

PR #7164, opened 2017-07-27
chromium: update to 61.0.3163.91.

ebfe:


ebfe:
musl still wonky, maybe something to do with the stack size estimator code... also something wrong with the shutdown_signal_handler thread on glibc .

ebfe:
> also something wrong with the shutdown_signal_handler thread on glibc .

also happens with google-chrome so i'd assume that one isn't our fault.

Duncaen:
> also happens with google-chrome so i'd assume that one isn't our fault.

Yes this happens for some time already

ebfe:
Updated to chromium 60.0.3112.90, on musl tabs still crash after a while.

ebfe:
glibc build broken by the glibc-2.26 update.

chneukirchen:
Does that help? https://github.com/OSSystems/meta-browser/blob/master/recipes-browser/chromium/files/0002-replace-struct-ucontext-with-ucontext_t.patch

ebfe:
update to 61.0.3163.91, glibc builds no energy left for musl for now.

ncopa:
this is the patch you need for musl: https://github.com/lluixhi/musl-extras/issues/10#issuecomment-328272349

ebfe:
@ncopa thanks. musl builds now too, otherwise untested.

ebfe:
81a77d6f5f7c3256f65fa29e47087fcbf05e5f69

PR #7168, opened 2017-07-28
sndio: don't listen on all interfaces by default

ebfe:


ebfe:
Hmpf, this doesn't work well since the permissions on the unix socket are too strict.

chneukirchen:
Perhaps this still makes sense, can't we make the socket group audio or something? (or group sndio)

ebfe:
> Perhaps this still makes sense, can't we make the socket group audio or something? (or group sndio)

It's not only permissions:

sndiod creates the socket in `/tmp/aucat-<uid>/aucat0` or
`tmp/aucat/aucat0` when run as root.

libsndio looks up the socket in `/tmp/aucat-<uid>/aucat0` `/tmp/aucat/aucat0`

We run sndiod as sndiod:audio, libsndio never even tries to connect to
it when run from a different user.

chneukirchen:
Hm, on OpenBSD, sndiod by default has a socket at `/tmp/aucat/aucat0` which is root:wheel and `srw-rw-rw-`.

chneukirchen:
> Connecting to the socket object requires read/write permission. This behavior differs from many BSD- derived systems which ignore permissions for UNIX domain sockets.

ebfe:
Dropping the chpst and letting sndiod drop the privileges itself makes sndiod
listen on /tmp/aucat/aucat0 world-rw. Connect from aucat succeeds but connection is dropped after the "hello" and no audio transferred through it. May a problem with the $HOME/.aucat_cookie...

ebfe:
Ok the problem is that sndiod when run as root switches to 'sndiod:sndiod' which doesn't have perms to access the /dev/snd/* devices...

ebfe:
No idea how we could solve this without patching. We need to either change the group or the socket path in sndiod.

ebfe:
`sndiod_groups="audio"` also doesn't work due to:
```
sndiod/sndiod.c:513:            if (setgroups(1, &pw->pw_gid) ||
```

ebfe:
Running sndiod with `-a` solves the sndiod user can't open alsa devices since it causes it to open it before dropping privs and keep it open all the. But this might not play well with everything else?

ebfe:
I think i misinterpreted the groups issue, and the problem was something else having the alsa device open. Adding sndiod to the users group and dropping the chpst seems to work. https://github.com/ebfe/void-packages/commit/c4733d4f20337ef312b0cfca2c5722125664a212

chneukirchen:
Alsa should use dmix by default, no?

ebfe:
I don't have any custom alsa config and as long and sndiod has the device open other applications fail with EBUSY

xYuusha:
I copied ebfe's template and the service returned me 'failed to open audio device'.

As far as i could understood, running sndiod as root drops privileges to sndiod:sndiod which does not have permission to read from the device, and the line "sndiod_groups" adds the audio group as a secondary group.

I solved that changing sndiod_groups to sndiod_pgroup and then it works fine.

Issue #6840, opened 2017-06-25
base-files: /usr/lib/sysctl.d/void.conf permanently disables kexec


ebfe:
```
[...]
# Turn off kexec, even if it's built in.
kernel.kexec_load_disabled=1
[...]
```

It's impossible to reenable once set. Either `void.conf` should be a `conf_file`, so it doesn't get overwritten on package updates, or the setting moved somewhere else.

the-maldridge:
I think the solution that had been discussed some time ago was to adjust the load order so that this could be overwritten later in the boot process.  Is that not an acceptable solution?

ebfe:
> I think the solution that had been discussed some time ago was to adjust the load order so that this could be overwritten later in the boot process.  Is that not an acceptable solution?

It's not a solution at all, because you can't reenable kexec once disabled.

chneukirchen:
If we used sysctl properly, you could just create a void.conf file in /etc?

xtraeme:
Files with the same filename in /etc have preference over the ones in
/usr/share

ebfe:
In theory yes, but `/etc/runit/core-services/08-sysctl.sh` loads files individually and not via `--system`

ebfe:
This commit dropped `--system` https://github.com/voidlinux/void-runit/commit/5f719f131c3d75a99f1a09d4cd7ccccc1534a100, apparently for busybox compat.

amak79:
> If we used sysctl properly, you could just create a void.conf file in /etc?

There is open issue regarding the order of sysctl file loading: https://github.com/voidlinux/void-runit/issues/70

the-maldridge:
I am reasonably sure that our current version of busybox cannot bring up the system.  I know for sure it is insufficient to run the install environment.  If that is truly the case, I am in favor of backing out the commit that dropped `--system`.

ebfe:
ping

PR #7170, opened 2017-07-28
base-files: Use IPv6 privacy extensions by default

Duncaen:


ebfe:
Can we put these things in a `conf_file`? #6840 etc.

the-maldridge:
This will break my network.  I would like to see the sysctl loading issue fixed before this is merged so it is possible to override back to what I had before.

cr6git:
This is not already set by dhcpcd ? 

```
# /etc/dhcpcd.conf
# Generate Stable Private IPv6 Addresses instead of hardware based ones
slaac private
```

cr6git:
In fact [RFC4941](https://tools.ietf.org/html/rfc4941) and [RFC7217](https://tools.ietf.org/html/rfc7217) are directly related.

lemmi:
I'm just going to mention that I had a lot of trouble using privacy extensions on networks with changing prefixes (as it is common for Germany).
Essentially what happens:
- On a prefix change, new temporary addresses will be generated for an interface. These are the newest addresses, so they will be used.
- Old addresses are kept for a while for open connections.
- After an address runs out of its lifetime, a new address is generated. 
- If that happens to be an address of the old prefix, this one will be the default address and now.
- bam

lemmi:
@voidlinux/pkg-committers can we close this?

the-maldridge:
Working as intended.

PR #7195, opened 2017-07-29
vim: update to 8.0.0823

cr6git:


chneukirchen:
It would be nice if one of the next vim bumps would enable the new builtin terminal emulator.

Gottox:
@chneukirchen cputime is cheap. just bump it :)

PR #7196, opened 2017-07-29
Luxrender

Piraty:
powerful renderer,
see http://www.luxrender.net/.

I tested it for example in combination with freeCAD, which has the ability to export 3D-models to this render engine.

during compilation i noticed that luxrays was imcomplete and did not include opencl-support.


cr6git:
Your `build_options=` are a mess.

Piraty:
That comment wasn't helpful, can you be more specific?

srcpkgs/luxrender/template
```diff
@@ -0,0 +1,38 @@
+# Template file for 'luxrender'
+pkgname=luxrender
+version=1.6
+revision=1
+_version=b3f85cf7742f
+wrksrc=luxrender-lux-${_version}
+build_style=cmake
+hostmakedepends="bison flex"
+makedepends="luxrays-devel boost-devel openimageio-devel tiff-devel libpng-devel libopenexr-devel fftw-devel embree-devel libgomp-devel MesaLib-devel glu-devel opencl-headers opencl-clhpp"
+nopie=yes
```
Gottox: Are you perfectly sure this is needed? Most of the times this is just an indicator, that our CFLAGS aren't applied correctly.

srcpkgs/luxrender/template
```diff
@@ -0,0 +1,38 @@
...
+distfiles="https://bitbucket.org/luxrender/lux/get/v${version/./}.tar.gz"
+checksum=9c4d21efbd9395fd31d528ea6e6272d29cd895be6fef2d901909786a0d39dd53
+
+post_install() {
+	if [ -d ${DESTDIR}/usr/lib64 ]; then
+		mkdir -p ${DESTDIR}/usr/lib/
+		mv ${DESTDIR}/usr/lib64/* ${DESTDIR}/usr/lib/
+		rm -r ${DESTDIR}/usr/lib64
+	fi
+}
```
Gottox: Move this function to the bottom.
* Definitions at the top
* buildinstructions at the bottom

srcpkgs/luxrender/template
```diff
@@ -0,0 +1,38 @@
...
+
+build_options_default="python qt"
```
Gottox: For the sake of readability, first define all build_options in one definition (`build_option="x y z"`) then add the default options (`build_options_default="x y z"`)

srcpkgs/luxrender/template
```diff
@@ -0,0 +1,38 @@
...
+build_options=" python"
+if [ "$build_option_python" ]; then
+	makedepends+=" python-devel"
+fi
+
+# luxrender executable won't be build without qt
+build_options+=" qt"
+if [ "$build_option_qt" ]; then
+	makedepends+=" qt-devel"
+fi
```
Gottox: Don't branch for build_options. Instead use `$(vopt_if qt qt-devel)`

Piraty:
@Gottox Done. I tested the new build again with rendering some models in FreeCAD, works good.

srcpkgs/luxrender/template
Piraty: done

srcpkgs/luxrays/template
```diff
@@ -1,10 +1,10 @@
 # Template file for 'luxrays'
 pkgname=luxrays
 version=3.1
-revision=6
+revision=7
 build_style=cmake
 hostmakedepends="flex bison boost-python"
-makedepends="openimageio-devel tiff-devel libpng-devel libgomp-devel boost-devel glew-devel opencl-headers libopenexr-devel python-devel libfreeglut-devel embree-devel"
+makedepends="openimageio-devel tiff-devel libpng-devel libgomp-devel boost-devel glew-devel opencl-headers opencl-clhpp libopenexr-devel python-devel libfreeglut-devel embree-devel"
```
the-maldridge: This line should be wrapped at 80 columns.

srcpkgs/luxrender/template
```diff
@@ -0,0 +1,32 @@
+# Template file for 'luxrender'
+pkgname=luxrender
+version=1.6
+revision=1
+_commit=b3f85cf7742f
+wrksrc=luxrender-lux-${_commit}
+build_style=cmake
+hostmakedepends="bison flex"
+makedepends="luxrays-devel boost-devel openimageio-devel tiff-devel libpng-devel libopenexr-devel fftw-devel embree-devel libgomp-devel MesaLib-devel glu-devel opencl-headers opencl-clhpp $(vopt_if qt qt-devel) $(vopt_if python3 python3-devel)"
```
the-maldridge: Wrap at 80 columns.

Piraty:
done @the-maldridge

Piraty:
This is currently broken, I'll have yet to decide if i revoke this PR or patch the luxrender

Piraty:
in order to continue, we need boost to be python3 compatible.

PR #7209, opened 2017-07-30
mdocml: add patch to fix segfault on `man mount`

pbui:
This adds a patch that fixes a segfault when the user performs `man mount`.  The cause of the segfault is an out of bounds array access, so the patch simply performs a check to ensure we only access items with the boundary.


chneukirchen:
I just reported this upstream and will give them a bit of time to deliver a patch of their own.

chneukirchen:
0addb10

Issue #7211, opened 2017-07-30
lxappearance depends on dbus


cr6git:
The package [lxappearance](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/lxappearance/template) actually depends on dbus : 
````
$ lxappearance
process 849: D-Bus library appears to be incorrectly set up; failed to read machine uuid: Failed to open "/etc/machine-id": Aucun fichier ou dossier de ce type
See the manual page for dbus-uuidgen to correct this issue.
  D-Bus not built with -rdynamic so unable to print a backtrace
Abandon
````

First option: 
`configure_args="--disable-dbus"`

Second option: 
`depends="dbus"`

PR #7214, opened 2017-07-31
lxappearance: disable dbus

cr6git:
To close [#7211](https://github.com/voidlinux/void-packages/issues/7211). It works perfectly without dbus.


cr6git:
git is overly complex and I've made mistakes, again... 
I give up, good night.

PR #7219, opened 2017-07-31
more: replace with less

tetero:
Given the recent breakage of mdocml due to the post_extract sed -i 's,"more -s","less -s",g' main.c and the likely future breakage from patching its source to use more, I see it fit to replace
more with less. Less has a more-compatibility mode (LESS_IS_MORE=1) which is automatically invoked if called from a link named 'more'.

This stops building more, and removes and replaces it with a hardlink to less.

### DIFF
--- a/srcpkgs/less/template
+++ b/srcpkgs/less/template
@@ -14,5 +14,5 @@ checksum=02a476dccd12e1b76e5d91259cba5dac1b8c8e5a730465c1c89c798b19741509
 
 post_install() {
 	rm -f /usr/bin/more
-	ln /usr/bin/less /usr/bin/more
+	ln -s ${DESTDIR}/usr/bin/less /usr/bin/more
 }
--- a/srcpkgs/less/template
+++ b/srcpkgs/less/template
@@ -1,7 +1,7 @@
 # Template file for 'less'
 pkgname=less
 version=494
-revision=1
+revision=2
 build_style=gnu-configure
 configure_args="--with-regex=pcre"
 makedepends="ncurses-devel pcre-devel"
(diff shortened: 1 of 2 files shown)


Vaelatern:
Obviously only one of these is POSIX. And it isn't less(1). http://pubs.opengroup.org/onlinepubs/9699919799/utilities/more.html
I get that you replace more with a symlink to less. This would necessarily require modifications to the base-system family, so systems can have a pager installed.

I am (not strongly) against this change. Not for reasons of size (our less package is just 266 kb, while `util-linux` is a whopping 11 megs), but because of the precedent set by using `dash` instead of `bash` with compatibility mode.

Vaelatern:
Before any of your changes can be considered, you'd need to follow proper commit messages. And fix that symlink `ln -s less ${DESTDIR}/usr/bin/more`.

skurtix:
Done. The build fails because the URL to less-494 404's. Looking at the site the latest version is 487, which makes me wonder why it's 494 in the repo?

Gottox:
looks okay. @Vaelatern for you too?

Gottox:
Is there any software that depends on more?

tetero:
None. In the off chance something even uses more, the symlink from more to less will put less in LESS_IS_MORE=1 compatibility mode allowing it to work with applications that might use it.

Gottox:
@Vaelatern can you approve or give any feedback?

Duncaen:
I don't like this change, lets just keep more. Less in more mode is **more** use**less** than more.

```
~$ du -h /usr/bin/{less,more}
660K    /usr/bin/less
40.0K   /usr/bin/more
```

This change doesn't improve anything and the reason in the first comment is wrong.

tetero:
Seeing as more isn't used for anything, or by anyone, I don't see why not to remove it. The fact that it's 40K isn't really much of an argument. What, exactly, is it that you think makes LESS_IS_MORE mode useless?

ayghor:
`less` will behave liek `more` wat is teh benefit of this change?

the-maldridge:
Closing as there is substantiated opposition to altering the base system.  The standard way to do this is to set the PAGER for your environment.

PR #6560, opened 2017-05-31
vidcutter: update to 3.5.0

cr6git:

PR #7222, opened 2017-07-31
opencv: enable openblas support

juliendehos:
"OpenCV now can use vendor-provided OpenVX and LAPACK/BLAS (including Intel MKL, Apple’s Accelerate, OpenBLAS and Atlas) for acceleration" (http://opencv.org/opencv-3-2.html)

I've run some tests and obtained a x8 speed-up.

### DIFF
--- a/srcpkgs/opencv/template
+++ b/srcpkgs/opencv/template
@@ -1,13 +1,13 @@
 # Template file for 'opencv'
 pkgname=opencv
 version=3.2.0
-revision=2
+revision=3
 build_style=cmake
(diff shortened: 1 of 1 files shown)


cr6git:
bug in Travis ?

NeoChontrix:
Looks so, error one is caused by Travis not being able to resolve the address and the 2nd error is Travis not being able to install a simple xbps package.

cr6git:
I remember having exactly the same problem on my PR [#6560](https://github.com/voidlinux/void-packages/pull/6560) ...

PR #7224, opened 2017-07-31
ppsspp: update to 1.4.2.

asiekierka:


Gottox:
dublicate of #7778, closing this as the other had no build issues with musl.

PR #7225, opened 2017-07-31
New package: ion-1.0.5

konimex:


srcpkgs/ion/template
```diff
@@ -0,0 +1,22 @@
+# Template file for 'ion'
+pkgname=ion
+version=1.0.5
+revision=1
+hostmakedepends="cargo rust"
+register_shell="/bin/ion"
+short_desc="The Ion Shell, written in Rust"
+maintainer="Muhammad Herdiansyah <herdiansyah@opmbx.org>"
+license="MIT"
+only_for_archs="i686 x86_64 i686-musl x86_64-musl"
```
Gottox: why shouldn't it run on arm?

konimex: The only reason is stripping binaries doesn't work in ARM. Should I use `nostrip=yes` instead?

PR #7233, opened 2017-08-01
New package: monero-0.10.3.1

Epictek:
WIP

### DIFF
--- a/srcpkgs/monero/template
+++ b/srcpkgs/monero/template
@@ -1,0 +1,14 @@
+# Template file for 'monero'
+pkgname=monero
+version=0.10.3.1
+revision=1
+build_style=cmake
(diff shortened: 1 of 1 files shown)


Epictek:
musl build is failing, seems like some patches are required.
https://git.alpinelinux.org/cgit/aports/tree/testing/monero

Cross compiling should be an easy fix by passing the arch to cmake

Will fix it later today, I need some sleep first

Epictek:
Should I include a service file similar to in the AUR repo?
https://aur.archlinux.org/cgit/aur.git/tree/monerod.service?h=monero

lemmi:
#7779 seems to be a newer and seems to build with less issues. I'm sorry this got burried, but if you want to add anything to the other PR, pleaso do so.

PR #7234, opened 2017-08-01
armadillo: update to 7.960.0

juliendehos:
- update to 7.960.0
- enable HDF5 support
- improve openblas wrapping


juliendehos:
Ok, fixed. I'm not an expert with bulid_option so let me know if I did something wrong.

PR #7241, opened 2017-08-02
New package: notepadqq

cr6git:
Template based on the Arch [PKGBUILD](https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/notepadqq).

### DIFF
--- a/srcpkgs/notepadqq/template
+++ b/srcpkgs/notepadqq/template
@@ -1,0 +1,28 @@
+# Template file for 'notepadqq'
+pkgname=notepadqq
+version=1.0.1
+revision=1
+_Code_commit=d790fc39c1a5f06aa66415110b8ebe3026df665a
+build_style=qmake
+hostmakedepends="pkg-config qt5-tools"
+makedepends="qt5-svg-devel qt5-webkit-devel"
+depends="desktop-file-utils hicolor-icon-theme"
+maintainer="cr6git <quark6@protonmail.com>"
+short_desc="Text editor for developers"
+homepage="http://notepadqq.altervista.org/s/"
+license="GPL-3"
+distfiles="https://github.com/notepadqq/notepadqq/archive/v${version}.tar.gz
+	https://github.com/notepadqq/CodeMirror/archive/${_Code_commit}.tar.gz"
+checksum="552f274b9c9d3d48e2e55462f33aea8df8ebd0f421b0d7a23b78a46693fbed34
+	f3b23a367f9070d19bbc3d2d48dab4839b436ada15f2b26f1b4a7a599883642c"
+create_wrksrc=yes
+build_wrksrc="notepadqq-${version}"
+
+if [ -n "$CROSS_BUILD" ]; then
+	hostmakedepends+=" qt5-svg-devel qt5-webkit-devel"
+fi
+
+post_extract() {
+	mv -Tv CodeMirror-${_Code_commit} ${build_wrksrc}/src/editor/libs/codemirror
+}
+


srcpkgs/notepadqq/template
```diff
@@ -0,0 +1,46 @@
...
+
+do_fetch() {
+	mkdir -p ${wrksrc} && cd ${wrksrc}
+	printf "\033[1;35m=> Fetching source from ${_notepadqq} ...\033[m\n"
+	git clone ${_notepadqq}
+	cd notepadqq && git reset --hard ${_release}
+	cd ..
+	printf "\033[1;35m=> Fetching source from ${_CodeMirror} ...\033[m\n"
+	git clone ${_CodeMirror}
+}
```
Gottox: Why not using the provided tarballs?

cr6git: Using the tarballs the git commands failed: 

pre_configure() {
	git config submodule.src/editor/libs/codemirror.url ../CodeMirror
	git submodule update --init
}

ebfe: You can fetch the corresponding codemirror tarball too and extract it manually

cr6git: Using the tarballs it fails: 
````
=> notepadqq-1.0.1_1: running pre_configure ...
/builddir/notepadqq-1.0.1/notepadqq-1.0.1
fatal: not in a git directory
````
Do you have a better idea ?

ebfe: Don't run git submodule stuff, fetch the corresponding codemirror tarball and extract it.

cr6git: Without git submodule stuff I get new errors: see [the log](https://bpaste.net/show/22af9bb9be34) ...

ebfe: The submodule lives at `src/editor/libs/codemirror` so you need to extract it there.
See https://github.com/notepadqq/notepadqq/blob/master/.gitmodules

cr6git: OK. I created this [new template](https://bpaste.net/show/56ea28cb10c8), moving CodeMirror to the right folder, but I get  new errors: see my [new log](https://bpaste.net/show/3ddc967a7683) ...

Gottox: Don't branch for hostmakedepends when on crossbuild! It's only valid in a few cornercases!

ebfe: @cr6git just use the revision referenced by the submodule 
`_Code_release=d790fc39c1a5f06aa66415110b8ebe3026df665a`

cr6git: @ebfe Exact!! this commit solves the problem, now I can build the package for my current arch x86_64, using [this template](https://bpaste.net/show/812beb762081). 
@Gottox I can't find another solution regarding cross-compilation.

cr6git: What should I do ? new git push with these changes ?

Gottox: just remove the branch around hostmakedepends.

cr6git: Done.

cr6git:
@Gottox I hope the changes are what you want.

cr6git:
@Gottox I removed `qt5-tools-devel` from this template, I think it's better.

cr6git:
I think this new template is cleaner.

cr6git:
Is there a reason why this PR is still blocked ?

Issue #7004, opened 2017-07-09
Package request: jgmenu


paoloschi:
Small X11 menu intended to be used with openbox and tint2 
<https://github.com/johanmalm/jgmenu>

Key Features
- It is a stand-alone, simple and contemporary-looking menu
- It is hackable with a clean, small code base
- Although written to be used with tint2 and openbox, it runs with other panels and window managers
- It has a config file to set alignment, margins, padding, transparency, etc
- It can synchronise with tint2's settings
- It does not depend on any toolkits such as GTK and Qt
- It uses cairo and pango to render the menu directly onto an X11 window
- It reads menu items from stdin in a similar way to dmenu and dzen2, but parses comma seperated fields for the name, command and icon

cr6git:
Today I created this [template](https://bpaste.net/show/742b767e0cac) for **jgmenu** , but I got errors during the build, see the [log](https://bpaste.net/show/34f5da6ae647).

paoloschi:
When I compiled it in Void in the traditional way (without xbps-src template) just for testing this sw, I've ran into the same issue which, if I remember correctly, I've solved by installing the package:

    pkg-config-0.29.2_1    System for managing library compile/link flags

cr6git:
@paoloschi Following your advice, I added `hostmakedepends="pkg-config"` in my template, but I got exactly the same error.

paoloschi:
I can assure you: the source compile in Void (x64, glibc)  without errors or even warnings in a couple of seconds!
After digging into my _.bash_history_ file, I am now able to tell you with certainty the commands that I ran over 3 weeks ago...

Looking the Void matches for the package names listed in the compilation guide for Debian, I have installed
```
libX11-devel libXinerama-devel cairo-devel pango-devel librsvg-devel libxml2-devel libglib-devel menu-cache-devel
```
at this point, running _make_ I've got your own same issue `fatal error: cairo.h: No such file or directory #include <cairo.h>`

Later, reading further, I came to the conclusion that I missed the `pkg-config` package.
I then tried again..
```
make
make install
```
..and two seconds later I had the `jgmenu` binary in my `~/bin` directory, working perfectly..
This way it went and more than so I would not know how to help you

johanmalm:
I'm not familiar with the Void Linux packaging system.

@cr6git, does a manual `git clone` + `make` work?

What does `pkg-config cairo --cflags` give you?

Does it include the directory where cairo.h is? Try `locate cairo.h`

[...]

PR #7242, opened 2017-08-02
New package: jgmenu

cr6git:
Closes [#7004](https://github.com/voidlinux/void-packages/issues/7004).


srcpkgs/jgmenu/template
```diff
@@ -0,0 +1,17 @@
...
+pkgname=jgmenu
+version=0.6.1
+revision=1
+build_style=gnu-makefile
+make_use_env=yes
+make_install_args="prefix=/usr"
+hostmakedepends="pkg-config python3"
+makedepends="cairo-devel libglib-devel librsvg-devel libxml2-devel
+ libX11-devel libXinerama-devel menu-cache-devel pango-devel"
+depends="python3"
```
ebfe: Where does the python3 dependency come from?

cr6git: You are right! python3 is not needed, I don't remember why I added it.

cr6git:
@ebfe It's ok for you ?

srcpkgs/jgmenu/template
ebfe: There are python 2 scripts in usr/lib/jgmenu/ and a python script that's run during build in scripts/. Sorry, inital version was right.

cr6git: I trust you. I'm glad to see this package accepted.

PR #7253, opened 2017-08-03
Update spice and spice-gtk

Hoshpak:


srcpkgs/spice/template
```diff
@@ -1,7 +1,7 @@
 # Template file for 'spice'
 pkgname=spice
-version=0.13.3
-revision=3
+version=0.13.90
```
Gottox: 0.13.90 is an unstable release. This won't be merged if there's no urgent reason.

Gottox: 0.13.90 is an unstable release. This won't be merged if there's no urgent reason.

Hoshpak: 0.13.3 is a development release as well. Since someone decided to put the development releases in the repository instead of using the stable releases, we might as well use the current development release - or revert that decision and downgrade to a stable release of course. Perhaps one argument in favor of merging would be that it fixes 3 CVEs (just noticed after looking up the release notes).

Gottox: Someone being me, looking at the logs. okay, we're following this update path. 😏

Issue #6797, opened 2017-06-20
Package request: qarma


AcarBurak:
Very functional and simple zenity replacement for a qt5 system like LXQt or a WM with qt5 packages.

https://github.com/luebking/qarma

By

`ln -s /usr/bin/qarma $pkgdir/usr/bin/zenity`

one can use zenity scripts/dialogues without any alteration as well such as this useful [script](https://github.com/stamatisxm/ZumDisks) (among many others) to very easily un/mount drives without any bloat like gvfs or the likes:

![qarma](https://user-images.githubusercontent.com/1426842/27468051-ff341a72-57ef-11e7-96d5-7c4d03047fed.png)

cr6gtx:
I created a [template](https://paste2.org/nMV2Gjsn) to build the latest git version, dated 2017-02-14.

AcarBurak:
Thanks very much.

cr6gtx:
Another [template](https://paste2.org/d4F4yVg4) to build the latest Release [t1](https://github.com/luebking/qarma/releases/tag/t1).

AcarBurak:
Thank you.

cr6git:
@AcarBurak As you can see, my first PR for `qarma` has been ignored... 
Is it worth it to open a new PR ?

AcarBurak:
It's very useful for a Qt-only desktop and now it's a recommended app for LXQt. Void seems more welcoming to Qt apps, so it would fit. Yet, naturally, it's up to you…

cr6git:
ok, let's try again.

AcarBurak:
Thank you.

cr6git:
it's done.

AcarBurak:
Thank you.

PR #7264, opened 2017-08-03
New package: qarma

cr6git:
To close [#6797](https://github.com/voidlinux/void-packages/issues/6797).


cr6git:
Modified template.

PR #7270, opened 2017-08-04
vim: update to 8.0.0858

cr6git:


Gottox:
Can you take https://github.com/voidlinux/void-packages/pull/7195#issuecomment-319348954 in account? Would be great :)

cr6git:
About the new builtin terminal emulator ? It's already enabled.

PR #7273, opened 2017-08-04
azpainter: update to 2.0.6

cr6git:
I copied the patch from [AUR](https://aur.archlinux.org/cgit/aur.git/tree/azpainter-2.0.6-translation.patch?h=azpainter).


srcpkgs/azpainter/template
```diff
@@ -1,19 +1,25 @@
...
-checksum=7c3bc35fff51b3529911e5d6ce6ee2ce565694c17755fe1e6f52e6d9ad0ce839
+homepage="https://osdn.net/projects/azpainter/"
+distfiles="http://osdn.dl.osdn.net/azpainter/67805/azpainter-${version}.tar.bz2"
+checksum=f7f1c0bff9f0fe2fe6864c051a91f6e2b931c527a0e75ecde91170142163a321
 
 CPPFLAGS="-I${XBPS_CROSS_BASE}/usr/include/freetype2"
 
+post_extract() {
+	# Fixing translations
+	patch -Np1 -i ${FILESDIR}/azpainter-2.0.6-translation.patch
```
Gottox: Why not using the patches dir and is there a urgent reason to patch the translation at all?

cr6git: @Gottox You are right, let's drop this patch.

PR #7280, opened 2017-08-04
New package: rtags-2.12

egorenar:


srcpkgs/rtags/template
```diff
@@ -0,0 +1,20 @@
+# Template file for 'rtags'
+pkgname=rtags
+version=2.12
+revision=1
+only_for_archs="i686 x86_64"
```
Gottox: Why that? - only_for_archs means that this package does not make sense to build on certain platforms. It's useful for bootloaders or platform dependant drivers.

If it is broken use nocross="<URL TO BUILDLOG>" or broken="<URL TO BUILDLOG>".

egorenar: hi,

the rates package doesn't build on other platforms because of llvm.
I wasn't sure what it the right way to handle such a situation.

egorenar:
the build fails due to issue #7324

srcpkgs/rtags/template
Gottox: In such cases it's reasonable.

egorenar:
What needs to be done to get it merged ?

PR #7282, opened 2017-08-05
New package: duktape-2.1.1

CMB:
This will be required as a dependency of edbrowse soon.


ebfe:
> This will be required as a dependency of edbrowse soon.

missing a common/shlibs entry then?

CMB:
> missing a common/shlibs entry then?

Good catch, thank you!

PR #7285, opened 2017-08-05
New package: madbomber

cr6git:
Template based on the AUR [PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=madbomber).


srcpkgs/madbomber/template
```diff
@@ -0,0 +1,38 @@
...
+makedepends="SDL-devel SDL_image-devel SDL_mixer-devel"
+maintainer="cr6git <quark6@protonmail.com>"
+short_desc="Clone of Activision's Atari 2600 game Kaboom!"
+homepage="http://www.newbreedsoftware.com/madbomber/"
+license="GPL-2"
+distfiles="ftp://ftp.tuxpaint.org/unix/x/madbomber/src/madbomber-${version}.tar.gz"
+checksum=951ec8cfde3965255ceafa92fdbbbb50a1981b71afcf359c58c7d22dbf122f48
+
+# non-PIE executable found in PIE build: /usr/bin/madbomber
+nopie=yes
```
Gottox: This usually means that our CFLAGS/LDFLAGS are ignored. `nopie` should only be set if it is needed for building.

srcpkgs/madbomber/template
```diff
@@ -0,0 +1,38 @@
...
+
+post_extract() {
+	sed -i 's|BIN_PREFIX=.*|BIN_PREFIX=/usr/bin/|' Makefile
+	sed -i "s|CFLAGS=|CFLAGS=$CFLAGS |" Makefile
+	gendesk -f -n --pkgname=${pkgname} --pkgdesc="${short_desc}" --name="Mad Bomber"
+}
+
+do_build() {
+	make ${makejobs}
```
Gottox: As you already have set `build_style=gnu-makefile` you can remove this function

srcpkgs/madbomber/template
```diff
@@ -0,0 +1,38 @@
...
+short_desc="Clone of Activision's Atari 2600 game Kaboom!"
+homepage="http://www.newbreedsoftware.com/madbomber/"
+license="GPL-2"
+distfiles="ftp://ftp.tuxpaint.org/unix/x/madbomber/src/madbomber-${version}.tar.gz"
+checksum=951ec8cfde3965255ceafa92fdbbbb50a1981b71afcf359c58c7d22dbf122f48
+
+# non-PIE executable found in PIE build: /usr/bin/madbomber
+nopie=yes
+
+post_extract() {
```
Gottox: iirc CFLAGS isn't set during extract phase, rename the function to do_configure(). Maybe this already solves the nopie issue.

cr6git:
without `nopie=yes` it fails :(

Gottox:
Yea, as I said, I guess our CFLAGS or LDFLAGS are ignored.

cr6git:
FLAGS are wrongly configured in the Makefile, but for now I can't find the solution to this problem, sadly.

Vaelatern:
@cr6git can you try a patch to the Makefile in which you change the `=` to `+=` or whatever?

newbluemoon:
I did a quick test and found that the target in the Makefile doesn’t use $LDFLAGS so adding the line

    sed -i 's/\($(CC) $(CFLAGS) madbomber.o -o $(TARGET) $(SDL_LIB) -lm\)/\1 $(LDFLAGS)/' Makefile

to the template did the trick.

cr6git:
Thank you all, but I have no longer time to work on this. 
I must take care of my real life.

Issue #7281, opened 2017-08-05
Package Request: cbatticon


ctucx:
Hello,
i'm wanted to try void-linux but im missing the lovely cbatticon utility!
wish to have a package for it. 

https://github.com/valr/cbatticon

greetings, 
ctucx

cr6git:
@ctucx Here's the GTK2 version, but maybe you prefer GTK3? I don't know ;)

cr6git:
Added GTK3 version! [#7291](https://github.com/voidlinux/void-packages/pull/7291)

cr6git:
@ctucx [Merged](https://github.com/voidlinux/void-packages/commit/52021cb92edc9723637a13af0a1c43d44d43cc95). You can close this ;)

maxice8:
as cr6git mentioned, resolved

@the-maldridge

PR #7290, opened 2017-08-06
New package: cbatticon-gtk2

cr6git:
Package Request [#7281](https://github.com/voidlinux/void-packages/issues/7281), based on the AUR [PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=cbatticon-gtk2).


Johnnynator:
You could use the gnu-makefile build style instead of none.

cr6git:
@Johnnynator I know there is a Makefile, but if I use `build_style=gnu-makefile` in this template it generates an error with `glib.h` , see [paste](https://bpaste.net/show/2b0278ced85d) ...

Johnnynator:
have you tried setting make_use_env ?

cr6git:
Hurray!! Yes it was that, you are good ;)

PR #7291, opened 2017-08-06
New packages: cbatticon-gtk3 & cbatticon-gtk2

cr6git:


Gottox:
Is it possible to join #7290 with this template and split them via subpackages?

cr6git:
@Gottox Done.

PR #7299, opened 2017-08-07
ranger: add python 3 version

NoctuaNivalis:


Gottox:
Is there a reason why we should have both variants of the very same software?

NoctuaNivalis:
I tried to write a plugin (in python 3, I hadn't checked the version) and it errored because ranger was python 2. I figured I'd rather update ranger to python 3 than rewrite my plugin to an older python version, especially since I knew the same is done for e.g. youtube-dl.

Gottox:
Yea, there are dependencies on `youtube-dl` as well as `python3-youtube-dl`. But there are none on ranger. Is there any reason why switching ranger to python3 only is a bad idea?

NoctuaNivalis:
No, I just assumed it would be preferred to minimize impact. Some people may have their own plugins written in python 2, which updating to python 3 would break.

Gottox:
Any opinions from @voidlinux/pkg-committers? Imho replacing python2 with python3 instead of providing both is reasonable. But I'm unsure about the plugins.

thypon:
IMHO we should check all the plugins first. Otherwise, not really a problem.

lemmi:
Since it's not a lib, it should be fine (provided it works of course).

NoctuaNivalis:
Should work now, I forgot to add rifle to the alternatives. I'll make another PR for just python 3, then you guys can decide which to merge. Thanks for the input.

PR #7305, opened 2017-08-08
Change in package name

Smatchcube:
Every distro use screenfetch instead of screenFetch for the packagename.


Vaelatern:
This does not meet our standards for package names.
Package names shall, when possible, match upstream package names.
This also does not change the directory as necessary to actually change a package name.
This change also seems useless. `xbps-query -Rs screenfetch` finds `screenFetch` without a problem.

PR #7317, opened 2017-08-08
icewm: update to 1.4.2

cr6git:
In response to [#6864](https://github.com/voidlinux/void-packages/issues/6864).


cr6git:
It does not build on musl, as you can see :(

Johnnynator:
`
diff --git a/src/base.h src/base.h
index 79e63be8..37d51b08 100644
--- a/src/base.h
+++ src/base.h
@@ -44,10 +44,11 @@ inline T abs(T v) {
 /*** String Functions *********************************************************/
 
 /* Prefer this as a safer alternative over strcpy. Return strlen(from). */
+#ifdef __GLIBC__
 size_t strlcpy(char *dest, const char *from, size_t dest_size);
 /* Prefer this over strcat. Return strlen(dest) + strlen(from). */
 size_t strlcat(char *dest, const char *from, size_t dest_size);
-
+#endif
 char *newstr(char const *str);
 char *newstr(char const *str, int len);
 char *newstr(char const *str, char const *delim);
diff --git a/src/globit.c src/globit.c
index 5fa5d170..5a397d35 100644
--- a/src/globit.c
+++ src/globit.c
@@ -29,6 +29,10 @@
 #include <unistd.h>
 #include "globit.h"
 
+#if !defined(__GLIBC__)
+#define GLOB_TILDE 0
+#endif
+
 /*
  * globit_best - return the best prefix for a pattern, or the sole match
  *
diff --git a/src/ylocale.cc src/ylocale.cc
index 5e440027..c3ed8b72 100644
--- a/src/ylocale.cc
+++ src/ylocale.cc
@@ -55,6 +55,8 @@ YLocale::YLocale(char const * localeName) {
     int const codesetItems[] = {
 #ifdef CONFIG_NL_CODESETS
 	CONFIG_NL_CODESETS
+#elif defined(CODESET)
+	CODESET
 #else
 	CODESET, _NL_CTYPE_CODESET_NAME, 0
#endif
`
This should fix musl (and I hope it breaks nothing else)

cr6git:
Thank you @Johnnynator 
Hmm... xbps-src was not happy with your patch: 
````
=> icewm-1.4.2_1: running post-extract hook: 00-patches ...
=> icewm-1.4.2_1: patching: musl.patch.
=> ERROR: icewm-1.4.2_1: failed to run post-extract_00-patches() at line 34.
````
OK, so I removed `pre_configure()` and `./autogen.sh` from the new template.

pullmoll:
The problem with musl is that `icewm` tries to define a local version of `strlcpy()` and `strlcat()` while these functions are already defined when including the musl libc headers.

I'll push an update which renames the functions to `my_strlcpy()` and `my_strlcat()` throughout the code, which avoids this type conflicts and works for both, gnu and musl libc.

PR #7319, opened 2017-08-08
base-system: add usbutils

cr6git:
USB is present on most computers.


chneukirchen:
It's about the same size as pciutils, and useful i guess.

cr6git:
I wonder why it's not merged.

cr6git:
Thanks!

PR #7331, opened 2017-08-09
geomyidae: update to 0.28.

m-cz:
geomyidae has moved to [bitreich][1] but since I don't know if `gopher://` links are allowed in template and the old links work fine, I'm keeping it as it is for now.

[1] gopher://bitreich.org/1/scm/geomyidae/log.gph


phikal:
What is the situation with `gopher://` links actually? I was thinking about adding [sacc](gopher://bitreich.org:70/1/scm/sacc/log.gph) ([proxy](https://gopher.floodgap.com/gopher/gw?gopher://bitreich.org:70/1/scm/sacc/log.gph)) and [clic](gopher://bitreich.org:70/1/scm/clic/log.gph) ([proxy](https://gopher.floodgap.com/gopher/gw?gopher://bitreich.org:70/1/scm/clic/log.gph)), which are two gopher clients, to the repository, but as far as I see, I would have to use a gopher proxy, like [here](https://gopher.floodgap.com/gopher/gw?gopher://bitreich.org:70/1/scm/sacc/tag), which probably wouldn't be something the port maintainers would like.

My idea was to use a `do_fetch`, which one could extend to this repository too, using curl, to download these directly from bitreich.org. Would this make sense?

chneukirchen:
Perhaps $XBPS_FETCH_CMD could be set?

I wouldn't mind using a half-way robust proxy to http tho.

phikal:
@chneukirchen I didn't know about that variable, but it seems to work, which is quite nice.

PR #7334, opened 2017-08-10
New package: mtpaint

cr6git:


chneukirchen:
We don't do `-git` packages.

cr6git:
OK @chneukirchen , I modified the template.

PR #7339, opened 2017-08-10
linux4.4: update to 4.4.80

cr6git:


cr6git:
`4.4.81` released, no need to keep this open.

PR #7353, opened 2017-08-12
libGLES: fix common/shlibs for Opera

m-wynn:
The new version of opera looks for libGLESv2.so


ebfe:
This is not a fix, opera ships it's own libGLESv2.so.

Gottox:
I'm not in favor to change libGL for some nonfree package. Especially I'm not in favor of adding unversioned solibs again.

PR #7358, opened 2017-08-12
New package: zulucrypt

Piraty:
Including it's library zuluplay, which is a form of tcplay (formerly truecrypt)

i guess enabling gnome's libsecret support by default is suitable, although zulucrypt uses lxqt-wallet by default for it's own keywallet (a bundled lib, since lxqt-wallet is not in void's repos)


Piraty:
I have corrected the issues i got as feedback in irc.
but moving *-gui into own subpackage still leaves qt5 as a dependency for the main package... ? i guess i could leave *-gui in the main package, as this is probably the intended use for most users.

Piraty:
updated to 5.4.0

Vaelatern:
I'd merge but I screwed up fixing the merge conflict, and it needs to be two commits.

PR #7371, opened 2017-08-14
New package: fontmatrix-0.6.0

newbluemoon:


newbluemoon:
Cleaned up dependencies.

PR #7380, opened 2017-08-15
vim: update to 8.0.0946

cr6git:


Vaelatern:
You do know that vim pushes a new release every commit, yes?

What's gained by this version that it's worth bumping?

cr6git:
Good question, but I'm not smart enough to give you the answer. 🃏

chneukirchen:
I guess a bump every 100 or 200 patches is enough. (Esp. in times of heavy development as currently.)

cr6git:
@Vaelatern & @chneukirchen No problem, I will slow down the pace! 
I tend simply to follow the AUR updates, even if it's not really necessary... 
I am at your service. :)

Gottox:
Merging this one nontheless, thanks mate!

PR #7381, opened 2017-08-15
New package: partclone-utils

cr6git:


cr6git:
Does not build on musl.

Vaelatern:
If I had twenty minutes for this, I'd look at the file `sysdep_posix.c:82:1: error: conflicting types for 'posix_close'` and see if there is a `#ifdef` that guards this functionality.

Vaelatern:
+1 @cr6git for letting it stay not building on musl. makes next steps easier to see.

cr6git:
I no longer have time to work on this, sorry. 
I must take care of my real life.

PR #7386, opened 2017-08-15
audacity: replace libsamplerate with libsoxr

cr6git:
It's `libsoxr` which is enabled in `audacity`, not `libsamplerate` !


chneukirchen:
Afaiu audacity supports both? is soxr better? why did you remove expat?

cr6git:
Audacity supports both, yes, but you can't enable the two at the same time. 
`libsoxr` is better (and recommended), see [the doc](https://stackoverflow.com/a/23205409)... 
I remove `expat-devel` because it's automatically pulled by `gtk+-devel`.

chneukirchen:
Makes sense. Thank you!

PR #7401, opened 2017-08-15
New package: fim-0.5

7185:
I was surprised there wasn't any fbi or fbida package, so I created a fim package.


7185:
I have no idea how to fix cross build and I'm no longer interested in maintaining this package (kodi is more convenient for me, even though it's so much heavier).
So it's probably better that I close this PR for now.

PR #7402, opened 2017-08-15
New package: goxel

cr6git:
It's [Goxel](https://guillaumechereau.github.io/goxel/). 
I can compile and run the program on x86_64, see [screenshot](https://prnt.sc/g8yo70). 
For now, I am forced to put `nocross=yes` and `nopie=yes` because I think FLAGS are not correctly set in the source file `SConstruct`. 
Correct me if I am wrong.


cr6git:
musl does not agree.

cr6git:
Failed, x86_64-musl does not agree.

Johnnynator:
You can try adding
```
	*-musl) makedepends+=" argp-standalone";;
esac
```
so argp gets installed for musl (glibc has it already integrated)

Johnnynator:
A patch which creates imgui.ini in ~/.config/goxel instead in the current work folder would probably nice

cr6git:
Nice work @Johnnynator !

PR #7405, opened 2017-08-16
New package: gtkedit

cr6git:


srcpkgs/gtkedit/template
```diff
@@ -0,0 +1,38 @@
...
+license="GPL-2"
+distfiles="https://github.com/drankinatty/gtkwrite/archive/v${version}.tar.gz"
+checksum=762f0d62f6733fd35d5d53003066a8eca1ad91d8724ec56d14573a251ac4b29d
+wrksrc="gtkwrite-${version}"
+
+# Cross-compilation fails on 32-bit i686 & armv7
+# /usr/include/glib-2.0/glib/gtypes.h: In function '_GLIB_CHECKED_ADD_U64':
+# /usr/include/glib-2.0/glib/gmacros.h:232:53: error: size of array '_GStaticAssertCompileTimeAssertion_0' is negative
+#
+only_for_archs="x86_64 x86_64-musl aarch64 aarch64-musl"
```
Gottox: **Never** use only_for_archs if you're not perfectly sure this application is unable to run on certian architecture at all, e.g. grub on arm will never work. If crossbuild is broken use nocross=<link to failed buildlog e.g. on travis>

From the manual:
> reserved for uses where the program really *only will ever work* on certain architectures, like binaries or when the program is written in assembly.

In doubt rather use nocross= or broken=.

srcpkgs/gtkedit/template
```diff
@@ -0,0 +1,38 @@
...
+
+# In file included from gtk_findreplace.c:2:0:
+# gtk_findreplace.h:4:21: fatal error: gtk/gtk.h: No such file or directory
+#  #include <gtk/gtk.h>
+#
+export CFLAGS="$(pkg-config --cflags gtk+-2.0)"
+export LDFLAGS="$(pkg-config --libs gtk+-2.0)"
```
Gottox: move these exports to pre_build().

srcpkgs/gtkedit/template
cr6git: I understand. 
Cross-compilation x86_64 --> 32-Bit (i686 / armv7) fails. 
OK, I will change this line and will put `nocross=yes` here.

srcpkgs/gtkedit/template
cr6git: OK.

cr6git:
I hope the change I made is correct ?

PR #7407, opened 2017-08-16
New package: monocypher

cr6git:
Based on the AUR [package](https://aur.archlinux.org/packages/monocypher/).


chneukirchen:
I don't think the .c files belong to `/usr/include`, perhaps there could be a static archive instead.

cr6git:
The AUR packager is doing so within his PKGBUILD, but I can change my template if you think it's necessary.

PR #7414, opened 2017-08-16
New package: subtitleeditor-0.54.0

newbluemoon:


newbluemoon:
The ‘undefined reference’ errors on the arm platforms are solved by rebuilding libxml++. I addressed the issue in #7409. (Probably should have included it in this PR?)

srcpkgs/subtitleeditor/template
```diff
@@ -0,0 +1,20 @@
+# Template file for 'Subtitle Editor'
```
Gottox: rather use the same as pkgname.

newbluemoon: Done.

PR #7418, opened 2017-08-17
New package: waterfox (x86_64)

cr6git:
Based on AUR [package](https://aur.archlinux.org/packages/waterfox-bin/).


Gottox:
I don't think we should use the binary if we can build it from source.

cr6git:
We can build it from source, you are right. 
I am working on it right now !

cr6git:
@Gottox Regarding the compilation, I found a slight difference between Waterfox [PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=waterfox-git#n98) and Firefox [template](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/firefox/template#L104). 

`make -j -f client.mk build` (Waterfox) 
`make -f client.mk build` (Firefox) 

According to the manual: 
````
$ man make
OPTIONS
-j [jobs], --jobs[=jobs]
            If the -j option is given without an argument, make will not limit
            the number of jobs that can run simultaneously.
````

What does it mean "no limit" ? ..."infinite" ? 
Is it safe ?

cr6git:
Also, I found something weird with Firefox: in [mozconfig](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/firefox/files/mozconfig#L44) we have `--enable-pie` but in the [template](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/firefox/template#L13) we have `nopie=yes`. 
Seems like there is a contradiction. 
Am I wrong ?

chneukirchen:
Plain `-j` is not safe and will bog down the machine.

FF has a few libs that are nopie for a reason, so we disable the check.

Gottox:
Actually, looking at the project, does this make sense to merge? It just looks like a rebranded firefox with minimal changes. Considering the buildtimes of FF, I'm not sure if having 99% of the source duplicated is reasonable. Even the changes they did aren't up to date: Sponsored tiles are gone since 2015 (https://arstechnica.com/information-technology/2015/12/firefox-retires-much-maligned-sponsored-tiles-hunts-for-new-revenue-streams/)

cr6git:
@Gottox Oups! I just opened a new PR for Waterfox (built from source) before seeing your last comment, I'm sorry.

cr6git:
OK, if you do not like this program, I close this PR. 
I can close [#7431](https://github.com/voidlinux/void-packages/pull/7431) too, as Travis was not able to build it from source...

PR #7419, opened 2017-08-17
keybase: update to 1.0.28-20170815161818+fed6e3db9

the-eater:
Since keybase don't really support their actual release tags and advices people to follow pre-release I feel we should too since you actually face problems if you don't

1: https://github.com/keybase/client/issues/7907
2: https://github.com/keybase/client/issues/8110
3: See their release log: https://github.com/keybase/client/releases , it's full of holes!

### DIFF
--- a/srcpkgs/keybase/template
+++ b/srcpkgs/keybase/template
@@ -1,8 +1,8 @@
 # Template file for 'keybase'
 pkgname=keybase
-version=1.0.22
-revision=2
-wrksrc="client-${version}"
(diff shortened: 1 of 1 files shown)


Gottox:
🙄 really? - did they get the concept of "releases"?

the-eater:
Yes, but you should only use their prebuilt releases, otherwise it's wrong. but luckily they use their short-commit hash in their release versions, so not hard to find which commit it is :)

Vaelatern:
One of these ways is much more work. If we wanted to fully go their route we'd package kbfs, which they refuse to tag.

PR #7420, opened 2017-08-17
postgresql: update to 9.6.4 and switch to use https

arete:

### DIFF
--- a/srcpkgs/postgresql/template
+++ b/srcpkgs/postgresql/template
@@ -1,7 +1,7 @@
 # Template file for 'postgresql'
 pkgname=postgresql
-version=9.6.3
-revision=2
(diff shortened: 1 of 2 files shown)


srcpkgs/postgresql/template
```diff
@@ -1,7 +1,7 @@
 # Template file for 'postgresql'
 pkgname=postgresql
-version=9.6.3
-revision=1
+version=9.6.4
+revision=2
```
Gottox: no need to increase the revision when the version is updated.

arete: ok,  fix

arete:
ji, i have update the commit with you changes requested.

PR #7421, opened 2017-08-17
Fritzing: fix cross-compilation

newbluemoon:


Johnnynator:
if it is only needed for cross compiling then you can use

```
        hostmakedepends+=" t5-serialport-devel qt5-svg-devel'
fi
```

newbluemoon:
To be honest, I was thinking about it, but wasn’t quite sure and decided against  ;-)

PR #7427, opened 2017-08-18
gcc: update to 7.2.0

pullmoll:
Some hackery was required to get `cross-i686-linux-musl` built and needs investigation: disabling atomic and gomp libs and not enabling default ssp.
Issues are related to a symbol `__stack_chk_fail_local` which is found in `libssp_nonshared.a`


chneukirchen:
Ok, we need to figure out what to do with pdftk, which needs gcj.

Also we need to review #6613 to see what default-pie/ssp break.

pullmoll:
As it is now the compilers are (still) built with nopie=yes, but the default to enable pie and ssp.
This should keep precompiled headers intact, no?
For pdftk I have no idea.

I thought I'd create a PR now as my tests suggest that all cross compilers are working,
except for the few issues with cross-i686-linux-musl and the missing symbol __stack_chk_fail_local.

chneukirchen:
I think gcc should remain `nopie=yes` for now.

Did you try x86_64-musl? does it work without https://github.com/voidlinux/void-packages/pull/6613/commits/0031ea3f8fe0b385d05df6ccb7eb81f32a36f768 ?

We'll also need to change hardening flags ala https://github.com/voidlinux/void-packages/pull/6613/commits/8f57795728671e64eccf0963afeabb403fa0a6d7

pullmoll:
I didn't know https://github.com/voidlinux/void-packages/commit/0031ea3f8fe0b385d05df6ccb7eb81f32a36f768 but it is for i686-musl, not x86_64-musl, no?
I'll take a look if that fixes building cross-i686-linux-musl with default ssp.

I built lots of packages for x86_64-musl and didn't see problems so far.
The repo at [voidlinux.de](http://repo.voidlinux.de/current/) has what I have built.

I didn't want to put the hardening flags change into this PR.
Can this be done in a separate step? Currently the flags still work as they are.

chneukirchen:
Yes, compilers are nopie currently.
(But without the hardening.sh patch, nopie doesn't work anymore. ;))

I think we can drop pdftk and replace it with something that fetches binaries from debian if needed.

chneukirchen:
Alternatively, an eager maintainer could strip down the gcc 6.3 template to only build gcj.

chneukirchen:
I'd put the hardening flags into this PR, as nopie= will not work else.

pullmoll:
I don't see the `${FILESDIR}/gcc-musl-libssp.diff` file anywhere. It isn't part of https://github.com/voidlinux/void-packages/commit/0031ea3f8fe0b385d05df6ccb7eb81f32a36f768 ?

chneukirchen:
I put it at http://sprunge.us/TfTB

pullmoll:
Strange. The order of my commits is different from what github tells.
I put the hardening.sh patch after gcc update where it should fit.

pullmoll:
I took the `gcc-7.2.0/libssp/ssp-local.c` file, removed the `#include "config.h"` and the `#if/#endif` instead of using the missing `${FILESDIR}/__stack_chk_fail_local.c`.

Now `cross-i686-linux-musl` should work while an `i686-musl` environment's `gcc` still reports the missing symbol `__stack_chk_fail_local`. Not sure, if I missed something or if the missing function should become part of the `musl` pkg as well.

pullmoll:
First gcc-7.2.0 built [ISOs uploaded](http://repo.voidlinux.de/live/) for testing.

pullmoll:
It looks like `i686-glibc` also needs the additional `-lssp_nonshared` in LDFLAGS, or packages like libaio won't link. I also think this lib should be in LDFLAGS regardless of nopie or pie, because may be required in either case.

I'll modify the hardening.sh script accordingly.

pullmoll:
Everything seems to work as expected now. I successfully built 
I'll leave merging to someone who reviews the changes and thinks they are good enough.

pullmoll:
Actually there are issues with nopie=yes packages now. :cry:
Investigating and also trying to include the glibc patches into the cross-* compilers.

pullmoll:
It seems the update is now seriously broken.

I can no longer build llvm for `x86_64` or `x86_64-musl`, while it builds for `i686`.
On the other hand i686 and i686-musl still have issues with not finding `__stack_chk_fail_local` in e.g. libaio, despite the `-lssp_nonshared` being in the `LDFLAGS`.

I'm not sure whether it comes from the updated patch set for glibc, which is applied to the cross compilers internal glibc as well, or rather from the `hardening.sh` changes.
The latter had to be modified again to keep the old way of using `-spec=` or otherwise the bootstrap would not work, because it uses the host `gcc` for a lot of packages, not just `gcc-7.2.0` itself.

pullmoll:
When building `./xbps-src -m /tmp/glibc-gcc7/ -a i686 -N pkg base-chroot` I now get this:
```
=> glibc: building (dependency of base-chroot) ...
=> glibc-2.26_2: installing host dependency 'bison-3.0.4_2' ...
=> glibc-2.26_2: installing host dependency 'perl-5.26.0_1' ...
=> glibc-2.26_2: running do-fetch hook: 00-distfiles ...
=> glibc-2.26_2: running do-extract hook: 00-distfiles ...
=> glibc-2.26_2: extracting distfile(s), please wait...
=> glibc-2.26_2: running post-extract hook: 00-patches ...
=> glibc-2.26_2: patching: glibc-upstream-01.patch.
=> glibc-2.26_2: patching: glibc-upstream-02.patch.
=> glibc-2.26_2: patching: glibc-upstream-03.patch.
=> glibc-2.26_2: patching: glibc-upstream-04.patch.
=> glibc-2.26_2: patching: glibc-upstream-05.patch.
=> glibc-2.26_2: patching: glibc-upstream-06.patch.
=> glibc-2.26_2: patching: glibc-upstream-07.patch.
=> glibc-2.26_2: patching: glibc-upstream-08.patch.
=> glibc-2.26_2: patching: glibc-upstream-09.patch.
=> glibc-2.26_2: patching: glibc-upstream-10.patch.
=> glibc-2.26_2: patching: glibc-upstream-11.patch.
=> glibc-2.26_2: running pre-configure hook: 00-gnu-configure-asneeded ...
=> glibc-2.26_2: running pre-configure hook: 01-override-config ...
=> glibc-2.26_2: running pre-configure hook: 02-script-wrapper ...
=> glibc-2.26_2: running do_configure ...
checking build system type... x86_64-unknown-linux-gnu
checking host system type... i686-pc-linux-gnu
checking for i686-pc-linux-gnu-gcc... i686-pc-linux-gnu-gcc
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether i686-pc-linux-gnu-gcc accepts -g... yes
checking for gcc... cc
checking for i686-pc-linux-gnu-readelf... i686-pc-linux-gnu-readelf
checking whether we are using the GNU C++ compiler... yes
checking whether i686-pc-linux-gnu-c++ accepts -g... yes
checking whether i686-pc-linux-gnu-c++ can link programs... no
checking for sysdeps preconfigure fragments... aarch64 alpha arm hppa i386 m68k microblaze mips nios2 powerpc s390 sh sparc tile x86_64 
configure: running configure fragment for add-on libidn
checking for -fstack-protector... yes
checking for -fstack-protector-strong... yes
checking for -fstack-protector-all... yes
checking for assembler and linker STT_GNU_IFUNC support... no
checking for gcc attribute ifunc support... yes
configure: error: --enable-multi-arch support requires assembler and linker support
=> ERROR: glibc-2.26_2: failed to run do_configure() at line 100.
```

Looks like something about multiarch support has changed.

chneukirchen:
I'll try to have a closer look at this on the weekend, I think perhaps we should drop PIE during bootstrap, and tell people to rebuild base-chroot once they've bootstrapped for the real thing.

pullmoll:
Main problem is that some things are currently moving targets.
It's difficult to give a definite state of the patch set without retrying bootstrapping again after changes like the glibc patches, modified hardening etc.

Gottox:
@pullmoll: How can we help?

pullmoll:
@Gottox:
The problems when building llvm have recently been solved.
Currently the remaining major issues are with `i686` and `i686-musl` arches.

For some packages they do not find the `__stk_chk_fail_local` symbol from `/usr/lib/libssp_noshared.a`, depsite the lib being explicitly listed on the gcc command line with
`-lssp_nonshared` and implicitly with the compiler internal flags (`LINK_SSP_SPEC`) for stack-protector enabled builds.

I think that perhaps the `-lssp_nonshared` needs to be the last item on a command line, not just the last item in the `LDFLAGS`, for the symbol to resolve? Or there is still something wrong with gcc's template.

Also for  builds using `gyp` with `python2.7` (e.g. qt5 or webkit/chromium) there is a prominent bug. The builds hang in a race condition in python's semaphore/multithreading code and don't ever finish. I mentioned that [elsewhere](https://github.com/voidlinux/void-packages/issues/7468).

I can, however, build many packages without problems, or rather with problems with gcc7 already being solved. Currently I'm trying without the `--with-default-libstdcxx-abi=gcc4-compat` again, just to see if that reduces  the number of remaining problems.

If you want to help, you could take the PR and modify gcc/template to add that flag again, as that is how we wanted to proceed. Then trying to bootstrap and build e.g. base-system for the various arches to see how well it goes.

I'll retry that as well after having built the live ISOs for testing.

pullmoll:
Oops! I see the `--with-default-libstdcxx-abi=gcc4-compat` is in place in srcpkgs/gcc/template, so the PR is in shape as far as I can tell.

Gottox:
Is our webkit affected by this bug? https://bugs.webkit.org/show_bug.cgi?id=173407

pullmoll:
> Is our webkit affected by this bug? https://bugs.webkit.org/show_bug.cgi?id=173407

I don't know. I don't yet have an environment where I could run any tests and if I had it, I wouldn't know exactly how to run a test from that bug description. Could you give an example command line?

What I can tell is that firefox-esr also doesn't build for i686* due to the python semaphore/mutex race condition.

chneukirchen:
The fix is trivial anyway, this is not a blocker.

Gottox:
@chneukirchen you mean for firefox or for webkit?

I tend to add the [webkit fix](https://git.archlinux.org/svntogit/packages.git/tree/trunk/qt5-webkit-gcc7.patch?h=packages/qt5-webkit) without revbumping, so that we can be sure it's there once we rebuild.

chneukirchen:
The webkit bug.

pullmoll:
A workaround for the python mutex/semaphore problems is the following patch.
After disabling the `nopie=yes` for firefox-esr it seems I can build it for i686 now.
```
It seems that avoiding to create local aliases for the self._rlock and
self._wlock methods prevents the strange errors on i686 related to
semaphore locking and unlocking.

One example of failing builds is firefox-esr running gyp in an i686 environment.

--- Lib/multiprocessing/queues.py	2017-09-16 19:38:35.000000000 +0200
+++ Lib/multiprocessing/queues.py	2017-10-24 19:49:06.291351206 +0200
@@ -369,13 +369,12 @@
 
     def _make_methods(self):
         recv = self._reader.recv
-        racquire, rrelease = self._rlock.acquire, self._rlock.release
         def get():
-            racquire()
-            try:
-                return recv()
-            finally:
-                rrelease()
+            self._rlock.acquire()
+            try:
+                return recv()
+            finally:
+                self._rlock.release()
         self.get = get
 
         if self._wlock is None:
@@ -383,11 +382,10 @@
             self.put = self._writer.send
         else:
             send = self._writer.send
-            wacquire, wrelease = self._wlock.acquire, self._wlock.release
             def put(obj):
-                wacquire()
-                try:
-                    return send(obj)
-                finally:
-                    wrelease()
+                self._wlock.acquire()
+                try:
+                    return send(obj)
+                finally:
+                    self._wlock.release()
             self.put = put
--- Modules/_multiprocessing/semaphore.c	2017-09-16 19:38:35.000000000 +0200
+++ Modules/_multiprocessing/semaphore.c	2017-10-28 10:49:56.944993401 +0200
@@ -378,7 +378,7 @@
             }
         }
 #else
-        int sval;
+        int sval = -1;
 
         /* This check is not an absolute guarantee that the semaphore
            does not rise above maxvalue. */
```

pullmoll:
@voidlinux/pkg-committers this PR now very much seems to be ready.
Only fallout after building 6400 pkgs so far is gnupg for target `i686` which segfaults in the checks.
Some packages, namely firefox* and mozjs52 need `nopie=yes` removed to link.
Live ISOs as always on http://repo.voidlinux.de/

LMK if there are any objections or when and how to update gcc.

chneukirchen:
If gnupg only fails on i686-musl, that is not a problem IMO because this is not a supported configuration currently.

Else, +1 from me.

the-maldridge:
LGTM

PR #7441, opened 2017-08-19
dhcpcd: enable wpa_supplicant hook by default

the-maldridge:
@voidlinux/pkg-committers 

We've talked about turning this hook on by default for a while.  Either this PR should be merged, or we need to take a look at voidlinux/void-mklive#125.  As it stands wifi fails for a reason that makes sense if you know how it works, but not if you're a first time user.

Since dhcpcd is smart enough to not invoke the hook if the interface isn't wireless, I see no reason not to just have it on all the time.


chneukirchen:
I don't mind it, but I use my own dhcpcd start scripts anyway.

Gottox:
Ok for me.

the-maldridge:
Merged in 356af2896f

PR #7444, opened 2017-08-19
New package: rapidjson-1.1.0

egorenar:


srcpkgs/rapidjson/template
```diff
@@ -0,0 +1,28 @@
+# Template build file for 'rapidjson'.
+pkgname=rapidjson
+version=1.1.0
+revision=1
+nocross=yes
+build_style=cmake
+configure_args="-DCMAKE_INSTALL_PREFIX=/usr"
+make_build_args="VERBOSE=1"
```
the-maldridge: Is this necessary to get a build out?  I'd rather this not be here in the merged template

srcpkgs/rapidjson/template
```diff
@@ -0,0 +1,28 @@
+# Template build file for 'rapidjson'.
+pkgname=rapidjson
+version=1.1.0
+revision=1
+nocross=yes
```
the-maldridge: This should have a comment or link to a build log explaining why this is nocross.

Vaelatern: Indeed, `nocross="The Fizz is broken when Baz happens"` would (except being nonsense) be a good way to do this.

srcpkgs/rapidjson/template
```diff
@@ -0,0 +1,28 @@
...
+build_style=cmake
+configure_args="-DCMAKE_INSTALL_PREFIX=/usr"
```
Vaelatern: This is already passed as needed. Please remove this line.

srcpkgs/rapidjson/template
egorenar: removed nocross

srcpkgs/rapidjson/template
egorenar: done

srcpkgs/rapidjson/template
egorenar: done

egorenar:
i addressed all raised issues, please check again. Thanks.

Vaelatern:
In this case, it seems the main package is only documentation without the contents of the `-devel` package. Was I just not reading carefully?

If the main package has no use without the -devel package, and there is indeed no `.so` built for this library, then this package needs to be `noarch=yes` and the `-devel` package would need to be removed in favor of all files being in the main package.

egorenar:
you are right, this library is a header-only one.

PR #7447, opened 2017-08-19
New package: pioneer

cr6git:
The game [Pioneer](https://pioneerspacesim.net).


cr6git:
x86_64-musl does not agree. :(

pullmoll:
The functions `feenableexcept()` and `fedisableexcept()` are not available in musl libc.
You need to devise a patch which wraps this code in e.g.
```
#if defined(__GLIBC__)
...
#endif
```

newbluemoon:
@cr6git, I hope it’t ok, just created a patch that’s working for me:

<pre>
--- src/posix/OSPosix.cpp.orig  2017-08-19 20:00:23.645790781 +0200
+++ src/posix/OSPosix.cpp       2017-08-19 20:00:44.153855997 +0200
@@ -48,7 +48,7 @@

 void EnableFPE()
 {
-#if defined(_GNU_SOURCE) && !defined(__APPLE__)
+#if defined(_GNU_SOURCE) && !defined(__APPLE__) && defined(__GLIBC__)
        // clear any outstanding exceptions before enabling, otherwise they'll
        // trip immediately
        feclearexcept(FE_DIVBYZERO | FE_INVALID | FE_OVERFLOW);
@@ -58,7 +58,7 @@

 void DisableFPE()
 {
-#if defined(_GNU_SOURCE) && !defined(__APPLE__)
+#if defined(_GNU_SOURCE) && !defined(__APPLE__) && defined(__GLIBC__)
        fedisableexcept(FE_DIVBYZERO | FE_INVALID | FE_OVERFLOW);
 #endif
 }
</pre>

cr6git:
My deepest thanks to both of you!

cr6git:
I just discovered a bad thing: 
````
$ sudo xbps-remove -R qt5-devel
qt5-devel-5.8.0_6 (remove) breaks installed pkg `libassimp-4.0.1_1'
Transaction aborted due to unresolved dependencies.

$ sudo xbps-remove -R libassimp
libassimp-4.0.1_1 (remove) breaks installed pkg `pioneer-20170813_1'
Transaction aborted due to unresolved dependencies.
````
Because of this `libassimp` dependency, `qt5-devel` is automatically installed with `pioneer`. 
This is really annoying! :(

newbluemoon:
Begs the question if `qt5-devel` is really a necessary dependency of libassimp. Checking with debian and arch (though they’re at libassimp version 3.3.1) they only list glibc, libstdc++, libgcc, minizip, zlib and boost as dependencies. I guess the question never arose, because when checking the reverse dependencies of assimp/libassimp no package depends on either of those packages.

pullmoll:
I think I was just being to lazy to set the depends= for the libassimp subpackage.
Only the libassimp-devel subpackage should depend on other *-devel subpackages.
I'll try to fix that.

cr6git:
Thank you for taking the time to look at this.

PR #7448, opened 2017-08-19
kodi: add fix-string.patch

Hoshpak:
replaces string with std::string in filesystem/FTPParse.cpp to fix a compilation error


pullmoll:
A `using namespace std;` at the top of the file doesn't work?

Hoshpak:
I didn't try that. Adding stdd:: was my first try and I later found out that that's the same the upstream developers did https://github.com/xbmc/xbmc/commit/634153c85115b44015cdb1ddd6263f5a87e4e41b .

pullmoll:
Upstream doing it like your patch is a strong argument.
My idea was just to have a smaller patch, but yours is good then.

PR #7450, opened 2017-08-19
freeciv: move freeciv.desktop into gtk3 subpackage

Johnnynator:
Somehow the gtk3 .desktop file is named freeciv.desktop and was therefore not moved into the correct subpackage.


bra1nwave:
I added this fix to my PR (#7434). Please close your PR.
Thanks!

PR #7451, opened 2017-08-19
New package: gscope (GTK2)

cr6git:


cr6git:
Failed **only** on x86_64-musl, because of an [obscure error](https://travis-ci.org/voidlinux/void-packages/jobs/266355269#L865)...

Johnnynator:
I wouldn't call it obscure since it is actually fairly simple: The glibc has this line `sys/stat.h:# define ACCESSPERMS (S_IRWXU|S_IRWXG|S_IRWXO) /* 0777 */` and musl doesn't.

Adding a patch which defines it also in MUSL is fairly simple:

```
#ifndef ACCESSPERMS
#define ACCESSPERMS (S_IRWXU|S_IRWXG|S_IRWXO) 
#endif
```

cr6git:
maybe "obscure" for me, but you are far more skilled than me!! 
thank you for the help @Johnnynator

Gottox:
Hmm... Is there a reason why we need the gtk2 version of this software if we can provide gtk3?

cr6git:
@Gottox No particular reason. I can close this one if you prefer.

PR #7453, opened 2017-08-19
New package: gscope3

cr6git:


cr6git:
grumble.. same error as [#7451](https://github.com/voidlinux/void-packages/pull/7451).

cr6git:
@Johnnynator provided the [theoretical solution](https://github.com/voidlinux/void-packages/pull/7451#issuecomment-323548800) to get rid of this error, but I don't know how to make use of it, I never wrote any patch...

Johnnynator:
Creating a patch isn't that hard:
First find the file that has to be patched (in this case packages/gscope/src/auto_gen.c) and make  a backup. Then just edit the original file as needed (adding the 3 lines).
The last step is running `diff -u backupfile editedfile` and put the output into the patches folder, you can also redirect the output directly into a file.
```
-- auto_gen.c  2017-08-20 17:26:08.340009975 +0200
+++ packages/gscope/src/auto_gen.c      2017-08-20 17:27:13.104011572 +0200
@@ -25,6 +25,11 @@
 #include "display.h"
 #include "app_config.h"

+//musl doesn't define ACCESPERMS
+#ifndef ACCESSPERMS
+# define ACCESSPERMS (S_IRWXU|S_IRWXG|S_IRWXO) 
+#endif
+
 //===============================================================
 //       Defines
 //===============================================================
```

cr6git:
I learn something new every day!

srcpkgs/gscope3/template
```diff
@@ -0,0 +1,32 @@
...
+build_style=gnu-configure
+hostmakedepends="automake pkg-config"
+makedepends="gtksourceview-devel"
+maintainer="cr6git <quark6@protonmail.com>"
+short_desc="Interactive source code browsing tool for C and other languages"
+homepage="https://github.com/tefletch/gscope"
+license="MIT"
+distfiles="https://github.com/tefletch/gscope/archive/${_release}.tar.gz"
+checksum=e530438a42998c3f4f61468f8d917377a22861f39c9b338e42c021a5efe925ed
+wrksrc="gscope-2.10-v3.10"
```
the-maldridge: for consistency please put  wrksrc and build_wrksrc lines below _release

the-maldridge: nvm, I just fixed it in the merge.

PR #7454, opened 2017-08-19
New package: palemoon-27.4.2

ColdPhoenix00:


Hoshpak:
I just pulled your branch, built and installed the package and it seems to work and not segfault right away like my earlier attempts at building palemoon. Thanks, that's amazing! Even if this ends up not being merged into the official repository, I will pull your branch into my private fork and test the resulting binary package some more.

the-maldridge:
Thank you for your interest in Void Linux.  Unfortunately we are not accepting any new firefox-sized browsers to the repo at this time.

PR #7458, opened 2017-08-20
linux4.4: update to 4.4.83

cr6git:


Gottox:
obsolete, we're at 4.4.85 atm. Sorry for letting you wait.

PR #7459, opened 2017-08-20
texstudio: fix checksum + template cleanup

Piraty:
( unfortunately, upstream changed the source-archive ..., I already complained about this.)

updated checksum, wkrsrc
+ consolidated build options while at it


Duncaen:
`nopoppler` is completely different from all buildoptions we have, I think it should be `poppler` and XOR.

Piraty:
you're right, i switched it to the opposite meaning.

PR #7466, opened 2017-08-21
s6 and related tools: allow cross-compilation.

CMB:
This is a followup to the work I started with skalibs
in May.  All of the s6 packages are now cross-buildable.


CMB:
The errors from travis are xlint, and they were pre-existing.

PR #7471, opened 2017-08-21
New package: emacs-ess-16.10

wagnerflo:
ESS is an addon for emacs. It includes a emacs-mode for R among other things.


srcpkgs/emacs-ess/template
```diff
@@ -0,0 +1,21 @@
+# Template file for 'emacs-ess'
+pkgname=emacs-ess
+version=16.10
+revision=1
+depends+=" emacs R"
```
chneukirchen: please just use depends="..." here

srcpkgs/emacs-ess/template
```diff
@@ -0,0 +1,21 @@
...
+makedepends+=" emacs wget"
```
chneukirchen: where is wget used?

srcpkgs/emacs-ess/template
wagnerflo: Done.

srcpkgs/emacs-ess/template
wagnerflo: They download a single file inside the package's [Makefile](https://github.com/emacs-ess/ESS/blob/v16.10/lisp/Makefile#L109).

Let's maybe include that file in the distfiles instead? See the amended template.

chneukirchen: I'd recommend to add it to distfiles, with a stable url (press y in github view)

chneukirchen: ah yes, perfect.

srcpkgs/emacs-ess/template
```diff
@@ -0,0 +1,26 @@
...
+depends="emacs R"
+makedepends="emacs"
+short_desc="Emacs Speaks Statistics: A Universal Interface for Statistical Analysis"
+maintainer="Florian Wagner <florian@wagner-flo.net>"
+license="GPL-2"
+homepage="http://ess.r-project.org"
+_juliarev=edaaff2e971a527460f838bebc0ef679f9a571b0
+_juliafile=julia-mode-${_juliarev}.el.txt
+distfiles="http://ess.r-project.org/downloads/ess/ess-${version}.tgz
+	https://raw.githubusercontent.com/JuliaEditorSupport/julia-emacs/${_juliarev}/julia-mode.el>${_juliafile}"
```
Duncaen: We usually indent in variables with one space (` `) instead of tabs.
```
distfiles="http://ess.r-project.org/downloads/ess/ess-${version}.tgz
 https://raw.githubusercontent.com/JuliaEditorSupport/julia-emacs/${_juliarev}/julia-mode.el>${_juliafile}"
```

wagnerflo: Fixed! Thanks for noticing.

PR #7473, opened 2017-08-21
mc: treat .xbps as .tar.xz.

mvf:
This allows browsing the contents of XBPS packages inside Midnight Commander.


chneukirchen:
Nice!

PR #7483, opened 2017-08-23
libvidstab: update to 1.1.0

newbluemoon:
I was preparing a template for vid.stab and after finishing I realised it’s already there as libvidstab. So I decided to open a pull request to update the package. But that leads me to a question about naming conventions. The [manual](https://github.com/voidlinux/void-packages/blob/master/Manual.md#libs) says:

> Libraries are packages which provide shared objects (*.so) in /usr/lib. They should be named like their upstream package name with the following exceptions:
> 
> The package is a subpackage of a front end application and provides shared objects used by the base package and other third party libraries. In that case it should be prefixed with 'lib'. An exception from that rule is: If an executable is only used for building that package, it moves to the -devel package.

The way I understand this is that the package should have the name ‘vid.stab’ instead of ‘libvidstab’ because it’s not a subpackage? It would be much easier to find.

P.S.: Sorry for being pedantic. ;-)

### DIFF
--- a/common/shlibs
+++ b/common/shlibs
@@ -2377,7 +2377,7 @@ libwiredtiger_snappy.so wiredtiger-2.9.0_1
 libwiredtiger_zlib.so wiredtiger-2.9.0_1
 libwiredtiger-2.9.2.so wiredtiger-2.9.2_1
 libwiredtiger-2.9.3.so wiredtiger-2.9.3_1
-libvidstab.so.0.9 libvidstab-0.98b_1
+libvidstab.so.1.1 libvidstab-1.1.0_1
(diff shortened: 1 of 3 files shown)


lemmi:
Have you checked it against ffmpeg? version 1.1 mentions something about an api change.

newbluemoon:
> Have you checked it against ffmpeg?

The commit includes a revision bump for ffmpeg and a local as well as the travis build don’t show any error. (If I understand you right and that’s what you mean.)

As for the naming: from now on I will look harder if a template already exists before starting one ;-) But now that it’s there I don’t know if it’s worth the trouble to rename it. The intension of my question was more generally, to get the details right. :)

PR #7485, opened 2017-08-23
kodi: update to 17.4

Hoshpak:


Gottox:
Can you update kodi-rpi alike?

Hoshpak:
I can do that later today. Usually I would not touch packages I'm unable to test but thinking about it, since I decomissioned my rbpi some time ago and I certainly have some spare SD cards, I could as well put a Void image on one of them and test the update.

Gottox:
I can test the package at the end of the week.

Hoshpak:
I added a commit updating kodi-rpi to 17.4 as well. The main issue I encountered is that the kodi developers now only support ffmpeg 3.1 and want us to use their patches version as documented in https://github.com/xbmc/xbmc/commit/5ba6eb7cadc41d5eb86ef94e3661b833b62953ff . Though cmake detects that kodi is being cross-compiled, cross-compiling ffmpeg fails (and I couldn't find out why yet). 

So the PR in it's current form builds kodi-rpi against the ffmpeg package from the repository which was done with 17.3 before and works fine from what I can tell. I just isn't officially supported by upstream any more. Is it preferable to use the library from the repository or rely on the upstream-patched and outdated version in this case? The kodi package for i686 and x86_64 currently builds the kodi-version of ffmpeg.

I tested the binary package on my Raspberry PI B (rev 1, armv6l) and after increasing video memory to 128MB, it works surprisingly well for such an old piece of hardware.

Gottox:
will test at the weekend.

PR #7490, opened 2017-08-23
WIP: New Package: Linphone an ever annoing to built sip client

Johnnynator:
~~I will check later if it is possible to let antlr3 generate the needed c code for belle-sip on another system and then use it on all the other systems where we don't have Java.~~

Cross compiling results in unusable devel packages for use when compiling natively (and vice versa)

The PR is based on #5354


Johnnynator:
Why is Travis never working as expected, both failing builds just compile fine on my system...

wagnerflo:
Are you planning to work on this again? I'd be a user/tester of your package...

Johnnynator:
I will try to fix some new bugs (doesn't compile anymore) tomorrow and the ask someone to finally review it (so it doesn't rot another 4 month here)

wagnerflo:
Thank you!

Johnnynator:
@wagnerflo If you want you can compile the linphone package yourself and test it.
The qt gui (linphoneqt package) does built but will not run before qt5.9 is merged into void.

Johnnynator:
Damm, it seems linke I broke it somewhere so it will take more time. It compiles fine but anything related to network doesn't work at all (which is everything useful)

Johnnynator:
Everything should be fine now, my problem was a problem in the new qt gui.
I will add the Qt gui in a later PR since we need qt>5.9 for that.

Vaelatern:
Oh, please do open this on the new fork

Johnnynator:
I will once I got everything in a more usable manner (aka working cross builds)

Issue #7494, opened 2017-08-24
nvidia-libs-384.69_1 fails to symlink /usr/lib/libnvidia-ptxjitcompiler.so.1 and /usr/lib/libnvidia-fatbinaryloader.so.1 → some CUDA applications will exhibit "unknown errors"


datenwolf:
The package nvidia-libs-384.69_1 fails to symlink /usr/lib/libnvidia-ptxjitcompiler.so.1 and /usr/lib/libnvidia-fatbinaryloader.so.1

The result is that a subset of CUDA applications, namely those which do not ship with preassembled GPU binary code will exhibit "unknown errors", because the PTX compiler fails to load which will result in a lot of nonsensical CUDA errors.

It just took me a couple of hours to figure out why the CUDA project I'm working on failed to launch after a system update.

Adding the symlinks

```
lrwxrwxrwx 1 root root       35 Aug 24 20:16 /usr/lib/libnvidia-fatbinaryloader.so.1 -> libnvidia-fatbinaryloader.so.384.69
lrwxrwxrwx 1 root root       34 Aug 24 20:16 /usr/lib/libnvidia-ptxjitcompiler.so.1 -> libnvidia-ptxjitcompiler.so.384.69

```

Solved the problem.

Hoshpak:
Are you sure that the fatbinaryloader-symlink is necessary? I just tested the installation of the current nvidia driver in a VM and the symlinks for libnvidia-ptxjitcompiler are indeed missing in the template. However no symlinks are created for libnvidia-fatbinaryloader.so by the installer and it is also not listed by the .manifest file in the installer.

datenwolf:
Yes, it's absolutely necessary. I've even wrote a MCVE test program, because I kept scratching my head about what was going on. The moment a program contains a CUDA kernel (either `__device__` or `__global__`) for which no compiled binary is cached (under `~/.nv`) it is attempted to compile it from the embedded PTX. Only if that's necessary `libnvidia-fatbinaryloader.so.1` and `libnvidia-ptxjitcompiler.so.1` are `dlopen()`-ed. And if `dlopen()` fails it will just fall through without doing anything, giving you CUDA errors on completely unrelated calls, like `cudaStreamCreate`.

After I had the MCVE (before I knew what was happening) I was able to `strace` it. Then to seal the deal I started in `gdb` put a breakpoint on dlopen and saw that it never tried to open a unversioned or full versioned shared object file.

Adding those symlinks solve this problem.

[...]

PR #7495, opened 2017-08-24
nvidia: add missing symlinks for libnvidia-ptxjitcompiler.so

Hoshpak:
to address #7494


the-maldridge:
This looks suspiciously like bogus compat symlinks.  Care to expand on why these symlinks should exist?

Hoshpak:
They are needed for certain CUDA applications. The libnvidia-ptxjitcompiler.so symlinks would be set by the nvidia installer if it was run manually, the libnvidia-fatbinaryloader.so would not be set by the installer but are also required according to #7494 .

datenwolf:
It seems like the changes are still not pulled into master. Yesterday I was again running into nonworking CUDA programs on my machine, after a full system update and it was again these symlinkes that were missing.

pullmoll:
nvidia has been updated and it does not contain the creation of symlinks.
@Hoshpak care to update your patch or should I try myself?

Hoshpak:
branch rebased, should be good to go now.

Issue #7467, opened 2017-08-21
nix should be rebuilt (perl update)


juliendehos:
```
$ nix-channel --update
perl/lib/Nix/Store.cc: loadable library and perl binaries are mismatched (got handshake key 0xdb80080, needed 0xde00080)
```

juliendehos:
Fixed in #7496.
Thanks.

PR #7496, opened 2017-08-24
nix: rebuild against perl-5.26.0

Hoshpak:
to address #7467 .


chneukirchen:
please add a `shlib_requires="libperl.so.5.26"`, so we don't oversee this on the next sobump.

Hoshpak:
done.

PR #7501, opened 2017-08-26
New package: git-crypt-0.5.0

cobaugh:


ebfe:
Please add the manpage/README to the package.

cobaugh:
Added man page and other docs, and now the build is failing on some archs. Not sure what's up with that.

ebfe:
> Added man page and other docs, and now the build is failing on some archs. Not sure what's up with that.

```make: xsltproc: Command not found```

This needs to go into `hostmakedepends` instead of `makepends`.

cobaugh:
Aha. Got it.

PR #7507, opened 2017-08-27
New package: mozc-2.20.2677.102

mvf:
This pull request adds [Mozc](https://github.com/google/mozc), the open-source version of [Google Japanese Input](https://www.google.co.jp/ime). It offers better accuracy than Anthy and features handwriting recognition and dedicated Emacs integration. The template is based on the [AUR PKGBUILD](https://aur.archlinux.org/packages/mozc) by ponsfoot. All functionality was tested on x86_64 glibc and musl.

New packages:
- `mozc` - Mozc server and configuration UI
- `ibus-mozc` - IBus client
- `emacs-mozc` - Emacs integration

New dependencies:
- `libzinnia` - Zinnia handwriting recognition runtime lib
- `libzinnia-devel` Zinnia lib and includes
- `zinnia` - Zinnia utilities
- `tegaki-zinnia-japanese` - Handwriting recognition data from the [Tegaki project](https://tegaki.github.io/). If there's demand, I could try to package all of Tegaki in another PR. All that would add though is a standalone application for handwriting recognition of Japanese and Chinese and a few developer-centric tools. Tegaki upstream is [no longer maintained](https://github.com/tegaki/tegaki/issues/13).

Mozc is a bit messy in that there are [no stable upstream versions](https://github.com/google/mozc#release-plan) and version numbers are only tracked in a [text file](https://github.com/google/mozc/blob/master/src/data/version/mozc_version_template.bzl#L31) instead of proper tags. It uses Google's own meta build system [GYP](https://gyp.gsrc.io/) for generating makefiles. GYP doesn't have numbered versions at all, so I didn't package it. I hope this is still acceptable. Mozc is quite a popular application.


mvf:
Looks like GYP calls `sem_open(3)`, which fails because on the glibc build servers, `/dev/shm` is not writable. On the musl ones it doesn't exist at all. I'll try to work around this, unless people feel that the build servers should support POSIX semaphores.

srcpkgs/mozc/template
```diff
@@ -0,0 +1,92 @@
...
+makedepends="gtk+-devel ibus-devel libzinnia-devel protobuf-devel qt5-devel"
+depends="tegaki-zinnia-japanese>=0.3"
+maintainer="Matthias von Faber <mvf@gmx.eu>"
+homepage="https://github.com/google/mozc"
+license="3-clause-BSD"
+short_desc="Japanese IME (Open Source version of Google Japanese Input)"
+build_wrksrc="mozc/src"
+create_wrksrc=yes
+distfiles="\
+https://github.com/google/mozc/archive/fc687ac2266c54f0c2f58b4c482486618d68786f.tar.gz \
```
Duncaen: All the backslashes in strings are not necessary.

mvf: Thanks, looks much cleaner now.

Vaelatern:
@Duncaen this okay?

the-maldridge:
merged

PR #7509, opened 2017-08-27
mumble: fix SEGV on musl.

### LINKED ISSUE #7425: Mumble segmentation fault with musl
After installing mumble in a X64 musl system, the only things, I get, is
 a window trying to open immediatly followed with a segmentation fault.
 
Running in gdb gives the following
(gdb) run
Starting program: /usr/bin/mumble 
[New LWP 2249]
CELT bitstream 8000000b from /usr/lib/mumble/libcelt0.so.0.7.0
CELT bitstream 80000010 from /usr/lib/mumble/libcelt0.so.0.11.0
Locale is "en_US" (System: "en_US")
[New LWP 2415]
Database SQLite: "3.19.3" 
[New LWP 2418]
[LWP 2418 exited]
Overlay: Removing old socket on "/home/hq/.MumbleOverlayPipe" 
Overlay: Listening on "/home/hq/.MumbleOverlayPipe" 
GlobalShortcutX: Using XI2 2.3
SocketRPC: Removing old socket on "/home/hq/.MumbleSocket" 

Thread 1 "mumble" received signal SIGSEGV, Segmentation fault.
0x00007fffda7cbb40 in ?? () from /usr/lib/mumble/libmumble.so
(gdb) bt full
#0  0x00007fffda7cbb40 in  () at /usr/lib/mumble/libmumble.so
#1  0x00007ffff7dc9b9f in do_init_fini () at /lib/ld-musl-x86_64.so.1
#2  0x000000000b007fea in  ()
#3  0x0000000000000476 in  ()
#4  0x0000000000000000 in  ()
(gdb) 

I have tried to install the debug version, but there is nothing more.

I have tried to compile with xbps-src. Compilation ends up, but I see strange
warmings like "mix of different architecture".
The result does not work better....

The version of mumble X64 glibc works without visible problems.

Some suggestions ?

mvf:
Fixes #7425. `libmumble` wraps `dlsym(3)` to intercept OpenGL calls for the overlay. On musl, its library constructor crashes while trying to resolve the original `dlsym`, at this line: https://github.com/mumble-voip/mumble/blob/1.2.19/overlay_gl/overlay.c#L699. The root cause is that the code relies on the addresses in the `struct link_map` returned by `dlopen(3)` to be absolute, but on musl they are relative to the load address. This PR adds a musl-only patch that reflects this.


pullmoll:
Instead of adding a musl specific patch, you could perhaps use a glibc aware patch like:
```
--- overlay_gl/overlay.c	2017-01-27 07:48:33.000000000 +0100
+++ overlay_gl/overlay.c	2017-08-28 12:19:14.862084750 +0200
@@ -687,6 +687,11 @@
 	} else {
 		int i;
 		struct link_map *lm = (struct link_map *) dl;
+#if defined(__GLIBC__)
+		const char *base = NULL;
+#else
+		const char *base = lm->l_addr;
+#endif
 		int nchains = 0;
 		ElfW(Sym) *symtab = NULL;
 		const char *strtab = NULL;
@@ -696,13 +701,13 @@
 		while (dyn->d_tag) {
 			switch (dyn->d_tag) {
 				case DT_HASH:
-					nchains = *(int *)(dyn->d_un.d_ptr + 4);
+					nchains = *(int *)(base + dyn->d_un.d_ptr + 4);
 					break;
 				case DT_STRTAB:
-					strtab = (const char *) dyn->d_un.d_ptr;
+					strtab = (const char *) (base + dyn->d_un.d_ptr);
 					break;
 				case DT_SYMTAB:
-					symtab = (ElfW(Sym) *) dyn->d_un.d_ptr;
+					symtab = (ElfW(Sym) *) (base + dyn->d_un.d_ptr);
 					break;
 			}
 			dyn ++;
```

That would avoid the template lines to patch only for musl.

xtraeme:
What @pullmoll said. Dont add musl specific patches.

mvf:
Thanks, this makes sense. Especially since this is a workaround for a glibc quirk and not a musl one.

PR #7513, opened 2017-08-28
setconf: Add setuptools to dependencies to avoid error

xyproto:
Hi,

I'm the author of setconf, but new to Void Linux, so please correct me and/or point me in the direction of some documentation if I have misunderstood anything.

On a blank Void Linux installation, installing and starting `setconf` results in an error:

```
Traceback (most recent call last):
  File "/usr/bin/setconf", line 6, in <module>
    from pkg_resources import load_entry_point
ImportError: No module named pkg_resources
```

However, when also installing `python-setuptools` and `python3-setuptools` (for setconf2 and setconf3, respectively), the required Python module is in place and everything works.

This pull requests bumps the package revision from `2` to `3` and adds `python-setuptools python3-setuptools` to `depends`.


pullmoll:
I think only the `setconf3` subpackage should have `python3-setuptools` added to its `depends=` line, while `setconf` should add only `python-setuptools`.

xyproto:
@pullmoll Thank you for the feedback, will fix.

PR #7515, opened 2017-08-28
New package: syncplay-1.4.0

KeepBotting:
http://syncplay.pl/
https://github.com/Syncplay/syncplay

Free software that synchronises media players so that faraway friends can watch videos together.

Edit: New to Void packaging, can anyone educate me on how to interpret Travis CI? I'm not sure why aarch64 and ARM builds are failing.


Johnnynator:
It seems like Shiboken a dependency of python-pyside can't be cross compiled for these Systems. So it is not you fault that the builds fail.

KeepBotting:
Thanks for the info! I've updated the template to only build for the architechtures that pass the check.

Johnnynator:
The one that just be marked as broken is Shiboken and not your package, since you package may or may not built correctly if all the dependecies would be there.

KeepBotting:
@Johnnynator thanks again. I've pushed another commit, everything should match up now.

Johnnynator:
You have always to expect that someone makes a typo and no one notices it 😄 , there is a H missing in MAC_H_INE.

KeepBotting:
@Johnnynator Right. Completely my fault, I copied what you posted without checking :P fixed.

Vaelatern:
Love it. I'm reviewing and intend to merge.

Vaelatern:
4742526fe86bf0a13f098d48e1568326e4bce1f7

Vaelatern:
My changes are as follows: only one commit per template, I removed the custom broken to indicate what the problem actually was (a mere packaging error, really, that will be fixed at some point). I also removed the bashism of `${pkgname^}` since that doesn't work in basically any other shell (even zsh) and github is not case sensitive.

PR #7519, opened 2017-08-28
mupdf: generate shared libraries

svenper:
The bulk of the binary sizes are font data that can be in a shared library instead.


chneukirchen:
These .so will need sonames.

chneukirchen:
And we need to fix the packages depending on mupdf I guess.

svenper:
Since the shared library is unofficial, what would be a suitable version? 0? 0.0.0? Both?

chneukirchen:
Why not `1.11`? We use this information to track rebuilds.

svenper:
I amended the pull request [like this gist](https://gist.github.com/svenper/6c543ac99195799c1c1dd92b8e396824).

Now for the depending packages. I assume the only ones are `fbpdf` and `zathura-pdf-mupdf` as their makedepends are the only ones that contain mupdf-devel. What kind of changes are needed besides making sure the makefiles know what to look for?

chneukirchen:
You need to check they still build (use proper -lmupdf), and then increase the revision to trigger a rebuild.

chneukirchen:
Also please pass the version via a variable into the Makefile, so we don't need to adjust the patch on every release.

svenper:
`fbpdf` and `zathura-pdf-mupdf` build without issues with no changes.

Gottox:
@chneukirchen ok to merge?

PR #7530, opened 2017-08-30
Fix building mlt with glibc-2.26 and unbreak synfig

Hoshpak:
Patch derived from https://github.com/mltframework/mlt/commit/2125e3955a0d0be61571cf43b674f74b4b93c6f8


newbluemoon:
In #5586 (kdenlive) @jnbr updated mlt to the latest version (6.4.1). He also added a patch to fix building with glibc-2.26. It’s just that the updated mlt package is tied to the not (yet) accepted kdenlive. But maybe that’s the way to go instead of using the quite old 0.9.8?

Hoshpak:
Perhaps both? This PR was intended as a hotfix to unbreak synfig again until mlt is updated.

newbluemoon:
Ah, ok. I just wanted to say that there’s already work done on mlt so that you don’t have to spend time on something that’s already there ;-)

Issue #7531, opened 2017-08-30
Package request: perl Math::Prime::Util


chneukirchen:
This will also need Math::Prime::Util::GMP.

Should be straight forward to do, but I don't have time atm.

ebfe:
6a4bc5aa0e06233a659b01b13bd75df78d4864e7

PR #7533, opened 2017-08-30
New package: perl-Math-Prime-Util-0.65

newbluemoon:
Package request #7531.


chneukirchen:
So quick, thanks!

newbluemoon:
Just wasn’t busy atm ;-)

newbluemoon:
Cross compilation of perl-Math-Prime-Util-GMP fails and perl-Math-Prime-Util doesn’t detect gmp because `Makefile.PL` uses the Devel::CheckLib module which compiles some code and then tries to execute it.

So I disabled the tests by adding to the perl-Math-Prime-Util-GMP template:
<pre>
	pre_configure() {
		sed -i -e "s/^check_lib_or_exit(lib => 'gmp', header => 'gmp.h');//" Makefile.PL
	}
fi
</pre>

and to the perl-Math-Prime-Util template:
<pre>
	pre_configure() {
		sed -i -e "s/^\(my \$have_gmp =\).*/\1 1;/" Makefile.PL
	}
fi
</pre>

From my point of view this should be OK, because gmp-devel is specified in `makedepends` and is there anyway. Should I open a new PR to add these changes?

chneukirchen:
Yes, this fix is fine. Please make a PR.

PR #7536, opened 2017-08-30
New package: lttng-tools-2.8.1

egorenar:


pullmoll:
Updated to 2.10.1, fixed -devel subpackages, fixed *-musl and cross building.
Have fun!

PR #7537, opened 2017-08-30
nfs-utils: patch to add missing include of stdint.h

lonwillett:
@xtraeme 

Trivial fix, but needed to build.


Vaelatern:
needs revbump at least.

Vaelatern:
and please use the real commit msgs.

lonwillett:
@Vaelatern 

Do you really want a revision bump? In theory, this change shouldn't result in an observably different package than one that compiled with headers that pulled in the missing include behind the scenes.

In actuality, I think there have been changes in the system environment that do result in a different package, or there have been changes to the package that should have had a revision bump but didn't. At any rate, that is a different issue than the one addressed here. I'll defer to your opinion, and add the bump, but let me know if you change your mind and want it retracted, or done as a separate commit.

Vaelatern:
hm, where is it needed to build? It built for the x86_64-glibc repo at least...

lonwillett:
It's missing for aarch64-musl, which I was just trying out. Of course, its nice to be able to rebuild packages anyway.

the-maldridge:
@lonwillett merged as written but please try to get this upstreamed.

PR #7538, opened 2017-08-30
diffutils: bumped revision to force recompile and reinstall.

lonwillett:
@xtraeme 

The version of diffutils for aarch64-musl in the repositories
segfaults for context diffs (-c, -u, and perhaps other things as
well). At least this is what I observed on my machine. But a rebuild
seems to fix the problem, so I haven't investigated further. But it
does seem that the revision should be bumped, to flush the bad version
from the repos and any systems that have it installed.


Vaelatern:
Does it fix it for cross builds? Really?

Vaelatern:
I don't have aarch64-musl gear yet, so I can't test.

lonwillett:
Oops! I forgot that the build system uses cross-compiling, while I had rebuilt this natively for my testing.

Just rectified this situation, and discovered that this change will *not* fix the issue. The real issue is that cross-compiling generates a bad version for aarch64-musl. So closing the PR, and changing it to an outstanding issue.

chneukirchen:
So is this package now broken or not?

chneukirchen:
Ah yes, #7539. Ok.

PR #7542, opened 2017-08-31
perl-Math-Prime-Util: fix cross compilation

newbluemoon:


newbluemoon:
I just don’t understand where that conflict came from. What can I do to prevent something like this in the future? I’m still learning git :-)

chneukirchen:
Good question, I think it's because you didn't pull before creating the branch, so the file was added in both branches, and not just merely updated.

newbluemoon:
Actually, what I did was

`git pull upstream master`
`git checkout -b perl-Math-Prime-Util`
modifyied the templates and then
`git push origin perl-Math-Prime-Util`

chneukirchen:
If you look at https://github.com/newbluemoon/void-packages/commits/perl-Math-Prime-Util , you see that you forked from 3e1adc69f4, which was before 6a4bc5aa0e which introduced Math::Prime::Util.

newbluemoon:
Ok, somwhere along the way I messed it up ;-) Thanks for your help, I will figure out where I made the mistake. But now all’s well anyway :)

chneukirchen:
You can also run `git pull --rebase origin master` on your topic branch to update it locally.

newbluemoon:
Ah, this might be it! Thank you! :)

PR #7553, opened 2017-09-02
New package: perl-Stow

harrigan:


srcpkgs/perl-Stow/template
```diff
@@ -0,0 +1,15 @@
...
+wrksrc="Stow-v${version}"
+build_style=perl-ModuleBuild
+hostmakedepends="perl perl-Module-Build"
+makedepends="${hostmakedepends}"
+depends="${makedepends}"
+short_desc="Perl module for managing symlinks"
+maintainer="Martin Harrigan <martinharrigan@gmail.com>"
+homepage="http://search.cpan.org/dist/Stow"
+license="Artistic, GPL-1"
+distfiles="http://search.cpan.org/CPAN/authors/id/A/AS/ASPIERS/Stow-v${version}.tar.gz"
```
the-maldridge: I'd be much happier merging this if you used CPAN_SITE here.

the-maldridge:
I'm reasonably sure this should be a noarch package as well.

harrigan:
I changed the distfiles to `${CPAN_SITE}/Finance/ASPIERS/Stow-v${version}.tar.gz` even though it is an odd location -- the module has nothing to do with finance. Also, I intend to follow this up with a package for Gnu Stow which wraps the above in a CLI.

the-maldridge:
Looks good but this should be only one commit, please rebase and squash.

PR #7556, opened 2017-09-04
xbps-src: use -1 to trigger errors on missing deps

Vaelatern:
This has been tested, and works as follows:

Set a magic flag on `-1` being passed in. If there are missing dependencies, fail hard instead of recursively building them.

In the steps where dependencies are computed and, if needed, built, this iterates over the arrays of missing dependencies and, if there is a problem, `msg_error`s (which in a non-informative-run `-i` causes an `exit`).

The `XBPS_BUILD_ONLY_ONE_PKG` variable is also respected if in the environment.

```
P  |B
K  |u
G  |i
   |l
H F|d
e l|
r a|o
e g|k
? ?|?
-----
N N|Y
N Y|N
Y N|Y
Y Y|Y
```


Vaelatern:
This is wanted for `dxpb` to ensure a failure in the graph (or remote builders not yet seeing a new sync of the repo) doesn't cause a recursive build as plague us today.

Side effect: if this is turned on for the builders we get absuvie but working graph dependency. I think the builders would require multiple pushes to rebuild a pkg after failure (so rebuilds could well take longer), but it would fix building pkgs out of order.

common/xbps-src/shutils/build_dependencies.sh
```diff
@@ -365,6 +365,21 @@ install_pkg_deps() {
         missing_rdeps+=("${_realpkg}")
     done
 
+    if [ -n "$XBPS_BUILD_ONLY_ONE_PKG" -a "$XBPS_BUILD_ONLY_ONE_PKG" != "0" -a "$XBPS_BUILD_ONLY_ONE_PKG" != "no" ]; then
+           for i in ${host_missing_deps[@]}; do
+                   msg_error "dep ${i} not built: instructed not to build\n"
```
the-maldridge: these messages should probably explicitly state the flag that prevented the build.

Vaelatern:
@the-maldridge the modified message to your liking?

the-maldridge:
Works for me, but my general preferred format is actually
```
Not building <x> (-1 specified)
```

Its splitting hairs and not worth changing again, but I personally prefer the parenthetical references.

PR #7570, opened 2017-09-05
tvheadend: update package to 4.2.3.

aleroot:
Upgrade to the version 4.2 that contains a lot of new features that can be seen at this page . The package has been released the 1st of July and is pretty stable.

PR #7571, opened 2017-09-05
tvheadend: update package to 4.2.3.

aleroot:
Upgrade to the version 4.2 that contains a lot of new features that can be seen at this page . The package has been released the 1st of July and is pretty stable.


Vaelatern:
Stop making new PRS!

Vaelatern:
#7567 
#7568 
#7569 
#7570 
stop. use this PR number. And just learn to use git. The spam is getting to me.

aleroot:
I'm sorry I thought it was the right procedure, closing and reopening. I'l avoid in the future. Sorry again.

Vaelatern:
Thank you.

Vaelatern:
@lemmi can you review? the new deps concern me that something may have changed upstream , or not at all.

srcpkgs/tvheadend/template
```diff
@@ -1,18 +1,18 @@
...
-revision=7
+version=4.2.3
+revision=1
 wrksrc=tvheadend-${version}
 build_style=gnu-configure
-configure_args="--disable-dvbscan"
+configure_args="--disable-dvbscan --disable-ffmpeg_static --disable-hdhomerun_static --disable-bintray_cache"
 hostmakedepends="pkg-config python"
-makedepends="avahi-libs-devel libressl-devel zlib-devel libcurl-devel"
+makedepends="avahi-libs-devel libressl-devel zlib-devel libcurl-devel cmake wget"
```
Johnnynator: Since the cross compile builds don't fail I highly doubt that cmake and wget are needed, and if they are needed they nearly always would belong into `hostmakedepends` and not `makedepends`

lemmi:
@aleroot I can do a proper review this weekend. Only then will I have the proper hardware for it. There is also version 4.3 available, but I haven't got any of that to work when I last tried.

srcpkgs/tvheadend/template
```diff
@@ -1,18 +1,18 @@
 # Template file for 'tvheadend'
 pkgname=tvheadend
-version=4.1
-revision=7
+version=4.2.3
+revision=1
 wrksrc=tvheadend-${version}
 build_style=gnu-configure
-configure_args="--disable-dvbscan"
+configure_args="--disable-dvbscan --disable-ffmpeg_static --disable-hdhomerun_static --disable-bintray_cache"
```
Johnnynator: If you disable static support for ffmpeg and hdhomerun shouldn't you supply these libraries so it can link dynamically against it, idk what functionality they serve inside tvheadend but I doubt that they shouldn't simply deactivated completely?

aleroot: These are new services implemented with the latest release, ffmpeg is needed just to transcode the stream, but if you don't have a very powerful machine you can't do it ...

aleroot:
@lemmi The 4.3 is the current development version so it rather unstable at the moment, the 4.2.3 is the latest stable release that has been released in July. Let me know what you think, I'm gonna test it on the respberry pi.

Vaelatern:
We prefer stable releases, so 4.2.3 is a good call.

PR #7578, opened 2017-09-06
postgresql: update 9.6.5

arete:
Signed-off-by: Gerardo  Di Iorio <arete74@gmail.com>


Gottox:
Microupdate. Should work without a db migration 👍

PR #7579, opened 2017-09-06
New package: vulkan-loader-1.0.57.0, convmv-2.03, pepper-flash-26.0.0.151

inefarius:
Also added an iwd build option for ConnMan

Aplologize for the mess. I realize that this should be four separate commits. Unfortunately I did it all in one branch and don't really know how to disentangle it. The commit messages are also not to spec; will make sure they are in the future.

### DIFF
--- a/srcpkgs/gnome-online-accounts/template
+++ b/srcpkgs/gnome-online-accounts/template
@@ -1,6 +1,6 @@
 # Template file for 'gnome-online-accounts'
 pkgname=gnome-online-accounts
-version=3.24.2
+version=3.24.3
 revision=1
 build_style=gnu-configure
 configure_args="$(vopt_enable gir introspection)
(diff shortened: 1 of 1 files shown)


Johnnynator:
The easiest way to split them up would probably creating new branches from master and then cherry picking the needed commits. 

`git branch temp` create a temporary branch that has al your current commit
`git reset --hard upstream/master` so you have a clean master branch, I would recomend alway having one
then checkout into new branches e.g. `git checkout -b convmw` and then pick the needed commit e.g. `git cherry-pick a88d731` and `git commit --amend` for changing the name of the last commit.

For e.g. merging to commit you can use e.g. `git rebase -i HEAD~2` for editing the last two commits.

And also try to never merge the upstream master branch, alway try to rebase you own branch on top of it e.g. `git pull upsteam master --rebase` or just `git rebase master` if your local master branch is up to date.

PR #7583, opened 2017-09-07
New package: scudcloud-1.62

Guppster:


Vaelatern:
`noarch=yes`

Duncaen:
Doesn't use the python-module build-style, something is off with the dependencies too.
Mixed python 2 and 3 packages only in makedepends and no runtime dependencies.

PR #7587, opened 2017-09-08
New package: m17n-lib-1.7.0

newbluemoon:
m17n is a library to process multilingual text. Emacs uses this library to display e.g. Indic fonts correctly. The issue arose in [this forum thread](https://forum.voidlinux.eu/t/font-issues-in-void-musl-emacs/3350).

In m17n-lib there is a directory `/usr/lib/m17n/1.0/` with some .a and .so files. I’m not sure if they belong in the devel package.

m17n uses a database (m17n-db) which is created by using glibc’s charmaps. I added the dependency libthai (which depends on libdatrie).

And finally, I added a new (default) build option ‘m17n’ to emacs.


newbluemoon:
I think Travis fails when cross compiling because libdatrie needs to be built for the host and for the target which causes some trouble.

newbluemoon:
Adjusted PR to latest emacs update.

emacsomancer:
I can confirm that this PR fixes font display issues in Emacs.

newbluemoon:
In case it’s my workaround to get anthy cross compiled that is stalling a review then I could easily change it into a build option or only enable it when not cross compiling since it’s only used in the sample program m17n-edit (for which it is recommended to install though).

newbluemoon:
I changed ‘anthy’ into a build option for m17n-lib, because I think it’s not essential (it’s recommended for the <i>sample</i> program ‘m17n-edit’) and removed my idea to cross compile anthy for now. This might be better in a separate PR.

PR #7602, opened 2017-09-09
linux4.9: update to 4.9.48

pbui:


Gottox:
obsolete, but thanks for the effort! We really need to improve the merge process 😞

Issue #7599, opened 2017-09-09
Chromium building errors


faulesocke:
Hey there,

I just tried building chromium from sources:
`./xbps-src pkg -j6 chromium`

and got the following errors during the build phase:
```
FAILED: obj/breakpad/client/microdump_writer.o 
g++ -MMD -MF obj/breakpad/client/microdump_writer.o.d -DV8_DEPRECATION_WARNINGS -DUSE_UDEV -DUI_COMPOSITOR_IMAGE_TRANSPORT -DUSE_AURA=1 -DUSE_PANGO=1 -DUSE_CAIRO=1 -DUSE_GLIB=1 -DUSE_NSS_CERTS=1 -DUSE_X11=1 -DNO_TCMALLOC -DDISABLE_NACL -DFULL_SAFE_BROWSING -DSAFE_BROWSING_CSD -DSAFE_BROWSING_DB_LOCAL -DCHROMIUM_BUILD -DENABLE_MEDIA_ROUTER=1 -D_FILE_OFFSET_BITS=64 -D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -DNDEBUG -DNVALGRIND -DDYNAMIC_ANNOTATIONS_ENABLED=0 -I../../breakpad -I../../breakpad/src -I../../breakpad/src/client -I../../breakpad/src/third_party/linux/include -I../.. -Igen -I../../breakpad/src -fno-strict-aliasing --param=ssp-buffer-size=4 -fstack-protector -Wno-builtin-macro-redefined -D__DATE__= -D__TIME__= -D__TIMESTAMP__= -funwind-tables -fPIC -pipe -B/usr/bin -m64 -march=x86-64 -pthread -O2 -fno-ident -fdata-sections -ffunction-sections -g0 -fvisibility=hidden -Wno-unused-local-typedefs -Wno-maybe-uninitialized -Wno-missing-field-initializers -Wno-unused-parameter -fvisibility-inlines-hidden -std=gnu++11 -fno-delete-null-pointer-checks -fno-rtti -fno-exceptions -Wno-deprecated -Wno-narrowing -c ../../breakpad/src/client/linux/microdump_writer/microdump_writer.cc -o obj/breakpad/client/microdump_writer.o
In file included from ../../breakpad/src/client/linux/microdump_writer/microdump_writer.cc:41:0:
[...]

PR #7605, opened 2017-09-09
Fixed chromium compilation with new glibc

faulesocke:
These two patches make chromium compile with the new glibc. See #7599 and #7164. I've briefly verified that the compiled browser works as intended.


Gottox:
Not revbumping is a good catch here. We'll rebuild it with the next major change 👍

PR #7608, opened 2017-09-09
lapack: fix checksum

liwakura:
Seems to have changed upstream. 3.7.1 is an official release and was updated after its official release, so being an snapshot release was not the issue here.


Vaelatern:
Needs a revbump. What changed between changes in the release tag?

PR #7610, opened 2017-09-09
rng-tools: update to version 6

lonwillett:
Note that this update required a minor patch for compilation with
musl, and that the upstream location has moved.

Also note the command line options for rngd have changed.


Vaelatern:
@thypon any comments?

Issue #7604, opened 2017-09-09
qucs: broken after glibc update from 2.25 to 2.26


pienjo:
(also reported to qucs upstream)

The recent update of libgc to 2.26 seems to have broken qucs. While the GUI seems to work just fine, the actual simulator ("qucsator") crashes, even on trivial schematics. I have narrowed this down to the update of glibc; if I downgrade glibc and glibc-devel to 2.25_6 qucsator works again. 

Downgrading glibc is not a feasible workaround; there are several binaries that no longer work (including sudo)

newbluemoon:
The build date of qucs is 2017-07-26 and glibc 2.26 was introduced on 2017-08-05. So I did a rebuild of qucs against glibc 2.26, but it still crashes.

pienjo:
I have narrowed this down; it looks like it's a bug in qucs (which triggers undefined behaviour). Apparently, the result of the undefined behaviour is different when using glibc-2.26 than it is with glibc-2.25, but that's not glibc's fault..

I'm working on a patch

PR #7611, opened 2017-09-09
qucs: Fix simulator crash

pienjo:
Avoid undefined behaviour while traversing and modifying a std::list.
Patch has been reported to Qucs upstream.

Fixes https://github.com/voidlinux/void-packages/issues/7604


ebfe:
Patch doesn't seem to apply. Wrong path?

ebfe:
xbps-src sefault is `-p0`, patch is in `-p1` format.

pienjo:
Odd, it wfm. I was just about to modify this patch (so it is identical to the change in upstream), I'll double check.

PR #7617, opened 2017-09-10
keybase: update to 1.0.29.

Vaelatern:
@the-eater is it really worth keeping track of commit messages for this?


the-eater:
Hopefully not anymore: https://github.com/keybase/client/pull/8113#issuecomment-323104302

PR #7623, opened 2017-09-10
android-tools: add bash-completion

xtremae:


Vaelatern:
Where did you get the bash completion file? Do we really need the file in the repo or is it available from somewhere else?

xtremae:
I'm not happy with said mechanism either. Unlike adb, google don't provide bash completion for fastboot, therefore i had to create and include <code>bash_completion.fastboot</code>. Maybe we should close or defer the PR until google handle it on their end.

P.S. Apologies for reopening. I wasn't up to speed with the guide.

the-maldridge:
iirc the repositories for the upstream components are open source.  Please submit the file you have produced as a PR there and update this PR with a reference to that request.

PR #7632, opened 2017-09-11
New package: pepper-flash-26.0.0.151

inefarius:
Should this be added to the nonfree repository? I'm not sure how the licensing works for Flash.


inefarius:
The `adobe-flash-plugin` package seems to provide the NPAPI plugin. This package provides the PPAPI version, intended for use with chromium and freshplayerplugin. Do you not want to include this in the repo?

Duncaen:
But why call it pepperflash, pepperflash is/used to be the version shiped with chrome.
I think something like `adobe-flash-ppapi` would better.

inefarius:
I went with `pepper-flash` because it was less verbose, but I could change it if you want.

crocket:
The download URL is broken.

The new URL is https://fpdownload.adobe.com/pub/flashplayer/pdc/29.0.0.140/flash_player_ppapi_linux.x86_64.tar.gz

PR #7625, opened 2017-09-11
New package: libdockapp-0.7.2

dbear6:

PR #3809, opened 2016-03-11
New package: arc-themes-20151214

pbui:

PR #7634, opened 2017-09-11
New package: arc-theme-20170302

dbear6:


Vaelatern:
#4002 #3809 #7625

dbear6:
@Vaelatern I see, quite sad. I looked forward to using Void Linux as a desktop after I packaged all of the software I need. After this, I am not sure if Void is an acceptable option considering that every other major Linux distribution packages this theme. Thank you for approving my pull requests thus far.

Vaelatern:
Themes can be installed into home dirs, so we avoid packaging them.

PR #7636, opened 2017-09-12
ETL: update to 1.2.1.

diogoleal:


Vaelatern:
This package should be `noarch=yes`

diogoleal:
Ok, added ```noarch=yes```

PR #7639, opened 2017-09-12
libcaca: add (optional) support for imlib2

newbluemoon:
According to `img2txt`’s manpage libcaca needs imlib2 to load other image formats than regular BMP files. 

Because `imlib2-devel` needs `libX11-devel` (else the build fails with `/usr/include/Imlib2.h:28:22: fatal error: X11/Xlib.h: No such file or directory`; it’s not a package dependency, though) and there already is a build option ‘x11’ I just added `imlib2-devel` to that option and didn’t enable it per default.

Alternatively one could add `libX11-devel` as a dependency to `imlib2-devel` (should it be added anyway because Imlib2.h includes Xlib.h?) and add an extra build option.

### DIFF
--- a/srcpkgs/libcaca/template
+++ b/srcpkgs/libcaca/template
@@ -1,12 +1,12 @@
 # Template file for 'libcaca'
 pkgname=libcaca
 version=0.99.beta19
-revision=7
+revision=8
 build_style=gnu-configure
 hostmakedepends="libtool automake pkg-config"
 short_desc="Graphics library that outputs text instead of pixels"
 maintainer="Juan RP <xtraeme@voidlinux.eu>"
-makedepends="ncurses-devel $(vopt_if x11 libX11-devel)"
+makedepends="ncurses-devel $(vopt_if x11 'libX11-devel imlib2-devel')"
 homepage="http://caca.zoy.org/wiki/libcaca"
 license="WTFPL"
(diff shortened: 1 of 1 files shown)


ebfe:
> Because imlib2-devel needs libX11-devel [...]

Please add it to imlib2-devel depends then.

newbluemoon:
Done. Should I then add a build option for imlib2 to libcaca or should I enable it per default (which would make the build option ‘x11’ superfluous)?

ebfe:
The correct way to disentangle x11 from imlib2 is building with -DX_MISSING_DISPLAY as libcaca master does. Enabling x11 by default doesn't sound right.

ebfe:
Eh, of course supporting imlib2 in the default would always pull in libX11. regardless of build flags.

newbluemoon:
> Enabling x11 by default doesn't sound right.

That’s what I thought, too. :)

So I would propose to just use one build option ‘imlib’ which includes the old ’x11’.

ebfe:
I think i liked the previous version better that is: build_option still named x11 but unconditionally enableing imlib2 when x11 is enabled. That way we don't break setups already having x11 enabled.

ebfe:
cc @voidlinux/pkg-committers

newbluemoon:
Ok. That sounds very reasonable. I changed it back. :)

PR #7640, opened 2017-09-12
New package: onboard-1.4.1

circhioz:


srcpkgs/onboard/template
```diff
@@ -0,0 +1,19 @@
...
+revision=1
+wrksrc="onboard-${version}"
+build_style=python3-module
+patch_args="-Np1"
+hostmakedepends="intltool pkg-config python3-distutils-extra"
+makedepends="dconf-devel eudev-libudev-devel gtk+3-devel hunspell-devel
+ libcanberra-devel libX11-devel libXi-devel libXtst-devel libxkbfile-devel
+ python3-devel"
+depends="iso-codes python3-cairo python3-gobject"
+nocross=yes
```
Vaelatern: Really?

srcpkgs/onboard/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'onboard'
+pkgname=onboard
+version=1.4.1
+revision=1
+wrksrc="onboard-${version}"
```
Vaelatern: Please remove line. This is the default.

srcpkgs/onboard/template
circhioz: onboard depends on python3-gobject which cannot be cross-compiled.
I'll remove ```nocross``` line since travis builds don't fail without it.

circhioz:
Thanks @Vaelatern, I also added the missing `pycompile_module="Onboard"`. Are there any other problems?

PR #7642, opened 2017-09-12
tracker: update to 2.0.0

Sir-Boops:
* Only tested on x86_64

### DIFF
--- a/srcpkgs/tracker/template
+++ b/srcpkgs/tracker/template
@@ -1,6 +1,6 @@
 # Template file for 'tracker'
 pkgname=tracker
-version=1.12.1
+version=2.0.0
 revision=1
 build_style=gnu-configure
 configure_args="--enable-libflac --enable-libvorbis --disable-unit-tests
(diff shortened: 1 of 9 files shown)


srcpkgs/tracker/template
```diff
@@ -32,6 +32,10 @@ if [ -z "$CROSS_BUILD" ]; then
 	build_options_default="gir"
 fi
 
+case "$XBPS_TARGET_MACHINE" in
+	arm*) broken="sqlite will not cross compile with fts5 support"
+esac
```
Gottox: sqlite has `--enable-fts5` enabled. Is this a noop?

Sir-Boops: For cross it seems like it. As non cross compiled versions work fine

Gottox: Strange. But ok. Do you have time to fix sqlite here? - I can give it a try tomorrow.

Sir-Boops: I have some time today to work at it.

ebfe:
This left the common/shlibs entries for libtracker-1.0. brasero/bijben need a revbump because they depend on the old sonames.

Gottox:
Even worse: It's an API change, not an ABI change.

Issue #4138, opened 2016-05-05
Package request: cura


chneukirchen:
3D model slicing software:

https://ultimaker.com/en/products/cura-software
https://github.com/daid/Cura

I think we have most deps already.

Vaelatern:
Duplicates #3586

Issue #3586, opened 2016-02-11
Package request: Cura


Vaelatern:
Someone in IRC suggested https://github.com/daid/Cura

ke7ofi:
That would be neat. Is there *any* slicer currently in a Void package?

karl-nilsson:
I'm working on packaging Cura and its dependencies:

- Uranium
- CuraEngine
- libArcus
- Clipper

any help is appreciated

PR #7645, opened 2017-09-13
new package: cura

karl-nilsson:
based on @equalsraf 's earlier work, fixes a few problems, and updated for v2.7.0
I've built and tested all the packages on x86, glibc

ref #4138, #3586


Vaelatern:
Fix the conflcit?

karl-nilsson:
@Vaelatern fixed

PR #7646, opened 2017-09-13
go-ipfs: use gx

lemmi:
@CMB I went ahead and packaged `gx`, since I'll probably need it for dht-node. I thought I work it into go-ipfs. Thoughts?


lemmi:
Don't know why travis doesn't run: https://travis-ci.org/lemmi/void-packages/builds/274912947
Cross compile works locally.

CMB:
lemmi <notifications@github.com> writes:

> Don't know why travis doesn't run: https://travis-ci.org/lemmi/void-packages/builds/274912947
> Cross compile works locally.

I couldn't get it to work reliably, so that's why I grab all the stuff
from gx before doing a go-ipfs package.

-- Chris

ebfe:
> Don't know why travis doesn't run: https://travis-ci.org/lemmi/void-packages/builds/274912947

This is a common issue and is so far only seen at travis. When packages, that are built as dependencies, share a common dependency, it is/might be removed in the "remove autodeps" step.

lemmi:
@CMB I went ahead and packaged gx. travis will fail since repo3 seems to have problems. If you don't like the current approach, I have one other idea: we could package a go-ipfs-deps package that includes all gx-downloaded sources. This could be noarch aswell. Essentially this will be the same setup for the ipfs package, but everything done on the buildservers.
But this might be unnecessarily complex. If you don't have anything against the current PR, I'd like to merge it, possibly introduce the dependency package if we can think of a good reason to do so.

CMB:
If you're cool with travis not working, go ahead and merge it.  This
looks fine otherwise.  I think it's pretty close to what I was playing
with last year, before I gave up on gx due to travis.

lemmi:
@CMB let's give it a try then. If it fails, we can revert.

lemmi:
https://github.com/voidlinux/void-packages/commit/9b1e9c091ca083c66800436a96c73efd1eae962c

PR #7651, opened 2017-09-13
wmmon: updated dependencies

dbear6:


dbear6:
This was done because when I do `xbps-install -Su wmmon` I get the following error

> MISSING: libdockapp>=0.72_1
> Transaction aborted due to unresolved dependencies.

PR #7652, opened 2017-09-13
libdockapp: Corrected incorrect entry in shlibs

dbear6:


Duncaen:
This would also requiere a rebuild of everything that links `libdockapp.so.3`.

PR #7654, opened 2017-09-14
New package: ghostwriter-1.5.0

abenson:


abenson:
This seems to suffer from the qt5-qmake cross compile problems that other packages have also had.

Johnnynator:
adding the qt5 devel packages also to hostmakepedends for cross compiling normally fixes cross compiling

abenson:
I tried that, but then I started getting other linking issues, specifically with hunspell, and couldn't even build it natively.

abenson:
Okay, I have cross-compiling working now. For some reason Travis is failing before it even gets to build this, any suggestions on that?

abenson:
Ooh, shiny green.

PR #7655, opened 2017-09-14
common/travis: switch to mirror for travis builds

Gottox:
This PR switches travis to an US mirror. This should result in less traffic on the main repository and faster build times on travis as the bits don't have to cross oceans anymore.


ebfe:
This broke travis builds because of:

- repo3. and repo. are out of sync
- binary-bootstrap still happens from repo.

PR #7660, opened 2017-09-14
sublime-text3: updated to build 3143

nam3less:
Hello,
because the sublime-text3 package was pretty outdated, I updated it.
I did bump the version number, so the right tarball gets fetched. Furthermore I updated the checksums.
On my machine both packages build using xbps-src and the 64bit version installed.

Because this is my first pull request on a big project and voidlinux, please correct me if I'm doing something wrong.


Vaelatern:
Exactly.

PR #7664, opened 2017-09-15
New package: higan-v104

RogueBeet:


srcpkgs/higan/template
```diff
@@ -0,0 +1,23 @@
...
+wrksrc=higan_$version-source
+nopie=yes
+do_build() {
+make -C higan compiler=g++
+make -C icarus compiler=g++
+}
+do_install() {
+make -C higan install  prefix=/destdir/${pkgname}-${version}/usr
+make -C icarus install prefix=/destdir/${pkgname}-${version}/usr
+}
```
Vaelatern: Functions at the bottom. Actually indent the body of those functions. Remove both `do_build` and `do_install` because they should not be used.

srcpkgs/higan/template
```diff
@@ -0,0 +1,23 @@
+# Template file for 'higan'
+pkgname=higan
+version=v104
+revision=1
+build_style=meta
+wrksrc=higan_$version-source
+nopie=yes
```
Vaelatern: Wrong. No reason given.

srcpkgs/higan/template
```diff
@@ -0,0 +1,23 @@
+# Template file for 'higan'
+pkgname=higan
+version=v104
```
Vaelatern: Please strip the v from this.

srcpkgs/higan/template
```diff
@@ -0,0 +1,23 @@
...
+revision=1
+build_style=meta
```
Vaelatern: No.
No.
No.
Use a real `build_style`

srcpkgs/higan/template
```diff
@@ -0,0 +1,23 @@
...
+make -C higan install  prefix=/destdir/${pkgname}-${version}/usr
+make -C icarus install prefix=/destdir/${pkgname}-${version}/usr
+}
+hostmakedepends="p7zip make pkg-config gcc"
+makedepends="gtk+-devel SDL-devel gtksourceview2-devel libXv-devel libao-devel freealut-devel libgomp-devel freetype-devel"
+short_desc="Multi-system emulator (SNES/Super Famicom and other older game consoles)"
+maintainer="RogueBeet <the@roguebeet.com>"
+license="GPL-3"
+homepage=:"https://byuu.org/emulation/higan/"
+distfiles="https://download.byuu.org/higan_$version-source.7z"
```
Vaelatern: Once you strip the `v` from the top, you need `distfiles="https://download.byuu.org/higan_v${version}-source.7z"`

srcpkgs/higan/patches/locate_shared_files.patch
```diff
@@ -0,0 +1,24 @@
...
+@@ -8,6 +8,9 @@
+   string location = {Path::program(), name};
+   if(inode::exists(location)) return location;
+ 
++  location = {Path::shared(), "icarus/", name};
++  if(inode::exists(location)) return location;
++
+   location = {Path::config(), "icarus/", name};
+   if(inode::exists(location)) return location;
+
```
Vaelatern: Why is this patch here?

RogueBeet: By default higan only tries to look for the required manifest files in ~/.local , this patch makes it so that users don't have to manually copy the files from /usr/share/higan to that location.

srcpkgs/higan/template
RogueBeet: I used this method because I need to use make two times and I couldn't figure out how to do that with make_builds_arg. Is it possible? Is the functions not even allowed if i use gnu-makefile as build style?

srcpkgs/higan/template
RogueBeet: The compilation failed when I tried to use -fPIE  with g++ that's why I included the nopie line

PR #7665, opened 2017-09-15
bluez: update to 5.47, added sixaxis support, removed pie flag

RogueBeet:
Hope this is acceptable, don't know if it's ok to modify and update packages at the same time

### DIFF
--- a/srcpkgs/bluez/template
+++ b/srcpkgs/bluez/template
@@ -1,10 +1,10 @@
 # Template file for 'bluez'
 pkgname=bluez
-version=5.46
+version=5.47
 revision=1
 build_style=gnu-configure
 configure_args="--with-udevdir=/usr/lib/udev --disable-systemd
- --enable-pie --enable-threads --enable-library"
+ --enable-sixaxis --enable-threads --enable-library"
 hostmakedepends="automake libtool pkg-config flex"
 makedepends="libglib-devel eudev-libudev-devel dbus-devel
  cups-devel libical-devel readline-devel"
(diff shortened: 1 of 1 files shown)


Gottox:
It's okay to update and change.

Please use a proper commit mesage `bluez: update to 5.47, added sixaxis support, removed pie flag`

Also, why was the enable-pie flag removed?

RogueBeet:
I already changed the typo (higan -> bluez) in the commit message: 
https://github.com/voidlinux/void-packages/pull/7665/commits/95e7e4160b432d128f1812c425b6d06c87af39a4

Wasn't that enough? If that wasn't what you're refering to please explain more.

The enable-pie flag was removed because I thought it was better for the build system to add -fPIE, can't remember if I confirmed that it did or not, but the package was built without complaints and went trough the checks.

RogueBeet:
never mind, seems all that was changed was the title on https://github.com/voidlinux/void-packages/pull/7665 , what is the easiest way to fix that with git?

Johnnynator:
```
git commit --amend
git push -f
```
Probably the easiest way of changing the commit message.

RogueBeet:
Not sure, but it seems like that would only apply if I didn't already push push the change to github?
See https://help.github.com/articles/changing-a-commit-message/

I will wait to try to fix everything until I get a response about the pie flag and final confirmation on what the proper way to fix the commit message is.

RogueBeet:
Sorry Johnnynator, didn't notice the -f flag, I suppose that is the the --force flag they are talking about in the link in my previous message. Still not sure if this will break anything or if I should go this route.

Johnnynator:
The force flag will break nothing in this case, you just have to be aware that the force flag will 'remove' any commit that is not in your local repo.

RogueBeet:
Ok tried it, hope everything works and I that I didn't make any mistakes.

This is the output from configure, without the pie flag:

checking for cc option to produce PIC... -fPIC -DPIC
checking if cc PIC flag -fPIC -DPIC works.. yes

Not sure if that means that it's using it when compiling or if the build script just checks if it would be possible.

Johnnynator:
@RogueBeet Does look fine to me.
To me it looks like it doesn't make a differnece if enable_pie is set or not since the build script just adds -fPIC to CFLAGS and -pie to LDFLAGS which is also already added by xbps.

ebfe:
Needs a revbump, bluez was already at 5.47 at time of merge.

Gottox:
done, thx for the hint!

PR #7667, opened 2017-09-15
New package: iridium-58.0

faulesocke:
Iridium is a modified version of chromium intended to enhance user privacy. The template is basically a copy of the chromium template with minor fixes to it.


ebfe:
The general consensus seems to be to not add any more browsers to repo for now.

faulesocke:
Okay, I accept that but would like to note, that Iridium is easy to maintain since it's *really* similar to chromium. And @Gottox tried to include it back in 2016 but it failed for some reason.

faulesocke:
Is there still interest in this? Otherwise I'll close the PR.

travankor:
@faulesocke Upstream is no longer in sync with chromium's release schedule. Maybe you can help package [this](https://github.com/Eloston/ungoogled-chromium/issues/375) instead?

avoidr:
I would have liked Iridium in the repos, as it's the only google-free Chromium-based browser. But since you reenable everything that is non-free, what is the point?

@faulesocke, does the template work?

faulesocke:
@avoidr it worked back when I created it. Not sure about any options I choose there. I don't use this browser anymore since Firefox 57 came out.

PR #7668, opened 2017-09-15
New package: arp-scan-1.9

faulesocke:
A small utility to discover and fingerprint hosts on the local network.


srcpkgs/arp-scan/template
```diff
@@ -0,0 +1,18 @@
+# Template file for 'arp-scan'
+pkgname=arp-scan
+version=1.9
+revision=1
+build_style=gnu-configure
+makedepends="automake libpcap-devel"
```
Johnnynator: automake should be in hostmakedepnds since it is needed for the build host.

faulesocke:
Okay, I've changed the commit, automake is now in hostmakedepends.

srcpkgs/arp-scan/template
```diff
@@ -0,0 +1,19 @@
...
+license="GPL-3"
+homepage="http://www.nta-monitor.com/tools/arp-scan/"
+distfiles="https://github.com/royhills/$pkgname/archive/$version/$pkgname-$version.tar.gz"
+checksum=b2a446a170e4a2feeb825cd08db48a147f8dffae702077f33e456c4200e7afb0
+
+pre_configure() {
+	autoreconf --install
+}
+
+# vim: set sw=4 ts=4 noet:
```
Vaelatern: no.

faulesocke: Sorry, I don't understand?

Johnnynator: He is probly speaking about you vim line.

faulesocke: Hmm, I grep'ed for vim modelines before commiting and found a few other templates containing some so I thought it's not a problem. If it is, I can of course remove it.

faulesocke:
Any progress? What should be changed in the template to get it merged?

Johnnynator:
You could either mark armv7* as broken or fix it, otherwise it looks fine to me.

faulesocke:
Fixed cross compilation by backporting a newer commit.

PR #7672, opened 2017-09-15
go-jira: update + golang dep support

thypon:


common/build-style/go.sh
```diff
@@ -3,12 +3,17 @@
...
 
 do_build() {
+	local path="${GOPATH}/src/${go_import_path}"
 	if [[ "${go_get}" != "yes" ]]; then
-		local path="${GOPATH}/src/${go_import_path}"
 		mkdir -p "$(dirname ${path})"
 		ln -fs $PWD "${path}"
 	fi
 
+	if [[ -x /usr/bin/dep ]]; then
```
ebfe: Can we rather detect this by checking for Gopkg.{lock,toml}? Or explicitly enable it with an variable?

PR #7676, opened 2017-09-15
new package: cura

karl-nilsson:
re-do #7645 , because I messed up and broke the previous PR.


Gottox:
looks okay at first glance, any other opinions?

PR #7679, opened 2017-09-16
New package: lksctp-tools-1.0.17_1

bitshark:
New package: lksctp-tools-1.0.17_1
A package with Linux Kernel Stream Control Transmission Protocol Tools
http://lksctp.sourceforge.net/
**I created this package because it is a build dependency needed in the process of bootstrapping openjdk7 on Void musl**


the-maldridge:
You've got some comments sitting in the template that makes me think this isn't fully baked.  Is this ready for review?

PR #7682, opened 2017-09-16
New package: gcompat-0.0.1

djames1:


silvernode:
Wow, I can't believe we didn't have this already. Seems pretty useful.

djames1:
@silvernode yes, I think that if this library makes it beyond 0.0.1 it will be extremely useful to musl users

djames1:
They don't have the file LICENSE in the tarball, but we can't build ISC licensed applications without LICENSE file... I may fork and make their current repo changes be 0.0.2. They've since added LICENSE since 0.0.1 was created.

djames1:
Fixing ARM issues. Appears that the issue is where gcc is hard coded into makefile.

djames1:
Sadly this package does not currently build against Void Linux's version of GCC for aarch64-musl.
However, it is possible because this package builds against aarch64-musl in Alpine Linux. I will investigate at a later date.
https://git.alpinelinux.org/cgit/aports/tree/testing/gcompat/APKBUILD

ebfe:
-1 on forking and tagging a new version to package it.

djames1:
@ebfe in that case what do you suggest? The 0.0.1 release does not have a LICENSE file and we need that file for ISC licensed packages. That, and 0.0.1 is missing a lot of beneficial changes, and if project doesn't start having more commits again I plan to fork it and work on my fork.

djames1:
I realized I can add the LICENSE file to files/LICENSE and then do `vlicense ${FILESDIR}/LICENSE`....

I guess I should have thought of that before. Reverting to upstream version 0.0.1

PR #7685, opened 2017-09-16
New package: julia-0.6.0

espectalll:
The package is not 100% clean, as some dependencies are compiled directly within Julia (not statically!) instead of having their own, independent packages. However, these are only being used for Julia at the moment, so for now it should be fine (I'll try to clean everything up in the next few days).

Here's the list of each library currently unavailable at `void-packages` and compiled within Julia:
- OpenLibm: https://github.com/JuliaLang/openlibm
- OpenSpecFun: https://github.com/JuliaLang/openspecfun
- DSFMT: http://www.math.sci.hiroshima-u.ac.jp/~m-mat/MT/SFMT/
- AMOS: http://netlib.org/amos
- SuiteSparse: http://faculty.cse.tamu.edu/davis/suitesparse.html
- utf8proc (seems not to be `libutf8proc`?): https://github.com/JuliaLang/utf8proc
- libosxunwind: https://github.com/JuliaLang/libosxunwind

Also, libuv is currently not being used from `void-packages` due to resulting compiling errors on first attempts. This could be fixed just by editing the `template`.


chneukirchen:
Using these vendored packages, esp. as they are patched is ok from my side.

For i686, please adjust the march flags to make the build pass.

I guess for musl some patching will be required (or may not work currently: https://github.com/JuliaLang/julia/issues/16792)

Please add `/etc/julia/juliarc.jl` as a conf_file.

chneukirchen:
Making this all cross compile is probably hard, feel free to add `nocross=yes`.

silvernode:
@espectalll Thanks again for packaging this

espectalll:
No ARM or Musl support for now. If anyone has machines running Void with any of these, please help to make the package compatible!

silvernode:
woot, hopefully it gets merged!

the-maldridge:
This is insanely special purpose software and unlikely to be ever totally compatible with our packaging standards.  This is being merged on a best effort basis with the understanding that parts of it may make the system less stable.

the-maldridge:
Merged with manual rebase in dfed30c462

PR #7686, opened 2017-09-16
roxterm: unbreak

newbluemoon:
roxterm’s configure stage fails with

    Searching for program po4a-gettextize... not found

`po4a-gettextize` is part of po4a for which I created a package with it’s dependencies. The sombok (line breaking algorithm) package has an optional dependency on libthai which I already included in PR #7587 so I didn’t include it in this one. It can be added later.


chneukirchen:
Thank you, i wanted to have perl-SGMLSpm anway. ;)

newbluemoon:
I made a couple of other perl packages I needed locally, but wasn’t sure if they should be added to the repository without any dependency. So if you need one of them, too, just let me know :). Here’s the list:

perl-CGI
perl-Class-Inspector
perl-DateTime-Format-Strptime
perl-Eval-Closure
perl-File-ShareDir
perl-MRO-Compat
perl-namespace-autoclean
perl-Package-DeprecationManager
perl-Params-ValidationCompiler
perl-Role-Tiny
perl-Specio
perl-Sub-Identify
perl-Sub-Name
perl-Test-Fatal

Or should I add them anyway? I’m not sure what the policy is.

chneukirchen:
at least perl-CGI and perl-Sub-Name are generally useful I think.

Maintaining pure Perl modules is zero hassle, and well, XS take one rebuild per major Perl version, which is not that bad either.

So feel free to add them.

newbluemoon:
I made those packages some time ago, but completely forgot about why: I just needed perl-CGI and perl-DateTime-Format-Strptime which depends on perl-DateTime. But Void’s perl-DateTime was lacking some dependencies of dependencies ... which are the above mentioned packages. I will clean up the templates and open a PR to fix perl-DateTime.

PR #7688, opened 2017-09-16
New package: numix-icon-theme-17.04.05

jearbear:
This set doesn't use versions, so I used the tag instead. Is that too much of a problem?


jearbear:
Looks like the tests failed because the files couldn't be listed quickly enough on arm?

the-maldridge:
This is a theme that is just unpacked to use, as a general rule we do not package these.

PR #7694, opened 2017-09-17
New package: thingylaunch-2.0.3

maxice8:


srcpkgs/thingylaunch/template
```diff
@@ -0,0 +1,19 @@
...
+configure_args="-DUSE_XCB=ON"
+hostmakedepends="pkg-config"
+makedepends="xcb-util-wm-devel xcb-util-keysyms-devel"
+short_desc="X11 application launcher with tab-completion, history, and bookmarks"
+maintainer="maxice8 <thinkabit.ukim@gmail.com>"
+license="BSD"
+homepage="https://gahr.ch/thingylaunch/"
+distfiles="https://github.com/gahr/${pkgname}/archive/${version}.tar.gz"
+checksum=4d9cc6b826bddb3d02836f721316bfc24c2b31f75c9c2d5077ae162eb3d60b60
+nopie=yes
```
Vaelatern: I don't trust this flag. What's the reason for it? I bet it isn't that great a reason, but has something to do with either our LDFLAGS or CFLAGS not being respected. Want to check for me if the compilation and linking happen in the same step? I bet that, in a pre_configure(), a `CFLAGS="${CFLAGS} ${LDFLAGS}"` might fix this.

Either way, the flag is almost definitely wrong.

maxice8: > I don't trust this flag. What's the reason for it?

Builds fails with an error during linking: [log](https://p.iotek.org/v0f)

>  Want to check for me if the compilation and linking happen in the same step?

According to logs both compilation and linking happen during `fdo_build()` 

> I bet that, in a pre_configure(), a CFLAGS="${CFLAGS} ${LDFLAGS}" might fix this.

Sorry in advance if i'm doing it wrong ( i'm new to this ), but adding:

```
pre_configure() {
    CFLAGS="${CFLAGS} ${LDFLAGS}"
}
```
to the template doesn't seem to fix it: [log](https://p.iotek.org/ab8)

Vaelatern: So our CFLAGS are not being respected at build time. This can be patched to work, in the build scripts.

maxice8: This [Line](https://github.com/gahr/thingylaunch/blob/master/CMakeLists.txt#L8) seems to be causing the problems by overriding the cflags that are set up, removing it allows the compilation to go smoothly *

\* tested locally on: `x86_64 x86_64-musl i686 aarch64 aarch64-musl armv7hf-musl

the-maldridge:
@maxice8 looks like you found the issue.  Can you amend this PR with the change you described above?

maxice8:
Yes just finish a loose end.

maxice8:
Pushed changes asked @the-maldridge

PR #7696, opened 2017-09-17
New package: transmission-remote-gtk

SolitudeSF:


srcpkgs/transmission-remote-gtk/template
```diff
@@ -0,0 +1,14 @@
...
+pkgname=transmission-remote-gtk
+version=1.3.1
+revision=1
+build_style=gnu-configure
+build_options="geoip"
+hostmakedepends="pkg-config intltool"
+makedepends="$(vopt_if geoip geoip) appstream-glib-devel libcurl-devel gtk+3-devel json-glib-devel libnotify-devel libproxy-devel libmrss-devel"
+short_desc="GTK client for remote management of the Transmission BitTorrent client"
+maintainer="SolitudeSF <solitudesf@protonmail.com>"
+license="GPL-3"
```
Gottox: This is not GPL-3.

srcpkgs/transmission-remote-gtk/template
```diff
@@ -0,0 +1,14 @@
+# Template file for 'transmission-remote-gtk'
+pkgname=transmission-remote-gtk
+version=1.3.1
+revision=1
+build_style=gnu-configure
+build_options="geoip"
+hostmakedepends="pkg-config intltool"
+makedepends="$(vopt_if geoip geoip) appstream-glib-devel libcurl-devel gtk+3-devel json-glib-devel libnotify-devel libproxy-devel libmrss-devel"
```
Gottox: 👍 for using build options 😄

ebfe: any reason to not enable geoip by default?

SolitudeSF:
added geoip as default and changed license to gpl-2

PR #5709, opened 2017-02-10
numix-themes: build it instead of getting it pre-built.

jcjordyn130:
Upstream doesn't provide pre-built tarballs anymore it seems.

PR #7698, opened 2017-09-17
New package: numix-gtk-theme-2.6.6

jearbear:
Fixed the issue where the theme wasn't actually being built and installed as reffered to in https://github.com/voidlinux/void-packages/pull/5709. I changed the name to be consistent with upstream (https://github.com/numixproject/numix-gtk-theme), but if this is a problem I'll change that back.


Vaelatern:
Is there a way for users to seamlessly migrate to the new name?

djames1:
Interesting, it was my understanding that themes were not allowed to be merged because xtraem refuses to merge those.

jearbear:
@djames1 Interesting, I wasn't aware of this. For what it's worth there are a bunch of themes in the repo already.

@Vaelatern is it possible to make the two packages conflict? I don't know if there are facilities for upgrading packages to a different name, but based on https://github.com/voidlinux/void-packages/pull/5192 and the fact that the old package (lemonbuddy) still exists I'm guessing no?

Vaelatern:
@djames1, this is legacy, so it's possible the rules under which it were merged apply.

@jearbear it's possible, but not straightforward. I think reverting to the old name will make this PR much smoother.

jearbear:
@Vaelatern Sounds good, reverted to the old name.

Duncaen:
Can't this use `sassc` instead of installing the gem?

jearbear:
I tried, but it seems that the Makefile calls `scss` with some options that are specific to it, `sassc` isn't really a drop-in replacement unfortunately. I was going to modify the Makefile to try to make it work, but I read somewhere that that is frowned upon.

jearbear:
Bumping this, anything else I need to change here?

the-maldridge:
We're okay with patches to the makefile that are needed for things like cross builds and the like, but in this case this is an acceptable solution.

jearbear:
Looks like this got taken care of already by https://github.com/voidlinux/void-packages/pull/5709, closing

PR #7701, opened 2017-09-17
New package: xcursor-vanilla-dmz-0.4.4

jearbear:


Vaelatern:
We have a general policy of not merging themes that can just as easily be installed in a home directory.

Is this such a theme?

jearbear:
@the-maldridge This is actually distinct from https://github.com/voidlinux/void-packages/pull/7702 as it's a separate theme (different color)

PR #7702, opened 2017-09-17
New package: xcursor-vanilla-dmz-aa-0.4.4

jearbear:


Vaelatern:
We have a general policy of not merging themes that can just as easily be installed in a home directory.

Is this such a theme?

jearbear:
Yeah that's fair, this can definitely be installed in the home directory. It does require a few extra steps since it's not packaged in a way that just requires simple copying.

the-maldridge:
Its a few extra K and it does require make to be available at build.

Merged.

PR #7704, opened 2017-09-17
caddy: update to 0.10.9.

SevereOverfl0w:


lemmi:
I'd like to wait for https://github.com/mholt/caddy/commit/545fa844bbd188c1e5bff6926e5c410e695571a0 to be included in a caddy update.

Vaelatern:
Seconded. NOMERGE until subsequent update, please.

chneukirchen:
Irrelevant, the EULA never applied to the open source code release.

This is primarily a big communication fuck up by them.

lemmi:
@chneukirchen not entirely irrelevent. You are right this never applied, but I have absolutely no interest on serving these ridiculous headers. The commit I linked was the last commit in a chain that reverted also the Sponsor header thing.

SevereOverfl0w:
Looks like this release contains a significant bug fix:
https://github.com/mholt/caddy/releases

Given that the eula doesn't apply, I see no reason to avoid merging this.

If you don't want to run this, pin the version in xbps.

I don't think void should be making value decisions about the packages
themselves.

lemmi:
I don't want the headers in my traffic. I'm fine with with a 0.10.9 version as long it comes with the headers removed.

Gottox:
Either we wait for a new version or include this patch: https://github.com/mholt/caddy/commit/b6e10e3cb22990eaa0d9ff28256054a25efd25be

KenjiTakahashi:
FYI: 0.10.10 released

SevereOverfl0w:
I will open a PR for that instead then

Issue #2987, opened 2015-11-16
Package Request: lilypond


sunaku:
http://lilypond.org/

> LilyPond is a music engraving program, devoted to producing the highest-quality sheet music possible. It brings the aesthetics of traditionally engraved music to computer printouts. LilyPond is free software and part of the GNU Project.

Vaelatern:
Dependency list fails during configure...

`ERROR: Please install required programs:  International New Century Schoolbook fonts (make sure the fc-list utility can see them, or use --with-ncsb-dir) mf-nowin mf mfw mfont mpost kpsewhich metapost CTAN package (texlive-metapost) guile-config (guile-devel, guile-dev or libguile-dev package) or guile1-config (guile1-devel package) libguile (libguile-dev, guile-devel or guile-dev package). GUILE-with-rational-bugfix t1asm pkg-config perl gs texi2html dblatex bibtex pdflatex pdfetex pdftex etex epsf.tex lh CTAN package (texlive-lang-cyrillic or texlive-texmf-fonts) pngtopnm convert zip rsync`

The biggest one is it needs texlive stuff. I'm waiting for my computer to get past `=> lilypond-2.18.2_1: installing target dependency 'texlive-bin-2015_1' ...` 20 minutes and waiting.

maxice8:
solved by 6d92c2b2cff3b19fc3b61068a593d97d112b3448 and ec8e5b7d97da86ef32636b4b74297f16924b31f1 

@the-maldridge

PR #7712, opened 2017-09-18
New package: lilypond-2.19.65

newbluemoon:
This PR addresses issue #2987.

Lilypond is built without documentation. It will be added if/when the required package `dblatex` (#6347) becomes available. Though the packages `extractpdfmark` and `texi2html` are only used for documentation building I included them anyway. Lilypond upstream offers a pre-built documentation tarball which I used to create a `lilypond-doc` package (I think it’s good to have some offline doc for a program like lilypond).

Lilypond also depends on texlive. Instead of adding the `texlive-bin` package to `depends` (with all its implications) I added a post install message that says a working texlive installation is needed (a basic installation as done by `texlive-bin` should be sufficient).

A problem still is cross compilation: it builds natively without any problem (tested on x86_64, i686, and armv7l), but when cross compiling make just stops with a ‘no rule to make target’ error. Comparing the output of the nativ/cross configure scripts doesn’t show any difference. Right now I have no idea how to solve this, but thought I should open this PR anyway ;)


newbluemoon:
I just noticed in the log: on Travis `automake` is installed but not `autoconf`

    ./autogen.sh: line 17: autoconf: command not found

But

    $ xbps-query -Rx automake
    perl>=0
    autoconf>=0

?

chneukirchen:
t1utils are included in TeXLive.

I'm not very happy about adding yet another outdated unmaintained release of guile. Does this release still not support guile 2.0?

newbluemoon:
Wasn’t aware that t1utils are in texlive. Will change that.

Guile 2 support is only mentioned as ‘highly experimental’. The last time I tried it didn’t work. But I can try again.

newbluemoon:
Is there a way for a normal user to cancel a travis job? A previous build of lilypond (job no. 12568) has been running for over four hours now and is blocking others.

newbluemoon:
Could you please cancel lilypond’s jobs on Travis no. 12571 and 12572? They have already been running for hours and all packages were built. But it seems that `show_files.sh` is inflating the log, because there are 27601 files in lilypond-doc and Travis somehow get stuck and does neither end nor abort.

Vaelatern:
testing guile cross build.

newbluemoon:
On x86_64 and i686 all packages were built, Travis seems to fail because the log gets too big (lilypond-doc contains more than 27000 files). The musl builds have to fail because there is no texlive.

lemmi:
I can confirm the build works on `x86_64`, but fails on `x84_64-musl`. However `guile` seems to be missing as a dependency.
```
Log level set to 287
GNU LilyPond 2.19.65
Relocation: from PATH=/home/lemmi/bin:/home/lemmi/.gem/ruby/2.3.0/bin:/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/sbin:/sbin:/bin:/usr/share/apache-maven/bin:/opt/texlive/2017/bin/x86_64-linux
argv0=lilypond
PATH=/usr/bin (prepend)
Setting PATH to /usr/bin:/home/lemmi/bin:/home/lemmi/.gem/ruby/2.3.0/bin:/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/sbin:/sbin:/bin:/usr/share/apache-maven/bin:/opt/texlive/2017/bin/x86_64-linux
Relocation: compile datadir=, new datadir=/usr/share/lilypond//2.19.65
Relocation: framework_prefix=/usr/bin/..
Setting INSTALLER_PREFIX to /usr/bin/..
PATH=/usr/bin/../bin (prepend)
Setting PATH to /usr/bin/../bin:/usr/bin:/home/lemmi/bin:/home/lemmi/.gem/ruby/2.3.0/bin:/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/sbin:/sbin:/bin:/usr/share/apache-maven/bin:/opt/texlive/2017/bin/x86_64-linux
Setting GUILE_MIN_YIELD_1 to 65
Setting GUILE_MIN_YIELD_2 to 65
Setting GUILE_MIN_YIELD_MALLOC to 65
Setting GUILE_INIT_SEGMENT_SIZE_1 to 10485760
Setting GUILE_MAX_SEGMENT_SIZE to 104857600
Setting GUILE_AUTO_COMPILE to 0
Setting GUILE_WARN_DEPRECATED to detailed
Setting XDG_CACHE_HOME to /usr/share/lilypond/2.19.65
Throw without catch before boot:
Throw to key misc-error with args ("primitive-load-path" "Unable to find file ~S in load path" ("ice-9/boot-9") #f)Aborting.
Aborted
```
Installing `guile` manually solved the problem and simple examples produced working pdfs.

newbluemoon:
Hmm, when I check my local build libguile got added as a dependency:

<pre>
$ xbps-query --repository=hostdir/binpkgs/lilypond -Rx lilypond
python>=0
ghostscript>=0
glibc>=2.25_1
libguile>=2.0.9_3
gc>=7.1_1
pango>=1.24.0_1
glib>=2.18.0_1
fontconfig>=2.6.0_1
freetype>=2.3.9_1
libstdc++>=4.4.0_1
libgcc>=4.4.0_1
</pre>

I also tested some examples using both guile2 and guile1.8 and the pdfs looked the same.

lemmi:
The dependency `("ice-9/boot-9")` is found in `guile`, not `libguille`:
```
# xlocate ice-9/boot-9
guile-2.0.14_1	/usr/lib/guile/2.0/ccache/ice-9/boot-9.go
guile-2.0.14_1	/usr/share/guile/2.0/ice-9/boot-9.scm
```

newbluemoon:
Ah, sorry. Had this fixed when using guile1.8 and reintroduced it when switching to guile2. But that makes me wonder: when using ‘libguile’ depends on files located in ‘guile’ shouldn’t ‘libguile’ depend on ‘guile’?

newbluemoon:
Just another thought: Support of guile2 is regarded as ‘highly experimental’. It builds and pdfs are created without a problem for now but this might not be the case for all score sheets. Debian e.g. has guile-1.8 bundled with lilypond just for internal use. So I was thinking to do something like that and add a build option for an internal guile-1.8?

lemmi:
Before this gets burried any deeper: Since we don't have guile-1.xx available, we might aswell try 2.0. If we do run into issues, we can still fix them.

newbluemoon:
I tried to combine guile 1.8 and lilypond in one template, but whatever I tried was just an ugly hack. That’s just something xbps-src isn’t supposed to do, I guess. The only reasonable way I came up with (when wanting to use 1.8 internally) is separately making a guile1.8-devel package with the only purpose to use when building lilypond and copying the binary guile1.8, the libs and the *.scm files to a location only lilypond knows about. But this would also add guile 1.8 to the repo.

So to sum it up: lilypond works best with guile 1.8 but if that’s not available I’m with @lemmi to give 2.0 a try.

chneukirchen:
Well, then package guile 1.8. As it is dead upstream, it probably won't be a maintenance burden. :P

newbluemoon:
Ok, I made lilypond use guile-1.8 again ;-)

lemmi:
Travis doesn't like it at all. The compilation output is gigantic. 
x86_64 built and worked.

lemmi:
@newbluemoon have you found a way to clean up the builds and have you had a chance to try on a musl system?

newbluemoon:
@lemmi I didn’t try it on a musl system because there’s currently no texlive available.
One thing that inflates the output is `common/travis/show_files.sh` because lilypond-doc contains about 27.000 files. Is there an option to disable `show_files.sh`? Or I could remove lilypond-doc for now.

lemmi:
As for the build log: Don't bother then.
Otherwise, I can't cross compile and musl doesn't want to build aswell. At least for now we should set nocross. For musl the reason seems to be texlive. For other archs I don't understand the error.
Please set nocross. Then we'll merge and hopefully come up with some solutions later on.

newbluemoon:
@lemmi: Done.
In the initial post, last paragraph, I mentioned cross compiling: It just stops for no apparent reason. Do you get a different error? I think that’s somehow due to lilypond’s off-the-wall building system, but right now have absolutely no clue.

Issue #7710, opened 2017-09-18
julia: Target architecture mismatch


espectalll:
Here's the entire log I get every time I try to run Julia:

    ERROR: Target architecture mismatch. Please delete or regenerate sys.{so,dll,dylib}.

The error is being reported on `x86_64`, after verifying the system was clean of system files from previous Julia manual builds.

silvernode:
https://github.com/JuliaLang/julia/issues/10639

I am looking into it, and I am building it now with some flags. I hope it works

silvernode:
@espectalll sent you an email

silvernode:
@espectalll Well the build is complete and the installed package works but that doesn't mean it will work for you. I'm not sure if I should send you the package directly to make sure it works or if I should just commit to the repo

espectalll:
I can confirm it works on my machine natively. Let me try out as `i686` and I'll confirm if it works there too.

espectalll:
...not working on `i686`, as shown on the next screenshot, with a chroot and a Virtualbox VM:

![captura de pantalla de 2017-09-18 17-52-54](https://user-images.githubusercontent.com/2456419/30551234-2ae352e8-9c9a-11e7-9968-c9c759727a6b.png)

This is likely easy to fix, so give me a moment... I wish it was easier to debug how things will be built in prod, though.

silvernode:
Looks like a solid fix, great work.

silvernode:
apparently, this is a thing in their README:  

*You can also set MARCH=native for a maximum-performance build customized for the current machine CPU.*

espectalll:
Yes, I'm aware. The problem is, I don't think it's properly detecting the architecture.

...in fact, it seems that the i686 builds are still failing???

wangp:
Unfortunately building for core2 on x86_64 means the binary does not work on a Phenom II processor which lacks SSSE3:
```
ERROR: The current host does not support SSSE3, but the system image was compiled for Core2.
Please delete or regenerate sys.{so,dll,dylib}.
```
For comparison, the official build does work.

PR #7720, opened 2017-09-18
julia: fix issue #7710

espectalll:


Gottox:
use a sane commit message and ping me, will merge afterwards 👍

PR #7724, opened 2017-09-18
unbound: update to 1.6.6.

dne:


chneukirchen:
Second commit is useless tho, as -d doesn't create a pidfile already.

dne:
Sorry, but it does -- I just double checked.

chneukirchen:
I couldn't find it... also my unbound didn't run. duh. ;) thanks.

PR #7725, opened 2017-09-18
New package: megacmd-3.2.0

jbgg:


jbgg:
I already fixed this problem: https://travis-ci.org/voidlinux/void-packages/jobs/272788281#L2137

srcpkgs/megacmd/template
```diff
@@ -0,0 +1,25 @@
...
+hostmakedepends="base-devel autogen doxygen"
+makedepends="libcurl-devel c-ares-devel libressl-devel crypto++-devel zlib-devel sqlite-devel freeimage-devel readline-devel libsodium-devel pcre-devel pcre2-devel libuv-devel fuse-devel"
+depends="pcre pcre2 libpcrecpp"
+short_desc="The official client for cloud service mega.nz"
+maintainer="jbgg <jbosco.gg@gmail.com>"
+license="2-clause-BSD"
+homepage="https://mega.nz"
+distfiles="https://github.com/meganz/sdk/archive/v${version}.tar.gz"
+checksum=4f4b607caa80fc7a91ba1cbb882b1c9b903d9b67b59db47fb7ec821e33de29f3
+only_for_archs="x86_64 i686 armv7hf"
```
Johnnynator: If it doesn't build on certain platforms then specify them as broken. The Musl builds can be fixed by including string.h where needed.

Vaelatern: Agreed. This list seems sparse, lacks armv6hf... and that's just an indicator that the line is wrong to begin with. Look up examples of marking musl builds as broken, or just patch to include string.h where needed.

PR #7733, opened 2017-09-19
qt5: add fix for font rendering issues in qt-webengine

Hoshpak:
Since the chromium environment shipped with qt-webengine also bundles
the skia library, it is affected by the same font rendering issues as
the mozilla products with freetype 2.8.1. This commit adds the patch
backported for Firefox ESR from
https://bug1393467.bmoattachments.org/attachment.cgi?id=8904705 which also
applies to the skia version bundled with qt-5.8.

### DIFF
--- a/srcpkgs/qt5/patches/skia-freetype.patch
+++ b/srcpkgs/qt5/patches/skia-freetype.patch
@@ -1,0 +1,148 @@
+# HG changeset patch
+# User Lee Salzman <lsalzman@mozilla.com>
+# Date 1504640559 14400
+#      Tue Sep 05 15:42:39 2017 -0400
+# Node ID 923246286b9858fb103e100f886c03714b97b5ec
+# Parent  3fff2b174212af40a7b7ba75a047db431f81c780
+clip FreeType glyph bitmap to mask in Skia (52 ESR)
+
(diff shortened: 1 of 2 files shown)


Gottox:
Yet another qt rebuild \o/

Anachron:
Really need this! My fonts look ugly right now.

https://i.imgur.com/QwYnts9.png

Gottox:
Bad Keming

Anachron:
Looks like I still have issues... is there something I am missing?

Tried reinstalling as well.

```
Name Action    Version           New version            Download size
qt5  reinstall 5.8.0_8           5.8.0_8                -
```

Hoshpak:
Did you also update the qt5-webengine package? It's a separate sub-package of qt5. What application are you still having issues with?

Anachron:
Yes I also updated that one (`using xbps-install -f`). 

I am having issues with `qutebrowser`. 

I just realized it uses `python3-PyQt5-webkit` which has not been fixed yet?

Hoshpak:
I think these python packages are just wrappers around the actual libraries so it shouldn't make a difference. ```python3-PyQt5-webkit``` indicates webkit which is not the same as qt-webengine which is not based on webkit but Chromium/Blink instead. Which rendering enige are you using? If it's webkit, please try installing the necessary packages and launch qutebrowser with qt-webengine to see if it make a difference. Since I don't have much time right now (will have more later, though), I briefly installed qutebrowser on my laptop and tried it with webengine and webkit which both look fine here so I wonder what is causing your issue.

PR #7740, opened 2017-09-19
New package: pam_rundir-1.0.0

maxice8:
Need some specific feedback on this one about:

- This package's activation* requires writing to a file owned by another package, it can either be done manually by the admin or automated by appending to said file, i decided to tell the admin via INSTALL.msg to and how to activate it, but i need a maintainer  on which one to do.

\* State after installation by the package manager but before being operational.


the-maldridge:
Usually its safe to assume the administrator will know how to use advanced PAM modules such as this.  I have removed the `INSTALL.MSG` at merge time as such.

PR #7749, opened 2017-09-20
mono: update to 5.2.0.224

Hoshpak:


Hoshpak:
i686 builds find locally so I think it just takes a little bit longer than the rest to build and thus times out on travis.

ebfe:
Let's find out.

PR #7755, opened 2017-09-21
pulseaudio: rebuild because of missing signature file

Johnnynator:


Johnnynator:
mh.. I was just to fast and managed to try to update my system in the 10seconds the signature was generated...

PR #7757, opened 2017-09-21
New package: ufraw-0.22

### LINKED ISSUE #7099: Package Request: ufraw
Ufraw is a library to convert digital camera raw image files (like CR2) to more common formats (png, jpeg). 

I tried building it myself but it fails for now, I believe some issue with glib. 

I'm traveling right now so can't share the template. If someone wants to join right in he's free to :)

Johnnynator:
closes #7099


lemmi:
I was just checking this out, but SourceForge doesn't seem to let me download the distfile atm. I just get a bunch of redirects. I will look at it the next couple of days again.

srcpkgs/ufraw/template
```diff
@@ -0,0 +1,20 @@
+# Template file for 'ufraw'
+pkgname=ufraw
+version=0.22
+revision=1
+build_style=gnu-configure
+configure_args="--enable-contrast --enable-mime --enable-extras --enable-openmp --with-gimp"
+hostmakedepends="pkg-config"
+makedepends="exiv2-devel gimp-devel glib-devel gtkimageview-devel lcms2-devel lensfun-devel"
```
lemmi: I'm not sure about the value, but you could add jpeg2000 support via libjasper. If you have gimp installed, that lib is on your system anyway.

Johnnynator:
@lemmi I just set extras as optional build_option since probably no one will use nikon-curve and also it is recommended in the README for packaging to not enable extras.
Builds fail because of SourceForge.

lemmi:
+1 for using build_options.

cr6git:
Good job !

PR #7768, opened 2017-09-21
New package: sanoid-1.4.14

emacsomancer:
Sanoid is a policy-driven snapshot management tool for ZFS filesystems


newbluemoon:
Instead of using `install -Dm644 ...` in `do_install()` you should use [global functions](https://github.com/voidlinux/void-packages/blob/master/Manual.md#global_funcs) e.g.:

    vlicense LICENSE
    vbin sanoid
    ...

And configuration files should also be listed in `conf_files` so that they won’t get overwritten when the package gets updated.

emacsomancer:
thanks, @newbluemoon; I I think I've got it fixed now.

newbluemoon:
Don’t use `${DESTDIR}` in `conf_files`, just the location where the package manager looks for them when installed:

    conf_files="/etc/sanoid/sanoid.conf /etc/sanoid/sanoid.defaults.conf"

You can also install them with a global function:

    vinstall sanoid.conf 644 etc/sanoid/

srcpkgs/sanoid/INSTALL.msg
```diff
@@ -0,0 +1,6 @@
+Syncoid depends on ssh, pv, gzip, lzop, and mbuffer.  It can run with reduced 
+functionality in the absence of any or all of the above. SSH is only required
+for remote synchronization.  On newer FreeBSD and Ubuntu Xenial 
+chacha20-poly1305@openssh.com, on other distributions arcfour crypto is the 
+default for SSH transport since v1.4.6. Syncoid runs will fail if one of them 
+is not available on either end of the transport.
```
Vaelatern: This is mostly incomprehensible. Only the first two sentences bear any resembalence to useful annoyances. Perhaps a single line: `Syncoid has more functionality when ssh, pv, gzip, lzop, and mbuffer are installed`

emacsomancer: I had taken the original message from the documentation, but I agree it's not as clear as it could be. Amended as suggested.

the-maldridge:
Merged with minor changes.  In the future make sure your files end with a newline.

PR #7771, opened 2017-09-21
New package: siji-font-20160513

vikrrrr:


srcpkgs/siji-font/template
```diff
@@ -0,0 +1,22 @@
+# Template file for 'siji-font'
+pkgname=siji-font
+_commit=95369afac3e661cb6d3329ade5219992c88688c1
```
the-maldridge: This is a -git package which we do not allow.  Please either use a real release or request the upstream to release a real tarball.

PR #7778, opened 2017-09-22
ppsspp: update to 1.4.2.

Johnnynator:


Johnnynator:
would replace #7224 which I may have overseen.
Using system ffmpeg fixes the issues with musl.

PR #7779, opened 2017-09-22
New package: monero-0.11.0.0

Hoshpak:


srcpkgs/monero/INSTALL
```diff
@@ -0,0 +1,9 @@
+# INSTALL
+case "$ACTION" in
+post)
+	mkdir var/lib/monero
+	mkdir var/log/monero
+	chown monero.monero var/lib/monero
+	chown monero.monero var/log/monero
```
lemmi: You don't need to do this manually. Look up the `make_dirs` variable in https://github.com/voidlinux/void-packages/blob/master/Manual.md#optional-variables

Hoshpak: Thanks. That's a feature I didn't come across yet. I ammended the commit.

lemmi:
LGTM. The only thing that left me a bit wondering are the directory permissions. They seem a little loose for an application like that, or is this fine?

Hoshpak:
The /var/lib/monero directory is used only by monerod. The daemon is responsible for retrieving the monero blockchain and communicating with the network. Since that is all publicly available information, I saw no reason to restrict it further, although 0750 certainly wouldn't hurt either. The private key for the wallet is created and used only by monero-wallet-cli which communicates with monerod via it's socket. The key files are created in the directory where monero-wallet-cli is executed, which usually would be a user's home directory.

lemmi:
If it doesn't hurt the functionality I like to keep permissions tight.

Hoshpak:
Alright, restricted it to 0750.

PR #7784, opened 2017-09-22
clementine: Add Spotify buildoption and fix musl

Johnnynator:


srcpkgs/clementine/template
```diff
@@ -18,13 +18,28 @@ wrksrc="Clementine-$version"
...
 checksum=f885931a9ab7c88607d07b50c64fcce46fc05f13dd2c0a04188c94eff938f37c
 
+if [ -n "$CROSS_BUILD" ]; then
+	hostmakedepends+=" qt-host-tools qt-devel protobuf-c"
+fi
+
+build_options="spotify"
+
+case "$XBPS_TARGET_MACHINE" in
+	x86_64 | i686 | armv6l | armv7l)
```
Vaelatern: missing aarch64? Did you perhaps to make a `*-musl` and a `*`?

Vaelatern:
Not sure why your patch replaces `NULL` with a compatible alternative? This seems like a patch for upstream, that does not affect usability in the slightest.

Johnnynator:
It is required for musl, otherwise I get a error when compiling.

srcpkgs/clementine/template
Johnnynator: It is due to the fact that the libspotify package is only for these Arches

ebfe:
broken on armv6 https://build.voidlinux.eu/builders/armv6l_builder/builds/5092/steps/shell_3/logs/stdio

ebfe:
also not sure if enabling a build_option that depends on a nonfree package is a good idea.

Johnnynator:
@ebfe Spotify was the default before so I didn't want to change it and also it can be installed without the non free part.

Issue #7656, opened 2017-09-14
meson cross builds


ebfe:
The build-style currently doesn't handle cross builds. meson has support for this but requires some setup:
http://mesonbuild.com/Cross-compilation.html

Gottox:
duplicate

Gottox:
whoops. Not ment to be closed.

PR #7793, opened 2017-09-23
fuse-sshfs: update to 3.3.0

pbui:
- Upstream has switched to meson
- Need python-docutils to build manpage


ebfe:
This breaks cross builds #7656

pbui:
Should work with #7801

PR #7798, opened 2017-09-23
New package: antibody-3.3.2

SolitudeSF:


srcpkgs/antibody/template
```diff
@@ -0,0 +1,14 @@
+# Template file for 'antibody'
+pkgname=antibody
+version=3.3.1
+revision=1
+build_style=go
+go_import_path="github.com/getantibody/antibody"
+hostmakedepends="git"
+depends="zsh"
+short_desc="A faster and simpler antigen written in Golang."
```
Duncaen: I dont really like the description, what is it what does it?

SolitudeSF: Changed desc to be more general and removed zsh from runtime dependencies since it doesnt actually requires zsh to function, its just build primarely to manage zsh plugins.

Vaelatern:
please actually follow the commit message standards.

SolitudeSF:
i think i corrected everything. didnt realize what was wrong initially.

Issue #7713, opened 2017-09-18
Tracking issue: Update Gnome 3.26


Gottox:
Tracking PR for merging Gnome 3.26. https://github.com/Gottox/void-packages/tree/gnome-3.26 is the main branch for this update.

## gnome

* [ ] NetworkManager 1.9.2 *unstable version*
* [x] adwaita-icon-theme 3.25.91 *unstable version*
* [x] at-spi2-atk 2.26.0
* [x] at-spi2-core 2.26.0
* [x] atk 2.26.0
* [x] atkmm 2.24.2
* [x] baobab 3.26.0
* [ ] cairomm 1.15.4 *unstable version*
* [x] cantarell-fonts 0.0.25
* [x] caribou 0.4.21
* [x] cheese 3.26.0
* [x] clutter 1.26.2
* [x] clutter-gst 3.0.24
* [x] clutter-gtk 1.8.4
* [x] cogl 1.22.2
* [x] dconf 0.26.0
* [x] eog 3.26.0
* [x] epiphany 3.26.0
* [x] evince 3.26.0
* [x] evolution-data-server 3.26.0
* [x] file-roller 3.26.0
* [x] folks 0.11.4 *unstable version*
* [x] gcab 0.7 *unstable version*
* [x] gcr 3.20.0
* [x] gdk-pixbuf 2.36.10
* [x] gdm 3.26.0
* [x] gedit 3.22.1
* [x] geocode-glib 3.25.4.1 *unstable version*
* [x] gfbgraph 0.2.3
* [x] gjs 1.50.0
* [x] glib 2.54.0
* [x] glib-networking 2.54.0
* [x] glibmm 2.50.1
* [x] gmime 2.6.23
* [x] gnome-backgrounds 3.25.4 *unstable version*
* [x] gnome-bluetooth 3.26.0
* [x] gnome-boxes 3.26.0
* [ ] gnome-calculator 3.25.92 *unstable version*
* [x] gnome-calendar 3.26.0
* [ ] gnome-characters 3.25.92 *unstable version*
* [x] gnome-clocks 3.26.0
* [x] gnome-color-manager 3.26.0
* [x] gnome-contacts 3.26
* [x] gnome-control-center 3.26.0
* [x] gnome-desktop 3.26.0
* [x] gnome-disk-utility 3.26.0
* [ ] gnome-documents 3.25.91 *unstable version*
* [ ] gnome-font-viewer 3.25.90 *unstable version*
[...]

PR #7801, opened 2017-09-23
meson: add preliminary support for cross compilation

pbui:
This should address (or begin to address) #7656.

I've tested this with armv6, armv6l-musl, armv7l for the ncmpc and fuse-sshfs packages.

### DIFF
--- a/common/build-style/meson.sh
+++ b/common/build-style/meson.sh
@@ -4,6 +4,54 @@
 do_configure() {
 	: ${meson_cmd:=meson}
 	: ${meson_builddir:=build}
+	: ${meson_crossfile:=xbps_meson.cross}
+
+	if [ "$CROSS_BUILD" ]; then
+		_MESON_TARGET_ENDIAN=little
+		_MESON_TARGET_CPU=${XBPS_TARGET_MACHINE}
+		case "$XBPS_TARGET_MACHINE" in
+			mips|mips-musl)
+				_MESON_TARGET_ENDIAN=big
+				;;
+		esac
+
+		# Record cross-compiling information in cross file.
+		# CFLAGS and LDFLAGS must be set as c_args and c_link_args.
+		cat > ${meson_crossfile} <<EOF
+[binaries]
+c = '${CC}'
+cpp = '${CXX}'
+ar = '${AR}'
+ld = '${LD}'
+strip = '${STRIP}'
+readelf = '${READELF}'
+pkgconfig = 'pkg-config'
+
+[properties]
+needs_exe_wrapper = true
+c_args = ['$(echo ${CFLAGS} | sed -r "s/\s+/','/g")']
+c_link_args = ['$(echo ${LDFLAGS} | sed -r "s/\s+/','/g")']
+
+[host_machine]
+system = 'linux'
+cpu_family = '${XBPS_MACHINE}'
+cpu = '${XBPS_MACHINE}'
+endian = 'little'
+
+[target_machine]
+system = 'linux'
+cpu_family = '${_MESON_TARGET_CPU}'
+cpu = '${_MESON_TARGET_CPU}'
+endian = '${_MESON_TARGET_ENDIAN}'
+EOF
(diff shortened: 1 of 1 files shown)


common/build-style/meson.sh
```diff
@@ -4,6 +4,55 @@
 do_configure() {
 	: ${meson_cmd:=meson}
 	: ${meson_builddir:=build}
+	: ${meson_crossfile:=meson.cross}
+
+	if [ "$CROSS_BUILD" ]; then
+		_MESON_TARGET_ENDIAN=little
+		_MESON_TARGET_CPU=${XBPS_TARGET_MACHINE}
+		case "$XBPS_TARGET_MACHINE" in
+			mips)
```
Vaelatern: `mips*` perhaps?

common/build-style/meson.sh
```diff
@@ -4,6 +4,55 @@
 do_configure() {
 	: ${meson_cmd:=meson}
 	: ${meson_builddir:=build}
+	: ${meson_crossfile:=meson.cross}
```
Vaelatern: can we make this perhaps `xbps_meson.cross` or something like that? Something less generic.

common/build-style/meson.sh
```diff
@@ -4,6 +4,55 @@
...
+				_MESON_TARGET_ENDIAN=big
+				;;
+
+		esac
+
+		# Record cross-compiling information in cross file.
+		# CFLAGS and LDFLAGS must be set as c_args and c_link_args.
+		cat > ${meson_crossfile} <<EOF
+[binaries]
+c = '${XBPS_CROSS_TRIPLET}-gcc'
```
Vaelatern: what's wrong with `${CC}`?

common/build-style/meson.sh
```diff
@@ -4,6 +4,55 @@
...
+cpp = '${XBPS_CROSS_TRIPLET}-g++'
+ar = '${XBPS_CROSS_TRIPLET}-ar'
+ld = '${XBPS_CROSS_TRIPLET}-ld'
```
Vaelatern: Maybe all this can just use the XBPS vars instead of reconstructing?

common/build-style/meson.sh
Johnnynator: Isn't mips big endian and mipsel little endian? Thefore `mips | mips-musl` would be better.

pbui: I've changed this to `mips|mips-musl`

common/build-style/meson.sh
pbui: I've renamed this to `xbps_meson.cross`.

common/build-style/meson.sh
pbui: Yes, I've switched this to `${CC}`, `${CXX}`, etc. and tested it on the ncmpc package with armv7l and arm7l-musl.

pbui:
I've made the changes suggested by @Vaelatern and @Johnnynator.  

I've pushed a squashed commit... I'm not sure I was suppose to do that since it collapsed the review above.

Gottox:
thanks, I'll give it a try with the latest Gnome update (https://github.com/voidlinux/void-packages/issues/7713).

common/build-style/meson.sh
```diff
@@ -4,6 +4,54 @@
...
+cpu = '${_MESON_TARGET_CPU}'
+endian = '${_MESON_TARGET_ENDIAN}'
+EOF
+		configure_args+=" --cross-file=${meson_crossfile}"
+
+		# Meson tries to compile natively with CC, CXX, so when cross
+		# compiling, we need to set those to the host versions.
+		export CC=${CC_host} CXX=${CXX_host}
+
+		unset _MESON_TARGET_CPU _MESON_TARGET_ENDIAN
```
Gottox: declare those vars with `local _MESON_TARGET_CPU= _MESON_TARGET_ENDIAN=` at the top.

Gottox:
I'd like to merge. Objections?

PR #7808, opened 2017-09-24
New package: masterpassword-cli-2.6

SolitudeSF:


srcpkgs/masterpassword-cli/template
```diff
@@ -0,0 +1,30 @@
+# Template file for 'masterpassword-cli'
+pkgname=masterpassword-cli
+version=2.6
+revision=1
+create_wrksrc=yes
+_file="mpw-2.6-cli-2-0-g30fdb54e.tar.gz"
+build_options="json color"
+build_options_default=${build_options}
+desc_option_json="Support JSON-based user configuration format"
+desc_option_color="Show a colorized identicon"
```
lemmi: I'm not sure it's worth adding all these options.

srcpkgs/masterpassword-cli/template
```diff
@@ -0,0 +1,30 @@
...
+hostmakedepends="libsodium-devel $(vopt_if json json-c-devel) $(vopt_if color ncurses-devel)"
+makedepends="libsodium-devel $(vopt_if json json-c-devel) $(vopt_if color ncurses-devel)"
+short_desc="CLI version of Master Password"
+maintainer="SolitudeSF <solitudesf@protonmail.com>"
+license="GPL-3"
+homepage="https://ssl.masterpasswordapp.com/"
+distfiles="${homepage}${_file}"
```
lemmi: Use github tarballs. They have a much nicer layout and can be checked for updates.
https://github.com/Lyndir/MasterPassword/releases
You will however change the wrksrc to point to ${pkgname}-${version}/platform-independent/cli-c

srcpkgs/masterpassword-cli/template
```diff
@@ -0,0 +1,30 @@
...
+checksum=c206e512d193a154814a7c5d1faa8dc89e3fa0817b059bd6d957a61dc8ee1e68
+
+do_extract() {
+	tar xfz ${XBPS_SRCDISTDIR}/${pkgname}-${version}/${_file} -C ${wrksrc}
```
lemmi: Remove, will work with github tarballs and proper wrksrc setup

srcpkgs/masterpassword-cli/template
```diff
@@ -0,0 +1,30 @@
...
+}
+
+do_build() {
+	./build
```
lemmi: Use cmake build style. Otherwise cross-compile will be difficult.

srcpkgs/masterpassword-cli/template
```diff
@@ -0,0 +1,30 @@
...
+}
+
+do_install() {
+	vbin mpw
```
lemmi: Should work with cmake directly.

srcpkgs/masterpassword-cli/template
SolitudeSF: i changed build style to cmake, but on building stage linker throws this (```relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a shared object; recompile with -fPIC```) error for every *.o* file. Have no clue how to fix this.

srcpkgs/masterpassword-cli/template
lemmi: @Vaelatern suggested to use
```
wrksrc="${pkgname}-${version}"
build_wrksrc="platform-independent/cli-c"
```

srcpkgs/masterpassword-cli/template
lemmi: One option is to patch the `Makefile` `cmake` generates in `post_configure` and add `-fPIC` to the `CFLAGS`. This can be done via sed or a patch. grep other packages for inspiration.

SolitudeSF:
Changed build_style to cmake. Still using vbin, because of this error "```make: *** No rule to make target 'install'.  Stop.```" on do_install(). Also still fetching source from official site, since github tarball doesnt have ```VERSION``` file which is required by cmake and it contains sources for all other platforms.

SolitudeSF:
Switched to github tarball, echo ```VERSION``` in pre_configure and used vlicense.

lemmi:
I think this won't get any prettier. We can merge this as is, but it would be nice if you made the bash-completion available as well. But I leave this up to you since the completion scripts aren't ready to ship as they are.
You either need to change the second line of `mpw.completion.bash` and rewrite the path to `bashcomplib` and install that script somewhere else than the mpw completion or concatenate the files into one (without sourcing bashcomplib).
Let me know when you are done.

SolitudeSF:
i *theoretically* added bash completion, but when i install package, neither completion script, nor license file are present in system and only file registered for the package is the binary, even tho when i check the destdir during packaging, all files are present. I guess im missing something on my end.

lemmi:
fwiw: ./xbps-src **-f** ... will overwrite the xbps package

Issue #7805, opened 2017-09-24
tbb: build loop


ebfe:
```
$ xbps-checkvers
[...]
pkgname: tbb repover: 2018_1 srcpkgver: 2018/_U/.}_1
[...]
```
triggers a rebuild every builder run. Marking broken for now.

yopito:
@ebfe : thanks for pointing this. I will check this ASAP.

yopito:
My thought: `xbps-checkvers` does not handle correctly shell variable substution in `version`.  
But variable substition is allowed in `version` variable ?

Details:  `tbb/template` uses the following construction:
```
_ver=2018
version=${_ver/_U/.}
```
This way, `_` character is not present in `version` and upstream numbering scheme is correcly handled (they use `_U` for their minor version ...) :
* 2017_U1 => 2017.1
* 2018 => 2018

Should I fill a bug about it for `xbps-checkvers` ?   
or formal version's rules should **not** allow shell variable subsitution, thus redisign `tbb/template` accordingly ?

ebfe:
I think the easiest solution for now would be to change the logic to something like:
```
version="2018.U"
_ver=${version/.U/_U}
```

xbps-checkvers (which is a bit of a mess) only understands trivial expansions, and it's safer to not rely on it.

ebfe:
xbps-checkvers only cares about pkgname, version, revision, reverts. So ideally these assignments are as logic free as possible.

yopito:
OK, let's go this way. thanks for your replies.

ebfe:
7809

PR #7811, opened 2017-09-24
Manual.md: no underscore or shell subst allowed in version

yopito:
following discussion in issue #7805


yopito:
I also work on a PR to xtools to enforce check of this rule too.

Gottox:
👍

Issue #1933, opened 2015-06-30
Package Request: Haxe


johnku1:
http://haxe.org/

teh-jazzman:
For the most part [Neko](http://nekovm.org/) is also required to use Haxe.

In the absence of proper packages, OpenFL provides [installer-scripts](http://www.openfl.org/builds/haxe/) that install both Haxe and Neko. They need a minor patch to work with xbps-install though.

PR #7813, opened 2017-09-25
New package: haxe-3.4.6

vikrrrr:
#1933


srcpkgs/haxe/template
```diff
@@ -0,0 +1,37 @@
...
+version=3.4.3
+revision=1
+hostmakedepends="nekovm ocaml camlp4 git"
+makedepends="pcre-devel zlib-devel"
+short_desc="The Haxe Cross-Platform Toolkit"
+maintainer="vikrrrr <vikr@protonmail.com>"
+license="GPL-2, MIT"
+homepage="https://haxe.org/"
+_repo="https://github.com/HaxeFoundation/haxe"
+wrksrc=${pkgname}-${version}
```
Johnnynator: Not needed since this is the default.

srcpkgs/haxe/template
```diff
@@ -0,0 +1,37 @@
...
+nostrip=yes
+nocross=yes
+nopie=yes
+
+do_fetch() {
+	git clone "${_repo}.git" "${wrksrc}"
+	cd "${wrksrc}"
+	git checkout ${version}
+	git submodule update --init --recursive
+}
```
Johnnynator: Generally it is preferred to just fetch the release tar and the sub modules tar's instead of using the git command.

srcpkgs/haxe/template
```diff
@@ -0,0 +1,37 @@
...
+
+do_build() {
+	make ${makejobs}
```
Johnnynator: Why don't you just use the gnu-makefile build_style?

srcpkgs/haxe/template
```diff
@@ -0,0 +1,37 @@
...
+makedepends="pcre-devel zlib-devel"
+short_desc="The Haxe Cross-Platform Toolkit"
+maintainer="vikrrrr <vikr@protonmail.com>"
+license="GPL-2, MIT"
+homepage="https://haxe.org/"
+_repo="https://github.com/HaxeFoundation/haxe"
+wrksrc=${pkgname}-${version}
+nostrip=yes
+nocross=yes
+nopie=yes
```
Johnnynator:
```
+nostrip=yes
+nocross=yes
+nopie=yes
```
You should try yo avoid these variables and if they really are needed put a reasoning behind them, in case of nocross the build log is normally enough.

vikrrrr:
@Johnnynator is this a proper way to fetch the submodules?

Johnnynator:
Yep, that looks OK (to me).

vikrrrr:
@andyli any idea as to why [this](https://travis-ci.org/voidlinux/void-packages/jobs/343448882#L1391-L1397) OCaml error occurs?
```
File "extString.ml", line 113, characters 11-24:
Error: This expression has type string but an expression was expected of type
         bytes
```

andyli:
That's because haxe 3.4.x (mainly the ocamllibs submodule) is not compatible with ocaml's relatively new "safe string" feature.
You may set the env var `OCAMLPARAM="safe-string=0,_"` to make it work. It is used in our Travis setup: https://github.com/HaxeFoundation/haxe/blob/3.4_bugfix/.travis.yml#L3-L4

PR #7822, opened 2017-09-25
New package: setroot-2.0.1

SolitudeSF:


srcpkgs/setroot/template
```diff
@@ -0,0 +1,18 @@
+# Template file for 'setroot'
+pkgname=setroot
+version=2.0.1
+revision=1
+build_style=gnu-makefile
+build_options="xinerama"
+build_options_default="xinerama"
+desc_option_xinerama="Multiple monitors support"
+CFLAGS+=" -fPIE -pie"
+LDFLAGS+=" -fPIE"
```
Vaelatern: no. These two are wrong.
Perhaps `make_use_env=compliant`?
We have CFLAGS already. The makefile should respect them.

SolitudeSF: i changed these lines to suggested but it reverts to throwing error ```non-PIE executable found in PIE build```

Vaelatern: Then somewhere it isn't respecting our LDFLAGS.

SolitudeSF:
Added ```export CFLAGS="$CFLAGS $LDFLAGS"```  to pre_build.

PR #7826, opened 2017-09-26
xbps-src: ignore files in patches/ without patch/diff suffix

ebfe:


Anachron:
@ebfe I think that's a good idea!

chneukirchen:
+100

PR #7832, opened 2017-09-26
gnustep-make: fix issue with LDFLAGS

abenson:


abenson:
As it turns out this might not be necessary.

PR #7834, opened 2017-09-26
New package: gnustep-base-1.25.0

abenson:


abenson:
There's no way this is going to work well that I feel comfortable pushing.

PR #7837, opened 2017-09-27
efibootmgr: update to 15.

maxice8:
- Remove GCC7 patch as it is merged upstream
- Reduce the default_to_grub2 patch by making use of [EFI_LOADER](https://github.com/rhboot/efibootmgr/commit/ce9dbb3e59bd6c5aea67aa18bb3c3629483a5722) and [EFIDIR](https://github.com/rhboot/efibootmgr/commit/2d9f9e3d574b3f0785fb6063c2e91b2d9ca7cdd8)


Vaelatern:
This might totally break systems with 32 bit efi but 64 bit hardware.
@the-maldridge, can you perhaps comment?

the-maldridge:
I do not believe this would directly affect hybrid boot, but I can test it later tonight.

the-maldridge:
This has no effect on hybrid boot.

PR #7838, opened 2017-09-27
btrfs-progs: update to 4.13.1.

maxice8:


lemmi:
```
...
	btrfs-restore zstd: no
...
```
`zstd` is being added to btrfs right now. Adding `zstd-devel` seems to enable the support.

maxice8:
Should i add a dependency on it when zstd is merged ?

lemmi:
I'm in favour. Do the @voidlinux/pkg-committers have other opinions?

lemmi:
@maxice8 no other opinions as it seems :) Please go ahead and and add zstd. We'll need it sooner or later anyway.

maxice8:
Done

PR #7851, opened 2017-09-27
New package: onerng-3.5

lonwillett:
Support for OneRNG hardware RNG devices.


lemmi:
#7839 Please do not close and reopen issues. Instead, amend (rebase, squash, fixup) your changes and force-push into your branch.

Vaelatern:
This mechanism for running it seems totally ludicrous. How do other distributions handle this?

This does not appear to fall under the purview of a runit service at all, and it's far too much shell for tooling that upstream probably already has handled.

lonwillett:
@Vaelatern

I'm not quite sure what part you consider ludicrous.

Upstream has udev launch a shell script. That then goes through a convoluted process using "at" and "systemctl" to kill off any existing rngd service and launch its own. I don't think we want to use that, even with the systemd stuff changed to runit stuff. There is really no point in using "at" when we can use "sv". And since "upstream" is basically just one guy, Paul Campbell, who is more of a hardware hacker than a coder, I don't think we can expect the upstream sources to be cleaned up very soon.

I basically just refactored the upstream script, putting the device handling in the "init.sh" script, and rewrote the system stuff for runit and put it in the "run" file. And changed the udev rule to use "sv once".

I did end up leaving the "run" script rather over-parameterised and over-commented and over-checked. I put a list of variables at the top to control everything, mainly because my simple port to Void turned into a major debugging session involving a problem with the firmware, so I wanted to easily control everything. I left it in place with the idea that someone might want it in order to repurpose the device, e.g. for a crypto application. So I suppose it's an artefact of my porting/debugging session.

And the urandom_reseed stuff doesn't really belong there, but it certainly doesn't belong with the device initialisation, and it seemed excessive to write yet another script just for it.

If it will make you happy, I can strip out the excess parameterisation and comments. You are probably correct that it confuses the situation more than it helps anyone.

I could also, of course, move everything to a new script, and just have "run" exec that. Which wouldn't particularly help anyone that wants to work with this, but would produce a nice tidy run script.

Or do you have a deeper objection to some part of this? Is there some reason that udev shouldn't activate a service? Or that this service (onerng) shouldn't be stopping/starting another one (rngd)? Let me know; its not clear to me if you have objections to the process, or just its implementation.

lonwillett:
New version:

Removed all unnecessary parameterisation, and dropped any pretence at being able to easily repurpose the device. Renamed the service, to more accurately reflect this.

I did not actually touch any of the underlying operations.

the-maldridge:
I still think this is a really complicated process to run through and seems very fragile across updates.

lonwillett:
@the-maldridge 

Fragile across updates: yes, very. Unfortunately, I don't see a good way to avoid this (outside of convincing upstream to restructure their code). I did just add a comment to the template to emphasize this. An ameliorating factor is that the bit that needs to be updated manually is fairly straight-forward (linear scan of init script, comparing to upstream script, but sadly not really amenable to using "patch").

I also just converted the 79-onerng.rules file from a provided file to a patch. It is a rather silly patch (replaces the entire file contents), but this should at least force a manual fix when the file is changed by an update. Is this an improvement?

Finally, I removed the urandom_reseed stuff, since it isn't really needed (it's now just a comment in the README.Void file). I had just moved this feature over from the upstream package, but its utility is doubtful.

As for "really complicated process", I don't suppose you can be more specific? The hardware is what it is, and I tried to select the most appropriate approach to manage it. Dealing with crypto stuff, even rngs, requires some care, so some complication is inevitable. I think we do want udev to handle the device, we do need a script to initialise the device, and do need a service to run rngd, and the components do need to synchronise their operation. How should this be done?

Certainly the interaction with the standard "rngd" service is questionable, but then the alternative is to instruct the users to disable rngd. They can do that anyway, and then there won't be any interaction, so including those few lines of code seems harmless enough, and quite possibly helpful when users don't follow the instructions.

PR #7856, opened 2017-09-28
New package: io-2017-09-06

Vaelatern:
@digoleal could you take a look at this and tell me how you would fix the `nopie` issue?

Otherwise it's working at least on native x86_64.


Vaelatern:
Not sure why travis fails.

logenkain:
I decided to try to build it myself. Cloned your repo and it fails for me at the same place travis fails.

Vaelatern:
I again built it locally and it succeeded, on a different box.

pullmoll:
Tested with gcc-7.2.0 now and it fails for various errors of the type "returning address of local variable." Methinks this needs some testing and patching, also because gcc is now default-pie.

PR #7857, opened 2017-09-28
python-gitlab: update to 1.0.1.

maxice8:


srcpkgs/python-gitlab/template
```diff
@@ -1,6 +1,6 @@
 # Template file for 'python-gitlab'
 pkgname=python-gitlab
-version=0.21.2
+version=1.0.1
 revision=2
```
Vaelatern: `revision=1`

maxice8: oops, fixed

PR #7861, opened 2017-09-28
bind: update to 9.11.2.

maxice8:


chneukirchen:
9.10 is "Current Stable", 9.11 is "Early Deployment".

maxice8:
oops, my bad then

PR #7862, opened 2017-09-28
New package: xss-lock-0.3.0

teldra:


teldra:
better version https://github.com/voidlinux/void-packages/pull/7964

PR #7873, opened 2017-09-29
python-sqlparse: update to 0.2.4.

maxice8:


maxice8:
Updated on e45812674d413bc17ca17bdd16e4b59e6c46095b

PR #7887, opened 2017-09-30
libixion11 and liborcus11 no longer needed

asergi:
@pullmoll


pullmoll:
Sure. I don't think we'll need them again. Preparing an update libreoffice-5.4.1.2...

PR #7890, opened 2017-09-30
orientdb: update to 2.2.28.

maxice8:


srcpkgs/orientdb/template
```diff
@@ -12,7 +12,7 @@ license="Apache-2.0"
 patch_args="-p1"
 homepage="http://orienttechnologies.com"
```
lemmi: The homepage doesn't exist anymore.

maxice8: Thanks, fixed

PR #7899, opened 2017-10-01
go-ipfs: update to 0.4.11.

CMB:
This does not crossbuild anymore and I can't figure out why.
Looks like it tries to rebuild the go runtime or something on
cross builds.
If someone figures that out before merging, that'd be great.


Vaelatern:
testing cross build fix now

Vaelatern:
eh, `collect2: fatal error: cannot find 'ld'`
Can you remove nocross so relevant logs are generated?

lemmi:
4.11 introduces plugins. I think we should make the subpackages. Currently there is a **git** and **ethereum** plugin.

CMB:
> Can you remove nocross so relevant logs are generated?

Yes.  Pushed without nocross.

CMB:
Fixed the cross-build issue and pushed.  Builds cleanly on travis for
everything.

PR #7907, opened 2017-10-02
bitlbee: add libpurple

the-eater:
This will add libpurple support and also fix the building of bitlbee with openssl

### DIFF
--- a/srcpkgs/bitlbee/accept_openssl.patch
+++ b/srcpkgs/bitlbee/accept_openssl.patch
@@ -1,0 +1,13 @@
+diff --git a/configure b/configure
+index e2357815..efef1a31 100755
+--- a/configure
++++ b/configure
+@@ -620,8 +620,6 @@ elif [ "$ssl" = "openssl" ]; then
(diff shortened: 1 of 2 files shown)


thypon:
Maybe a disabled option is better in this case.

the-eater:
for what?

thypon:
`libpurple` was disabled on purpose by `bitlbee` developers. 

This is what they say about the `libpurple` support the developers and one of the reason it is disabled by default.

> The downside is it adds a lot of dependencies (on most Debian systems it means installing around 100MBytes of extra packages), and it makes the program more resource-hungry at runtime. Also, its huge codebase may have more stability/scalability issues when running on large installations.

the-maldridge:
nomerge until a default-off build option is in place.

libpurple's questionable support for protocols is infamous.

jcjordyn130:
Would a subpackage be any better?

Gottox:
Not possible with bitlbee. I'm in favor of adding it as build option with default off.

PR #7915, opened 2017-10-02
SDL2: add -rpath=/opt/vc/lib on rpi archs 

ebfe:
Linking against SDL2 currently fails on armv[67] because ld apparently can't find libbcm-host.so. 

Better solutions?

### DIFF
--- a/srcpkgs/SDL2/template
+++ b/srcpkgs/SDL2/template
@@ -1,7 +1,7 @@
 # Template file for 'SDL2'
 pkgname=SDL2
 version=2.0.6
-revision=1
+revision=2
 build_style=gnu-configure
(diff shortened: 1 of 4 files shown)


ebfe:
cc @voidlinux/pkg-committers

pullmoll:
Oh, I hadn't seen this and patched around for smpeg2, SDL2_image, SDL2_mixer and SDL2_net as well as rocksndiamonds, which failed to cross build for the same reason.

I guess using -rpath here is the correct fix. Then the my patching around can be reverted.

Gottox:
👍

ebfe:
@thypon what are you doing? ;)

thypon:
@ebfe sorry, I was trying the github inline editing. Didn't thought gives these results.

PR #7923, opened 2017-10-02
New package: ms-core-fonts-2.0

faulesocke:
The Microsoft core fonts as published on Sourceforge.


ebfe:
https://github.com/voidlinux/void-packages/blob/master/srcpkgs/msttcorefonts/template ?

faulesocke:
Ah, shit, it's nonfree and restricted, therefore I didn't found it. Sorry.

PR #7924, opened 2017-10-02
yakuake: update to 3.0.4

djames1:


Chocimier:
Do you plan to add also other kde5 apps to repo?

djames1:
@Chocimier yakuake is the only KDE program that I use, but I do have more applications I plan to add to repo such as SeaMonkey and Remmina

phikal:
Will/can SeaMonkey be added? Or is there any progress on that?

djames1:
@phikal I tried to package seamonkey, but ran across various issues. I will try again at some point. Until then, if you are using x86 or x86_64 GLIBC builds, there are prebuilt binaries available on the official SeaMonkey home page.

PR #7929, opened 2017-10-03
New package: joker-0.8.5

SevereOverfl0w:
Cannot use the go build style due to joker being a bootstrapped application
which requires some double building.


Vaelatern:
What is the procedure if joker can be depended on to already exist?

SevereOverfl0w:
@Vaelatern I _believe_ that it would work with the normal go build style. As I think it calls generate on your behalf.

SevereOverfl0w:
It seems that this package is incomplete due to the bootstrap process only being partial.

Apparently the correct way to bootstrap is to download a recent-ish x86 build & use that under a normal go build, and cross-compile for any architectures.

Could I have some advice on how to _only_ allow cross compilation for a package? I suspect it isn't possible. In which case I can mark this package as only for x86_64, which would be a bit of a shame.

SevereOverfl0w:
Joker author has removed need for bootstrapping. I will update this package in response.

PR #7931, opened 2017-10-04
zfs: update to 0.7.2.

maxice8:


maxice8:
updated on 31fd9288a1fcb21509809007a3f0bbc3fec149a6

PR #7939, opened 2017-10-04
okular: update to 17.08.1

Chocimier:


the-maldridge:
@voidlinux/pkg-committers we already have okular, do we want to add a new one for qt5 or start aging out qt4?

I'm in favor of aging out qt4.

lemmi:
okular(-qt4) doesn't work with cups anymore. If this package works, there is no reason to keep the old okular. I see that I test this today.

Gottox:
We should replace okular4 imho.

@Chocimier can you change this pr or open a new one that updates okular-4 instead?

lemmi:
So I finally got around to try it out. The Cups issue isn't resolved, so maybe that's another issue.
@Chocimier I'm missing some icons, the rest works.

Chocimier:
Now updates old okular package. Added icons dependency. 
Libokular as dependency for calligra isn't yet tested.

PR #7943, opened 2017-10-05
ghostscript: update to 9.22

Hoshpak:
the update includes fixes for the following security issues:
CVE-2017-9611, CVE-2017-9612, CVE-2017-9726, CVE-2017-9727, CVE-2017-9739,
CVE-2017-9835 and CVE-2017-11714


chneukirchen:
What's the rationale for the cross changes?

configure_args should include --build etc no?

Hoshpak:
Thanks. I didn't know that --build and --host were common for configure scripts and already included by the build style, so conditionally running different configure commands was completely unneccessary, I removed it.  Between the two releases, the ghostscript developers have improved the cross-compilation support and I wanted to use that now instead of compiling binaries manually.

The CCAUX and CFLAGSAUX are to set the host compiler and it's flags to use to compile binaries which are executed during the build and therefore need to be compiled for the host architecture.

chneukirchen:
Much better now, thanks.

PR #7948, opened 2017-10-06
poppler: update to 0.60.1. (soname bump)

ebfe:
build switched to cmake


ebfe:
796132fd5cdb85fe2ee9fc3b6da67ffbb79e9c39

PR #7950, opened 2017-10-06
ghostwriter: add install message for required packages

abenson:


Vaelatern:
Install message is very long, too many lines to be easily parsed and ignored if need be. I don't like it.

abenson:
I can see that. I removed the extra lines.

PR #7918, opened 2017-10-02
New package: gnustep-base-1.25.0

abenson:

PR #7956, opened 2017-10-06
New packages: gnustep-{base,gui,back}

espectalll:


Anachron:
You may want to squash your commits.

espectalll:
OK, but I can't ATM. If someone else can do it sure, otherwise it'll take me some hours.

KenjiTakahashi:
Related: #7918 

Funny how multiple people jumped into this at the same time. I have my own, too, but refrained from submitting, seeing there are already multiple other floating around.

Yours is definitely the most complete, though 👍 .

espectalll:
Oh, didn't see the other PR! Seems like interest is growing a bit right now, from what I can tell on Twitter, Reddit, Google Trends, etc. In particular, I decided to package it as I saw people talking about it, I wanted to try it out (again), and now I know how to package stuff.

My packages are not perfect. It's still mixing up both development and runtime libraries. But hey, it works~

common/shlibs
```diff
@@ -2872,3 +2872,5 @@ libvala-0.38.so.0 libvala-0.38.1_1
 libvaladoc-0.38.so.0 libvala-0.38.1_1
 libphodav-2.0.so.0 phodav-2.2_1
 libgfshare.so.2 libgfshare-2.0.0_1
+libgnustep-base.so.1.25 gnustep-base-1.25.0_1
+libgnustep-gui.so.0.25 gnustep-gui-0.25.1_1
```
Vaelatern: newline at end error

srcpkgs/gnustep-back/template
```diff
@@ -0,0 +1,21 @@
+# Template file for 'gnustep-back'
+pkgname=gnustep-back
+version=0.25.1
+revision=1
+build_style=gnu-configure
+configure_args="--prefix=/usr --sysconfdir=/etc/GNUstep"
```
Vaelatern: `s:--prefix=/usr::`

srcpkgs/gnustep-back/template
```diff
@@ -0,0 +1,21 @@
...
+hostmakedepends="pkg-config gcc-objc"
+makedepends="gnustep-make gnustep-base gnustep-gui cairo-devel libXt-devel
+ libXmu-devel"
+depends="gnustep-base gnustep-gui cairo libXt libXmu"
+only_for_archs="i686 x86_64"
```
Vaelatern: Really?

Vaelatern: Please remove the line. It is wholly uninformative and does not let us see what could be very easy fixes to add support.

espectalll: OK then! I thought it was the only way to exclude broken platforms, but turns out there's the `broken` variable. OK, I'll change it asap.

KenjiTakahashi: @Vaelatern There are no ObjC compilers available for ARM in Void. I presume (from reading the gcc template) that it's because they cannot be (easily?) cross-compiled.
And musl doesn't work, because GNUStep requires iconv with TRANSLIT support. It is possible to compile without iconv support, but I don't know if it's desired.

common/shlibs
espectalll: Fixed

the-maldridge:
This PR is an umergeable trainwreck right now.  You need to rebase out where you merged in master, that corrupted the branch.  After that we can look into what can be done or how to accurately set the broken status for the other architectures.

PR #7963, opened 2017-10-07
Update yarn{,-bin} to 1.1.0

the-eater:


Vaelatern:
Is there a reason we keep the `-bin` package around?

the-eater:
@Vaelatern https://github.com/voidlinux/void-packages/pull/6949#issue-240273378

I vote for keeping it in tho, otherwise it would break building the repo from scratch

Vaelatern:
Then perhaps we should have a bootstrap build option that defaults to off, which makes us want `yarn-bin` instead of `yarn`.

the-eater:
We may be able to use that also for `go` I imagine?

the-maldridge:
@Vaelatern bootstrap should be simple, I'm fine with having both normal and bin, and there are cases where it may be preferable to use -bin.

PR #7966, opened 2017-10-07
New package: keybase-bin-1.0.34.20171006165932.d0664a9cb

SevereOverfl0w:
Pretty sure this doesn't meet packaging requirements.

Doing it this way because it is a number of separate modules rolled into one release.
* One is not released
* The others are a combination of go modules & electron builds, these I think are (technically) not built
* I couldn't find any thorough build instructions for them all

This is mostly for my own personal uses, and I am happy to update it for others
too.


Vaelatern:
Any reason why the regular keybase package is not enough?

Vaelatern:
There is enough oddity around the binary releases of keybase that I'd like to avoid those.

SevereOverfl0w:
I need access to the kbfs/chat functionality, which aren't present/fully functional in the regular keybase package.

I agree, they're definitely odd. But I need to use them. Happy to keep this just to myself.

Vaelatern:
yes.

https://github.com/keybase/kbfs/issues/686

considering packaging kbfs anyway, but I really hate to without a version tag. Upstream doesn't seem to understand that people like building from source.

With kbfs, chat and kbfs are present. Would that functionality preclude your need for these pre-built binaries?

SevereOverfl0w:
I think the only convenient way to chat is via the electron app, but that might be easier to package anyway, as it's part of the client. I haven't actually taken a hard look at that yet.

I see you're familiar with the odd state that is kbfs/client split.

Vaelatern:
If you want to try packaging the electron app for that, I'll happily review a PR. But good luck.

SevereOverfl0w:
One concern is perhaps the `run_keybase` script, which is intended to orchestrate all of these loosely coupled services.

I'm guessing that would need packaging into (yet another) separate package?

Vaelatern:
I am not familiar with the script.
It might make sense to include it in a keybase package.

SevereOverfl0w:
It doesn't do too much, but it's the official way to make stuff happen around kbfs & such: https://github.com/keybase/client/blob/master/packaging/linux/run_keybase

Looking closer, I feel like I might replace it with sv services that come up with login or something.

PR #7979, opened 2017-10-08
cups: include various fixes in order to standardize the usage of cups

thypon:


srcpkgs/cups-filters/files/cups-browsed/run
```diff
@@ -1,3 +1,2 @@
 #!/bin/sh
-sv check avahi-daemon >/dev/null || exit 1
```
Vaelatern: Maybe check for if the service is supposed to be enabled?

thypon: avahi-daemon is not needed anymore if you have already configured the remote printer or you are not using DNS-SD protocol

srcpkgs/cups/template
```diff
@@ -27,6 +27,7 @@ build_options="gssapi"
 
 # Package build options
 system_accounts="cups"
+system_groups="lpadmin"
```
Vaelatern: can we start system groups, especially new ones, with an underscore?
https://github.com/voidlinux/void-packages/commit/78548dba2a5a4c74d5940107a6b2c9a0aa8e1c9f

thypon: I dunno. lpadmin is more or less standard in debian/slackware systems. It is also consistent with the lp and cups group/user.

Maybe we should discuss a bit more this particular case with the @voidlinux/pkg-committers .

the-maldridge: While I agree that many systems use the lpadmin group, it is generally used for legacy reasons from when you were running both lpd and cupsd simultaneously and desired to manage them with the same group.  Since cups can fill the function that lpd once had, and since we have a general policy of not altering the upstream configuration of upstream packages, I would prefer to have cups run as cups:root and be managed by users in the cups group.

thypon: @the-maldridge https://github.com/apple/cups/blob/b94f7488e861c3119037efdf08257749fcce06b8/config-scripts/cups-defaults.m4#L236 many system uses lpadmin because cups expects to found that group in its system groups. What instead we have is the root and the sys group that are too generic to represent the lpadmin. 

We are not altering the upstream config, but instead we are forcing lpadmin when lpadmin is not already installed on top of the distro.

PR #7997, opened 2017-10-09
libvirt: enable zfs storage backend

mporrato:
https://www.redhat.com/archives/libvir-list/2016-January/msg00005.html


Vaelatern:
I think the change currently made to `hostmakedepends` is better made to `makedepends`, can you please verify?

mporrato:
Moving zfs-devel to makedepends seems to break cross builds.

Vaelatern:
What about adding zfs to hostmakedepends and zfs-devel to makedepends?

ebfe:
Please make it optional and default to off. Installing libvirt shouldn't pull in out-of-tree kernel modules, at least by default.

mporrato:
I'm not sure I fully understand your concern @ebfe . The resulting binary package does not require and does not pull in zfs at all. This patch only builds an extra .so file which is dinamically loaded when/if the actual feature is needed at runtime.

ebfe:
Ok, but building libvirt still involved building the zfs dkms module, so
maybe the solution is to split off the dkms parts into a separate
package.

Vaelatern:
Agreed, will investigate.
Separately, @mporrato, does the build work if you replace `zfs-devel` with `zfs`?

mporrato:
Both `zfs` and `zfs-devel` in makedepends break the builds.
`zfs` in hostmakedepends looks ok.

ebfe:
The building libvirt involves building zfs kernel modules still is an issue.

pullmoll:
@ebfe please elaborate on that for someone who doesn't understand what you mean.
libvirt built locally and on the builders without depending on or building zfs kernel modules?
Installing libvirt-devel still does not require zfs kernel modules being built AFAICT.

pullmoll:
Closing this was too quick, because of 1695b451a6ed2fb1219f46fa1dd97f968380742b

ebfe:
> Installing libvirt-devel still does not require zfs kernel modules being built AFAICT.

Installing (or configuring) zfs triggers dkms.

pullmoll:
So let's make zfs support a default disabled option?

ebfe:
@Vaelatern did you run into any problems with splitting out the dkms part into a separate package? Or just didn't get a chance to look at it?

Follpvosten:
I have looked at this because i want to switch to zfs for my main file system - in my own branch for this, Follpvosten:libvirt-zfs, i added the build option and added zfs to the makedepends, and that seems to work without any problems. Should i create a new PR? Or is there something else i can do?

Vaelatern:
Have not had a chance to work on the dkms parts of this.

Vaelatern:
If anybody wants to continue this work on the new void-linux group, that would be welcome.

PR #8000, opened 2017-10-10
New package: python-jmespath (usefull for ansible)

yopito:
NB: this package is almost required by `ansible`, since `ansible`'s `json_query` rely on it.


the-maldridge:
I swore I packaged this ages ago, but I probably never pushed it.  Thanks for pushing your copy!

yopito:
orphaned commits are so sad ;)

PR #8001, opened 2017-10-10
evolution-data-server: unbreak crossbuild

yopito:
related to issue #7713


yopito:
unexpected build fails with x86_64-musl (native). need deeper investigation.

yopito:
here it is.

PR #8002, opened 2017-10-10
qt5: update to 5.9.2

pullmoll:
New components: `qt5-networkauth`, `qt5-purchasing`, `qt5-remoteobjects`
Moved `lconvert`, `lupdate` and `lrelease` from `qt5-tools` to `qt5-host-tools`

After this qt5-doc will be updated and most probably also qt5-webkit.


pullmoll:
The current show stopper is this: we will need the `armv*hf{,-musl}` builders running in an `i686` environment, not in `x86_64` as it is now!?

I verified it is possible to cross build qt5-5.9.2 for `armv*hf` from `i686` but **not** from `x86_64`.

It may be possible to get `armv[67]hf` cross built in an `x86_64` environment with the help of `gcc-multilib` and some `*-32bit` libraries, but I could not yet solve this path. In any case this can't work for cross `armv*hf-musl` because no `gcc-multilib-musl` exists and noone would want to write it.

The reason why a 32 bit build environment is necessary for 32 bit targets (which `arm*` are) is in `chromium` and its `v8` ECMA-script engine, which is part of `qtwebengine`, and the way chromium builds some host binaries to be run during the build.

**In short the prerequisite before updating qt5 to 5.9.2 is: make the arm builders run in an i686 environment.**

Note that `i686-musl` has its own issues and won't work as an environment for armv*hf-musl due to other issues. Lucky for us i686 can cross build for *-musl as well.

yopito:
Hello there
* could you add a `qt5-quickcontrols2-devel`  subpackage ?
  Probably needed for subsequent PR beyond #8268.
* some other distros don't use a monolithic packaging for qt5 , but instead separated ones  (downloads from http://download.qt.io/official_releases/qt/5.9/5.9.2/submodules/).
  might it make sense for us ? I'm remember you taked about it, but i fail to retrieve what was your feeling about it.

Thx - Yopito

pullmoll:
@yopito I can easily add a qt5-quickcontrols2-devel, but does that make sense? Are there header files etc.?  I'll see when I try.

For splitting up qt5 I spend literally weeks on trying to do that with 5.9.0 IIRC but entirely failed to coss compile for any target, which makes that try pointless in the first place.

yopito:
@pullmoll : thanks for the details about monolithic packaging. worse than ever ...
 
about `qt5-quickcontrols2-devel` : yes, please add. I need it for a new kde framework package, it's confirmed.

pullmoll:
qt5-quickcontrols2-devel is there now and contains headers, pkgconfig etc.
I didn't originally add it to the template because of being under the impression that quickcontrols2 was just like quickcontrols which has no development files.

pullmoll:
+ Fixed building of qmake for the host architecture.
The way it was before, doing a clean and rebuild with Void's CXXFLAGS and LDFLAGS resulted in a qmake which would print just empty strings for most fields in a `qmake -query` call. After the fix now it prints the expected (configured) strings and building e.g. lxqt with qt5-5.9.2 succeeds again.

pullmoll:
Qt5-5.9.3 is out. Closing this one and will open PR for 5.9.3 once it was locally built for all targets.

PR #8007, opened 2017-10-10
Bear: update to 2.3.7.

maxice8:
- Add libBear subpkg for libear.so
- Add patch that removes a include to xlocale.h, fixes build, no
side-effects
- add libear.so to common/shlibs


ebfe:
Since nothing else links to "libbear.so" and it isn't versioned, i think a common/shlibs entry + subpkg is a bit overkill for now.

maxice8:
I'll scale it down

maxice8:
Removed libBear subpkg and entry from common/shlibs

PR #8008, opened 2017-10-10
onboard: add missing runtime dependency

circhioz:
`onboard` fails to start in some circumstances without `python3-dbus`.


liwakura:
I cannot reproduce, it works for me without `python3-dbus`.

circhioz:
It fails to start for me under Plasma 5.

PR #8017, opened 2017-10-10
New package: recoll 1.23.3

equalsraf:
The python modules are not built, and x11 monitor support is disabled.


equalsraf:
I cant fix this today musl is failing likely because of utils/pxattr.cpp, it needs xattr.h

srcpkgs/recoll/template
```diff
@@ -0,0 +1,23 @@
...
+distfiles="https://www.lesbonscomptes.com/$pkgname/$pkgname-${version}.tar.gz"
+checksum=15125871424ef32ac584eb1c3e4fdd4ac7ff32894db60a3ebff32fa7a45d2eb6
+
+pre_configure() {
+	autoreconf -fi
+}
+
+post_install() {
+	# remove manpage since the GUI binary is not installed
+	rm ${DESTDIR}/usr/share/man/man1/recoll.1
```
Vaelatern: why not install the gui binary? Seems strange to disable all of the things you do....

equalsraf: its kind of annoying to install the whole gui deps (x11, qt5webkit) when you just want the cli maybe i should just break these into subpackages. recoll has 3 binaries

- recollindex: the indexer, just a cli tool (but may depend on x11 depending on the config flags --disable-x11mon)
- recollq a command line search interface
- recoll: the gui frontend (this calls recollindex)

Updated the PR. I've added the gui binary, but kept the x11mon support in recollindex disabled to avoid the dependency.

pullmoll:
Fixed *-musl and cross building: https://github.com/voidlinux/void-packages/commit/6b16960b2eb10e19e90d8c54f841dad48b4c5c09

PR #8024, opened 2017-10-11
mongoose: update to 6.9.

maxice8:
Big changes

- Makefile was removed on https://github.com/cesanta/mongoose/commit/c0af018e205a08618ed91367c0e783afcea1f5c0 so we restore it via patch, using a simplified Makefile from a [debian maintainer](https://gist.githubusercontent.com/jpleau/327b2653633e97d955fe/raw/e2c87ddf7f51d5e1571cf1df90f31e87052286b7/Makefile) and void's [changes](https://github.com/voidlinux/void-packages/blob/72035e6b43fa26cc0228a106a49a9c7e5db85a07/srcpkgs/mongoose/patches/mk-tweaks.patch), afterwards change build_style to gnu-makefile and use make_use_env
- The binary was removed (now available on their [website](https://cesanta.com/binary.html)), so remove libmongoose as there is no more separation necessary, mongoose is now the lib itself
- add mongoose-doc package that includes doc/ from upstream and examples
- update homepage from dead code.google.com/p/mongoose to the company's website
- update distfile to use their github
- License was changed to [dual licensed GPL-2 Commercial](https://github.com/cesanta/mongoose/commit/218428662e6f8d30a83cf8a89f531553f1156d25) , change license to GPL-2


the-maldridge:
Thanks for the well formed PR message!

PR #8025, opened 2017-10-11
Fix a few xlint complains

Johnnynator:


chneukirchen:
We try to avoid these mass commits, and instead fix things when we touch the templates.

Also, changes of short_desc require a revbump.

Johnnynator:
Would you prefer me to just close this (and  or just split it into single commits?

chneukirchen:
I'd leave this for the maintainers to pick up upon.

Thanks, still.

PR #8044, opened 2017-10-11
pioneer: update to 20171001

cr6git:


Johnnynator:
Can you add this commit https://github.com/pioneerspacesim/pioneer/commit/ce10affe481b57f57f2cd1831d3734ee1d51b4b3
currently pioneer just segfaults because of SDL 2.0.6 and this should fix it

cr6git:
Thank you for the waste of time... now I must revert this commit again.

PR #8047, opened 2017-10-11
fwup: update to 0.16.0.

maxice8:


maxice8:
updated on 70355325e1baf23594789ddd60b1b640ca03d16d

PR #8061, opened 2017-10-12
node_exporter: Fix outdated run script

dne:
Remove `-collectors.enabled` flag -- replaced by individual boolean flags in v0.15.0.

See prometheus/node_exporter#640.


xtraeme:
revision must be increased due to different binpkg contents

dne:
Of course, sorry about that!

PR #8063, opened 2017-10-12
youtube-viewer: update to 3.3.1

cr6git:


maxice8:
If memory serves right adding perl to `makedepends` should fix cross-compilation

cr6git:
I noticed 'perl-module-build' in the AUR [PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=youtube-viewer-git)

So I modified the Void template to include: 
````
build_style=perl-ModuleBuild
hostmakedepends="perl-Module-Build"
````
It fixes the cross-compilation errors ! :)

cr6git:
Looks like it broke off something in [#8184](https://github.com/voidlinux/void-packages/issues/8184) ? 😕

PR #8067, opened 2017-10-12
mbuffer: update to 20171011.

diogoleal:


chneukirchen:
I reported some issues upstream and they were fixed, so we can simplify the template even further.

PR #8070, opened 2017-10-12
CopyAgent: removed, because is was discontinued

diogoleal:
Copy was discontinued on May 1, 2016.


chneukirchen:
> - If you've removed a package use `"<pkgname>: removed ..."`.

PR #8075, opened 2017-10-12
New package: remmina

cr6git:
Package request [#8134](https://github.com/voidlinux/void-packages/issues/8134).


srcpkgs/remmina/template
```diff
@@ -0,0 +1,18 @@
+# Template file for 'remmina'
+pkgname=remmina
+version=1.2.0rcgit.21
```
the-maldridge: This is not a release version.  Please back up to the last stable release and update this PR.

srcpkgs/remmina/template
```diff
@@ -0,0 +1,18 @@
...
+revision=1
+_gitver=1.2.0-rcgit.21
```
the-maldridge: As general policy we do not accept packages at git revisions.

cr6git:
Back to the latest stable release 1.1.2 (dated 2014-12-08).

srcpkgs/remmina/template
cr6git: ok I modified the template.

cr6git:
@the-maldridge why it is still blocked ?

the-maldridge:
Because life is busy. I'll try to get to this tomorrow but maybe @Vaelatern
can get to it faster.

PR #8082, opened 2017-10-12
libXfont2: update to 2.0.2.

maxice8:


maxice8:
updated on a01e1903bb558d51dc553383933b61296f9cb50e

PR #8085, opened 2017-10-12
kdenlive: update to 17.08.2.

jnbr:


lemmi:
```
-- Qt5 WebKitWidgets not found. You cannot use your Freesound.org credentials, only preview files can be downloaded from the Online Resources Widget
```

```

 * LibV4L2, Collection of video4linux support libraries, <http://freecode.com/projects/libv4l>
   Required for better webcam support

-- The following OPTIONAL packages have not been found:

 * Qt5WebKitWidgets
```
Could you add the v4l stuff? I don't know how important the webkit dependency is.

lemmi:
Since travis doesn't like kdenlive, please add `[ci skip]` to the commit message.

jnbr:
Thank you for the review. I added both, v4l and webkit.

PR #8096, opened 2017-10-13
botan: update to 2.3.0. (soname bump)

ebfe:


ebfe:
bcde472a3545ab75cec6ef672c7206d4ef6a56a8

PR #8019, opened 2017-10-11
New package: pulsemixer-1.3.0

z1lt0id:
Re-submitting with vlicense. Had to resubmit due to corrupt repo.

PR #8097, opened 2017-10-13
New package: pulsemixer-1.3.0

z1lt0id:
Referencing old pull request:

[New package: pulsemixer-1.3.0 #8019](
https://github.com/voidlinux/void-packages/pull/8019/files/4502d83f7529eafbece57702c807d110f6f1dc7b)

Removed fuseiso (obselete) and updated pulsemixer with appropriate source from author with appropriate licence file.

I haven't closed the other one off.  I did look into trying to merge this pull request with the other one but I couldn't find any clear information or if it was even possible after deleting the previous repository (I really do I apologise).


the-maldridge:
You say this is a curses program but it doesn't seem to depend on curses?  Does it use an internalized version?

z1lt0id:
You are right, not sure why I thought it requires curses.  I think I saw it as a dependency in Arch and for some reason I put it in there.  I'll remove the dependency from the package and amend the package.

z1lt0id:
I think the creators title is a bit-misleading it curses-like without using the curses library.

the-maldridge:
Ah, that makes sense.

Merged.

PR #8098, opened 2017-10-13
New package: pan

cr6git:
Forum request.


newbluemoon:
Good to see you back, @cr6git :-)

Don’t want to be a smart ass, but just thought that it might be worth mentioning that you can omit introducing a new variable `_release=0_142` when all you need is replacing the dot with the underscore:

    distfiles="https://github.com/GNOME/pan2/archive/PAN_${version/./_}.tar.gz"
    wrksrc="pan2-PAN_${version/./_}"

PR #8105, opened 2017-10-13
libvdpau-va-gl: fix xlint warnings, make it nocross.

maxice8:


the-maldridge:
What is the impact of making this nocross, i.e. what will break?

maxice8:
On Sun, Oct 15, 2017 at 10:35:29PM +0000, Michael Aldridge wrote:
> What is the impact of making this nocross, i.e. what will break?
> 

Nothing depends on it apparently (xbps-query -RX libvdpau-va-gl) so i don't expect anything will break, it will just stop failing to cross-compile when future version bumps are made

PR #8115, opened 2017-10-13
lz4: add missing license.

maxice8:


asergi:
> Relevant license is reminded at the top of each source file,
> and with presence of COPYING or LICENSE file in associated directories.

So there is no need to install a stub LICENSE file just to make a linter happy. Thanks anyway.

PR #8117, opened 2017-10-13
xbps-src: preserve proxy environment variables

### LINKED ISSUE #8083: xbps-src: fetching files fails when running behind a proxy
`xbps-src` does not always respect the global proxy settings specified using the `http_proxy` and `https_proxy` environment variables.
The `binary-bootstrap` command does not seem to be affected but `pkg` is: https://gist.github.com/anonymous/af06c19421af9cb965ee11000e3b182e

The only workaround I found is adding the proxy environment variables in `common/environment/setup/misc.sh` but it's definitely not a good solution.

mporrato:
Fixes #8083


Duncaen:
any reason to not do something like `env -- ${http_proxy+http_proxy=${http_proxy}} ...`

mporrato:
No special reason. IMO, this implementation is easier to extend in case there's a need to add more variables to the list.

Vaelatern:
Please at least use `$()` instead of `` to denote a subshell.

ebfe:
I'm not sure if passing through environment variables into the chroot is a good idea at all. Explicit configuration via an XBPS_PROXYS variable or similar in /etc/conf keeps the chroot independent.

xtraeme:
Exactly

2017-10-16 11:16 GMT+02:00 Michael Gehring <notifications@github.com>:

pulux:
I'm work behind a proxy, set XBPS_PROXYS in /etc/conf dosn't work.

This PR works well.

ebfe:
> I'm work behind a proxy, set XBPS_PROXYS in /etc/conf dosn't work.
> This PR works well.

I neither claimed that this PR doesn't work nor that setting XBPS_PROXYS currently works. I just don't think that unconditionally passing through environment variables is a good idea.

mporrato:
What about using a sudo-like approach where you can define the list of the variables you want to passthrough in a conf file? (`env_keep` in `/etc/sudoers`)

maxice8:
any updates on this ?

Issue #8010, opened 2017-10-10
GNOME metapackage broken


ghost:
The Metapackage for the GNOME desktop is broken due to the move to version 3.26.
It depends in newer versions of gnome-calculator and gnome-calendar than there are in the repo.

It is looking for gnome-calculator>=3.25.92 and gnome-calendar>=3.26.0 .
The versions available in the repo are: gnome-calculator=3.24.0 and gnome-calendar=3.24.3.

freedomrun:
$ sudo xbps-install -Su
[*] Updating `https://repo.voidlinux.eu/current/x86_64-repodata' ...
[*] Updating `https://repo.voidlinux.eu/current/multilib/nonfree/x86_64-repodata' ...
[*] Updating `https://repo.voidlinux.eu/current/multilib/x86_64-repodata' ...
[*] Updating `https://repo.voidlinux.eu/current/nonfree/x86_64-repodata' ...
MISSING: gnome-calculator>=3.25.92
MISSING: gnome-calendar>=3.26.0
MISSING: gnome-multi-writer>=3.26.0
MISSING: nautilus-sendto>=3.8.6
Transaction aborted due to unresolved dependencies.

Sir-Boops:
Are you guys still having the issue?

freedomrun:
Yes, there are still holdups:

MISSING: gnome-multi-writer>=3.26.0
MISSING: nautilus-sendto>=3.8.6
Transaction aborted due to unresolved dependencies.

greywolf01:
For me it is Wirkung fine now.
Did a fresh install this morning and I'm now running GNOME 3.26.
The metapackage die work withkut a Problem.

TBH I don't know why it would ask for multi-writer and sendto as a dependency.
At least it didn't do that in my end.

greywolf01:
For me it is Wirkung fine now.
Did a fresh install this morning and I'm now running GNOME 3.26.
The metapackage die work withkut a Problem.

TBH I don't know why it would ask for multi-writer and sendto as a dependency.
At least it didn't do that in my end.

circhioz:
`gnome-apps` depends on `nautilus-sento` and `gnome-multi-writer`.
I tried to build multi-writer, but it requires an unreleased version of `polkit` (it hasn't got a revbump in 2 years...)

circhioz:
Unresolved dependencies have been removed in f1f0a21db165f7a9dfbd59d1fe33264d5cfff154.
[...]

PR #8119, opened 2017-10-13
gnome-calendar: update to 3.26.2.

circhioz:
This PR addresses issue #8010 together with #8118.

Switched to `build_style=meson`.


circhioz:
@the-maldridge got the meson build working

PR #8141, opened 2017-10-14
caddy: update to 0.10.10.

SevereOverfl0w:


srcpkgs/caddy/template
```diff
@@ -1,6 +1,6 @@
 # Template file for 'caddy'
 pkgname=caddy
-version=0.10.8
+version=0.10.10
 revision=2
```
KenjiTakahashi: Reset revision?

SevereOverfl0w: Oops, yes, fixed.

PR #8144, opened 2017-10-14
xlockmore: add missing license

cr6git:
Added license from the Arch [package](https://git.archlinux.org/svntogit/community.git/tree/trunk?h=packages/xlockmore).


the-maldridge:
Please refactor this to obtain the license from an official source.  Versioning the license with void-packages will undoubtedly lead to an out of date license.

Vaelatern:
It looks like it comes from a source file, so maybe judicious `sed` can get the license out for you.
See the spim template, line 18, for an example.

cr6git:
@Vaelatern OK, I used `sed` to get the license text from `xlock/xlock.h`

srcpkgs/xlockmore/template
```diff
@@ -21,4 +21,7 @@ do_install() {
 		bindir=${DESTDIR}/usr/bin mandir=${DESTDIR}/usr/share/man/man1 \
 		prefix=${DESTDIR}/usr install
 	chmod u+s ${DESTDIR}/usr/bin/xlock
+
+	sed -n 56,74p xlock/xlock.h > LICENSE
```
the-maldridge: this looks really fragile.  Is it possible to do this with anchors in the text?

cr6git: anchors ? excuse-me, what is it ?

the-maldridge: like clipping from a point in the text by the start of license to the end of the license.  Though I doubt xlockmore chances much anymore.  If that is infeasible I am satisfied to merge it as is.

cr6git: me I go to sleep.

PR #8170, opened 2017-10-14
New package: rssguard

cr6git:


cr6git:
Sometimes Travis is not working properly. 😕

PR #8183, opened 2017-10-15
gcompat: update Web URL and source URL

awilfox:
* 0.1.0 is released
* Do not use GitHub as it is not the authoritative Web URL for gcompat
* Official tarballs are on distfiles.adelielinux.org
* LICENSE is included in the tarball now

### DIFF
--- a/srcpkgs/gcompat/template
+++ b/srcpkgs/gcompat/template
@@ -1,18 +1,15 @@
 # Template file for 'gcompat'
 pkgname=gcompat
-version=0.0.1
+version=0.1.0
 revision=1
-wrksrc="${pkgname}-${pkgname}-${version}"
+wrksrc="${pkgname}-${version}"
 build_style="gnu-makefile"
 short_desc="Compatibility layer to allow running glibc binaries on musl systems"
 maintainer="Daniel James <djames@orcadian.net>"
-homepage="https://github.com/AdelieLinux/gcompat"
+homepage="https://code.foxkit.us/adelie/gcompat"
 license="ISC"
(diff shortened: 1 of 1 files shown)


srcpkgs/gcompat/template
```diff
@@ -41,7 +38,3 @@ make_install_args="LINKER_PATH=/usr/lib/${_musl} LOADER_NAME=${_glibc}
 pre_build() {
 	sed -i 's/gcc/$(CC)/g' Makefile
 }
-
-post_install() {
-	vlicense ${XBPS_SRCDISTDIR}/${pkgname}-${version}/LICENSE
```
maxice8: use `vlicense LICENSE`

licenses aren't automatically installed.

srcpkgs/gcompat/template
```diff
@@ -1,18 +1,15 @@
 # Template file for 'gcompat'
 pkgname=gcompat
-version=0.0.1
+version=0.1.0
 revision=1
-wrksrc="${pkgname}-${pkgname}-${version}"
```
maxice8: wrksrc can be removed as ${pkgname}-${version} is default behaivour

ebfe:
merged with above issues fixes

ebfe:
@awilfox apparently the sysctl syscall is deprecated an not available on aarch64, used a7aa924e41f885dec6daa046bb80a2dca5434244 as a a workaround for now.

awilfox:
Fixed in upstream 03c6a554; you will be able to drop the patch when 0.2.0 is released.  Thanks for reporting.

Issue #8134, opened 2017-10-14
Package request


alexey-nemovff:
Would love to have this packages within the void linux repos to have a better experience with this amazing distro:
- waterfox
- transcode
- pale moon
- remmina
- syncthing-gtk
- mediainfo-gtk

cr6git:
- waterfox => [#8186](https://github.com/voidlinux/void-packages/pull/8186)

chneukirchen:
Remaining package is at https://bitbucket.org/france/transcode-tcforge

PR #8186, opened 2017-10-15
New package: waterfox 55.2.2

cr6git:
Package request [#8134](https://github.com/voidlinux/void-packages/issues/8134).


cr6git:
When I reread [this comment](https://github.com/voidlinux/void-packages/pull/7418#issuecomment-323376543) I think it's not necessary, `waterfox` will never be merged.

PR #8188, opened 2017-10-15
New package: perl-Sort-Versions-1.62

WantToHelp:


WantToHelp:
Travis is failing because the repo mirror is timing out.

cr6git:
All the builds failed today. 😢

WantToHelp:
It seems to be the same mirror causing issues for the other builds as well.

PR #8189, opened 2017-10-15
notepadqq: update to 1.2.0

cr6git:
Removed unneeded CodeMirror stuff.


cr6git:
````
ERROR: [reposync] failed to fetch file `http://198.204.250.219/current/musl/x86_64-musl-repodata': Operation timed out
````

Piraty:
(Following up on the discussion about `qt*-svg` as a runtime dependency here: https://github.com/voidlinux/void-packages/pull/8458)

(I couldn't come up with the source of this information in short time, but i ran into this issue once before)
It needs to be added to `depends` manually, as it is done in other templates too, because it is not recognized as a runtime-dependency automatically when `qt*-svg-devel` is in `makedepends`.
Otherwise, if the build doesn't depend on `qt*-svg-devel`, leave both out.
See https://github.com/voidlinux/void-packages/search?q=qt5-svg for example as reference.

cr6git:
As you can [see here](https://prnt.sc/h22g9u) I don't need `qt5-svg` to run `notepadqq`.

cr6git:
@Piraty for information, what's your arch? 

I'm currently using i686 and x86_64 glibc, but I have not tested notepadqq-1.2.0 on musl.

cr6git:
How much longer must we wait to merge this ?

PR #8206, opened 2017-10-16
perl-rename: use vbin instead of vinstall

maxice8:


srcpkgs/perl-rename/template
```diff
@@ -18,5 +18,5 @@ depends="${hostmakedepends}"
 
 do_install() {
 	vinstall blib/man1/rename.1p 644 /usr/share/man/man1/ perl-rename.1p
```
asergi: `vman blib/man1/rename.1p perl-rename.1p`

srcpkgs/perl-rename/template
```diff
@@ -17,6 +17,6 @@ makedepends="${hostmakedepends}"
 depends="${hostmakedepends}"
 
 do_install() {
-	vinstall blib/man1/rename.1p 644 /usr/share/man/man1/ perl-rename.1p
-	vinstall bin/rename 755 /usr/bin/ perl-rename
+	vman blib/man1/rename.1p perl-rename.1
```
asergi: `s/perl-rename.1/perl-rename.1p/`

PR #8210, opened 2017-10-16
nautilus-CopyAgent: remove broken symlink to CopyAgent

cr6git:
`CopyAgent` does not exist.


asergi:
And `caja-CopyAgent` too. Follow up on #8070.

PR #8218, opened 2017-10-16
rofi-pass: update to 1.5.3.

xYuusha:


maxice8:
There is a 1.5.3 version available.

xYuusha:
Damn, this project gets updated faster than it gets merged.

maxice8:
On Sat, Oct 28, 2017 at 09:14:27PM -0700, Yuusha Spacewolf wrote:
> Damn, this project gets updated faster than it gets merged.
> 

Happens very frequently with things from carnager like clerk but maintainers are also particularly slow that even slow-updating things sometimes get updates.

Issue #8113, opened 2017-10-13
linux-firmware: conflicts with alsa-firmware


ebfe:
```
linux-firmware-20170622_1.noarch <-> alsa-firmware-1.0.29_1.noarch:
	/usr/lib/firmware/ctefx.bin
	/usr/lib/firmware/ctspeq.bin
	/usr/lib/firmware/ess/maestro3_assp_kernel.fw
	/usr/lib/firmware/ess/maestro3_assp_minisrc.fw
	/usr/lib/firmware/korg/k1212.dsp
	/usr/lib/firmware/sb16/alaw_main.csp
	/usr/lib/firmware/sb16/ima_adpcm_capture.csp
	/usr/lib/firmware/sb16/ima_adpcm_init.csp
	/usr/lib/firmware/sb16/ima_adpcm_playback.csp
	/usr/lib/firmware/sb16/mulaw_main.csp
	/usr/lib/firmware/yamaha/ds1_ctrl.fw
	/usr/lib/firmware/yamaha/ds1_dsp.fw
	/usr/lib/firmware/yamaha/ds1e_ctrl.fw
	/usr/lib/firmware/yamaha/yss225_registers.bin
```

cr6git:
Our package is dated 2017-06-22. 
The Debian Sid [package](https://packages.debian.org/sid/firmware-linux) is newer: 2017-08-23. 
The Arch [package](https://www.archlinux.org/packages/core/any/linux-firmware/) is newer: 2017-09-07. 
Maybe we should update our package in the process...

maxice8:
resolved at https://github.com/voidlinux/void-packages/commit/2c177d6ee1f7d8d05f39a3f269dfbe140af930fc

@the-maldridge

PR #8219, opened 2017-10-16
alsa-firmware: prevent conflict with linux-firmware

cr6git:
To help [#8113](https://github.com/voidlinux/void-packages/issues/8113)

### DIFF
--- a/srcpkgs/alsa-firmware/template
+++ b/srcpkgs/alsa-firmware/template
@@ -1,7 +1,7 @@
 # Template file for 'alsa-firmware'
 pkgname=alsa-firmware
 version=1.0.29
-revision=1
+revision=2
 build_style=gnu-configure
 configure_args="--with-hotplug-dir=/usr/lib/firmware"
 nostrip=yes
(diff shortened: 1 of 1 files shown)


cr6git:
Error, again. 
````
[*] Downloading binary packages

ERROR: [trans] failed to download `xbps-triggers-0.102_3' package from `http://198.204.250.219/current': Operation timed out

=> ERROR: Failed to install bootstrap packages!
````

jcjordyn130:
Try another mirror.

cr6git:
it's okay now ! 😃

jcjordyn130:
Nice

Duncaen:
I think the files should be removed from `alsa-firmware` instead.
`alsa-firmware` is newer and users that use on of those firmwares only have `linux-firmware` installed and would need to switch to `alsa-firmware`.

cr6git:
@Duncaen ok I modified `alsa-firmware` instead.

cr6git:
Remove the unneeded files during `post_install()` , do you agree @Duncaen ?

PR #8256, opened 2017-10-17
wine: add x86_64 target

Nicop06:
I tried to run a Steam game using wine64 and it works perfectly.


Sir-Boops:
>64-bit Wine built without 32-bit support will not be able to run ANY 32-bit applications - [src](https://wiki.winehq.org/Building_Wine#Shared_WoW64)

There are still alot of 32 bit windows programs out there

cr6git:
@Sir-Boops Very precise documentation!

Nicop06:
It should work using `WINEARCH`. The only inconvenient is that you can't have both 32 bits and 64 bits applications installed. What I do is that I install each application in a different wine folder and I create a script to run my applications:

```
export WINEARCH=win32
export WINEPREFIX=~/.wineprefix/my_application
wine ./setup.exe
echo -e "#!/bin/sh\nWINEARCH=win32 WINEPREFIX=~/.wineprefix/my_application wine /path/to/my_application.exe" > ~/bin/run_my_application.sh
chmod +x ~/bin/run_my_application.sh
```

I mostly use wine for Windows games.

Also, I noticed that I have a 64 bits wine running Steam for Windows and I could successfully run all the games I have, some of them being pretty old. But maybe they are 64 bits game, although Steam is installed in the `Program Files (x86)` directory. I will do more tests to confirm that 32 bits applications cannot run in wine64.

But I believe that it's better to be able to run 64 and 32 bit binaries alongside with this setup than only 32 bits binaries. And why not support WOW64 in the near future.

Sir-Boops:
You _could_ install both it's just a very dirty way of getting it to work.

> And why not support WOW64 in the near future.

It's a problem with XBPS iitself as to compile multilib wine you need both 64 and 32 bit packages but XBPS does not have the ability to cleanly pull multilib packages in.

Nicop06:
I just ran some tests and having both version the way I did it does not work. I was lucky as the way I installed wine (32 bits version then 64 bits version) worked. Although, some files are included in both versions, and I am surprised that XBPS didn't warn me before overwriting the files...

But ignoring this, with both versions of wine installed I am actually able to run 32 bits programs in a 64 bits wine installation. That explains why Steam is running.

I do want to do it cleanly and I am open to suggestions. I just checked how other distributions are doing it and here is what I found:
* Gentoo is using 2 versions of wine as [portage does not support WOW64](https://gitweb.gentoo.org/repo/gentoo.git/tree/app-emulation/wine/wine-2.3.ebuild#n484)
* Arch Linux [is using the --with-wine64](https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/wine#n141)

Which way do you prefer? It seems that the Gentoo way would be easiest right now considering what you said about XBPS and multilib. I could check mainstream distributions too if needed. 

Anyway, I do believe that having wine 64 would benefit to people using wine for some games and some productivity tools as it will give them performance improvement. But it seems that you don't want to have a 64 bits wine running in Void and I can't understand why.

Sir-Boops:
Arch is doing the multilib setup using 32bit libs, So you'll have to try doing it that same way Gentoo is.

And i'm not against having 64-bit wine i just don't want to make it a big hassle for people to use it.

Nicop06:
I understand I am all for it. I'll work on this to make it as hassle free as possible.
Is it really a problem to have 2 wine installed then? In my opinion this is fine considering you'll need multilib anyway to run 32 bits wine applications.
Also, should I close the branch and reopen it when I am done or is it better to leave it opened?

Sir-Boops:
Imo it's fine to have two versions installed as long there is no 
conflicting files.

Just use this PR and force push to the branch.

Nicop06:
OK thank you. I'll try to fix this issue in the coming days.

Of course, this is not my first PR and I know how to work with github.

Vintodrimmer:
Good day,

Were you able to make bot versions independent, so they don't overwrite each other or was there any other solution?

Nicop06:
No sorry I didn't find the time to work on that. I'll update the PR as soon as I made any progress.

skrzyp:
What's the status of that?

I would like to have WINE running in native 64-bit mode too, for some modern applications which don't run in 32-bit anymore.

The last commit seems to be actually not-that-large patch, and the only Travis error is that because the i686 build exceeded its time to build.

bobwya:
@Nicop06 

> Gentoo is using 2 versions of wine as portage does not support WOW64

**Gentoo** supports pure **32-bit Wine**, pure **64-bit Wine** or true **SysWOW64 Wine multilib 32-bit/64-bit** - on 64-bit x86 hardware / OS. In addition it supports mutli-slotted Wine... Allowing multiple versions of Wine to be installed on a single system (all built with different patch-sets and USE flags).

Please don't spread misinformation!

Duncaen:
> Please don't spread misinformation!

Don't ignore dates.

bobwya:
@Duncaen 

I've maintained an Overlay fork of the main **Gentoo** packages for 2-3 years. So I can state categorically that unless one steps back to 2015 - building **SysWOW 64-bit / 32-bit Multi-lib** Wine has been fully supported by **Gentoo**!

**NP-hardass** depreciated the older **app-emulation/wine** package, because this package does not support the new-mutlislotting feature **Gentoo** has introduced (i.e. multiple Wine versions installed in parallel). This has nothing to do with support for true **SysWOW64** support - which has been available since 2015(ish) (when the **Gentoo** **emul-linux** multilib packages were finally officially depreciated).

See: [Gentoo Wiki: Wine: New Packaging](https://wiki.gentoo.org/wiki/Wine:_New_Packaging).

Frankly IHMO **Gentoo** has the best support, for flexible and customisable packaging of Wine, of any Linux distribution.

bobwya:
@Duncaen 

I see my comments as being relevant - because you guys are clearly looking around at how build chains on alternative source-based distributions build Wine.

I'll poke around - but I'm not familiar with **Void** Linux...
There's a guy on the WineHQ forums trying to build SysWOW Wine on **Void**.
So I'm on the end of a Google search about this...
I can state straight off... That building **Syswow Multi-lib Wine** isn't possible without being to refer to the 64-bit build files from the 32-bit build step.
**Gentoo** uses a 2 pass approach. The 32-bit and 64-bit build directories are able to reference the other architecture's build directory, during the build process.

Do I gather correctly that on **Void** - you end up with 2 separate packages?
I.e. a pure 32-bit Wine package and a pure 64-bit Wine package - that don't "talk to each other"?
Which is mainly an issue with your packaging tool-chain... Right?

I'd warn you that it's probably taken the **Gentoo** project around a decade to stabilise good quality multilib package building support.

Nicop06:
@bobwya 
Indeed, I was confused by this comment in the ebuild file:

> Note: using --with-wine64 results in problems with multilib.eclass
> CC/LD hackery. We're using separate tools

Also, it turns out that when I switched to Void in 2017 I was still using app-emulation/wine. I do not know the current state of the wine package but I don't recall any elog / ebuild warning for the switch to virtual/wine.

Back to Void, it seems there is some kind of multilib support but I am unsure how it works. Surprisingly, using this patch does enable me to run either a 32 bits wineprefix or a 64 bits wineprefix with support for 32 bits applications (I guess this is what is called Wow64).

I just need to install both the 64 bits wine built locally using this patch and the 32 bits one from the multilib remote repository. I never had to build both wine versions together. The only problem I have is that both wine share some files such as `/usr/bin/wine` or `/usr/share/wine/*`.

So moving the common files to a new shared package would solve the issue and allow user to run 32 bits applications in 64 bits wineprefix, or still using a 32 bits wineprefix. Unless someone has a better idea I will try to do that.

Nicop06:
The repo moved. This PR needs some work anyway so I will submit a new one once it's ready.

PR #8260, opened 2017-10-18
iio-sensor-proxy: update to 2.3.

circhioz:


Gottox:
already at 2.4. Thanks for your effort anyway!

PR #8266, opened 2017-10-18
xfce4-weather-plugin: update to 0.8.10.

maxice8:


xtraeme:
already updated, sorry.

PR #8269, opened 2017-10-18
Add package erlang19

arete:
Couchdb 1.6 and 2.0 have need of erlang 19.
for couchdb 2.0 is necessary only to compile time.


arete:
Couchdb 1.7.0 support erlang 20.1

PR #8274, opened 2017-10-18
base-system: add rfkill

cr6git:
From void-mklive [#75](https://github.com/voidlinux/void-mklive/issues/75). 
On some laptops (as it is the case with my HP dv6) the Wi-Fi is **blocked** by default and `rfkill` is needed to **unblock** it.


chneukirchen:
I'd like to delay this for a few weeks until next util-linux comes out, which includes rfkill already.

cr6git:
Good news! :)

Gottox:
closing, obsolete with up to come util-linux.

Issue #7572, opened 2017-09-05
Package request: cadence


jarbowski:
Running PulseAudio on top of JACK is a pain and QJackCtl doesn't help much.
From the official jackaudio site: "remove QJackCtl if you use that and install Cadence from the KXStudio repositories. From within Cadence you can enable the ALSA and PulseAudio bridges with the GUI. Done."

http://kxstudio.linuxaudio.org/Applications:Cadence

jack_capture and zita-ajbridge are the only missing (but optional) dependencies in the repo

cr6git:
````
=> cadence-0.8.1_1: running do_build ...
make -C c++/jackmeter
make -C c++/xycontroller
pyrcc4 -py3 resources/resources.qrc -o src/resources_rc.py
make: pyrcc4: Command not found
=> ERROR: cadence-0.8.1_1: failed to run do_build() at line 9.
````

As I understand it: `pyrcc4` must be part of `python3-PyQt4-devel`, but this package is not available in Void repos. 
😕

Johnnynator:
@cr6git It is in python-PyQt4-devel-tools.

cr6git:
OK! I'll tray again...

cr6git:
@jarbowski I wonder what the Void devs are waiting for...

cr6git:
@jarbowski it's done!

PR #8287, opened 2017-10-18
New package: cadence-0.8.1.20170707

cr6git:
resolves [#7572](https://github.com/voidlinux/void-packages/issues/7572)


cr6git:
It sounds like this package cannot be cross compiled.

cr6git:
The build failed on `x86_64-musl` **only**, I don't know why...

Johnnynator:
The probem is that the sys/prctl.h header in glibc just inlcudes the linux/prctl.h header while on musl sys/prctl.h redefines everything. Therefore the linux/prctl.h doesn't have to be included manually (at least I think so).

```
--- c++/jackbridge/CarlaUtils.hpp	2013-07-16 06:36:57.000000000 +0200
+++ c++/jackbridge/CarlaUtils.hpp	2017-10-18 15:56:12.684964988 +0200
@@ -36,7 +36,6 @@
 # include <kernel/OS.h>
 #elif defined(CARLA_OS_LINUX)
 # include <sys/prctl.h>
-# include <linux/prctl.h>
 #endif
```

cr6git:
Thank you @Johnnynator 
I added your patch for `x86_64-musl`.

Johnnynator:
#8305 ladish cross compilation is fixed in this PR, but then you run into the problem that python-gobject can't be cross compiled. (And you don't have to flag cadence as nocross because you don't know if it cross compiles or not)

cr6git:
ok I will change my template.

cr6git:
The build is ok now! 😃

cr6git:
Changed the source to github.

jarbowski:
@cr6git 

thank you very much but I think it's missing python dbus to be functional:

Traceback (most recent call last):
  File "/usr/share/cadence/src/cadence_session_start.py", line 5, in <module>
    import dbus, sys
ModuleNotFoundError: No module named 'dbus'


jackdbus doesn't launch

cr6git:
@jarbowski I probably mixed up `python-dbus` (already pulled by `jack`) with `python3-dbus` ? 

@pullmoll fixed the issue!

PR #8320, opened 2017-10-18
psutils: fix typo in patch

cr6git:


cr6git:
Why not merged ? This is just a typo...

PR #8325, opened 2017-10-19
New package: vrms-void-0.1.1

abenson:


Vaelatern:
Any reason you needed to develop your own, instead of just patches on top of Debian's code?

abenson:
Debian's vrms just seems to check what repo it was installed from or if the name matches a specific package. Mine is checking the project's stated license for GPL-compatibility. 

Mine's also highlighted some issues with template quality regarding the license field, such as misspellings ("Propietary"), incorrect or ambiguous ("opensource" when it should be "MIT"), and even missing license fields.

cr6git:
With all due respect for RMS, I do not feel like this program may be useful in Void linux. 
About the licenses, for example we have `xlint` to test the validity of our templates.

abenson:
> With all due respect for RMS, I do not feel like this program may be useful in Void linux.

That's fair. It was written in response to a request/comment from someone on #voidlinux.  I think my approach might be off, as well: I was looking at what was GPL compatible, and there is a ton of software that isn't GPL compatible (MPL, for example). 

If people don't feel this has much value added, that's fine, I was just offering it up.

>  About the licenses, for example we have xlint to test the validity of our templates.

I don't think that's as relevant as an argument, as `xlint` doesn't check for correctness or accuracy of the reported license, just that the license is reported.

cr6git:
I'm not expert in licenses, maybe you are right. 😉

abenson:
After some thought and discussion, this might not really be something to be included directly in void.

PR #8358, opened 2017-10-19
xiate: update to 17.10.

maxice8:


diogoleal:
this is not necessary. 
#8326

PR #8399, opened 2017-10-20
yoshimi: fix build

newbluemoon:
In the previous PR the check for `random_r` wasn’t put through to the source code and the build segfaults because of a function name collision. This is now fixed. Sorry! (I will never fully trust Travis again, even when all checks have passed ;-))


newbluemoon:
Hm, worked locally. Working on it.

newbluemoon:
While the patch for musl works, the GUI won’t come up on arm, command line mode seems to work. But I believe this is not related to the patch.

newbluemoon:
The GUI problem on musl systems is now fixed. All patches are already merged upstream and can be removed with the next relase. Sorry again for any inconvenience.

PR #8414, opened 2017-10-20
New package: yq-2.3.3.

maxice8:


asergi:
`yq` is not a library, so there is no need to have both variants. Please change `pkgname` to `yq`, `build_style` to `python3-module`, etc.

maxice8:
Done, thanks for the feedback.

Issue #7741, opened 2017-09-19
qt5 on musl is broken


Johnnynator:
If I start any program that is compiled against qt5 gui (not qml) then it will hog one cpu core and do nothing.

https://forum.voidlinux.eu/t/chromium-qt-issues/3460

mvf:
Works over here on x86_64-musl (Qt Creator, mozc_tool, bunch of PyQt5 apps)

Johnnynator:
Mkay... Can you give me a list of your manually installed package and running services. 
Do you have changed anything qt5 configs/themes?

mvf:
I run `i3` under `nodm` with `widgetStyle=Windows`. Pure Qt without `kdelibs` (last time I checked those sneak into all Qt apps). The only relevant self-built packages are `ibus-mozc`+deps and `freetype`. Running services shouldn't matter. Does Qt Creator work for you? It seems to ignore the `widgetStyle` setting, maybe this could help to exclude those. The first thing I'd try is starting a QtGui app under gdb and see which library it loops in.

Johnnynator:
@mvf No only qml applications work. And yes running services shouldn't matter since I have the default ones minus NetworkManager. I created a live Iso and qt5 works there so at least I have one system that works for comparison for myself.

I don't have any knowledge how to use gdb, 
I only made had run strace ones

mvf:
Nevermind, after a reboot I now see this too. In this Qt Creator callstack `QXcbBackingStore::endPaint` does not return:
```
#0  0x00007ffff6cda669 in QImage::rgbSwapped_inplace (this=this@entry=0x7fffffffd6c0) at image/qimage.cpp:3320
#1  0x00007fffef3d0558 in QImage::rgbSwapped() && (this=0x7fffffffd6c0) at ../../../../include/QtGui/../../src/gui/image/qimage.h:284
#2  QXcbBackingStore::endPaint (this=0x555556db3420) at qxcbbackingstore.cpp:565
#3  0x00007ffff7240dc8 in QWidgetBackingStore::endPaint (beginPaintInfo=<synthetic pointer>, backingStore=<optimized out>, cleaned=..., this=0x555555f98f00) at kernel/qwidgetbackingstore.cpp:364
#4  QWidgetBackingStore::doSync (this=this@entry=0x555555f98f00) at kernel/qwidgetbackingstore.cpp:1392
[...]

PR #8444, opened 2017-10-22
keepassxc: update to 2.2.2.

maxice8:


Vintodrimmer:
Can you check the package on musl, please?

It just opens an empty frame and loads one of my cores 100%. This is both for 2.2.1 and 2.2.2.

lemmi:
@Vintodrimmer could you report the issue upstream? You need to provide a little more context aswell. Does is happen with new databases? Do you use a Yubikey? Have you tried downgrading and does that fix the issue?

maxice8:
Problably related to https://github.com/voidlinux/void-packages/issues/7741

Vintodrimmer:
@lemmi It happens any time I start the programm. Either by itself or by opening a database. I don't use anything extra, just a plain database, created with keepassxc 2.2.0.

I have not tried downgrading yet and I'm not sure the problem is with the package itself, but I'll write report there as well.

@maxice8 Looks alot like my issue, thanks.

PR #8441, opened 2017-10-22
lxqt: update to 0.12.0.

maxice8:
- [x] lixqtxdg-3.1.0
- [x] lxqt-build-tools-0.4.0
- [x] liblxqt-0.12.0
- [x] libfm-qt-0.12.0
- [x] libsysstat-0.4.0
- [x] lxqt-l10n-0.12.0
- [x] lxqt-globakeys-0.12.0
- [x] lxqt-qtplugin-0.12.0
- [x] lxqt-runner-0.12.0
- [x] lxqt-panel-0.12.0
- [x] lxqt-sudo-0.12.0
- [x] lxqt-session-0.12.0
- [x] lxqt-about-0.12.0
- [x] lxqt-config-0.12.0
- [x] lxqt-notificationd-0.12.0
- [x] lxqt-powermanagement-0.12.0
- [x] lxqt-policykit-0.12.0
- [x] lxqt-openssh-askpass-0.12.0
- [x] lxqt-admin-0.12.0
- [x] lximage-qt-0.6.0
- [x] obconf-qt-0.12.0
- [x] pavucontrol-qt-0.3.0
- [x] qtermwidget-0.8.0
- [x] qterminal-0.8.0
- [x] compton-conf-0.3.0
- [x] pcmanfm-qt-0.12.0
- [x] lxq-themes-0.12.0
- [x] lxqt-0.12.0

PR #8449, opened 2017-10-22
lxqt-build-tools: update to 0.4.0

cr6git:
This update is needed, in order to build `compton-conf` [version 0.3.0](https://github.com/lxde/compton-conf/releases).


maxice8:
https://github.com/voidlinux/void-packages/pull/8441

cr6git:
Fine, I'll do that then.

maxice8:
There is no need since lxqt-builds-tools and obconf-qt are already being updated together with the rest of the lxqt suite on #8441 , just some cross-compilation issues need to be ironed out

cr6git:
Ah yes, exactly!

PR #8454, opened 2017-10-22
Adobe flash plugin

z1lt0id:


cr6git:
[#8315](https://github.com/voidlinux/void-packages/pull/8315).

asergi:
`pycp` -> #8227

PR #8457, opened 2017-10-22
obconf-qt: update to 0.12.0

cr6git:


maxice8:
https://github.com/voidlinux/void-packages/pull/8441

cr6git:
I hadn't noticed, thanks.

PR #8458, opened 2017-10-22
notepadqq: update to 1.2.0.

Piraty:
I added qt5-svg as runtime dependency, as this always needs to be set manually


cr6git:
@Piraty I updated `notepadqq` 7 days ago in [#8189](https://github.com/voidlinux/void-packages/pull/8189).

Piraty:
Sorry, didn't see it. You should add qt5-svg to `depends`.

cr6git:
No, `qt5-svg` is not a runtime dependency of `notepadqq`.

Issue #8451, opened 2017-10-22
Package Request: tunesviewer


icf20:
project webpage http://tunesviewer.sourceforge.net/

this is a project that allow linux users access content from ITunes-Universirty while the RSS subscriptions can be handled by gpodder

cr6git:
⚠️ At the moment, in [files](https://sourceforge.net/projects/tunesviewer/files/) there is only BETA version.

icf20:
@cr6git look like the "correct" version for example ubuntu has 1.4.99 https://packages.ubuntu.com/zesty/tunesviewer

cr6git:
@icf20 As you can see in my PR, there was a problem with the license. It should be `GPL-2` but @Vaelatern  did not agree.

PR #8459, opened 2017-10-22
[NOMERGE] New package: tunesviewer

cr6git:
[#8451](https://github.com/voidlinux/void-packages/issues/8451).


srcpkgs/tunesviewer/template
```diff
@@ -0,0 +1,30 @@
+# Template file for 'tunesviewer'
+pkgname=tunesviewer
+version=1.5.4
+revision=1
+only_for_archs="i686 x86_64"
```
Vaelatern: why?

Vaelatern:
This is not a source build for an AGPL project?

cr6git:
@Vaelatern would you agree with `license="AGPL-2"` ?

cr6git:
No response? I close.

Vaelatern:
I'm awful about checking my git notifications. With that license, I can accept `GPL-2.0`

cr6git:
Unfortunately, the latest release 2.1 does not build. 😕

Issue #8462, opened 2017-10-22
 Package Request: Dukto


4goodapp:
[Dukto](http://www.msec.it/blog/?page_id=11)

cr6git:
@4goodapp As you can see in my PR, we couldn't find the tarballs source... 
Do you have an idea where they are ??

4goodapp:
@cr6git Is it this? https://sourceforge.net/projects/dukto/files/latest/download

cr6git:
@4goodapp did you try to unzip this file ? 
````
$ unzip DuktoR6-Portable.zip
Archive:  DuktoR6-Portable.zip
   creating: DuktoR6/
  inflating: DuktoR6/dukto.exe       
  inflating: DuktoR6/libgcc_s_dw2-1.dll  
  inflating: DuktoR6/libstdc++-6.dll  
  inflating: DuktoR6/mingwm10.dll    
  inflating: DuktoR6/QtCore4.dll     
  inflating: DuktoR6/QtDeclarative4.dll  
  inflating: DuktoR6/QtGui4.dll      
  inflating: DuktoR6/QtNetwork4.dll  
  inflating: DuktoR6/QtScript4.dll   
  inflating: DuktoR6/QtSql4.dll      
  inflating: DuktoR6/QtXmlPatterns4.dll
````
You'll get `dukto.exe` , this is not the source code.

PR #8463, opened 2017-10-22
New package: dukto

cr6git:
[#8462](https://github.com/voidlinux/void-packages/issues/8462).


srcpkgs/dukto/template
```diff
@@ -0,0 +1,25 @@
...
+short_desc="Easy file transfer tool designed for LAN use"
+homepage="http://www.msec.it/blog/?page_id=11"
+license="GPL-2"
+
+if [ "$CROSS_BUILD" ]; then
+	hostmakedepends+=" qt-devel"
+fi
+
+do_fetch() {
+	svn checkout svn://svn.code.sf.net/p/dukto/code/trunk@111 ${pkgname}-${version}
```
Vaelatern: uhhhhhhhh

Vaelatern: tarball plz?

cr6git: tarball not available. 
on the [download page](https://software.opensuse.org/download.html?project=home:colomboem&package=dukto) there are only .deb or .rpm packages. 
should I use the .deb package, in your opinion ?

cr6git: @Vaelatern I modified the template to use .deb packages instead of svn. I added an option to use ubuntu-16.04 packages instead of debian-8.0.

Vaelatern:
https://sourceforge.net/projects/dukto/ The source files are available...
If you need a download of type zip, just add `unzip` to `hostmakedepends`

cr6git:
Aaah yes!! The link to sourceforge was not in "Download", but I can see it now! 
Thank you.

cr6git:
I close.

PR #8476, opened 2017-10-23
harfbuzz: update to 1.6.1.

maxice8:


Duncaen:
This update results in segfaults on random pages with chromium #8483.

PR #8480, opened 2017-10-23
New package: wrk-4.0.2

nilium:


chneukirchen:
Does this do anything but crash with `PANIC: unprotected error in call to Lua API (attempt to index a nil value)`?

nilium:
I’ve not encountered that, so yes. I’d appreciate it if you’d provide more information if you’ve encountered a problem with the package.

chneukirchen:
```
% wrk http://vuxu.org/                                 
PANIC: unprotected error in call to Lua API (attempt to index a nil value)
```

According to strace it looks for wrk.lua and wrk.so at a few places, but can't find it.

chneukirchen:
on x86_64 glibc.

nilium:
Thanks. I’ll take a look later today — weird that it’d be looking for a wrk.so though. Might be an odd LuaJIT case based on your strace, so it’s possible the bytecode object isn’t being linked correctly (or at all). Entirely guessing until I have more time, though.

nilium:
@chneukirchen I may have a fix for this in #10108 -- if you have some free time, would you be able to build that and see if it's working for you? (I'd like to avoid another "it works for me" blunder.)

chneukirchen:
Now it works fine, thanks.

PR #8481, opened 2017-10-23
lilypond: update to 2.19.80

newbluemoon:
The build for x86_64-musl fails because there’s no texlive on musl systems. Is there something like `nocross` for musl? Like `nomusl=yes` to prevent such errors?


cr6git:
Why not `only_for_archs="i686 x86_64 armv7l aarch64"`

newbluemoon:
@cr6git see `xbps-src --help`, option `-a`: there are a lot more targets listed. Would I have to list all of the non-musl ones then? Something just doesn’t feel quite right. ;-)

newbluemoon:
Just found #3109. This would definitely help.

srcpkgs/lilypond/template
```diff
@@ -1,6 +1,6 @@
 # Template file for 'lilypond'
 pkgname="lilypond"
-version="2.19.65"
+version="2.19.80"
 revision=1
 _tlversion=2017
```
yopito: `_tlversion` is unused, remove it ?

newbluemoon:
I was being stupid. All the time it was before my very eyes: just set it `broken` when `$XBPS_TARGET_MACHINE` is *-musl!

lemmi:
I'm confused. You  set `nocross` and add decisions whether there is a cross build in [line 27](https://github.com/newbluemoon/void-packages/blob/2a784250a191af5833af90c8537680864229f5cb/srcpkgs/lilypond/template#L27).
Could you remove the unnecessary part? If you need it again it should be easy to rewrite/pull out of git again.

newbluemoon:
@lemmi Sorry, it’s just that I’m still trying to find out why it fails to cross compile, so I kept those lines just for convenience ;-) Removed them now.

PR #8491, opened 2017-10-24
mate-themes: update to 3.22.14.

maxice8:


Vaelatern:
You, erm, got a tarball stuck in that one.
It's gone now https://git.io/vdpjg

maxice8:
my bad, should really get a separate environment to test packages besides the dir my current shell is in.

Anachron:
Just setup a container and do it all in there.
Isn't too hard. Ping me if you need help or ask leah2/Duncean.

maxice8:
A Container would be a massive overkill, the problem is file staying around ~/src/void-packages after i finish testing stuff like (curl and yaml, which created a tar file and a yaml file i downloaded to test the parser).

the simplest solution is use [tempsh](https://p.iotek.org/u25)

PR #8493, opened 2017-10-24
nginx: update to 1.13.6.

maxice8:


Duncaen:
We use the stable version, this is mainline.

PR #8531, opened 2017-10-25
New package: pywal-0.7.1

xYuusha:


maxice8:
Cross compilation seems to be failing due to no `python3-setuptools` on `hostmakedepends`

pretty sure you can replace 

```
hostmakedepends="python3"
makedepends="python3-devel python3-setuptools"
```

with

```
```

xYuusha:
On Wed, Oct 25, 2017 at 02:13:43AM +0000, maxice8 wrote:
> Cross compilation seems to be failing due to no `python3-setuptools` on `hostmakedepends`
> 
> pretty sure you can replace 
> 
> ```
> hostmakedepends="python3"
> makedepends="python3-devel python3-setuptools"
> ```
> 
> with
> 
> ```
> hostmakedepends="python3-setuptools"
> ```

Thanks for the input, completely missed that. I really hate python when packaging.

maxice8:
The `hostmakedepends` and `makedepends` are very easy most of the time, the only annoying thing is `depends` since you have to run the python binary or import the module just for it to fail so you can see what it actually depends on.

PR #8532, opened 2017-10-25
gcompat: removed unneeded pre_build function

djames1:
pre_build is no longer needed as they accepted my pull request specifying to use the CC variable rather than hard-coded gcc


Duncaen:
Its not necessary to increase the revision as it doesn't change the resulting binaries.

djames1:
@Duncaen I have removed the revision bump

Vaelatern:
Did they accept this PR for the version we have, or a newer version?

djames1:
@Vaelatern it is for the version 0.1.0 that we currently have packaged. The pre_build was needed when I originally wrote the template for version 0.0.1.

PR #8538, opened 2017-10-25
sqlite: update to 3.21.0.

maxice8:


xtraeme:
Updated via https://github.com/voidlinux/void-packages/commit/5c22f99ea1254776204f1aa386fd9594f847547e

PR #8545, opened 2017-10-25
cegui: unbreak arm + aarch64

newbluemoon:
freeimage and irrlicht use an internal libpng which isn’t configured at build time. ARM Neon is enabled per default, but in the source tree of libpng the subdirectory ‘arm’ with platform specific code is missing. This leads to an ‘undefined reference’ error when linking to freeimage or irrlicht on arm/aarch64. The fix disables ARM Neon explicitly (as it is in the system’s libpng) instead of adding the missing ‘arm’ directory (which would work, too, btw.).

Some additions to freeimage’s narrowing patch were also necessary. The adjustments made are according to libraw’s latest release. (I learned to be careful with the `char` data type: it is signed on x86 and unsigned on arm :-))

ois needs to be rebuild to fix an undefined reference error when building cegui.


newbluemoon:
> The job exceeded the maximum time limit for jobs, and has been terminated.

Should I split this PR and add cegui if/when the dependencies are accepted?

Gottox:
no, it's ok.

PR #8549, opened 2017-10-25
New package: ascii-patrol (x86_64)

cr6git:


newbluemoon:
I don’t think it’s GPL-2. There’s no source code and LICENSE.txt states:

> The Software MAY NOT be redistributed over public channels.

:-(

cr6git:
You are absolutely correct! I was confused, I did not look closely... 
It's sad. :(

newbluemoon:
Looked pretty cool, though. You always find interesting stuff. :-)

Issue #5502, opened 2017-01-10
Package request: open-vm-tools


jsumners:
https://github.com/vmware/open-vm-tools

`open-vm-tools` provides the guest OS drivers and services for systems running in the VMWare environment. These tools are necessary for proper management of virtual machines within the VSphere environment.

jsumners:
I started investigating how I would contribute this, but only got as far as the following steps:

1. autoreconf -i
2. ./configure —without-x —without-ssl —disable-deploypkg
3. make

Step 2 doesn't finish because the configure script can't find "glib2". And I can't find a "glib2" anything in the repository.

Duncaen:
This is what i have so far, we need to package dnet (libdumbnet1 in debian), `--without-dnet` does not work. 

```
# Template file for 'open-vm-tools'
pkgname=open-vm-tools
version=10.1.0
revision=1
wrksrc="${pkgname}-stable-${version}"
build_wrksrc=${pkgname}
build_style=gnu-configure
configure_args="--without-xmlsecurity --without-dnet --without-kernel-modules"
hostmakedepends="automake gettext libtool pkg-config"
makedepends="fuse-devel glib-devel gtk+-devel gtkmm2-devel libmspack-devel libressl-devel libX11-devel libXext-devel libXinerama-devel libXi-devel libXrender-devel libX
randr-devel libXtst-devel libxerces-c-devel pam-devel procps-ng-devel"
depends=""
short_desc="Set of services and modules that enable several features in VMware"
maintainer="Duncaen <duncaen@voidlinux.eu>"
license="GPL-2, LGPL-2.1, BSD"
homepage="https://github.com/vmware/open-vm-tools"
distfiles="https://github.com/vmware/open-vm-tools/archive/stable-${version}.tar.gz"
checksum=2b76d8461022774f2c077b032fbe870f86046f20724eb24824cf9b6ccc4b5355

pre_configure() {
        sed -i -e 's|.*common-agent/etc/config/Makefile.*||' configure.ac
        mkdir -p common-agent/etc/config
        autoreconf -fi
}
```

jsumners:
Source for libdnet is https://github.com/dugsong/libdnet . Debian has it as "libdumbnet" because they already had a "libdnet" ([ref](https://sources.debian.net/src/libdumbnet/1.12-7/debian/patches/01rename_library.sh.dpatch/)).

jsumners:
[...]

PR #8558, opened 2017-10-25
New package: open-vm-tools-10.1.15

memchk:
This closes issue #5502, and obsoletes PR #5347. I want to acknowlage @ottbot, this PR is heavily based on his, with an update and tweaks. 

I examined the possiblillity of using nmap's libdnet and for now the observed improvments do not outweigh the difficuitly of packaging it. This could be considered for down the road.

### DIFF
--- a/srcpkgs/libdnet/template
+++ b/srcpkgs/libdnet/template
@@ -1,0 +1,33 @@
+# Template file for 'libdnet'
+pkgname=libdnet
+version=1.12
+revision=1
+hostmakedepends="automake libtool autoconf pkg-config"
+wrksrc="libdnet-libdnet-${version}"
+build_style=gnu-configure
+short_desc="A portable interface to low level networking routines"
+maintainer="Carson Page <pagem.carson@gmail.com>"
+homepage="https://github.com/dugsong/libdnet"
+distfiles="https://github.com/dugsong/libdnet/archive/libdnet-${version}.tar.gz"
+license="BSD"
+checksum=b6360659c93fa2e3cde9e0a1fc9c07bc4111f3448c5de856e095eb98315dd424
+
+pre_configure() {
+	autoreconf -I config --force --install
+}
+
+post_install() {
+	vlicense LICENSE
+}
+
+libdnet-devel_package() {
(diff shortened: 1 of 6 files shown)


srcpkgs/libdnet/template
```diff
@@ -0,0 +1,33 @@
+# Template file for 'libdnet'
+pkgname=libdnet
+version=1.12
+revision=1
+makedepends="automake libtool autoconf pkg-config"
```
maxice8: automake, libtoo, autoconf and pkg-config should be in hostmakedepends to allow cross-compilation.

memchk:
open-vm-tools I believe was never designed to work with ARM. For musl, additional work will need to be done as musl does not have the Sun RPC implementation. Would it be OK to set only_archs="i686 x86_64"?

maxice8:
Alpine has the package and 15 patches bundled with it

memchk:
Ok, I will see what I can do.

maxice8:
The patches are over 9k lines and need to be ported to a new version, i'd honestly just mark only x86_64 i686 and work on fixing musl later.

Vaelatern:
Please do not use `only_for_archs`. Better to mark where it is broken.

memchk:
What is the appropriate flag for excluding certain Arch's? I did not see it
in the manual.

Johnnynator:
@memchk mostly a case statement for marking broken builds get used e.g.
```
    *-musl) broken="Musl is broken because stuff";;
    arm*) broken=yes
esac
```

jsumners:
Awesome! This looks great. Thank you very much @memchk

memchk:
Ok, I removed only_archs and added the suggested case statement. It looks like everything is building just fine at this point.

memchk:
@maxice8 Is there anything further you need to see / I need to do before this merges?

maxice8:
LGTM, now just need to wait for a Member to merge this.

memchk:
Gotcha. One more thing, I have been updating my local copy of void-packages with `git pull --rebase upsteam master`, that is the proper way to do it right?

PR #8560, opened 2017-10-26
arduino: update to 1.8.5.

maxice8:


the-maldridge:
Last I checked this does not actually work.

PR #8572, opened 2017-10-26
go: update to 1.9.2.

ebfe:
Don't merge yet. xbps-uhelper fetch doesn't like the certificate (or the redirects?)


ebfe:
49af20662c2fda146ab9166765540950e6b064e8

PR #8573, opened 2017-10-26
mixxx: remove duplicate dependency

cr6git:
duplicate `libmp4v2-devel` !


Duncaen:
This doesn't change the resulting build/binaries, no need to increase the revision.

cr6git:
Oups! Thank you. I'm not at home at the moment but I will change that.

cr6git:
Problem with `armv7hf` and `aarch64` :(

PR #8576, opened 2017-10-26
guile1.8: fix musl build

newbluemoon:
In some cases (e.g. compiling it natively on a Raspberry Pi) guile1.8 uses `LONG_LONG_MAX` which musl doesn’t know. Replaced it with `LLONG_MAX` (`LONG_LONG_MIN` respectively). Doesn’t occur when cross compiling.


Gottox:
@newbluemoon can you report this upstream too?

newbluemoon:
@Gottox Bug report sent. :-)

PR #8577, opened 2017-10-26
dhcpcd: add missing license

cr6git:
````
$ xlint srcpkgs/dhcpcd/template
srcpkgs/dhcpcd/template:13: license 'BSD', but no use of vlicense
````


maxice8:
very creative

cr6git:
To be honest, I simply copied the Arch [PKGBUILD](https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/dhcpcd) ...

maxice8:
On Thu, Oct 26, 2017 at 12:41:53PM +0000, cr6git wrote:
> To be honest, I simply copied the Arch [PKGBUILD](https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/dhcpcd) ...


Ok then, atleast it works *shrugs*

cr6git:
@maxice8 Have you noticed on the [homepage](https://roy.marples.name/projects/dhcpcd) the upcoming 7.0.0 release? 

In the tarball `dhcpcd-7.0.0-rc2.tar.xz` the LICENSE is present. 

I'm currently using this version, it works very well. 
But I noticed a detail in `/etc/dhcpcd.conf` : 
````
# Most distributions have NTP support.
#option ntp_servers
````
This option is now **disabled** by default. 
I'm happy to see that, because I always disable this option...

maxice8:
On Fri, Oct 27, 2017 at 07:06:55AM -0700, cr6git wrote:
> @maxice8 Have you noticed on the [homepage](https://roy.marples.name/projects/dhcpcd) the upcoming 7.0.0 release? 
> 
> In the tarball `dhcpcd-7.0.0-rc2.tar.xz` the LICENSE is present. 
> 

No i didn't notice, i don't keep looking around package homepage for updates, i look at void-updates and my rss feed which is loaded with github releases.atom files.

PR #8582, opened 2017-10-26
ponysay: make sure we use python3

pbui:
It requires python3, but uses `/usr/bin/env python` in the she-bang,
which is ambiguous.  Setting the environment to `python3` fixes the
current run-time error.


Vaelatern:
Maybe if it uses the `python3-module` build style the shebang hook will catch it?

pbui:
The build_style hook doesn't work since we need to specify the additional paramaters to setup.py (maybe there is a way to do this).  Moreover, simply replaying the she-bang will not work since the final executable is actually a zipped archive (I had tried this earlier).  The `--with-custom-env-python=python3` forces setup.py to correct the she-bang before zipping up the executable.

PR #8595, opened 2017-10-27
New package: tor-messenger

cr6git:
They have it in [AUR](https://aur.archlinux.org/packages/tor-messenger-bin/). 

The AUR package license is `"MIT"` but I can't find where is the license file! 
I put `license="unknown"` , but this is not correct, I know...


cr6git:
drop it in the trash.

PR #8603, opened 2017-10-27
firejail: unbreak aarch64

newbluemoon:
The patch is taken from [upstream](https://github.com/netblue30/firejail/issues/1504) where the issue is already resolved. (I just didn’t check for closed ones at first.)

Or maybe it’s better to just wait for the next relase?


newbluemoon:
New version available.

PR #8612, opened 2017-10-28
physfs: update to 3.0.1.

maxice8:
Untested


Anachron:
Why is physfs getting updated without rev-bumping hedgewars?

PR #8624, opened 2017-10-28
Update netsurf and dependencies

Hoshpak:


Hoshpak:
The cross-compilation failures on Travis seem to be caused by failure to install pkg-config (which is in hostmakedepends). For my local builds, pkg-config gets installed and the packages cross-compile just fine.

PR #8630, opened 2017-10-28
wallpaperd: add missing license, shorten short_desc

cr6git:


cr6git:
Why `Connection timed out` ?? Travis is talking nonsense.

cr6git:
Same story as [#8631](https://github.com/voidlinux/void-packages/pull/8631) : web.archive.org server is too slow. 
I give up.

Issue #8642, opened 2017-10-28
Package Request: Kaffeine


anarchomoh:
Hi 

Could you please package Kaffeine for Void?

Website: https://www.kde.org/applications/multimedia/kaffeine/

https://github.com/KDE/kaffeine

cr6git:
There aren't any [releases](https://github.com/Me-TV/Me-TV/releases) ?

anarchomoh:
@cr6git, Hi - you're right. I got confused with the older Me-Tv which is on Launchpad and is no longer being developed.  I have found a better option to watch TV and have updated the package request accordingly.

cr6git:
Done.

anarchomoh:
Thanks Heaps!!!

cr6git:
You can close.

PR #8649, opened 2017-10-29
New package: kaffeine

cr6git:
[#8642](https://github.com/voidlinux/void-packages/issues/8642).


cr6git:
the coffee will get cold :/

PR #8746, opened 2017-10-31
dbus: update to 1.12.0

Sir-Boops:
* [Changelog](https://lists.freedesktop.org/archives/dbus/2017-October/017345.html)
* I've done some basic testing on a x86_64 gnome desktop and nothing i can see seems broken.


maxice8:
```
[Unix] The --enable-abstract-sockets and --disable-abstract-sockets
  options no longer exist. See the release notes for 1.11.20.
```

You can safely --enable-abstract-sockets and ac_cv_have_abstract_sockets=yes from the configure_args, see https://github.com/maxice8/void-packages/commit/b3f2c25569f3c60f964354fd22e3723363bac079

Sir-Boops:
* @maxice8 Nice catch! Removed them. 
* Also removed the ${homepage} from distfiles and set both to use https

PR #8748, opened 2017-10-31
New package: xjadeo

cr6git:


maxice8:
Musl is stricter with stuff, O_RDNONLY is missing because there isn't an include to fcntl.h

```
--- src/xjadeo/mqueue.c
+++ src/xjadeo/mqueue.c
@@ -23,6 +23,7 @@
 
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <sys/fcntl.h>
 #include <unistd.h>
 #include <stdarg.h>
 ```

Should fix

cr6git:
You are fast! ok I will try your fix...

Johnnynator:
@maxice8 the correct path would be fcntl.h and not sys/fcntl.h.

cr6git:
Oups...

PR #8749, opened 2017-10-31
codecrypt: update to 1.8.

maxice8:
Wasn't thoroughly tested


cr6git:
At least we have packaged this thing [before](http://pqcrypto.org/) somebody builds a quantum computer.

maxice8:
On Wed, Nov 01, 2017 at 05:44:54PM -0700, cr6git wrote:
> At least we have packaged this thing [before](http://pqcrypto.org/) somebody builds a quantum computer.
> 

Always good to be one step ahead

PR #8774, opened 2017-11-01
New package: ansi-1.0.0.

### LINKED ISSUE #4898: Package request: ansi
ANSI escape codes in pure bash - change text color, position the cursor, much more
https://github.com/fidian/ansi

maxice8:
resolves #4898


avoidr:
I think `noarch=yes` makes sense here.

maxice8:
Yeah, good catch, didn't even remember to set it so. Thanks.

PR #8776, opened 2017-11-01
vidcutter: update to 4.0.5

cr6git:
[#commit](https://github.com/ozmartian/vidcutter/commit/4338684edcdf44f4078f5c56ee3025bc487eaf20)
Fix for 32-bit Linux users.


cr6git:
Hmm... at closer inspection, the commit provides this fix: 
````
allow Python versions below 3.5.3 without typing.Optional to still work
````
As Void linux uses python 3.6.3 , I doubt whether we're concerned... 
Correct me if I am wrong.

PR #8779, opened 2017-11-01
simple-scan: fix cross-compilation.

maxice8:
My bad, didn't notice travis crying.


lemmi:
Needs a revbump

maxice8:
revbumped.

PR #8780, opened 2017-11-01
perl-X11-Protocol-Other: update to 30.

enguerrand:


maxice8:
https://github.com/voidlinux/void-packages/pull/8743

maxice8:
Updated on https://github.com/voidlinux/void-packages/commit/797757b6fd6266101c2b1548df6b1e7bac44892b

PR #8785, opened 2017-11-02
New package: refind-0.11.2

julio641742:


Vaelatern:
This package does a lot. The first things I'd like is to have the INSTALL.msg at least shrunk if not gone completely (it's shown on all upgrades!), the second is I'd like that image file to, at most, be put in void-artwork instead of included in this package, here.

I have not reviewed the rest of the package yet.

julio641742:
I removed the extra stuff, please check the package again 👍

Duncaen:
`only_for_archs` does not contain `x86_64-musl`.

Issue #6460, opened 2017-05-20
Package Request: chkrootkit


ZweiSpeedruns:
http://www.chkrootkit.org/

Utility to check locally for signs of a rootkit

cr6git:
It's not easy... seems like chkrootkit needs to be patched, see this [discussion](https://forum.voidlinux.eu/t/chkrootkit-template/2610)...

ZweiSpeedruns:
Huh, that's interesting. Looking at the discussion and doing my own research, the error would probably be fixed by adding `#include <fcntl.h>` to the file with the main function. `#include <sys/types.h>` and 
`#include <sys/stat.h>` may also be good for portability. I think these should be patched in at the `__linux__` ifdef in chklastlog.c

zqqw:
There is more than one way to fix a problem, but I think the underlying trouble is in the headers wherever it is failing to build - MUSL, 64 bit GLIBC? It builds on 32 bit GLIBC so it is getting the correct defines from somewhere.
e.g. from /usr/share/include/unistd.h:
/* NOTE: These declarations also appear in <fcntl.h>; be sure to keep both
   files consistent.  Some systems have them there and some here, and some
   software depends on the macros being defined without including both.  */
So it would make more sense (to me) to patch the headers that are missing something so all software builds, not patch this particular code and then have to repeat the exercise.

cr6git:
@zqqw It's really sweet of you to try to help us :wink:

zqqw:
Well it's better than shouting answers at a quiz show on TV. I do try, how much I succeed is something else!
I emailed the bug about the hidden file scan failing on BTRFS to the maintainer, who is going to look into it by the way. I wouldn't know what to do about that except ask the BTRFS devs.
Perhaps the need to port things to MUSL is a security feature so MUSL shouldn't be 'fixed'.
It is getting fcntl.h on my system through sys/file.h, I couldn't check on that before as the forum was down and I couldn't remember your exact error message. Including fcntl.h would thus only explicitly add what is already implicitly added so wouldn't matter for GLIBC users.

avoidr:
I just looked at the code. Weirdly, I get a different error than you. But the code has more problems than that. It would have to be "updated" a little bit. Some things changed since '97. Looking at it, I wouldn't run it as root.

It's simply missing headers, and using weird ifdefs. Like checking __linux__ for standard library headers… makes no sense.

And what is this?
```
#if __FreeBSD__ > 9
int main () { return 0; }
[...]

PR #8796, opened 2017-11-02
New package: chkrootkit

cr6git:
[#6460](https://github.com/voidlinux/void-packages/issues/6460).


cr6git:
musl cannot bear it.

maxice8:
Another case of lack of `fcntl.h` and now PATH_MAX which is from `limits.h`

First patch, fixes missing fcntl.h

```
--- chklastlog.c
+++ chklastlog.c
@@ -41,6 +41,7 @@ int main () { return 0; }
 #include <stdlib.h>
 #endif
 #include <sys/stat.h>
+#include <fcntl.h>
 #include <unistd.h>
 #include <string.h>
 #include <signal.h>
--- chkproc.c
+++ chkproc.c
@@ -62,6 +62,7 @@ int main (){ return 0; }
 #include <string.h>
 #include <errno.h>
 #include <sys/types.h>
+#include <fcntl.h>
 #include <dirent.h>
 #include <ctype.h>
 #include <stdlib.h>
--- chkwtmp.c
+++ chkwtmp.c
@@ -25,6 +25,7 @@ int main () { return 0; }
 #include <stdio.h>
 #include <stdlib.h>
 #include <unistd.h>
+#include <fcntl.h>
 #include <string.h>
 #include <utmp.h>
 #include <time.h>
```

second patch, fixes PATH_MAX

```
--- chkdirs.c
+++ chkdirs.c
@@ -33,6 +33,7 @@
 #include <sys/syslimits.h>
 #endif
 
+#include <limits.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <sys/types.h>
```

Tried compiling with `-m ~/src/masterdir-musl` and it worked

maxice8:
Also needs `binutils` on `depends` otherwise it will error out with

```
chkrootkit: can't find 'strings'.
```

cr6git:
Thank you @maxice8 , I will try right now.

PR #8806, opened 2017-11-02
New package: ramroot

cr6git:
Experimental. 
I haven't tested it yet (and I have never used `mkinitcpio` in Void) but that sounds interesting... 

I quote the README : "This package is designed specifically to work with the Arch Linux boot process. I suspect these scripts may work with slight modifications on other distributions as well, however this is currently untested." 
Who wants to give it a try ?


cr6git:
not yet ready anyway.

PR #8817, opened 2017-11-02
New package: cppzmq-4.2.2.

### LINKED ISSUE #6723: Package Request: cppzmq
It's very small, but useful.

https://github.com/zeromq/cppzmq

C++ binding for 0MQ http://www.zeromq.org

maxice8:
resolves #6723


srcpkgs/cppzmq/template
```diff
@@ -0,0 +1,18 @@
+# Template file for 'cppzmq'
+pkgname=cppzmq
+version=4.2.2
+revision=1
+build_style=cmake
+hostmakedepends="pkg-config"
+makedepends="zeromq-devel"
+depends="zeromq"
```
Vaelatern: That really shouldn't be necessary.

PR #8818, opened 2017-11-02
brisk-menu: update to 0.5.0.

maxice8:
It uses meson now


cr6git:
I think we need to add the license: 
````
license="GPL-2, CC-BY-SA-4.0"

post_install() {
	vlicense LICENSE.CC-BY-SA-4.0
}
````

PR #8846, opened 2017-11-03
New package: gifski

cr6git:
Same issue as [#8835](https://github.com/voidlinux/void-packages/pull/8835) with cross compilation.


cr6git:
grumble... 
My local `xlint` is ok with `AGPL-3` , but (oddly) Travis found an error: 
````
$ common/travis/xlint.sh
srcpkgs/gifski/template:10: license 'AGPL', but no use of vlicense
````

Vaelatern:
Travis's xlint is newer.

PR #8851, opened 2017-11-03
Seafile and al: update to 6.1.3

yopito:
partial duplicate with #8849 : wait its merge first then reconsider these commits.


yopito:
this PR is fine now, ready for merge.

PR #8867, opened 2017-11-04
New package: tinc-pre-1.1pre15

faulesocke:
This is the "prerelease" version.


faulesocke:
Looks like it does not build with musl. The missing header file (linux/if_tun.h) is from the package kernel-libc-headers. Is there a musl equivalent to this package?

faulesocke:
Okay not sure why it still fails. The file should be there...

maxice8:
Cross-compilation is annoying, try adding kernel-libc-headers to hostmakedepends sometimes the build system only detects if it's installed in the host.

Gottox:
We try to avoid unstable releases whenever possible. Please wait for stable and update tinc.

faulesocke:
It's not an unstable release, it's just not marked as being stable. See the stable release as some LTS thing and the prerelease as the one with the features.

Gottox:
Can you point me to documentation that explains the release behavior?

faulesocke:
I think the release behaviour is probably best explained here:

https://tinc-vpn.org/goals/

So 1.0 is a "stable" release (in the debian way of thinking, i.e. they will mostly fix bugs on this version) where 1.1 is the active branch and 2.0 is the future branch.

PR #8900, opened 2017-11-05
xorg-server: enable xfree86-utils

cr6git:
This flag is enabled in Debian and Arch, this is the default setting. 
````
--enable-xfree86-utils  Build xfree86 DDX utilities (default: enabled)
````


Gottox:
Do we need those?

cr6git:
For instance, on the forum someone complained that `gtf` was missing, to solve his video issues.

Gottox:
for reference: https://forum.voidlinux.eu/t/local-install-several-video-issues/3752

PR #8907, opened 2017-11-05
quota: fix musl build

Piraty:
Since it didn't build on musl before.


maxice8:
DIdn't fix cross

OpenEmbedded seems to have a patch http://lists.openembedded.org/pipermail/openembedded-core/2015-April/103378.html

also no clue sys/cdefs.h is added at all since it's an internal glibc header that should be [avoided](https://wiki.musl-libc.org/faq.html#Q:.When.compiling.something.against.musl..I.get.error.messages.about..code.sys.cdefs.h..code.).

Piraty:
i applied the type patch which is more sane indeed.
cdefs.h fixed the __P issue, but that is now patched too.

thanks.

PR #8914, opened 2017-11-06
nilfs-utils: update to 2.2.7.

maxice8:
Untested


xtraeme:
Already updated... sorry

PR #8936, opened 2017-11-06
[NOMERGE] [Need License] New package: python-supervise-1.1.1.

maxice8:


Vaelatern:
Can't be merged until upstream adopts a license, unless it's marked `restricted=yes`. I'll just wait on upstream in this case.

Vaelatern:
It's ISC license, but there is no copy of the license included as per the license conditions.

maxice8:
PKG-INFO on the distfile tells us:

`License: ISC`

but there is no license on the distfile tarball

PR #8938, opened 2017-11-06
kompare: update to 17.08.2

Chocimier:


maxice8:
Cross is broken add qt5-devel to hostmakedepends

maxice8:
oops, my bad i misread the error message, replacing qt5-devel with qt5-qmake fix the error message but it ends up hitting another one.

PR #8940, opened 2017-11-06
New package: python-envdir-0.7.

maxice8:


cr6git:
@maxice8 I do not understand why `xlint` says: 
````
$ xlint srcpkgs/python-envdir/template
srcpkgs/python-envdir/template:8: custom variables should use _ prefix: checkdepends="python-pytest"
````

maxice8:
> I do not understand why xlint says

wait new release

https://github.com/chneukirchen/xtools/issues/69

> Not sure but... when I build the package, it looks like this part is not executed:

need to

`echo XBPS_CHECK_PKGS=yes >> etc/conf`

otherwise

`=> libuev-2.1.0_1: skipping check (XBPS_CHECK_PKGS is disabled) ...`

do above, then

`./xbps-src check $pkgname`

if done correctly

`=> libuev-2.1.0_1: running do_check ...`

cr6git:
Thank you for the explanation!

PR #8941, opened 2017-11-06
New package: caudec

cr6git:
Multiprocess audio converter. 
Packages in `depends` are in the list of [optdepends](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=caudec#n14) on the PKGBUILD.


cr6git:
I think it's best to wait a while... until `mac` problem [#8967](https://github.com/voidlinux/void-packages/issues/8967) is fixed.

cr6git:
Anyway, I found a better method to convert my audio files. I close this.

PR #8943, opened 2017-11-06
New package: way-cooler-0.6.2.

maxice8:


srcpkgs/way-cooler/INSTALL.msg
```diff
@@ -0,0 +1,5 @@
+This is a wayland compositor so it needs to be ran as root or if running
```
Piraty: (language)
ran=run
A configuration

maxice8: Fixed, thanks

maxice8:
I saw that sway uses u+s to allow a normal user to run the compositor, should i add here too ?

pullmoll:
> I saw that sway uses u+s to allow a normal user to run the compositor, should i add here too ?

IMO using the setuid bit is no good idea while other team members' opinion may differ.
I'll merge it at is for now.

Timidger:
You need the `u+s` in order to properly use the compositor in a TTY. Without it, it's only usable in X or nested in another Wayland compositor, which isn't very useful.

Also there's a new version (0.7.0) out.

maxice8:
On Wed, Dec 27, 2017 at 12:38:38PM -0800, Preston Carpenter wrote:
> You need the `u+s` in order to properly use the compositor in a TTY. Without it, it's only usable in X or nested in another Wayland compositor, which isn't very useful.
> 

INSTALL.msg tells about that.

> Also there's a new version (0.7.0) out.
>

Yeah i'm working on 0.7.0 but xcb-0.8.1 is failing.

see: https://travis-ci.org/maxice8/void-packages/builds/322054526

Timidger:
> INSTALL.msg tells about that.

Is there a reason it's not part of the build process? Is Void generally against setuid binaries? Seems annoying, but hey you can package this however you want. Just pointing out I get a lot of bug reports due to this not being enabled on void (see [this thread](https://github.com/way-cooler/way-cooler/issues/351) and the many informal bug reports on my gitter page).

> Yeah i'm working on 0.7.0 but xcb-0.8.1 is failing.

Way Cooler requires libxcb now, it should be added as a dependency to this package.

maxice8:
On Wed, Dec 27, 2017 at 09:41:17PM +0000, Preston Carpenter wrote:
> > INSTALL.msg tells about that.
> 
> Is there a reason it's not part of the build process? Is Void generally against setuid binaries? Seems annoying, but hey you can package this however you want. Just pointing out I get a lot of bug reports due to this not being enabled on void (see [this thread](https://github.com/way-cooler/way-cooler/issues/351) and the many informal bug reports on my gitter page).
> 

Very fair, i have made the setuid binary changes, sorry for causing the trouble.

> > Yeah i'm working on 0.7.0 but xcb-0.8.1 is failing.
> 
> Way Cooler requires libxcb now, it should be added as a dependency to this package. 
> 

tried that, turns out that python3 was also required but all is solved, i have made a PR at #10235 for updating to 0.7.0

PR #8963, opened 2017-11-07
New package: pstate-frequency

cr6git:


cr6git:
See my [comment](https://github.com/voidlinux/void-packages/pull/9170#issuecomment-344189633) ...

PR #8966, opened 2017-11-07
Update libtirpc and rpcbind

Hoshpak:


Hoshpak:
Travis tries to build the libtirpc twice for cross-compiled architectures and fails on the second try. Works fine locally.

PR #8972, opened 2017-11-07
New package: kfind-17.08.3

Chocimier:


Chocimier:
Travis error do not look like problem with template. Can I do something to test building again?

PR #8977, opened 2017-11-07
New package: immortal-0.17.0.

maxice8:


srcpkgs/immortal/template
```diff
@@ -0,0 +1,34 @@
...
+	vlicense LICENSE
+}
+
+immortal-example_package() {
+	noarch=yes
+	short_desc+=" - example configuration"
+	pkg_install() {
+		vmove usr/share/${sourcepkg}/example
+	}
+}
```
Vaelatern: Unless this is very large, please don't break this out.

PR #9003, opened 2017-11-08
New package: ZeroNet-0.6.0

inefarius:


srcpkgs/zeronet/files/zeronet/run
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+exec /opt/zeronet/zeronet.py
```
Vaelatern: Must it run as root?

srcpkgs/zeronet/template
```diff
@@ -0,0 +1,22 @@
...
+homepage="https://zeronet.io/"
+distfiles="https://github.com/HelloZeroNet/ZeroNet/archive/v${version}.tar.gz"
+checksum=87f810d9558a2cb92cdbf737d800cc81d27527a82c2e786517829329e0b49661
+
+do_install() {
+	vmkdir /opt/zeronet
+	vcopy plugins /opt/zeronet
+	vcopy src /opt/zeronet
+	vcopy tools /opt/zeronet
+	vcopy zeronet.py /opt/zeronet
```
Vaelatern: Why do you use `/opt`?

srcpkgs/zeronet/template
```diff
@@ -0,0 +1,22 @@
+# Template file for 'zeronet'
+pkgname=zeronet
```
Vaelatern: `pkgname` should be `ZeroNet` since that seems to be what upstream names it.

srcpkgs/zeronet/files/zeronet/run
inefarius: I believe it does. I got errors trying to run it as a regular user.

srcpkgs/zeronet/template
inefarius: Honestly, no idea. Just copying how the Arch AUR package did it. I don't know what the proper convention is for this. Should I chuck it in /usr/lib/zeronet instead?

srcpkgs/zeronet/template
inefarius: Alright, will change it.

srcpkgs/zeronet/template
Vaelatern: Persistent files like this belong under `/usr`

inefarius: What directory should it go in under `lib`? Also, ZeroNet downloads lots of data, since it's based on BitTorrent. I am going to try and put data in `/var/lib`.

PR #9027, opened 2017-11-09
New package: python-unittest-mixins-1.4.

maxice8:
Required for testing on python-coverage


srcpkgs/python-unittest-mixins/template
```diff
@@ -0,0 +1,25 @@
...
+license="Apache-2.0"
+homepage="https://github.com/nedbat/unittest-mixins"
+distfiles="${PYPI_SITE}/u/unittest-mixins/unittest-mixins-${version}.tar.gz"
+checksum=3979159f81e379416b8f10c3db7bbe9aaaefb7ebb8dcf778e6e5de5150c8281b
+noarch=yes
+
+python3-unittest-mixins_package() {
+	noarch=yes
+	pycompile_module="unittest_mixins"
+	depends="python3-six"
```
Gottox: No description in this subpkg.

maxice8: oh my bad, i i thought the subpkg would just copy the short_desc from the sourcepkg

Gottox: They do, but it's an antipattern not to change it 😃

PR #9035, opened 2017-11-09
atom: update to version 1.22

teajay-fr:
This corrects the build for atom and updates to version 1.22


the-eater:
You're missing a curl revision bump :)

PR #9044, opened 2017-11-09
quazip: fix cross compilation

newbluemoon:


Vaelatern:
fine, there was a commented line that should have been removed. But I felt bad about it taking so long to merge an obviously good fix.

newbluemoon:
Sorry, should have seen it.

PR #9049, opened 2017-11-10
postgresql : update 9.6.6

arete:


arete:
I have rebase on last master

PR #9058, opened 2017-11-10
perl-Mail-Message: Add tests

maxice8:
- `perl-Mail-Message` can use `perl-Email-Simple` and `perl-Email-Abstract`
- `perl-Email-Simple` requires `perl-Email-Date-Format`
- `perl-Email-Abstract` requires `perl-Email-Simple` and `perl-Module-Pluggable`
- `perl-Module-Pluggable` can use `perl-App-FatPacker`


pullmoll:
I don't know the policy for perl module names like Email::Simple in short_desc. I think they are considered redundant.
This PR needs to remove perl-Module-Pluggable as it is already there.

PR #9081, opened 2017-11-11
go-ipfs: update to 0.4.12.

maxice8:
Untested


lemmi:
https://github.com/ipfs/go-ipfs/blob/master/docs/plugins.md#other mentions how plugins can be built in. I tried that locally for 4.11 a couple of days ago and it seems to work. Unless they add uncountable plugins, I'd like to ship them directly. Building them isolated seems a lot of work. It might also fix cross-compiliation since the loader isn't needed, but haven't tried that.

PR #9102, opened 2017-11-11
python-Jinja2: Add tests.

maxice8:


maxice8:
Package was updated on https://github.com/voidlinux/void-packages/commit/5c51d3ca23177fd93483a1c63d9b4503fc65905a but tests weren't added, PR adapted

PR #9104, opened 2017-11-11
New package: enchive

kojicomics:
We should make a patch for the Makefile to avoid the use of "nopie=yes", but I don't know how to do that.


ebfe:
Seems to work fine without `nopie=yes`. UNLICENSE needs to be installed (use vlicense).

kojicomics:
I added the function to install the license.
If I run `./xbps-src pkg enchive` without `nopie=yes`, it returns:

```
ERROR: enchive-3.3_1: non-PIE executable found in PIE build: /usr/bin/enchive
ERROR: enchive-3.3_1: failed to run post-install_06-strip-and-debug-pkgs() at line 58.
```

ebfe:
> If I run `./xbps-src pkg enchive` without `nopie=yes`, it returns:
> 
> ```
> ERROR: enchive-3.3_1: non-PIE executable found in PIE build: /usr/bin/enchive
> ERROR: enchive-3.3_1: failed to run post-install_06-strip-and-debug-pkgs() at line 58.
> ```

Hmm, works fine here. Try again after `./xbps-src bootstrap-update`,
maybe you still have the old gcc 6 in the masterdir.

kojicomics:
I tried after bootstrap-update, and even did a zap + reinstall the binary bootstrap, but neither of them seems to solve the issue.
I chroot into the masterdir, and the gcc version is the 6.3.0

ebfe:
Current gcc version is 7.2.0. Using a mirror?

kojicomics:
I built void-packages from scratch and its compiling fine now.

PR #9106, opened 2017-11-11
New package: calc-2.12.6.0

Chocimier:


pullmoll:
This template doesn't work at all. I think the Makefile needs heavy configuration or even patching, as it happens many times with the gnu-makefile style.

Chocimier:
This compiles well on my machine, but only with one concurrent job (-j 1). Is it possible and acceptable to force it?

Johnnynator:
You can set disable_parallel_build=1 for forcing single threaded building

PR #9108, opened 2017-11-11
wxWidgets: fix cross compilation

newbluemoon:
When cross compiling wxWidgets some files/directories carry the cross triplet in their names and there are some references in the wx-config script. The fix boils down to removing all occurrences of the cross triplet.

Furthermore the build is marked as a cross build in wx-config which I changed to no, else the tool tries to append/prepend the triplet to library names. 

And last I made wx-config ignore the `--host` option, because configure scripts use it automatically when cross compiling which breaks the scripts (there is no configuration under that name and an error message is returned).

I added `wx-config` to the list of script-wrappers (in `common/hooks/pre-configure/02-script-wrapper.sh`).

Then I tried to make as many packages using wxWidgets as possible fit for cross compilation. 
- Audacity needed a configure patch because it used AC_CHECK_FILE to find some files.
- Openbabel is broken for ARM (I already contacted the maintainer). 
- Golly builds for ARM, but needs GLX which is currently disabled for that platform in libGL. Maybe set it broken, too?
- Codeblocks and vba-m execute a target built binary so I set them nocross.
- Cmake seems to ignore the wx-config wrapper. I couldn’t find out why.

I tested cross as well as native builds on a Raspberry Pi 3 (armv7l-musl). The only caveat is that some packages need `-DGLIBCXX_USE_CXX11_ABI=1` when building natively.


newbluemoon:
The  problem with Travis (apart from reaching the maximum log size) is that some packages need wxWidgets-devel as hostmakedepends (the `wxrc` utility in particular), but cannot build it in the cross environment.

newbluemoon:
I removed the revbumps for packages where the package itself isn’t changed (where only the cross-built package is added).

And I removed filezilla and poedit because they also need wxWidgets in hostmakedepends (which Travis can’t handle). They will be added separately if this PR is accepted.

newbluemoon:
Audacity has been updated.

cr6git:
But `libsoxr` was already [enabled](https://i.paste.pics/a9f8cb3ecef3cb5865db2d2902e7b1b5.png).

newbluemoon:
@cr6git The original `libsoxr` uses cmake but the internal `libsoxr` is built via `audacity`’s build system. It builds and runs test binaries which can easyly be disabled with a cmake option when cross compiling (cf. `libsorxr`’s template), but not when building the internal version. So it just seemed to be the easiest/cleanest solution to add the system’s `libsoxr` instead of using another hack (though it might be a simple one, can’t really tell now). :)

cr6git:
Looks good! 
However, I think I've found a kind of conflict between `audacity` and my other new package `guitarix2` [#9963](https://github.com/voidlinux/void-packages/pull/9963), as you can see: 
````
$ audacity
lilv_world_add_plugin(): error: Duplicate plugin <http://guitarix.sourceforge.net/plugins/gx_detune_#_detune_>
lilv_world_add_plugin(): error: ... found in file:///usr/lib/lv2/gx_detune.lv2/
lilv_world_add_plugin(): error: ... and      file:///usr/lib/lv2/gx_detune.lv2/
lilv_world_add_plugin(): error: Duplicate plugin <http://guitarix.sourceforge.net/plugins/gx_barkgraphiceq_#_barkgraphiceq_>
lilv_world_add_plugin(): error: ... found in file:///usr/lib/lv2/gx_barkgraphiceq.lv2/
lilv_world_add_plugin(): error: ... and      file:///usr/lib/lv2/gx_barkgraphiceq.lv2/
lilv_world_add_plugin(): error: Duplicate plugin <http://guitarix.sourceforge.net/plugins/gxechocat#echocat>
lilv_world_add_plugin(): error: ... found in file:///usr/lib/lv2/gxechocat.lv2/
lilv_world_add_plugin(): error: ... and      file:///usr/lib/lv2/gxechocat.lv2/
...
````
😠

cr6git:
Fixed now. 😃

PR #9129, opened 2017-11-12
gsmartcontrol: update to 1.1.3.

yopito:


emacsomancer:
The message this displays:

```
 gsmartcontrol must be run with high privileges. So its desktop menu
  requires a running polkit authentication agent. GNOME or KDE DE
  provide it.  Otherwise, install lxpolkit that is a lightweight
  polkit auth agent.
```
Doesn't seem as helpful as it could be given that `lxpolkit` does not appear in the Void repos, but is rather part of `lxsession` (it seems to me a parenthetical "(as part of the `lxsession` package)" could be added).

yopito:
@emacsomancer : thanks for the pointer, I wasn't aware of this change.

PR #9162, opened 2017-11-13
New package: perl-Clipboard-0.13

lukrop:
Optional dependency for kpcli.


srcpkgs/perl-Clipboard/template
```diff
@@ -0,0 +1,16 @@
...
+version=0.13
+revision=1
+wrksrc="Clipboard-${version}"
+build_style=perl-module
+hostmakedepends="perl"
+makedepends="perl"
+depends="perl"
+short_desc="Clipboard - Copy and paste with any OS."
+maintainer="Lukas Kropatschek <lukrop@hiddenbox.org>"
+homepage="http://search.cpan.org/~king/Clipboard/"
```
maxice8: change ~king to dist

srcpkgs/perl-Clipboard/template
```diff
@@ -0,0 +1,16 @@
...
+license="The Perl 5 License (Artistic 1 & GPL 1)"
```
maxice8: `Artistic, GPL-1` is used when referring to the Perl license

PR #9167, opened 2017-11-14
strace: update to 4.20.

maxice8:


Gottox:
👍  for adding do_check functions!

maxice8:
There isn't much point currently since XBPS_CHECK_PKGS isn't enabled on the buildbots and it doesn't work for cross-compiled packages (understandably)

Would be nice if it could be enabled and get it to run do_check even when cross-compiling if the package is noarch ( perl-, python- ) packages.

but it helps with my testing so i add them.

PR #9174, opened 2017-11-14
firefox-esr: update to 52.5.0

cr6git:
Removed some dependencies from `makedepends` because they are automatically pulled by `gtk+-devel`. 
For the same reason, removed `rust` from `hostmakedepends` because it's automatically pulled by `cargo`.


cr6git:
'The job exceeded the maximum time limit for jobs, and has been terminated.'

PR #9176, opened 2017-11-14
New package: cdetect

cr6git:


Vaelatern:
please use xbump so the commit message is properly formatted.

cr6git:
Oopss... yes, and you already warned me about that... 
ok, I've learned my lesson! 😉

cr6git:
@Vaelatern look at that: 
````
$ cdetect /usr/bin/leafpad
GCC 4.9.3
````
4.9.3 ? it's really old.

cr6git:
I just recompiled `leafpad` with GCC 7.2.0, but now `cdetect` claims: 
````
$ cdetect /usr/bin/leafpad
GCC 6.3.0
````
How is that possible ??

PR #9187, opened 2017-11-14
New package: rambox-0.5.13

R13N:


srcpkgs/rambox/template
```diff
@@ -0,0 +1,21 @@
...
+revision=1
+build_style=fetch
+short_desc="Free and Open Source messaging and emailing app"
+maintainer="Rien Maertens <maertensrien@gmail.com>"
+license="GPL-3"
+homepage="http://rambox.pro/"
+distfiles="https://github.com/saenzramiro/rambox/releases/download/${version}/Rambox-${version}-x64.tar.gz"
+checksum="eb506db7e1b412dbdbf9e0ab69686cc10f60e3c7f2e338db5c477e45eb05d198"
+nopie=yes
+only_for_archs="x86_64"
```
Vaelatern: Really?
This app is free and open source.
So build it from that open source.

srcpkgs/rambox/template
```diff
@@ -0,0 +1,21 @@
+pkgname=rambox
+version=0.5.13
```
Vaelatern: Start the template with a comment.

srcpkgs/rambox/template
```diff
@@ -0,0 +1,21 @@
...
+revision=1
+build_style=fetch
```
Vaelatern: No. It's FOSS. Treat it that way.

Vaelatern:
Is electron. Do we care?

srcpkgs/rambox/template
R13N: Should this be a check added to `xlint`?

R13N:
You are right about building this the FOSS way, but it is electron so building it is a hassle.
I'm going to try building an alternative app, which does the same, but is easier to build.

PR #9196, opened 2017-11-14
xbps-src: update-check looks into version directories

Chocimier:
Usual path for sources is like https://example.org/pub/${pkgname}-${version}.tar. But some projects, especially KDE, uses per-version directories like https://example.org/pub/${version}/${pkgname}-${version}.tar. This change makes possible to find updates on servers with such path layout.
It looks into all higher version directory rather than the highest to not miss anything (ex. kea, phonon-backend-gstreamer).
Here is a [list](https://gist.github.com/Chocimier/6d9dc3be795a24f10f3b28e3aa11a796) of updates of packages having `/${version}/` string in distfiles found with updated script.


yopito:
nice idea. I was thinking about that too ;)

Regarding KDE stuff: could you extend your code to deal with directories that contains only a subpart of `${version}` but which name is still with digits ?
    Example: package `extra-cmake-modules` (and all KDE framework packages) uses "`distfiles=http://download.kde.org/stable/frameworks/${version%.*}/${pkgname}-${version}.tar.xz`"  
But it costs much more curl checks (and code) since have to check several versioned-directories ?

ebfe:
The problem with this is that it makes a slow process even slower. So maybe this could be made optional by setting some flag in the update file.

Chocimier:
@yopito: Well, it already looks for numeric-named directories, as update-check work on processed distfiles value, so it can't easily check whether unprocessed value has `${version%.*}`, `${version}`, `5` or anything else.

@ebfe: More requests are done only if there is numeric directory, so it is likely that newest packages are in another directory. For non-numeric path, slowdown comes only from additional shell code.
Should this option be enabled by default? If not, should we have script that checkes whether flag is enabled for packages having `${version` in distfiles?

ebfe:
> @ebfe: More requests are done only if there is numeric directory, so
> it is likely that newest packages are in another directory.
Ok, i missed that part.

> Should this option be enabled by default? If not, should we have
> script that checkes whether flag is enabled for packages having
> `${version` in distfiles?

I'd just add an option the update file
(https://github.com/voidlinux/void-packages/blob/master/Manual.md#checking-for-new-upstream-releases)
that enable this.

Not sure how many false positives you'd actually get with this enabled
by default.

ebfe:
Regarding slowdown/false positives. Did you try running [void-updates](https://github.com/voidlinux/void-updates) on master vs. this change?

Chocimier:
Oh, I didn't know that script. I can run for timing, but do you have any idea to automatically spot false positives? ;)

ebfe:
> I can run for timing, but do you have any idea to automatically spot false positives? ;)

Automatically no. But you get a rough idea by comparing void-updates output with/without .this.

Chocimier:
I can't get reliable time comparision for all packages now, sorry. Checking packages selected as described in first post takes about 1.5 more time after modification.

Chocimier:
Here is a [comparision](https://gist.github.com/Chocimier/cf2ab56e71bb1200a9ba223a94d1d024#file-updates-diff) of scanning with master and updated script. Lower time may be caused by declining traffic in all network during scanning.

PR #9201, opened 2017-11-14
 refind: support for x86_64-musl arch

julio641742:


Duncaen:
Not necessary to increase the revision, it was never built for `x86_64-musl` therefor `_1` is the latest, this avoids rebuilding and updating `refind  for the architectures that are already there.

julio641742:
Alright, I set the revision number back to 1

PR #9217, opened 2017-11-15
libevhtp: update to 1.2.13.

maxice8:


srcpkgs/libevhtp/template
```diff
@@ -1,17 +1,15 @@
...
 pkgname=libevhtp
-version=1.2.12+1
+version=1.2.13
 revision=1
-_upver="${version/+/-}"
-wrksrc="${pkgname}-${_upver}"
 build_style=cmake
 makedepends="libevent-devel libressl-devel"
 short_desc="Create extremely-fast and secure embedded HTTP servers with ease"
 maintainer="Toyam Cox <Vaelatern@gmail.com>"
```
Vaelatern: please change this to `@voidlinux.eu` while you are here.

maxice8: .

PR #9222, opened 2017-11-15
New Packages: kiwix

Johnnynator:
Cross Compiling seems to fail only on Travis, it compiled fine locally.


pullmoll:
I had some trouble merging this and hope I've done it right.

Johnnynator:
Now we have a couple of double entire in common/shlibs

pullmoll:
Ok, I'm going to clean this up.

PR #9224, opened 2017-11-15
New package: cross-mips-linux-muslhf-0.24 

hipperson0:
Added cross build profile for MIPS32 BE hard float and added support for the mipshf-musl arch to packages which have arch specfic requirements.

Not sure if the changed packages require revbumps or not.


chneukirchen:
What's the target hardware for this architecture?

hipperson0:
It's the same as that of mips toolchain but with support for using the floating point unit.

Such as CPUs based on MIPS 24kf or kc like the MIPS Malta board.

srcpkgs/qt/template
```diff
@@ -85,6 +85,10 @@ do_configure() {
 				_opts+=" -arch mips"
 				_spec="mips-linux-musl-c++"
 				;;
+			mipshf-musl)
+				_opts+=" -arch mips"
+				_spec="mips-linux-musl-c++"
```
pullmoll: Shouldn't that be `mips-linux-mulshf-c++`?

hipperson0: Yes, I missed that. Changing it now

pullmoll:
LGTM. Revbumps are not necessary because AFAICT nothing would change for the existing builders architectures.

hipperson0:
Cool, that's what I thought, just wanted to be sure!

pullmoll:
I updated the musl libc to 1.1.18 and binutils to 2.29.1 before merging. I tested that locally and the binary-bootstrap plus quite a number of packages built [can be found here](http://repo.voidlinux.de/current/).

hipperson0:
Thank you very much pullmoll.

PR #9241, opened 2017-11-16
scrot: add debian patchset

crater2150:
This PR adds the debian patches for scrot (found at http://scrot.sourcearchive.com/downloads/0.8-17/, same source that is currently used for distfiles).
Besides minor bug and typo fixes, this also adds a mode for taking a screenshot of the currently focused window.

I'm not sure, what the usual procedure for this is, but as the patches in the upstream archive are in `-p1` format instead of `-p0`, I converted them manually and added them to the patches dir instead of adding the archive as a distfile and applying them in the template.


Vaelatern:
Why is this patch not upstreamed?

crater2150:
I think, there is no original upstream any more. The homepage entry in the current template already is the debian mirror. Upstream version 0.8 seems to be from 2003, the Debian patches 0.8-1 to 0.8-17 (all included in the PR) range from then to 2015.

cr6git:
The Arch [package](https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/scrot) is using the Debian patches, too.

Vaelatern:
Where does scrot development take place, or is it just on archives now?

crater2150:
The only development I found was at Debian, but it's not very active (about a patch per year, http://metadata.ftp-master.debian.org/changelogs/main/s/scrot/scrot_0.8-18_changelog).

crater2150:
I checked several more distros, all of them use the Debian patches (Arch, Gentoo, Fedora, Ubuntu, Alpine). As the original upstream does not exist anymore and the patches also include new features in addition to bugfixes, this seems to be the closest existing equivalent to an upstream version.

PR #9244, opened 2017-11-16
linux3.18: update to 3.18.82.

cr6git:


cr6git:
`The job exceeded the maximum time limit for jobs, and has been terminated.`

maxice8:
It is a well know fact that compiling Linux is too long a time for the time allotted to us by Travis.

cr6git:
The changes in the kernel are so rapid that 3.18.81 is already out of date :/

PR #9253, opened 2017-11-16
haveged: update to 1.9.2.

maxice8:


cr6git:
Dude, you are unstoppable!!

maxice8:
My battery just stopped me, had to shelf the notebook and use the phone.

PR #9273, opened 2017-11-17
environment: set _FILE_OFFSET_BITS=64 for 32bit

pullmoll:
To avoid mismatch between the environments autoconf cache settings
and the actual size of the glibc types off_t, ino_t and blkcnt_t
set them to 8 (64 bit) and define _FILE_OFFSET_BITS=64 in the
CFLAGS and CXXFLAGS.

Don't do this when building glibc itself or when building cross-*
packages with their embedded glibc.


chneukirchen:
I think a better way to do this is by patching glibc to use 64-bit types by default, but i don't know what's the recommended way to set this (and what we need to recompile afterwards).

pullmoll:
Interesting comments and some hints what we may have to watch out for: https://lwn.net/Articles/590815/
But it'll be really difficult to tell which programs or libraries may have problems with a change in sizeof(off_t). Most probably in case of really old code which hasn't been updated for a very long time.

pullmoll:
When we patch glibc to use 64-bit types by default we can't disable that easily to build gilbc or the cross-*gnu* packages.
I've cross compiled a lot (thousands) of packages for armv5tel and armv{6,7}l with this and the only problem which occured so far is that `libimobiledevice` needs `ac_cv_sys_file_offset_bits=${ac_cv_sys_file_offset_bits=64}` in autoconf_cache/arm-linux to not try to run a cross compiled binary to detect the size.

pullmoll:
Closing - looking for a better solution.

PR #9278, opened 2017-11-17
hedgewars: update to 0.9.23

Anachron:
Tested locally on x86_64 using `Void 4.13.12_1 x86_64 GenuineIntel uptodate rrFFF`.

Couldn't find any issues.


Anachron:
Musl is broken because `fpc` for `musl` is broken. 
Had to change distfiles-mirror since `gna.org` doesn't have it yet.

the-maldridge:
If that were set broken it would not have caused this build to go red.  Please investigate further and either patch fpc or mark it broken on the appropriate architectures.

Anachron:
@the-maldridge I was asking in irc and people said I should not mark it broken because it's the dependency (`fpc`) being broken, not hedgewars.

If you want me to exclude those arch I'll happily do that.

the-maldridge:
Per my previous comment you need to investigate fpc and mark fpc broken if necessary.

Not to be merged until travis marks the build green.

cr6git:
@the-maldridge I wonder why the `fpc` [template](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/fpc/template  is using an `.rpm` source? 

The Free Pascal [download page](https://www.freepascal.org/down/source/sources-netherlands.var) only offers `zip` / `.tar.gz` tarballs. 

The Arch [package](https://www.archlinux.org/packages/community/x86_64/fpc/) does not make use of any `.rpm` source. 
Arch applies a very [big patch](https://git.archlinux.org/svntogit/community.git/tree/trunk/fpc-gdb.patch?h=packages/fpc), however...

Anachron:
I'll get back to this shortly.

Since hedgewars has been [marked broken](https://github.com/voidlinux/void-packages/blob/cd0e387d8eb540358bf6e56e3f6bbdcc8826183b/srcpkgs/hedgewars/template#L21) I I guess I can just merge on that?

lemmi:
@Anachron: Please rebase

asergi:
e9013414d097c0ebce9a0443cf25e69b5ba2ee1a

PR #9286, opened 2017-11-17
New package: ddcutil-0.8.5

cr6git:
[Fixed] The build fails on musl, with the same kind of error as [#9252](https://github.com/voidlinux/void-packages/pull/9252).


cr6git:
I just noticed it: 

````
=> WARNING: ddcutil-0.8.5_1: libddcutil.so.0 not found in common/shlibs!
````

Seems like `xbump` forgot to include `common/shlibs` in my PR... 
:disappointed:

**Comment by pullmoll** [comment]
xbump cares about the srcpkg you specify and nothing else. You can do
```
git add common/shlibs
git commit --amend
git push -f
```

**Comment by newbluemoon** [comment]
With the following two patches the musl package should be built.

```
--- src/util/debug_util.c.orig	2017-11-16 07:13:16.000000000 +0100
+++ src/util/debug_util.c	2017-11-17 21:26:18.177917082 +0100
@@ -26,7 +26,9 @@
  */
 
 /** \cond */
+#ifdef __GLIBC__
 #include <execinfo.h>
+#endif
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -90,15 +92,21 @@
       printf("(%s) Starting.  stack_adjust = %d\n", __func__, stack_adjust);
 
    int j, nptrs;
+#ifdef __GLIBC__
    const int MAX_ADDRS = 100;
    void *buffer[MAX_ADDRS];
-   char **strings;
+#endif
+   char **strings = NULL;
 
+#ifdef __GLIBC__
    nptrs = backtrace(buffer, MAX_ADDRS);
+#endif
    if (debug)
       printf("(%s) backtrace() returned %d addresses\n", __func__, nptrs);
 
+#ifdef __GLIBC__
    strings = backtrace_symbols(buffer, nptrs);
+#endif
    if (strings == NULL) {
       perror("backtrace_symbols unavailable");
    }

```

```
--- src/util/edid.h.orig	2017-11-16 07:13:16.000000000 +0100
+++ src/util/edid.h	2017-11-17 21:29:05.725854175 +0100
@@ -36,6 +36,7 @@
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <sys/types.h>
 /** \endcond */
 
 #include "coredefs.h"

```

cr6git:
@newbluemoon Hmm... Are you sure your patches are correct? 
````
=> ddcutil-0.8.5_1: running do_build ...
make  all-recursive
make[1]: Entering directory '/builddir/ddcutil-0.8.5'
Making all in src
make[2]: Entering directory '/builddir/ddcutil-0.8.5/src'
Making all in util
make[3]: Entering directory '/builddir/ddcutil-0.8.5/src/util'
  CC       data_structures.lo
  CC       debug_util.lo
  CC       edid.lo
  CC       device_id_util.lo
debug_util.c: In function 'show_backtrace':
debug_util.c:102:12: error: implicit declaration of function 'backtrace'; did you mean 'show_backtrace'? [-Werror=implicit-function-declaration]
    nptrs = backtrace(buffer, MAX_ADDRS);
            ^~~~~~~~~
            show_backtrace
debug_util.c:108:14: error: implicit declaration of function 'backtrace_symbols' [-Werror=implicit-function-declaration]
    strings = backtrace_symbols(buffer, nptrs);
              ^~~~~~~~~~~~~~~~~
debug_util.c:108:12: error: assignment makes pointer from integer without a cast [-Werror=int-conversion]
    strings = backtrace_symbols(buffer, nptrs);
            ^
cc1: all warnings being treated as errors
make[3]: *** [Makefile:519: debug_util.lo] Error 1
make[3]: *** Waiting for unfinished jobs....
make[3]: Leaving directory '/builddir/ddcutil-0.8.5/src/util'
make[2]: *** [Makefile:1677: all-recursive] Error 1
make[2]: Leaving directory '/builddir/ddcutil-0.8.5/src'
make[1]: *** [Makefile:619: all-recursive] Error 1
make[1]: Leaving directory '/builddir/ddcutil-0.8.5'
make: *** [Makefile:486: all] Error 2
=> ERROR: ddcutil-0.8.5_1: failed to run do_build() at line 13.
````

newbluemoon:
It worked for me locally. Checking...

newbluemoon:
Seems like the last part of the debug_util.c patch got lost.

cr6git:
Same error again.

newbluemoon:
Strange. I just copied the patch above into a new file and it worked. We had something like this some time ago, do you remember? Maybe some encoding issue?

newbluemoon:
You can try to transfer the modifications manually to the source and then create a diff.

cr6git:
Of course! I will do that.

cr6git:
Interesting, while I was trying to update [widelands](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/widelands/template) to version 19, I found this patch: 
````
$ cat srcpkgs/widelands/patches/fix-execinfo.patch
Disable execinfo.h and backtrace() for musl

--- src/backtrace.cc	2015-06-01 10:36:22.655599851 +0200
+++ src/backtrace.cc	2015-06-01 10:35:04.179600083 +0200
@@ -22,15 +22,17 @@
 #ifndef _WIN32
 #ifndef __APPLE__
 #include <cstdlib>
-
+#ifdef	__GLIBC__
 #include <execinfo.h>
 #endif
 #endif
+#endif
 
 std::string get_backtrace() {
 	std::string result("Backtrace:\n");
 #ifndef _WIN32
 #ifndef __APPLE__
+#ifdef	__GLIBC__
 #define BACKTRACE_STACKSIZE 24
 
 	void * stack[BACKTRACE_STACKSIZE];
@@ -43,5 +45,6 @@
 	free(list);
 #endif
 #endif
+#endif
 	return result;
 }
````

pullmoll:
Yep, musl lacking execinfo.h and backtrace() is a major PITA. Someone ought to write a muls-execinfo package to add for musl builds. Don't look at me, though.

Are you sure you put the patch in `<pkgdir>/patches`. I sometimes accidentially put patches in the `<pkgdir>` and later wonder why they are not applied :)

cr6git:
I just made a new attempt, I did what @newbluemoon advised me: I modified manually the source files and then create a diff, but I got the same error again! 
I don't know what more I can do.

pullmoll:
Did you `./xbps-src clean` and have you seen your patch name being printed after the package was extracted?

cr6git:
Yes, I even did `./xbps-src zap` , then `./xbps-src binary-bootstrap` to get a clean masterdir. 
Yes, I've seen the patch name being printed during post-extract, with no error.

pullmoll:
Then I have no idea what goes wrong, sorry.

newbluemoon:
@cr6git Can you please post your command line?

cr6git:
@pullmoll , @newbluemoon I just figured out the problem! Let me explain... 

On this computer I have two root partitions: 
one x86_64-glibc 
one x86_64-musl 

Since the beginning of our discussion, I was using `xbps-src` in the **glibc** system. 

I've rebooted into the **musl** system to give it a try, and... Tada! the build error no longer appear. 

So, as I understand it: the patch works well with musl, but breaks glibc !? 
It's the opposite of the inital situation :/

newbluemoon:
> So, as I understand it: the patch works well with musl, but breaks glibc !?

The patch just makes lines depending on glibc “invisible“ to musl.

But you can build musl packages on a glibc system when using a musl masterdir. See [here](https://github.com/voidlinux/void-packages/blob/master/README.md#building-packages-natively-for-the-musl-c-library). I have a hunch you didn’t do it this way, hence the error? ;-)

cr6git:
I think you did not understand correctly. 

WITH THE PATCH: 
`./xbps-src pkg ddcutil` in glibc system: it fails. 
`./xbps-src pkg ddcutil` in musl system: it works. 

WITHOUT THE PATCH:
`./xbps-src pkg ddcutil` in glibc system: it works. 
`./xbps-src pkg ddcutil` in musl system: it fails. 

This is absurd.

newbluemoon:
> I think you did not understand correctly.

Ah, you’re right, sorry :)

But then the error would suggest that `execinfo.h` is not included because this is where the prototype for `backtrace()` is defined. This would imply that `#ifdef __GLIBC__` is false on a glibc system?
Can’t test it right now, but will do later.

cr6git:
> The patch just makes lines depending on glibc “invisible“ to musl.

But these lines are still visible to glibc, right ?

maxice8:
> Yep, musl lacking execinfo.h and backtrace() is a major PITA. Someone ought to write a muls-execinfo package to add for musl builds. Don't look at me, though.

@pullmoll does the package from Alpine fits the bill ?

https://pkgs.alpinelinux.org/package/edge/main/x86/libexecinfo

newbluemoon:
I think one has to include `features.h`. Will modify the patch...

cr6git:
@newbluemoon you nailed it ! :)

PR #9297, opened 2017-11-18
libsodium: add do_check.

maxice8:


xtraeme:
Please stop sending this kind of PRs. Definitely a common do_check() is the way.

PR #9306, opened 2017-11-18
common/xbps-src: make docheck use run_step. 

Gottox:
This allows a build_style to define a generic do_check function.


Gottox:
closed by https://github.com/voidlinux/void-packages/commit/2059f871fd7006d538a9b46b5e258f007622e994

PR #9307, opened 2017-11-18
common/environment/setup/sourcepkg.sh: unset {pre,do,post}_check

ebfe:


Gottox:
Note that pre_check and post_check only exist in conjunction with #9306

PR #9319, opened 2017-11-18
New package: yubioath-desktop-4.2.0

ColdPhoenix00:


the-maldridge:
The update files should not be necessary.  Please remove them.

ColdPhoenix00:
@the-maldridge Done.

srcpkgs/python3-pyotherside/template
```diff
@@ -0,0 +1,25 @@
...
+maintainer="ColdPhoenix <coldphoenix@interia.pl>"
+license="ISC"
+homepage="https://thp.io/2011/pyotherside"
+distfiles="https://github.com/thp/pyotherside/archive/${version}.tar.gz"
+checksum=00049d5f42cac448368bc2a521edb8de36bb6d2a624e195b7f1004236758b805
+
+if [ -n "$CROSS_BUILD" ]; then
+	hostmakedepends+=" qt5-host-tools qt5-declarative-devel qt5-svg-devel"
+	PATH="$PATH:${XBPS_CROSS_BASE}/usr/bin"
+	CFLAGS="-I=/usr/include/python3.6m"
```
Vaelatern: which of course breaks on updating our python to 3.7

ColdPhoenix00: `python3-pyotherside` doesn't like being cross compiled, so I had to force python include path. As a workaround, I can suggest disabling cross compiling completely.

ColdPhoenix00: I've set it to nocross if you like that better.

PR #9323, opened 2017-11-19
argyllcms: update to 2.0.0.

maxice8:
COMPLETELY UNTESTED


cr6git:
`srcpkgs/argyllcms/template:11: license 'AGPL', but no use of vlicense`

You should add: 
````
do_install() {
	vlicense License.txt
}
````

PR #9345, opened 2017-11-19
stow: add checkdepends and make_check_target.

maxice8:


pullmoll:
Already there https://github.com/voidlinux/void-packages/blob/master/srcpkgs/stow/template

PR #9356, opened 2017-11-20
coreutils: fix musl checks.

maxice8:
Also took the liberty to simplify the disabling of tests


srcpkgs/coreutils/template
```diff
@@ -11,6 +11,7 @@ homepage="http://www.gnu.org/software/coreutils"
 license="GPL-3"
 distfiles="${GNU_SITE}/coreutils/coreutils-${version}.tar.xz"
 checksum=1117b1a16039ddd84d51a9923948307cfa28c2cea03d1a2438742253df0a0c65
+make_check_args="RUN_EXPENSIVE_TESTS=1"
```
Gottox: I'm not sure if we want expensive tests to run.

maxice8: just to be sure we can crank up to RUN_VERY_EXPENSIVE_TESTS /s

Gottox: how long does a test run take with and without these options?

PR #9361, opened 2017-11-20
xbps-src: make it easier to run the check stage

ebfe:
Since the default is to not run checks, having '-Q' enable running checks seems more useful. This allows to
build with checks enabled via `./xbps-src -Q pkg ...`  in the default config.

The second commit always enables XBPS_CHECK_PKGS if the check target is given on the commandline.


ebfe:
Btw. is there any reason to deviate from the logic of the other variables (that is unset/empty = false everything else true)?

pullmoll:
I think it was because -Q set XBPS_CHECK_PKGS=0 instead of empty or undefined.
IMO we could make xbps-src unset the variable and remove the == 0 or == no checks.
So :+1: for this patch.

ebfe:
1a5b47bcc3c4795e5b2d495d8b3719e2b4b371e4

PR #9364, opened 2017-11-20
bitlbee-steam: fix cross compilation

newbluemoon:
Automake macros weren’t found when cross compiling.


newbluemoon:
This just got me thinking: I didn’t find anyting relevant in `common/`, so would it be a good idea if 

    ACLOCAL_PATH=${XBPS_CROSS_BASE}/usr/share/aclocal

was set automatically when cross compiling and build style is `gnu-configure`?

PR #9366, opened 2017-11-20
rpi-kernel: update to 4.9.65

pbui:


Duncaen:
Do you have any idea why this would break wireless support, the modules are not related at all and `CONFIG_CONNECTOR` is enabled as module by default and loaded by default, changing it to `CONFIG_CONNECTOR=y` allows to enable `CONFIG_PROC_EVENTS` which depends on connector being built in.

pbui:
Hmm.  I'm not sure what is going on.  I tested with the following kernels:

```
4.9.61_1 = Wireless works
4.9.62_1 = Wireless does not work
4.9.65_1 = Wireless sometimes work
4.9.65_2 = Wireless works
```

Attached are logs for each kernel version.  I don't really see anything different between .61 and .62 and .65 other than the wireless not working (the device is seen, but wpa_supplicant never connects).

I'm not sure why 4.9.65_1 (which has the CONFIG changes) works on one RPI 3, but not another.  4.9.65_2 (which does not have the CONFIG changes) always seems to work however.

[log.4.9.61_1.txt](https://github.com/voidlinux/void-packages/files/1507552/log.4.9.61_1.txt)
[log.4.9.62_1.txt](https://github.com/voidlinux/void-packages/files/1507553/log.4.9.62_1.txt)
[log.4.9.65_1.txt](https://github.com/voidlinux/void-packages/files/1507554/log.4.9.65_1.txt)
[log.4.9.65_2.txt](https://github.com/voidlinux/void-packages/files/1507555/log.4.9.65_2.txt)

pbui:
Closing b/c I'm going to update to 4.9.66

PR #9386, opened 2017-11-20
qbittorrent: update to 4.0.0.

maxice8:
Untested on glibc systems, Musl system produces only a blackbox as expected per https://github.com/voidlinux/void-packages/issues/7741

needs someone to test glibc.


pullmoll:
4.0.1 is already out and I'll take care once the current qt5 problems are solved.

PR #9391, opened 2017-11-20
build-style/cmake.sh: add generic do_check function.

maxice8:
https://cmake.org/Wiki/CMake/Testing_With_CTest


common/build-style/cmake.sh
```diff
@@ -47,6 +47,24 @@ do_build() {
 	${make_cmd} ${makejobs} ${make_build_args} ${make_build_target}
 }
 
+do_check() {
+	if [ -z "$make_cmd" ] && [ -z "$make_check_target" ]; then
+		if make -q test 2>/dev/null; then
+			:
+		else
```
Vaelatern: `if ! make -q test 2>/dev/null; then`

maxice8: why ?

Vaelatern: Save the useless 2 lines of confusion

chneukirchen: NACK. This will mingle $? = 1 and ? = 2

chneukirchen: Please revert.

chneukirchen:
As explained, this is broken. `! make` will never set `$? -eq 2`. Please fix.

PR #9392, opened 2017-11-21
xbps-src: reduce python invocations

Vaelatern:
Greatly eases the dxpb import process, and doesn't appear to break
current functionality. More testing would be nice, but the 4
calls to python on every read_pkg really slows things down, makes
dxpb's imports CPU bound instead of IO bound.


common/environment/setup/python.sh
```diff
@@ -8,14 +8,31 @@ __python3="/usr/bin/python3"
...
+	while IFS= read -r line; do
+		case "${state}" in
+		0) py2_ver="$line";;
+		1) py2_lib="$line";;
+		2) py2_sitelib="$line";;
+		3) py2_inc="$line";;
+		esac
+		state=$((state+1));
+	done <<EOF
+$(${__python2} -c 'import sys; print(sys.version[:3]); from distutils.sysconfig import get_python_lib, get_python_inc; print(get_python_lib(0, 1));  print(get_python_lib()); print(get_python_inc())')
```
Gottox: Can be simplified to:

```bash
IFS=$(printf "\t") read py2_ver py2_lib py2_sitelib py2_inc <<< $(${__python2} -c \
	'import sys; from distutils.sysconfig import get_python_lib, get_python_inc; print("\t".join([sys.version[:3], get_python_lib(0, 1), get_python_lib(), get_python_inc()]))
')
```

chneukirchen:
Even better would be to get rid of the python calls at all... can't we just figure out the major.minor version of python? We control the paths...

Gottox:
```
python --version
```

But honestly, I guess this behavior just as fast and more error prone in case we ever change the paths

ebfe:
fwiw, we have  XBPS_GCC_VERSION_* already  https://github.com/voidlinux/void-packages/blob/master/xbps-src#L410

Gottox:
how about removing these vars and replace the with calls in the template?

ebfe:
Note that the variables will only be set to the correct values if called chrooted in the masterdir. Putting these into a IN_CHROOT conditional gets rid of the calls for dxpbs usage and shouldn't break anything since python isn't a bootstrap package anyways.

common/environment/setup/python.sh
Vaelatern: @asergi you have self assigned this, so I shan't implement the above, pending your comments.

pullmoll:
I'd vote for merging this since especially the x86_64-musl build environment recently suffers from noticable slowdowns.

Every pkg check or install step now takes about a second while it used to be milliseconds. I don't exactly know where it comes from. I'd definitely appreciate any kind of speed up.

asergi:
I have hardcoded the values of those variables for now.
That means at the next feature release of Python 3, `py3_ver` needs to be increased manually (just like XBPS_GCC_VERSION_* variables).
If you know a better way to have those info (especially the 'major.minor') available in templates, let me know.
Shipping .pyc files into the .xbps packages and fixing cross detection of python2/3 at xbps-src level (i.e., python_wrapper, build styles) and not in every single template would be the ideal solution instead.

ebfe:
> Every pkg check or install step now takes about a second while it used to be milliseconds. I don't exactly
> know where it comes from. I'd definitely appreciate any kind of speed up.

If you are building templates that have no recent commits, the most likely cause for long delays is XBPS_USE_GIT_REVS=yes

ebfe:
Or rather not using XBPS_USE_BUILD_MTIME=yes

PR #9395, opened 2017-11-21
librsvg: remove test that is know to fail.

maxice8:


Vaelatern:
Link to upstream reference?

PR #9399, opened 2017-11-21
New package: qarma-t1

cr6git:
resolves [#6797](https://github.com/voidlinux/void-packages/issues/6797)
basically, I reused [#7264](https://github.com/voidlinux/void-packages/pull/7264)


cr6git:
hmm... seems like I missed something, and `qt5` exists only for `64-bit` arch, not `i686` / `armv7` ?

newbluemoon:
See #9403.

cr6git:
@newbluemoon it's good now! 🎉

PR #9418, opened 2017-11-21
New package: WoeUSB-3.1.4

cr6git:
forum request.


srcpkgs/woeusb/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'woeusb'
+pkgname=woeusb
```
Gottox: why not call the package ```WoeUSB```?

cr6git: yes, why not? I will change that right now.

cr6git:
Done.

PR #9421, opened 2017-11-21
New package: firefox-esr-i18n-52.5.0

Chocimier:


Gottox:
To clarify: firefox-i18n does not work for ff-esr anymore?

Chocimier:
Not checked, as firefox-i18n-* depend on firefox, which conflicts with firefox-esr.

Gottox:
but firefox-esr provides firefox, so this dependency is satisfied.

Gottox:
Ah, no, it does not. This is an error.

Chocimier:
Anyway, 52 refuses to load and use translation of 56.

asergi:
Merged manually in 0fec68021d9dc19d1a96f989b189146932c25bd4 (removed nonexistent `firefox-esr-i18n-be` and `firefox-esr-i18n-my`).

PR #9432, opened 2017-11-22
opera: update to 49.0.2725.39.

cr6git:


common/shlibs
```diff
@@ -124,7 +124,7 @@ libGLESv2.so.2 libGLES-1.0_1
 libEGL.so rpi-userland-0.0.0.0.20150907_1
 libGLES.so rpi-userland-0.0.0.0.20150907_1
 libGLESv2.so rpi-userland-0.0.0.0.20150907_1
-libGLESv2.so opera-48.0.2685.52_1
```
ebfe: why?

cr6git: You're right, I was probably just overreacting!

PR #9441, opened 2017-11-22
zimg: update to 2.6.3.

maxice8:


lemmi:
Sorry, didn't see that. But there's usually no need to do these bumps for my packages.

PR #9452, opened 2017-11-22
pstate-frequency: update to 3.9.7.

maxice8:


cr6git:
@maxice8 correct me if I am wrong but I think we should add this: 
````
conf_files="
	/etc/pstate-frequency.d/00-auto.plan
	/etc/pstate-frequency.d/01-powersave.plan
	/etc/pstate-frequency.d/02-balanced.plan
	/etc/pstate-frequency.d/03-performance.plan
	/etc/pstate-frequency.d/04-max.plan
	"
````

Gottox:
@cr6git Since a few month we support wildcards in config_files.

cr6git:
@Gottox Thanks! I was not sure.

Issue #9454, opened 2017-11-22
xbps-src: using build_options drops default flags if CFLAGS are set


ebfe:
```
[...]
CFLAGS=foo
do_build() {
        echo $CFLAGS
}
[...]

```

```
$ ./xbps-src build a
[...]
=> a-1.0_1: running do_build ...
-D_FORTIFY_SOURCE=2 -mtune=generic -O2 -pipe foo
```

```
$ ./xbps-src -g build a
[...]
=> a-1.0_1: running do_build ...
-D_FORTIFY_SOURCE=2 -mtune=generic -O2 -pipe foo -g
```

Add a build_option
```
[...]
build_options="bar"
CFLAGS=foo
do_build() {
        echo $CFLAGS
}
[...]

```

```
$ ./xbps-src -g build a
[...]
=> a-1.0_1: running do_build ...
-D_FORTIFY_SOURCE=2 foo
```

ebfe:
Using build_options causes the template to be sourced a second time, so this could be called a bug in the template. However tons of templates use this...

pullmoll:
This is due to the `XBPS_BUILD_OPTIONS_PARSED` hack in `common/xbps-src/shutils/common.sh:set_build_options()`. Perhaps this recursive call can be run in a separate shell and the result echoed and assigned in the call instead of exported to `PKG_BUILD_OPTIONS ` at the end of the function!?

ebfe:
I'm not sure if it is easy, because i think that part is also responsible for evaluation all the `$(vopt_if ...)`.

pullmoll:
Hmm.. if the last line of that function is just
```
echo "$PKG_BUILD_OPTIONS"|tr ' ' '\n'|sort|tr '\n' ' '
```
and the call like
```
export PKG_BUILD_OPTIONS="$(set_build_options)"
```
shouldn't the separate shell inherit everything necessary and the echoed string contain everything that's actually changed, i.e. just the sorted PKG_BUILD_OPTIONS?
I admit I haven't tried...

xtraeme:
A subshell for this case seems to be a good idea...

thanks!

2017-11-22 21:56 GMT+01:00 Jürgen Buchmüller <notifications@github.com>:

ebfe:
Hmm, maybe i miss something, but how does configure_args get the right value if the second evaluation happens in a subshell in this case:

```
configure_args=$(vopt_if foo bar)
build_options="foo"
build_options_default="foo"
```

[...]

PR #9455, opened 2017-11-22
musl: fix debug symbols

### LINKED ISSUE #7734: musl: no debugging symbols?
Where are the debugging symbols for musl itself? It says "not stripped" but doesn't have any.

This really sucks for debugging.

ebfe:
See #9454
Fixes #7734


Gottox:
is this PR superseeded by #9460?

ebfe:
Unless that turns up other problems, yes.

PR #9457, opened 2017-11-22
discord: add libatomic to dependencies

Coding-Doctors:
Prevents this message from appearing because of missing dependencies


Coding-Doctors:
Going to note this also fixes a screenshare crash in the app, which is why this is required in the first place.

PR #9462, opened 2017-11-22
plantuml: update to 1.2017.19

juliendehos:


srcpkgs/plantuml/template
```diff
@@ -1,6 +1,6 @@
 # Template file for 'plantuml'
 pkgname="plantuml"
-version=8059
+version=1.2017.19
```
Gottox: This will break the our version comperator.

PR #9479, opened 2017-11-23
New package: cross-arm-none-eabi 7.2.0

teajay-fr:
This package adds the bare metal arm version of the gcc compiler.
The newlib version of the C library is included.


teajay-fr:
I'll reopen this pull request once I have checked all the platforms.

Issue #5816, opened 2017-02-24
build-style/go.sh: please move part of do_build() to post_extract() for nonstandard builds.


ackalker:
To explain: after regular setup (`GOPATH` etc.), `github.com/docker/runc` needs to be built by running `make` from its source directory, or else the build will Git clone and import the entire Docker ecosystem and then some (and finally fail).
See: https://github.com/docker/docker/blob/v1.13.1/hack/dockerfile/install-binaries.sh#L23-L26

I'm referring specifically to these lines: https://github.com/voidlinux/void-packages/blob/1c977b0ba08bb50b6040d0f5af51cda07c4b4b33/common/build-style/go.sh#L6-L12

I think these lines technically don't belong to the `do_build` step itself, they set up the import directory and `go_package` default value.
I suppose it would be better to place them in a a `post-extract` hook. I do realize that that would mean the hook gets run for every package (and should bail immediately if the package is not a Go buildstyle package), but currently the only alternative is to override `do_build()` with most of its original contents from `build-style/go.sh` and add the non-standard build commands, `make` in this case.

Any comments on this are very welcome.

ebfe:
If the build system is `make`, why do you want to use `build_style=go`?

ackalker:
@ebfe Because `go` build style also sets up and exports environment variables specifically for use by `go build`, Also, the `gnu-makefile` build-style exports a conflicting `LDFLAGS` which interferes with yet another package i'm creating (Go uses `CGO_LDFLAGS`).
See `common/environment/build-style/go.sh`.

Also, I think that it is still possible that `docker` could become a `go get`able package like the others...

ackalker:
To be clear: `github.com/docker/runc` _is_ a Go package, it is just not yet properly vendored in and requires special treatment. (it is an forked, unreleased, commit-pinned version of `github.com/opencontainers/runc` required by Docker 1.13.1, don't ask me, I'm not a Docker dev...).

ackalker:
Working package template, illustrating the issue:
```sh
# Template file for 'docker-runc'
pkgname=docker-runc
version=1.0.0.rc2.20170129
revision=1
_githash=9df8b306d01f59d3a8029be411de015b7304dd8f
wrksrc="runc-${_githash}"
build_style=go
go_import_path=github.com/docker/runc
hostmakedepends="git pkg-config"
makedepends="libseccomp-devel"
short_desc="runc container cli tools"
[...]

PR #9480, opened 2017-11-23
Tick cleanup

the-maldridge:
Significant cleanup of the templates.

Not to be merged until tests confirm this all still works.


common/build-style/go.sh
```diff
@@ -9,6 +9,11 @@ do_build() {
 		ln -fs $PWD "${path}"
 	fi
 
+	if [ -x /usr/bin/gdm ] && [ -f Godeps ] ; then
```
pullmoll: There is a name clash between packages `gdm` and `go-gdm`: both of them install `/usr/bin/gdm`.

the-maldridge: Yes, this is a known issue.  They should conflict if they do not already.

pullmoll: Sorry, they do indeed conflict. I hadn't seen that and was thinking "You have seen this name elsewhere..." :)

the-maldridge: No worries, its a good catch as it means that a Greeter written in Go for GDM won't work with our current build system.  Fortunately GDM seems to be somewhat dead now.

srcpkgs/telegraf/template
```diff
@@ -1,36 +1,19 @@
...
-	mkdir -p "$(dirname ${path})"
-	ln -fs $PWD "${path}"
-}
-
-pre_build() {
-	gdm restore
-}
-
-do_build() {
-	go install -ldflags="-X main.version=${version} -X main.branch=${_branch} -X main.commit=${_commit}" ./...
```
ebfe: I think at least `-X main.version=${version}" should be kept, otherwise it identifies as "version: unknown". Can be set via the `go_ldflags` variable. Same for the other packages.

srcpkgs/telegraf/template
```diff
@@ -1,36 +1,19 @@
...
-revision=1
-_branch=release-1.4
-_commit=b473b6a
+revision=2
 build_style="go"
-nocross=https://travis-ci.org/kulinacs/void-packages/jobs/258706435
-go_import_path="github.com/influxdata/${pkgname}"
-build_wrksrc="_build-${pkgname}-xbps/src/${go_import_path}"
+go_import_path="github.com/influxdata/telegraf"
+go_package="./..."
```
ebfe: This also builds/installs `/usr/bin/stress_test_write` which looks like a test binary to me. Also part of the current package so maybe I'm wrong and it's useful anyways?

common/build-style/go.sh
ebfe: The underlying issue that these can't simpy be ran from a pre_build is essentially #5816?

the-maldridge: Yes, that is essentially correct.  I would not be averse to this taking the place of a configure stage, but I think that might be a little silly.

srcpkgs/telegraf/template
the-maldridge: It can be useful to know that your server is configured correctly, but isn't strictly required.  I could remove it without a loss of functionality.

common/environment/build-style/go.sh
```diff
@@ -14,6 +14,7 @@ case "$XBPS_TARGET_MACHINE" in
 esac
 
 export GOPATH="${wrksrc}/_build-${pkgname}-xbps"
+export GOSRCPATH="${GOPATH}/src/${go_import_path}"
```
ebfe: No need to export Maybe s/GOSRCPATH/go_srcpath/ to match the other variables relevant to and defined by the go build style?

the-maldridge: the srcpath though isn't settable.  This matches the other variables that are provided by the style and are read only.

ebfe: The other flags in the same style are for the go _compiler_, that's why they are exported.

the-maldridge:
Looks like 2 nodejs builds are too much for travis to handle.  I can split that out as a separate PR, but that will still be too long to run on Travis.

srcpkgs/kapacitor/template
```diff
@@ -18,21 +18,9 @@ homepage="https://www.influxdata.com/time-series-platform/kapacitor/"
...
-	go build -ldflags="-X main.version=${version} -X main.branch=${_branch} -X main.commit=${_commit}" ./cmd/kapacitord
-}
-
-do_install() {
-	vbin kapacitor
-	vbin kapacitord
-}
+# This package has a Gopkg.lock and a Gopkg.toml, but for build the
+# required files are part of the dist tarball.  Thus we don't run dep
+# ensure for this build.
```
Vaelatern: `dep ensure` should always be safe, is it not in this case?

the-maldridge: Should and are are very different things.  Running dep ensure results in an unbuildable source directory.

PR #9498, opened 2017-11-24
New package: fmit-1.1.13

cr6git:
Free Music Instrument Tuner 🎼


cr6git:
Oopss... According to [INSTALL.txt](https://github.com/gillesdegottex/fmit/blob/master/INSTALL.txt#L26) I forgot to add the qmake options. 
Sorry, I will change that right now.

cr6git:
it's done.

srcpkgs/fmit/template
```diff
@@ -0,0 +1,31 @@
...
+do_configure() {
+	qmake "CONFIG+=acs_qt acs_alsa acs_jack acs_portaudio" \
+		PREFIX=/usr \
+		LIB=/usr/lib \
+		QMAKE_CC=$CC QMAKE_CXX=$CXX QMAKE_LINK=$CXX \
+		QMAKE_CFLAGS="${CFLAGS}" \
+		QMAKE_CXXFLAGS="${CXXFLAGS}" \
+		QMAKE_LFLAGS="${LDFLAGS}" \
+		fmit.pro
+}
```
Vaelatern: Do you need to add `fmit.pro` to the end? Or can you fit the `"CONFIG+` string and `fmit.pro` into `configure_args` and use the build style?

cr6git: I tried using `configure_args="CONFIG+=acs_qt acs_alsa acs_jack acs_portaudio"` but it doesn't work: 
````
=> fmit-1.1.13_1: running do_configure ...
Cannot find file: acs_alsa.
Cannot find file: acs_jack.
Cannot find file: acs_portaudio.
=> ERROR: fmit-1.1.13_1: failed to run do_configure() at line 17.
````
seems like `xbps-src` is not managing the quotes properly.

newbluemoon:
What does work is

    configure_args="CONFIG+=acs_qt CONFIG+=acs_alsa CONFIG+=acs_jack CONFIG+=acs_portaudio"

But it looks a bit strange.

cr6git:
@newbluemoon thanks for the information! 

@Vaelatern do you agree?

newbluemoon:
Found the right sequence to escape the quotation marks:

    configure_args="\\\\"CONFIG+=acs_qt acs_alsa acs_jack acs_portaudio\\\\""

Does this look better? ;-)

cr6git:
Personally, I prefer your previous proposal...

PR #9504, opened 2017-11-24
lumina: update to 1.4.0

hipperson0:
Also claim maintainership as package is orphaned.


hipperson0:
Travis is failed as only the signature files for `cross-aarch64-linux-musl-0.24_2` and `cross-armv7l-linux-musleabihf-libc-0.24` are present in the repo.

PR #9542, opened 2017-11-26
Orphan kulinacs packages

kulinacs:
As I am no longer using Void Linux, I will no longer be able to maintain these packages.


the-maldridge:
Future reference, if you are making a PR to a project which does CI with Travis and you know your PR should not go through CI (as it shouldn't have here) you can add `[skip ci]` to the commit description and this won't schedule your job on the CI servers.

PR #9543, opened 2017-11-26
CONTRIBUTING.md: Add reference to ESR Rule #5

the-maldridge:
Though we'd appreciate people come and stay, that isn't always
possible either due to the level of commitment it takes to maintain
packages, or the personal interests of individual maintainers.  This
commit adds a reference to the ESR paper "The Mail Must Get Through"
which encourages maintainers to find a suitable replacement if they
decide they need to move on.


the-maldridge:
I could see this going better at the bottom of the Contributing file.  I kind of forgot we had that file.

chneukirchen:
While I agree on the sentiment, I'd rather not we quote that guy on anything. Also IMO orphaning packages is enough.

the-maldridge:
I can wholeheartedly agree on the not directly quoting him, that's a bit more association than we probably want.  What do you think of a paraphrased version?  Right now none of our documentation explains what to do if you're not going to maintain packages anymore, and I think there should be something there.

avoidr:
I suppose I hadn't lurked long enough, but may I know what the problem with ESR is?

>Right now none of our documentation explains what to do if you're not going to maintain packages anymore

If I may; I find it 100% natural what to do and feel as if this shouldn't have to be a rule. But then again…

PR #9547, opened 2017-11-26
atom: update to version 1.22

teajay-fr:
Bumped the curl package revision which was missing in the previous pull request.


teajay-fr:
Didn't get that this had been fixed and merged. Thank you !

PR #9553, opened 2017-11-26
New package: sfeed

kojicomics:


srcpkgs/sfeed/template
```diff
@@ -0,0 +1,22 @@
...
+short_desc="RSS and Atom parser"
+maintainer="Rubén Santos <kojicomics@cocaine.ninja>"
+license="ISC"
+homepage="https://git.codemadness.org/sfeed"
+
+do_fetch() {
+	git clone git://git.codemadness.org/sfeed ${wrksrc}
+	cd ${wrksrc}
+	git reset --hard ${version}
+}
```
Vaelatern: Is there no better way?

kojicomics: I simplified it by using a oneliner. There's no tarball, so we have to clone the last tag version.

Johnnynator: http://codemadness.org/releases/sfeed/ this does look like tarballs to me

Vaelatern: Please use the tarballs, @kojicomics

kojicomics: @Johnnynator Thanks, I will change it

PR #9563, opened 2017-11-27
New package: arena-1.1

cr6git:
[Arena](http://www.playwitharena.com) for chess players.


Johnnynator:
Shouldn't restricted be set?

> However, distribution or offering of Arena Chess GUI, SOS for Arena, Dragon, AnMon, Nejmet, Ruffian, Betsy, Chispa, Hermann, Frenzee, Spike, Aice and ELOstat on other webpages, newspaper, CDs, print and other media is only allowed in unmodified form and with the written permission of their respective authors

PR #9569, opened 2017-11-27
rpi-kernel: enable F2FS_FS_SECURITY config

ElementW:
This enables the use of the disabled-by-default "security." namespace in
extended attributes, allowing for storage of capabilities on files.

In my specific use case I'm trying to use capabilities with `sslh` to allow transparent proxying without running the daemon as root. As my root FS is F2FS, I can't do that yet.


chneukirchen:
LGTM, we already do that on our main kernels.

ElementW:
Just noticed I forgot to bump up the package's revision number. This is done now.

PR #9580, opened 2017-11-27
rdiff-backup: update to 1.3.3.

maxice8:


Duncaen:
> Version 1.3.3, released March 16th 2009, is the new development/unstable version.

Issue #9526, opened 2017-11-25
[Request] Free Orion 


codingHahn:
FreeOrion is a 4x space strategy sim. Github repo: https://github.com/freeorion/freeorion Website: http://www.freeorion.org/index.php/Main_Page
The page about compiling: http://www.freeorion.org/index.php/Compile

cr6git:
I created a template. 
I am currently compiling the program on my laptop, but this is so heavy that my CPU is in trouble! 
Please be patient. 😉

codingHahn:
Thank you. No worry, I'll be

cr6git:
@codingHahn done!

PR #9583, opened 2017-11-27
New package: freeorion-0.4.7.1

cr6git:
resolves [#9526](https://github.com/voidlinux/void-packages/issues/9526)


srcpkgs/freeorion/template
```diff
@@ -0,0 +1,17 @@
+# Template file for 'freeorion'
+pkgname=freeorion
+version=0.4.7.1
+revision=1
+_release=${version}_2017-09-03.139ffd9
+wrksrc="src-tarball"
+build_style=cmake
+hostmakedepends="boost-python cppcheck doxygen python-pycodestyle"
```
Vaelatern: sure that `cppcheck` isn't a `checkdepends`?

cr6git: not sure. I added it in the `hostmakedepends` because of this error during `do_configure()` : 
````
-- Found GLEW: /usr/include  
-- Found PNG: /usr/lib/libpng.so (found version "1.6.34") 
-- Could NOT find cppcheck (missing: CPPCHECK_EXECUTABLE) 
-- Found pycodestyle: /usr/bin/pycodestyle (found version "2.3.1") 
-- Found Doxygen: /usr/bin/doxygen (found version "1.8.13") found components:  doxygen missing components:  dot
-- Configuring done
````
what do you think ?

cr6git:
Unfortunately, `armv7` and `aarch64` failed.

srcpkgs/freeorion/template
Vaelatern: sure, it can stay.

Vaelatern:
Patch to the makefile is needed... it ALMOST figures out how to find `libboost_python.so`

cr6git:
I found another solution: moved `boost-python` to the `makedepends`. 
On my system I was able to cross build the package for `armv7`.

codingHahn:
Thank you.

PR #9584, opened 2017-11-27
New package: mdBook-0.0.21

hwangh95:


cr6git:
`hostmakedepends="cargo rust"`

`rust` is automatically pulled by `cargo`.

the-maldridge:
I'm suspicious of nocross=yes.  If this is something systemic with rust I would rather set it on the build style.

hwangh95:
I don't know if it is systematic, I saw that other rust packages had it so I wrote it that way. Thought it might break without it.

the-maldridge:
Please remove it and force push, if it turns out it needs it, you'll be able to see things break on Travis.  Since this just processes files though I don't think it should be architecture dependent.

srcpkgs/mdBook/template
```diff
@@ -0,0 +1,20 @@
+# Template file for 'ripgrep'
```
the-maldridge: This doesn't look like ripgrep to me.

srcpkgs/mdBook/template
```diff
@@ -0,0 +1,20 @@
...
+pkgname=mdBook
+version=0.0.21
+revision=1
+hostmakedepends="cargo rust"
```
the-maldridge: rust is a dependency of cargo, please remove it from this list.

srcpkgs/mdBook/template
```diff
@@ -0,0 +1,20 @@
...
+short_desc="Create book from markdown files. Like Gitbook but implemented in Rust"
+maintainer="Huidong Hwang <hwangh95@gmail.com>"
+license="MPL-2.0"
+homepage="https://github.com/rust-lang-nursery/${pkgname}"
+distfiles="https://github.com/rust-lang-nursery/${pkgname}/archive/${version}.tar.gz"
+checksum=bffb39106ab9f546c7a02667aa3616c06f05dfb566ef4830f51fd244c7986239
+#nocross=yes
```
the-maldridge: This should indeed be nocross.  Perhaps provide a brief comment that this cannot be cross compiled due to limitations in cargo at this time.

srcpkgs/mdBook/template
```diff
@@ -0,0 +1,20 @@
...
+
+do_build() {
+	cargo build --release
+}
+do_install() {
```
the-maldridge: newline between the end of the previous function and the start of this one.

srcpkgs/mdBook/template
hwangh95: Yeah it doesn't lol

Issue #9575, opened 2017-11-27
firefox-i18n update to 57.0


SevereOverfl0w:
This package is out of date, and doesn't work with the new firefox 57.

cr6git:
@SevereOverfl0w done.

PR #9589, opened 2017-11-28
firefox-i18n: update to 57.0.

cr6git:
resolves [#9575](https://github.com/voidlinux/void-packages/issues/9575)


cr6git:
`srcpkgs/firefox-i18n/template:18: indent with tabs`

Thank you very much Travis! 
Now I have to fix this dirty template...

PR #9601, opened 2017-11-28
qutebrowser: update to 1.0.4

NoctuaNivalis:


srcpkgs/qutebrowser/template
```diff
@@ -1,6 +1,6 @@
 # Template file for 'qutebrowser'
 pkgname=qutebrowser
-version=1.0.3
+version=1.0.4
 revision=1
 _srcrev=1
```
The-Compiler: You can probably remove this now?

cr6git:
Seems like there's trouble with musl, in [#9814](https://github.com/voidlinux/void-packages/issues/9814).

PR #9606, opened 2017-11-28
snoopy: fix cross compilation

newbluemoon:
Just removed `nocross`. Seems to be ok now.


Vaelatern:
Please remove the revbump. Removing nocross will have no effect on existing packages, and it will only create new packages in cross building.

newbluemoon:
Removed revbump. Sometimes I’m not sure if a revbump is necessary or not. Maybe some guideline could be added to the manual?

PR #9609, opened 2017-11-28
MEGAsync and new MEGAcmd

Johnnynator:
Since the way the megacmd package is build will be deprecated after the next sdk release I created a new package. 
Should I let it named megacmd or use the github repo name MEGAsync? 
If I let the name be the old I would have to set it revert so it replaces the old package since the new Version number is lower then the old one.


Johnnynator:
Damn, forgot to check the MEGAcmd package after i excperimented with MEGAsync... Will fix it.

thypon:
Have you tried using it?
It was incompatible with libressl an year ago.

Johnnynator:
@thypon megacmd is already in the repo since 1month, I haven't yet tested all the tools completely but all I tested so far was working (e.g. fuse mountpoint of my MEGA account)

pullmoll:
@Johnnynator could you rebase common/shlibs?
I want to avoid trouble merging this ;)

Johnnynator:
@pullmoll rebased now.

srcpkgs/MEGAcmd/template
```diff
@@ -0,0 +1,42 @@
+# Template file for 'MEGAcmd'
+pkgname=MEGAcmd
+version=0.9.6
+revision=1
+build_style=gnu-configure
+make_build_args='LIBTOOLFLAGS="--tag=CXX"'
+hostmakedepends="autoconf-archive autogen gettext automake libtool pkg-config"
+makedepends="MEGAsdk-devel readline-devel pcre-devel"
+short_desc="BSD-2"
```
Anachron: This is wrong, you need to add a speaking description for your package, mostly the title of a github repo, like this one:
`Command Line Interactive and Scriptable Application to access MEGA `.

PR #9615, opened 2017-11-28
libreoffice: fix desktop files on i686

### LINKED ISSUE #9614: Libreoffice: faulty desktop files on i686
### System

* xuname:
Void 4.13.15_1 i686 GenuineIntel uptodate r
* package:  
  libreoffice-5.4.3.2_1

### Expected behavior
provide desktop files that do link to the correct path for executable and icon
### Actual behavior
Desktop files on i686's version of libreoffice (glibc) mention `libreoffice5.4` instead `of libreoffice`, as is in x86_64's package. Therefore using an desktop-file-aware application launcher fails starting libreoffice
- install libreoffice in i686
- check entries in an application menu of your choice (whiskermenu for example) and the corresponding *.desktop files in `/usr/lib/libreoffice/share/xdg/`

Piraty:
This PR fixes #9614


pullmoll:
It's probably best to apply this patch without revbump and wait for the next update of libreoffice. They appear every couple of weeks or few months. I don't think it's worth hogging the builders just for this bugfix.

Piraty:
thanks. commit now avoids revbump

PR #9617, opened 2017-11-29
wine: nopie_files

thypon:


common/hooks/post-install/06-strip-and-debug-pkgs.sh
```diff
@@ -99,7 +99,13 @@ hook() {
...
 				fi
 				echo "   Stripped executable: ${f#$PKGDESTDIR}"
-				if [ -z "$nopie" ]; then
+				for x in ${nopie_files}; do
+					if [ "$x" = "$fname" ]; then
+						nopie_found=1
+						break
+					fi
+				done
+				if [ -z "$nopie" ] || [ -z "$nopie_found ]; then
```
Gottox: Rather use the fullpath here, so instead of wine, use /usr/bin/wine.

Gottox:
Workaround is okay for me.

common/hooks/post-install/06-strip-and-debug-pkgs.sh
```diff
@@ -99,7 +99,13 @@ hook() {
 					return 1
 				fi
 				echo "   Stripped executable: ${f#$PKGDESTDIR}"
-				if [ -z "$nopie" ]; then
+				for x in ${nopie_files}; do
+					if [ "$x" = "$fname" ]; then
+						nopie_found=1
```
pullmoll: You probably need to `unset nopie_found` somewhere, or it may still be set for an unexpected nopie binary?

PR #9640, opened 2017-11-29
libGL: update to 17.2.6.

maxice8:
Untested


cr6git:
`srcpkgs/libGL/template:14: license 'MIT', but no use of vlicense`

Just add: 
````
post_install() {
	vlicense docs/license.html
}
````

cr6git:
Merged, omitting the license ?

PR #9645, opened 2017-11-30
New package: urxvtconfig-1.0

cr6git:
I used a custom `do_configure()` because of this problem: 
````
=> urxvtconfig-1.0_1: running do_configure ...
Usage: /usr/lib/qt5/bin/qmake [mode] [options] [files]

=> ERROR: urxvtconfig-1.0_1: failed to run do_configure() at line 17.
````


maxice8:
can't it be added to configure_args ?

cr6git:
@maxice8 you're right, thanks!! (if you have time, maybe you could look at [#9498](https://github.com/voidlinux/void-packages/pull/9498) too?)

newbluemoon:
`qmake` just didn’t find the .pro file. What also works is to just set

    build_wrksrc="source"

;-)

cr6git:
@newbluemoon yes, I guess this should work too!

cr6git:
That's good! 🎉

PR #9646, opened 2017-11-30
New package: discord-canary-0.0.31

gingyr:


Vaelatern:
How is this diferent from `discord`?

gingyr:
Discord Canary is Discord's alpha testing program that receives updates before PTB, which receives updates before Stable.

Vaelatern:
We do not generally merge alpha quality software.

Coding-Doctors:
Discord prefers that regular users don't use canary as it is prone to breakages. If any "testing version" of Discord would be considered, ptb would be it. This is best left as discord prefers, where users who are intrested go out and grab it themselfs, rather than being overly public about it. 
**Canary isn't intended as a general consumer program.**

gingyr:
Understandable. I will close this PR.

PR #9662, opened 2017-12-01
New package: loksh-6.1

renatoaguiar:


srcpkgs/loksh/template
```diff
@@ -0,0 +1,31 @@
...
+make_install_args="BIN_NAME=loksh"
+hostmakedepends="pkg-config"
+makedepends="libbsd-devel"
+short_desc="A Linux port of OpenBSD's ksh"
+maintainer="Renato Aguiar <renato@renag.me>"
+license="Public Domain"
+homepage="https://github.com/dimkr/loksh"
+distfiles="https://github.com/dimkr/loksh/archive/${version}.tar.gz"
+checksum=988a9ae4c33aa8ff9e9dbc42ad22678ce7bd46892b9150c37b0e656f8a6ee7f1
+nopie=true
```
Vaelatern: really?

renatoaguiar: It doesn't make sense to me either. I even tried adding '-fPIC', but it didn't help.

renatoaguiar: I just found out that '-pic' is also needed in LDFLAGS.

Vaelatern: Odd. `gcc` now defaults to `PIE` on. Maybe it's just a problem local to you?

Johnnynator: @renatoaguiar is it possible that you chroot isn't up to date? For updating it run `./xbps bootstrap-update`. For me loksh complies fine as pie without the additional flags

renatoaguiar: @Johnnynator That's exactly what happened. Afer updating my chroot it worked fine :)

I'll open a new PR fixing that.

Issue #6557, opened 2017-05-31
Package Request: tor-browser, zuluCrypt and pulseeffects


kevinsmia1939:
Tor-broswer
I noticed that there is tor and torsocks package in the repo, but there aren't tor-browser in it. https://www.torproject.org/download/download.html.en

zuluCrypt
zuluCrypt is currently Linux only and it does hard drives encryption and it can manage PLAIN dm-crypt volumes, LUKS encrypted volumes, TrueCrypt encrypted volumes and VeraCrypt encrypted volumes.
zuluCrypt can manage encrypted volumes that are hosted in image files, lvm, mdraid, hard drives, usb sticks or any other block device.
zuluCrypt can also encrypt stand alone files (zuluCrypt menu -> zC -> encrypt a file).
https://mhogomchungu.github.io/zuluCrypt/

Pulseeffects
Limiter, compressor, reverberation, stereo equalizer and auto volume effects for Pulseaudio applications
https://github.com/wwmm/pulseeffects

Thanks for your hard work.

direwolfie:
> Tor-broswer
> I noticed that there is tor and torsocks package in the repo, but there aren't tor-browser in it.

I might be wrong, but from the website it looks like the tor-browser come with its own tor package already. So you can just download and use that.

If you like to use the tor that you install from repo. You can just use ESR version of Firefox with it(also in report). It is the versioin of firefox used by Tor-browser. All they do is add some plugins and modify config.

kevinsmia1939:
@direwolfie It is likely that the user will misconfigure tor, that is why tor bundle exist. I for example never configure tor manually, because I might mess up, tor-bundle is made for this purpose. I live in extreme lese majest law country here.

direwolfie:
No worries. :)
Yes, that is what I was trying to say - you can just unarchive the tor-browser from their website and use it. And it autoupdates itself.

Also, I highly suggest that you verify the downloaded tor-browser archive with gpg (if not already) - https://www.torproject.org/docs/verifying-signatures.html.en

cr6git:
@kevinsmia1939 For PulseEffects, take a look at my attempt [here](https://forum.voidlinux.eu/t/swh-plugins-for-musl/2703/2?u=cr6).

Piraty:
I have a template for zulucrypt, it is not PR-ready yet, but you can edit to your liking. take special look into the configure_args

The software runs stable.

kevinsmia1939:
[...]

PR #9684, opened 2017-12-01
New package: pulseeffects-3.0.9

cr6git:
And its depencency `swh-plugins`, resolves [#6557](https://github.com/voidlinux/void-packages/issues/6557).


Duncaen:
@cr6git shit should be 3 commits, one for each template, with a reason for the change as commit message.

cr6git:
@Duncaen sorry if I did something wrong.

PR #9693, opened 2017-12-01
monero-core: rename to monero-gui

Piraty:
because
1. it is more intuitive
2. upstream names it this way

I also added monero explicitly as a dependency.

@Hoshpak ok?


Hoshpak:
The upstream rename happened only recently. I planned to rename the package once the next version is released. Of course we could just as well do it now but I don't think it should be done this way. Just renaming the package and declaring a replace would lead to xbps not automatically installing the "new" package on systems that already have the "monero-core" package installed, leaving the users without updates. I think it would be better to turn "monero-core" into a transitional meta package and adding "monero-gui" as a new package, replacing the old one. Also the dependency on monero is something I deliberately didn't add when creating the package since the gui can be used as a "light wallet" by using a remote node, no local monerod required. This would make the monero package more of a candidate for "recommends/suggests" which xbps doesn't have. Perhaps it would warrant an install time message to notify the user but at the time I just assumed that someone using monero would know about the infrastructure behind it and know to install the right packages.

Piraty:
i talked to Hoshpak and we agreed to wait for the next release, and also make a transitional meta-package for the old package-name

PR #9701, opened 2017-12-02
fcitx: update to version 4.2.9.5

### LINKED ISSUE #9681: fcitx is oudated
### System

* package:  `fcitx-4.2.9.1`

### Expected behavior
should be up to date (the latest version is 4.2.9.5).
### Actual behavior
is outdated (and seems to be orphaned).

Thunderbottom:
Updated possibly orphaned package.

Closes #9681


Thunderbottom:
ugh I'm unable to fix the armv7 and aarch64 builds, any help is appreciated

srcpkgs/fcitx/template
```diff
@@ -1,18 +1,18 @@
...
 license="GPL-2, LGPL-2.1"
 homepage="https://fcitx-im.org/"
 distfiles="http://download.fcitx-im.org/fcitx/fcitx-${version}_dict.tar.xz"
-checksum=5337c8226417a1d4e6f88b3b1048aa73695d281ed483568a7b6a9ee849ad6c77
+checksum=be398ac4433f6f4599d2728e20615e4f4602a5f67e9c29df4a3d4c1126d376ca
 
 hostmakedepends="pkg-config doxygen gobject-introspection extra-cmake-modules"
 makedepends="iso-codes gettext-devel enchant-devel libxml2-devel
- libxkbfile-devel icu-devel qt-devel dbus-devel gtk+-devel gtk+3-devel"
+ libxkbfile-devel icu-devel qt qt-qmake dbus-devel gtk+-devel gtk+3-devel"
```
Johnnynator: Your probably need qt-devel here and qt-qmake normally belongs into hostmakedepends.

Johnnynator: qt-host-tools is also needed for cross compiling. (And then you will run into troubles with gobject-introspection)

Thunderbottom: alright, changing

Thunderbottom:
sorry for being so sloppy with commits

Johnnynator:
I investigated cross compiling further and disabling gir gets you through configuration but fcitx compiles a few binaries with the cross compiler while they are needed for the host and therfore compiling failes. I would propose setting nocross=yes for now.

Log: https://bin.disroot.org/?df402a001b8993e7#UZ2Nyap4dMP2k50GRsHBYyRRvryr0xr9c0aZ5unJXJo=

Thunderbottom:
I could just disable QT_im for now and that will work too

Johnnynator:
It looks like you have acidently added the package shotcut to your commit.

Johnnynator:
@Thunderbottom the problem isn't qt, if you would add qt-host-tools to hostmakedepends it wouldn't error because of qt (as I mentioned 2 commetns ago). But you run into other problems then that are harder to fix, and atleast I don't want to fight my way through  all the used CMakeLists and try to fix them.

This is [my works so far](https://github.com/Johnnynator/void-packages/commit/bc3f59dfba615588700fd1130049b6bccfd596f0), musl is fixed and cross compiling errors at the point it tries to execute a cross compiled tool, and not while configureing anymore.

Thunderbottom:
Okay then, my bad. I tried doing it and I too am getting the same errors you got.

I do not doubt your works, I just thought it'd be worth trying. Guess I was wrong. I have disabled cross compiling for now and I hope it works! Thanks for your help :) appreciate it!

Johnnynator:
@Thunderbottom if you want you can add my musl fix. (Check my last link)

Thunderbottom:
@Johnnynator no I'm sorry but you should do it, because it is you who made it right. I'll try using qt-devel in hostmakedepends one last time before looking into some other way of getting this working.

Johnnynator:
I wasn't yet able to get it cross compiled and I don't want to take the time and rewrite half of their CMake scripts. So I would say just let nocross set, it never cross compiled so we will not have an outdated package because of that.

the-maldridge:
In general for cross compilation you just need to turn off the host tests and then the cross compilation will work.

Johnnynator:
@the-maldrige these aren't tests that error, while there would be some tests somewhere else these didn't get executed so far.

Thunderbottom:
as @Johnnynator said disabling gir would only get you so far, and it still fails. Our best bet is to disable cross compilation for now.

the-maldridge:
@Johnnynator I'm well aware of the nature of the failing tests.

@Thunderbottom then you shall need to wait for a maintainer that is willing to do so, as I am not.

wangp:
I investigated the cross compiling situation a bit as well and here is what I found:
- fcitx-scanner generates header files that are consumed in the build so cannot be skipped
- other binaries are built and executed to generate data files: fcitx-po-parser, createPYMB, txt2mb, comp-spell-dict, maybe others

The easiest way to make these executables available during the build might be to make use of the CMake [EXPORT command](https://cmake.org/Wiki/CMake_Cross_Compiling#Using_executables_in_the_build_created_during_the_build).
1. First, build the necessary executables for the host architecture (in a separate build directory, say `hostbuild`) and generate an `ImportExecutables.cmake` file using `EXPORT`.
2. When cross-compiling for the target architecture, include that `ImportExecutables.cmake` file and force custom commands to run the binaries built for the host architecture instead of the target architecture. (Assuming it works, the necessary cmake changes should be pushed upstream.)

Also, fcitx 4.2.9.5 includes a [change](https://github.com/fcitx/fcitx/commit/b0a18300a4ac27ed1c1f5259085c2a3eea643f05)
that makes use of `pkg_get_variable` which, if I have understood correctly, currently [does not work in a cross-compiling setup](https://gitlab.kitware.com/cmake/cmake/issues/15805).

Thunderbottom:
ok that is a bit confusing

Johnnynator:
@Thunderbottom do you want to claim maintainership of the package? The original maintainer hasn't done anything since 3y.

Thunderbottom:
I would love to, but what about cross compilation? I still can't figure it out

Johnnynator:
@Thunderbottom IMO cross compilation can be ignored for now.

Thunderbottom:
I'll take over the package maintainence then

Thunderbottom:
I'll take over the package maintainence then

PR #9727, opened 2017-12-02
New package: libexecinfo-1.1

Johnnynator:


pullmoll:
So now we can start removing the musl no-execinfo patches and link against libexecinfo?

PR #9753, opened 2017-12-03
New package puppetserver

the-eater:
This PR also enables jruby support in facter


pullmoll:
Please read how to format commit messages.
Please install xtools and use xlint to clean up your package template (or take a look at the Travis-CI warnings).
Please no default dependency of facter on openjdk. Make it a build_option and default to off, or facter can't be built for target architectures where openjdk is not available.

PR #9764, opened 2017-12-04
qt5: update to 5.9.3 [ci skip]

pullmoll:
New components: qt5-networkauth, qt5-purchasing, qt5-remoteobjects
New devel pkg: qt5-quickcontrols2-devel
Moved lconvert, lupdate and lrelease from qt5-tools to qt5-host-tools

IMPORTANT: This version needs an i686 environment to cross build for
the arm* architectures! This is due to qtwebengine, WebKit's v8 engine
build rules which otherwise try to build with -m32 and fail doing so.


pullmoll:
After removing the "-hostdir host" from the configure args, there are now some more tools built for the host which need to be rebuilt for the target to get a complete set of cross packages.

I could use some help here :) Currently `qmlcachegen` won't link for the target arch because of missing symbols. This probably also (like some other tools) needs a fixed `${FILESDIR}/qmlcachegen.pro` file based on the one found in the source tree with additional modules listed in `QT += ...`.

pullmoll:
Closing because I'm going to try to package qt5-5.10.0 as separate modules instead of a monolithic package with submodules.

New year new luck :)

PR #9769, opened 2017-12-04
New package: orion 1.6.1

### LINKED ISSUE #9524: Package request: Orion
I dont know if anyone is interested in picking up https://alamminsalo.github.io/orion/ but here it is

Thunderbottom:
Qt-based client for Twitch.tv.

closes #9524


srcpkgs/orion/template
```diff
@@ -0,0 +1,20 @@
...
+build_style=qmake
+hostmakedepends="gcc qt5-qmake qt5-svg-devel qt5-quickcontrols2-devel qt5-graphicaleffects mpv-devel"
+makedepends="qt5-declarative-devel qt5-quickcontrols qt5-svg qt5-quickcontrols2 mpv"
+short_desc="Qt-based client for Twitch.tv"
+maintainer="Chinmay Pai <chinmaydpai@gmail.com>"
+license="GPL-3"
+homepage="https://github.com/alamminsalo/orion"
+distfiles="${homepage}/archive/v${version}.tar.gz"
+checksum=eaa8ace822d3381a1915db82c4203d4ce4e56ec324d4ee18544bd68b00bce749
+only_for_archs="i686 x86_64 x86_64-musl"
```
Vaelatern: Please remove this line

Thunderbottom: But it doesn't build for aarch64 and armv7

srcpkgs/orion/template
```diff
@@ -0,0 +1,20 @@
+#template file for 'orion'
+pkgname=orion
+version=1.6.1
+revision=1
+build_style=qmake
+hostmakedepends="gcc qt5-qmake qt5-svg-devel qt5-quickcontrols2-devel qt5-graphicaleffects mpv-devel"
```
Johnnynator: gcc is default so you don't have to pull it in manually. Also all the *-devel packages should be in makedepends and in makedepnds you normally don't have no devel packages.

hostmakedends is normally for tools that do get executed on the host while compiling smth e.g. qmake.
makedepnds should inlcude the packages that do get linked at / the coresponding headers (devel packages)
In some cases there are devel pacakges in hostmakedepends needed if cross compiling because e.g. qmake does also search for them for the host arch (or so) but if they are only needed for cross_compiling they should only be pulled in when the package does get cross compiled. (look in pacakges that do get build with qmake).

Thunderbottom: shit my bad. didn't really see that.

srcpkgs/orion/template
Johnnynator: if it doesn't cross compile you should set nocross and not only_for_archs. 
only_for_archs is only for cases were the program is only designed for only these archs e.g. if it is only avaible as a binary or if it is written is assembler.

Vaelatern: and it probably will cross compile if you let @Johnnynator see the result and tell you about qt5-host-tools.

Thunderbottom: alright I removed the only_for_archs for now
Thanks for help :) appreciate it!

srcpkgs/orion/template
```diff
@@ -0,0 +1,19 @@
+#template file for 'orion'
+pkgname=orion
+version=1.6.1
+revision=1
+build_style=qmake
+hostmakedepends="qt5-qmake"
+makedepends="qt5-declarative-devel qt5-quickcontrols qt5-svg qt5-quickcontrols2 mpv qt5-svg-devel qt5-quickcontrols2-devel qt5-graphicaleffects mpv-devel"
```
Johnnynator: this line should only include the devel packages, e.g. qt5-svg does get pulled in by qt5-svg-devel, and you forgot mpv-devel

srcpkgs/orion/template
Thunderbottom: [this is the error that pops up] (https://travis-ci.org/voidlinux/void-packages/jobs/311374864)

srcpkgs/orion/template
Thunderbottom: mpv-devel is included

Johnnynator: Ok, I missed the the scroll bar, generallly you should start a new line after around I think 70 charactrs (white spaces included)

Thunderbottom: okay will do, by the way, did you look at the error?

https://travis-ci.org/voidlinux/void-packages/jobs/311374864

Johnnynator: Yep, the problem is that qmake searches for the libraries in the wrong place (in the compilation itself later the correct path gets used). The generally 'accepted' workaround here is that the needed qt*-devel packges do get added to the hostmakedepends (if qt*-host-tools is not enough) 
for this package smth like
```
if [ "$CROSS_BUILD" ];then
        hostmakedepends+=" qt5-declarative-devel qt5-svg-devel qt5-quickcontrols2-devel"
fi
```
should work

srcpkgs/orion/template
Johnnynator: @Vaelatern I would be happy if only adding qt5-host-tools would work at more then ~1% of the packages... qmake never find any module if it is not installed also for the host.

srcpkgs/orion/template
Thunderbottom: okay let me try for once
thanks :)

srcpkgs/orion/template
```diff
@@ -0,0 +1,23 @@
+#template file for 'orion'
+pkgname=orion
+version=1.6.1
+revision=1
+build_style=qmake
+hostmakedepends="qt5-qmake"
+makedepends="qt5-declarative-devel qt5-quickcontrols qt5-svg-devel qt5-quickcontrols2-devel qt5-graphicaleffects mpv-devel"
```
Johnnynator: qt5-quickcontrols is probably not needed, otherwise the template looks good

icf20:
is there a dependency missing ? 
```
$ orion
Selected default network configuration
Segmentation fault
```

cr6git:
@icf20 👍 `Segmentation fault` for me too.

adomas187:
Same issue, running x86_64-musl
screenshot_20171219_033311

Thunderbottom:
I had it working on void. Can't debug until next week though, I'm extremely sorry!

pullmoll:
Just a hint how to get more information for failing packages (the first command is required once):
```
$ sudo xbps-install -S void-repo-debug gdb
$ sudo xbps-install -S orion orion-dbg
$ gdb orion
Copyright (C) 2017 Free Software Foundation, Inc.
...
(gdb) run
<crash happens>
(gdb) bt
```
The bt command is a backtrace which should tell where the binary is crashing.
As a side note it works for me with x86_64 glibc, so it's most likely musl specific.

adomas187:
Thanks @pullmoll
Here is output running with **gdb**:

```
(gdb) run
Starting program: /usr/bin/orion 
[New LWP 1894]
[New LWP 1895]
[New LWP 1897]
[New LWP 1898]
[New LWP 1899]
[New LWP 1900]
Selected default network configuration

Thread 1 "orion" received signal SIGSEGV, Segmentation fault.
0x00007ffff627c2eb in QObjectPrivate::addConnection(int, QObjectPrivate::Connection*) ()
   from /usr/lib/libQt5Core.so.5
(gdb) 
```

_xuname_ output: Void 4.14.6_1 x86_64-musl GenuineIntel uptodate r

Johnnynator:
Probably relates to the general qt5.8 musl issues that are there. It works fine with qt5.9.3

PR #9770, opened 2017-12-04
harfbuzz: update to 1.7.2.

maxice8:
DO NOT MERGE YET, i am still testing on Firefox and Chromium.

Just put this here so to not let anyone else waste time bumping.


maxice8:
Firefox is working fine.

PR #9788, opened 2017-12-05
php: update to 7.2.0.

maxice8:


cr6git:
@maxice8 maybe `replaces` should have been added to the template: 
````
php-sodium_package() {
	replaces="php-mcrypt>=0"

}

````
see [#9820](https://github.com/voidlinux/void-packages/issues/9820).

PR #9789, opened 2017-12-05
xbps-triggers: conform to new gtk-immodules cache locations.

### LINKED ISSUE #9767: xbps-triggers: GTK IM modules cache
It seems since GTK 2.24.20 the immodules cache was moved, and should be updated with `gtk-query-immodules-2.0 --update-cache` (similarly for GTK+ 3.0). Then `/etc/gtk-2.0/gtk.immodules` and `/etc/gtk-3.0/gtk.immodules` can probably be removed.

wangp:
Attempt to fix #9767

### DIFF
--- a/srcpkgs/xbps-triggers/template
+++ b/srcpkgs/xbps-triggers/template
@@ -1,7 +1,7 @@
 # Template file for 'xbps-triggers'
 pkgname=xbps-triggers
-version=0.102
-revision=3
+version=0.103
+revision=1
 noarch=yes
 bootstrap=yes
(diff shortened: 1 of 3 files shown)


srcpkgs/xbps-triggers/template
```diff
@@ -1,7 +1,7 @@
 # Template file for 'xbps-triggers'
 pkgname=xbps-triggers
 version=0.102
-revision=3
+revision=4
```
xtraeme: Please reset `revision` to 1 and increase `version` instead.

Thanks!

PR #9790, opened 2017-12-05
fcitx: run gtk-immodules triggers

### LINKED ISSUE #3168: libfcitx-gtk, libfcitx-gtk2 and ibus don't update GTKs IM cache
GTK keeps a cached list of available input modules that GTK programs use to figure out which IMs are available.

They can be updated with ` gtk-query-immodules-{2,3}.0 --update-cache`. This will update the files in `/usr/lib/gtk-2.0/2.10.0/immodules.cache` and `/usr/lib/gtk-3.0/3.0.0/immodules.cache`.

When I ran these commands as root, the cache was updated correctly but the files couldn't be read by non-root programs. Doing a `chmod a+r` on the cache files made ibus and fcitx work correctly in GTK applications (tested: Firefox, Chromium, Gedit).

wangp:
Fixes #3168


cr6git:
Did you take into account [#9681](https://github.com/voidlinux/void-packages/issues/9681) and [#9701](https://github.com/voidlinux/void-packages/pull/9701) ?

wangp:
No, bumping the version should be orthogonal to this change.

wangp:
Rebased for fcitx 4.2.9.5

PR #9794, opened 2017-12-05
virtualbox-ose-dkms: fix build on linux4.14

renatoaguiar:


freedomrun:
It would be very nice if you could pull: VirtualBox 5.2.2 that would solve the problem.
Modules for Kernel 4.14 don't compile -> fixed in 5.2 releases after November 18 2017

Please take a look in here: https://www.virtualbox.org/ticket/17267

Best regards

PR #9807, opened 2017-12-06
axel: update to 2.16.1.

maxice8:


pullmoll:
See https://build.voidlinux.eu/changes/13636

PR #9809, opened 2017-12-06
skype: remove, client no longer supported

crater2150:
This is officially a "preview version", but any non-preview version for linux is no longer able to connect to the skype servers.

This update no longer packages a license file for the application itself, it seems like there is no plaintext license document. The full license is split up into several HTML documents: https://www.skype.com/legal/
Should I add a file containing that link or should I scrape that site and package the HTML files?


crater2150:
Ok, I noticed they switched to Electron, which should not be packaged in void as far as I understand (and I support that, Electron is bad).
So I changed this pull request to delete skype, as the non-electron versions are no longer able to connect.

Gottox:
electron is bad, but there are already exceptions to that. I'm still considering the update okay to merge. Any objections?

crater2150:
FWIW, I kept my original commit with the update in [a separate branch](/crater2150/void-packages/tree/skypeforlinux-electron) and am using it locally.

asergi:
#9896
#10457

PR #9817, opened 2017-12-07
New package: Quaternion-0.0.4

julio641742:


srcpkgs/Quaternion/template
```diff
@@ -0,0 +1,29 @@
+# Template file for 'Quaternion'
+pkgname=Quaternion
+version=0.0.4
+revision=1
+only_for_archs="i686 x86_64 x86_64-musl"
```
Vaelatern: Remove.

Vaelatern:
Refactor to use the `build_style=cmake`.

srcpkgs/Quaternion/template
```diff
@@ -0,0 +1,19 @@
...
+build_style=cmake
+makedepends="cmake qt5-declarative-devel qt5-quickcontrols"
```
Johnnynator: cmake would  belong in hostmakedepnds, but the build_style is already pulling it in so it shouldn't be needed at all.

PR #8910, opened 2017-11-05
New package: ternimal

cr6git:

PR #9822, opened 2017-12-07
New package: vpnfailsafe-git-20171207

raedwulf:
vpnfailsafe openvpn to help prevent IP leaks


newbluemoon:
As far as I know git versions are somewhat discouraged, but if there is some benefit in adding the package I think the prefered naming/versioning is

    pkgname="vpnfailsafe"
    version=0.0.20171207

raedwulf:
Did as suggested although it's a tricky one because there are existing packages that do follow different /mixture of conventions:

- compton-git for instance uses the format I used previously, "-git" and ISO date version without 0.0.
- bloaty uses the format you suggest
- plan9port uses ISO date with 0.0. but is a git package

Unfortunately the author for vpnfailsafe doesn't make releases and just maintains his git version on the Arch User Repository.

newbluemoon:
Yes, I know what you mean, sometimes things seem to be ambiguous ;). But I guess the few -git packages are remains from before. There is a post by a maintainer which recommends the versioning I mentioned, I just can’t find it anymore.

cr6git:
@newbluemoon it's in this [comment](https://github.com/voidlinux/void-packages/pull/8910#issuecomment-342137589) ;)

newbluemoon:
That’t it! Thanks @cr6git :)

Issue #1781, opened 2015-06-06
Package request: crystal-lang


thypon:
http://crystal-lang.org/

A native ruby-like language.

thypon:
crytstal-lang requires crystal-lang compiler to compile itself: someone knows how to solve?

beefcurtains:
With just the information from that site:
* Use the Ruby compiler with the latest version that worked with it
* Use that Crystal compiler to build crystal 0.1.0
* Use that Crystal compiler to build crystal 0.2.0
...and so on.

thypon:
A little overkill.

teh-jazzman:
As far as I can tell at a glance [from the documentation](http://crystal-lang.org/docs/installation/from_source_repository.html) they want others to at least use the compiler from their binary releases to build the sources but they only provide it for x86_64. It is alpha software, so that may change in the future.

beefcurtains:
Or wait for the gentoo ebuild and port that.

aeosynth:
bump. how are other self hosted languages built?

LVMBDV:
I had the same question before I even saw this issue. I asked it in the IRC channel and it seems the correct approach would be to make the `crystal-bin` package first, then the source package `crystal`. I guess I will try my hand at it.

```
08:32 <lvmbdv> i have a question, it says "Compiled: The software needs to be compiled before being used, even if it is software that is not needed by the whole system."
08:33 <lvmbdv> what if a package needs a binary of itself to be compiled? (like language compilers written in their own language e.g. crystal)
08:34 <maldridge> generally you have thing-bin and thing
08:34 <maldridge> yarn and yarn-bin are examples of this
08:34 <lvmbdv> oh i see
08:34 <lvmbdv> -bin doesn't have to be statically compiled, right?
08:34 <maldridge> -bin just has to work
08:35 <maldridge> the only thing that's permitted to depend on it is the build of the native one
08:35 <lvmbdv> that's how i should do it then
08:36 <lvmbdv> first, make the package crystal-bin then make the package crystal which depends on crystal-bin
08:36 <maldridge> yes, that would be the ideal
08:36 <maldridge> you could also look at what go does if that makes sense
08:37 <lvmbdv> alright, thanks maldridge 
```

Vaelatern:
[...]

PR #9831, opened 2017-12-07
New package: crystal-bin-0.24.0

LVMBDV:
This precompiled binary is meant to be used to build Crystal from source (requested #1781) as it requires a binary of itself to compile. The script workaround was needed because the precompiled binary doesn't look for the standard library in the right places.


LVMBDV:
https://travis-ci.org/voidlinux/void-packages/jobs/313157340
Sorry, missed this one since I was working only on x86_64. I don't think there's a clean solution since files in distfiles are meant to be arch-agnostic source files. Maybe choose only one arch?

aeosynth:
one issue with crystal on void is libressl support: https://github.com/crystal-lang/crystal/issues/4676. not sure if you want to handle that somehow.

LVMBDV:
Yeah, that seems like a problem. Let's hold off until that issue is resolved/closed. I want crystal on void but I don't want it _that_ bad.

aeosynth:
i mean that issue deals with feature detection, which the void package doesn't have to worry about; a few `sed` calls should be enough. but i haven't looked deeply into it.

LVMBDV:
So I looked into it and the solution doesn't seem so complicated after all :)
Since Void Linux supports only  LibreSSL and LibreSSL currently supports only the 1.0.1 API, we could overwrite the checks for the 1.0.2 and 1.1.0 APIs to return false. Or we could add a pkg-config expression that assumes the library is LibreSSL if libssl version is greater than 2.0.0.

I prefer the first change as it would require us to update the `sed` line only if LibreSSL API compatibility changes which doesn't seem to happen very often.

Also, should I add `pkg-config` to `depends` even though it is not listed as [required](https://github.com/crystal-lang/crystal/wiki/All-required-libraries)?

Vaelatern:
+1 for the overwrite checks, maybe write a pkg-config that checks for `libtls` (which is a libressl only library). Don't bother with adding `pkg-config` to depends unless it's really required at runtime.

LVMBDV:
Added a check for `libtls` as @Vaelatern suggested. CI will fail due to xlint misinterpreting backticks in the sed replacement string as "command" backticks.

PR #9834, opened 2017-12-07
libinput: Update hardware database information

wesbarnett:
libinput put a file in `/usr/lib/udev/hwdb.d` so `udevadm hwdb --update` should be run.


Vaelatern:
When should this be run?

Vaelatern:
And a source for usage please.

cr6git:
I read that somewhere... not sure `post_install()` is the right place to run `udevadm` however. 
Maybe an `INSTALL` script is better? 
````
case "$ACTION" in
post)
	udevadm hwdb --update
	;;
esac
````

**Comment by wesbarnett** [comment]
@Vaelatern `man 8 udevadm` documents this usage and when it should be run. it states:

```
   udevadm hwdb [options]
       Maintain the hardware database index in /etc/udev/hwdb.bin.

       -u, --update
           Compile the hardware database information located in
           /usr/lib/udev/hwdb.d/, /etc/udev/hwdb.d/ and store it in
           /etc/udev/hwdb.bin. This should be done after any update to the
           source files; it will not be called automatically. The running udev
           daemon will detect a new database on its own and does not need to
           be notified about it.
```

Since `libinput` places `/usr/lib/udev/hwdb.d/90-libinput-model-quirks.hwdb`, it should be run.

wesbarnett:
@cr6git Good suggestion. I can update the pull request for that.

However is there a way to have a generic hook/trigger such that anytime any package updates, adds, or removes a file from `/usr/lib/udev/hwbd.d/` that `udevadm hwdb --update` is run?

Vaelatern:
The `post)` action will do what you want there, @wesbarnett.

wesbarnett:
But this would only run `udevadm hwdb --update` if libinput is updates (other packages that add files would need to run this as well). I was thinking more along the lines of [something like this](https://github.com/wesbarnett/void-packages/commit/d13cefc9658772942efc6e4f9bb90aed720c5b84). I'm not sure if I'm on the right track with that.

xtraeme:
A new trigger for `xbps-triggers` could be created that execs `udevadm hwdb --root=. -u`.

Why with `--root` because this needs to work when rootdir != /

wesbarnett:
Closing this in favor of #9858 .

PR #9854, opened 2017-12-08
glew: update to 2.1.0

SolitudeSF:
im not sure if this is a correct way to do this


newbluemoon:
glew has just been updated, see https://github.com/voidlinux/void-packages/commit/74c5cd982ceea2e27201dfafa52ee22d2850a8fe

PR #9860, opened 2017-12-08
New package: mimic

kojicomics:


kojicomics:
The checks fail for aarch64 and arm when trying to detect the ICU installation, but icu-devel is defined in the hostdepends, and it installs succesfully.

Any idea?

newbluemoon:
You have to move `icu-devel` to `makedepends`. But also add `alsa-lib-devel` to `makedepends` and drop `depends`. Else `configure` won’t detect an audio driver and the whole package is built without sound output (only writing to disk). Xbps-src makes the package depend on linked libraries automatically.

So it should look like this:

    makedepends="alsa-lib-devel icu-devel"

newbluemoon:
I think you should also create a -devel subpackage.

kojicomics:
Thank you!
I'll make the changes and take a look at the wiki about making devel subpackages.

newbluemoon:
When you make changes to your template use `git commit --amend` or squash additional commits into one so that in the PR there’s just one commit per affected package.

Regarding the -devel package a good starting point is also to have a look at the various examples in `srcpkgs`. In this case here I would move `usr/include` and `usr/lib` (contains only static libraries and pkg-config data) to mimic-devel. :)

kojicomics:
Thank you @newbluemoon 
Finally this is ready to merge :)

PR #9866, opened 2017-12-09
cura: update to 3.1.0.

maxice8:


karl-nilsson:
Almost everything looks good, except:
- Uranium needs to be updated
- common/shlibs needs to point to the new libArcus
- in cura-engine, the R_X86_64_32 relocations may have been fixed by these commits: 
https://github.com/Ultimaker/CuraEngine/commit/0bdd2e5ac92d1788858bc986e80979fb39e1e835
https://github.com/Ultimaker/CuraEngine/commit/2d0c284e4a7f1d202ef8d6cd8794a314e533f162

I'm not entirely sure about the last point, maybe one of the devs can chime in?

maxice8:
> Uranium needs to be updated

I bumped by mistake alone and this PR is actually to finish bumping what i started.

536ce58b61da113e95ce22be4d67131db218382a

> common/shlibs needs to point to the new libArcus

Found no such new lib soname, the symlink remained so.3, you mean updating the version pointer to 3.1.0_1 ?

> in cura-engine, the R_X86_64_32 relocations may have been fixed by these commits:
Ultimaker/CuraEngine@0bdd2e5
Ultimaker/CuraEngine@2d0c284

Didn't found this problem during Build and neither Travis, is this a runtime error ?

karl-nilsson:
> Found no such new lib soname, the symlink remained so.3, you mean updating the version pointer to 3.1.0_1 ?
Yes

> Didn't found this problem during Build and neither Travis, is this a runtime error ?
What I was trying to say is you may be able to get rid of the pre_configure section of cura-engine. does that make sense?

maxice8:
ok, updated the pointer in common/shlibs, testing the cura-engine thing

pullmoll:
Any idea what [could be the cause for this error](https://build.voidlinux.eu/builders/x86_64_builder/builds/6532/steps/shell_3/logs/stdio) on the builders?
It built fine for x86_64 locally.

maxice8:
Uranium looks for the cmake version and installs accordingly

if i rebuild `Uranium` with `cmake-3.10.x`

```
/usr/share/cmake-3.10/Modules/UraniumTranslationTools.cmake
```

the original package has

```
/usr/share/cmake-3.9/Modules/UraniumTranslationTools.cmake
```

pullmoll:
It seems I did it wrong: moved /usr/share/cmake-3.10/Modules to /usr/share/cmake/Modules
I should have moved to /usr/share/cmake, right?

pullmoll:
Nope, still doesn't work. So I guess it's best to leave UraniumTranslationTools.cmak where it is.

pullmoll:
Not my day... need a break, but it's solved :)

PR #9876, opened 2017-12-09
kdevelop: update to 5.2.1

yopito:
including addon packages for python and php languages support.


yopito:
build and runtime are ok on my side, but too much for Travis CI: 50+ min of build is needed.
```
The job exceeded the maximum time limit for jobs, and has been terminated.
```

PR #9884, opened 2017-12-09
linux4.4: update to 4.4.105.

cr6git:


cr6git:
4.4.106 released today.

PR #9897, opened 2017-12-10
clang: add the clang-c headers

cyril-colin-2:
Those are needed to install the irony completion server, I’m not sure there is a reason to leave them out.


chneukirchen:
Merging this into the llvm5 branch #7612

PR #9906, opened 2017-12-11
New package: bootchart2-0.14.8

Coding-Doctors:
The last release for this was in 2015, however there have been a few bugfixes to land on master, should I port those over with a patch or should they be ignored.


pullmoll:
Instead of downloading the release, you could download the latest commit and modify the version to be like: `version=0.14.8.20170531` (i.e. append the git commit date)

Specify the git commit's revision like `_gitrev=331ada031f1d65f6d934d918f896e1c708c64bf7` in your template. The URL to put into distfiles would then read:
`https://github.com/xrmx/bootchart/archive/${_gitrev}.tar.gz>${pkgname}-${version}.tar.gz`

The wrksrc usually becomes:
`wrksrc="${pkgname}-${_gitrev}"`

srcpkgs/bootchart2/template
```diff
@@ -0,0 +1,16 @@
...
+makedepends="python"
+depends="pygtk git busybox"
+short_desc="Startup graphing tool"
+maintainer="Benjamin Hoffmeyer <hoffmeyer25@gmail.com>"
+license="GPL-2"
+homepage="https://github.com/xrmx/bootchart"
+_gitrev=331ada031f1d65f6d934d918f896e1c708c64bf7
+distfiles="https://github.com/xrmx/bootchart/archive/${_gitrev}.tar.gz>${pkgname}-${version}.tar.gz"
+checksum=a8140cc690bd7d08ab9c030325cb075e154b7545083036ffb65bda8d740d85b4
+wrksrc="bootchart-${_gitrev}"
```
Coding-Doctors: The downloaded archive is bootchart, so this is correct.

Coding-Doctors:
Alright updated with those changes, everything should be good! 👍

Vaelatern:
Why is it bootchart2 when upstream seems to call it bootchart?

Coding-Doctors:
Its called bootchart2 upstream, the git repository is named bootchart which is why the wrksrc is different.

Issue #9811, opened 2017-12-06
docker-17.11.0_1 problems with cgroups


iaroki:
### System

* xuname:  
Void 4.13.15_1 x86_64 AuthenticAMD notuptodate hold rrFF
* package:  
docker-17.11.0_1

### Expected behavior
Docker should create and run containers from images
### Actual behavior
Docker daemon starts ok but `docker run` command return error:

```
root:/ >>> docker run -it --rm ubuntu:16.04 bash
docker: Error response from daemon: cgroups: cannot found cgroup mount destination: unknown.
```

### Steps to reproduce the behavior
* Install latest docker-17.11.0_1
* Start docker daemon `sv start docker`
* Run container `docker run -it alpine sh`

### PS

On Void foum users suggested disable and remove `cgmanager` service symlink and this actually solve the problem. But if you do this LXC containers refuse to start and freeze at startup:

```
root:/ >>> lxc-start -n jira -F
Failed to mount cgroup at /sys/fs/cgroup/systemd: Operation not permitted
systemd 219 running in system mode. (+PAM +AUDIT +SELINUX +IMA -APPARMOR +SMACK +SYSVINIT +UTMP +LIBCRYPTSETUP +GCRYPT +GNUTLS +ACL +XZ -LZ4 -SECCOMP +BLKID +ELFUTILS +KMOD +IDN)
Detected virtualization lxc.
Detected architecture x86-64.

Welcome to CentOS Linux 7 (Core)!

Cannot determine cgroup we are running in: No such file or directory
Failed to allocate manager object: No such file or directory
[!!!!!!] Failed to allocate manager object, freezing.
```

I had to downgrade and hold old __docker-17.09.0_1__ which is perfectly work.
So new __docker-17.11.0_1__ can't work with running __cgmanager__ and __LXC__ can't work without running __cgmanager__.

hipperson0:
When using LXC without `cgmanager` the `/sys/fs/cgroup/systemd` is needed, see this comment on how to create it: https://github.com/voidlinux/void-packages/issues/4018#issuecomment-208353453

iaroki:
This allow me to run docker and lxc without cgmanager, thanks

```
mkdir /sys/fs/cgroup/systemd
mount -t cgroup -o none,name=systemd cgroup /sys/fs/cgroup/systemd
```

Service `run` cgmanager file contain:

```
#!/bin/sh
sv check dbus >/dev/null || exit 1
[...]

PR #9915, opened 2017-12-11
New package: lxd-2.20

iaroki:


Anachron:
Build will fail on aarch64 as he tested locally, we will have to investigate why.

newbluemoon:
Move acl-devel from `hostmakedepends` to `makedepends`?

iaroki:
Can't understand the difference between `hostmakedepends` and `makedepends`

newbluemoon:
The difference is when cross compiling: `hostmakedepends` contains tools which run on the host that builds the package and `makedepends` contains packages against which is linked (in the targat binary format), in many/most cases the -devel packages.

newbluemoon:
In more detail: You have

```
hostmakedepends="git bzr lxc-devel pkg-config acl-devel"
makedepends="lxc-devel pkg-config acl-devel"
depends="liblxc acl acl-progs rsync squashfs-tools xz"
```

`hostmakedepends` should contain `git`, `bzr`, and `pkg-config` because those packages are only needed to build lxd and run on the host that builds the package. They are in the build host binary format.

`makedepends` should contain `lxc-devel` and `acl-devel` because those packages provide the includes and libraries needed to compile and link lxd. They are in the target binary format.

When creating the final package xbps-src automatically adds the packages that contain the linked against libraries to `depends`, no need to specify them manually. 

Therefore `depends` should only contain packages which are needed at runtime, but cannot be detected when creating the package. Here this would be `acl-progs`, `rsync`, `squashfs-tools`, and `xz` (I didn’t check if they’re essential, just took them as given).

So that as a result you’d have

```
hostmakedepends="git bzr pkg-config"
makedepends="lxc-devel acl-devel"
depends="acl-progs rsync squashfs-tools xz"
```

iaroki:
At the moment to get it working a few steps needed:
* install `lxd` package
* add mappings `echo "root:1000000:65536" | sudo tee -a /etc/subuid /etc/subgid`
* start runit service `ln -s /etc/sv/lxd /var/service`
* run `lxd init` to configure
* add himself to `lxd` group `usermod -aG lxd $USER`

LXD depends on `lxc` package to start containers. And for now to start container with systemd init inside you need `/sys/fs/cgroup/systemd` mounted on the host machine. Here #9811  suggested create and mount manually:

```
mkdir /sys/fs/cgroup/systemd
mount -t cgroup -o none,name=systemd cgroup /sys/fs/cgroup/systemd
```

The problem is this doesn't survive reboot and i don't know how to put all this pieces together.

iaroki:
Added creating and mounting `/sys/fs/cgroup/systemd` to `run` file and mapping/user info to `INSTALL.msg`

Anachron:
Hey, why don't you put the mount command from here
https://github.com/voidlinux/void-packages/pull/9915/commits/8d49edc9507b0272d62786ec5233c1e3d3907175#diff-ef330111a87ff3ce6846733e7fca9657R3
as command on here?
https://github.com/voidlinux/void-packages/pull/9915/commits/8d49edc9507b0272d62786ec5233c1e3d3907175#diff-ef330111a87ff3ce6846733e7fca9657R10

There is no need to do it inside the ifs.

iaroki:
Right, thanks!

PR #9917, opened 2017-12-11
dkms: add libelf to dependencies

jnbr:
On linux4.14 building dkms modules (at least wireguard and exfat-dkms) fails:

```
[*] Configuring unpacked packages
wireguard-0.0.20171211_1: configuring ...
Added DKMS module 'wireguard-0.0.20171211'.
Building DKMS module 'wireguard-0.0.20171211' for kernel-4.14.5_1... FAILED!
DKMS module 'wireguard-0.0.20171211' failed to build, please check /var/lib/dkms
for errors in the log file.
wireguard-0.0.20171211_1: installed successfully.
```

trying to build the module manually:
```
# pwd
/var/lib/dkms/wireguard/0.0.20171127/build
# make
Makefile:938: *** "Cannot generate ORC metadata for CONFIG_ORC_UNWINDER=y, please install libelf-dev, libelf-devel or elfutils-libelf-devel".  Stop.
make: *** [Makefile:32: module] Error 2
```

after installing `libelf`, dkms modules can be built. So I think it should be a dependency of `dkms`.


chneukirchen:
Doesn't it need libelf-devel?

jnbr:
there is no libelf-devel. In case you mean elfutils-devel, no. I don't have it installed, libelf makes the difference.

jnbr:
If it makes more sense, I don't mind changing it to elfutils or elfutils-devel. Reason to stick with libelf was just to keep it as small as possible.

chneukirchen:
Ok, the actual issue is `/usr/src/kernel-headers-4.14.4_1/tools/objtool/objtool` depends on libelf but it's not detected.

jnbr:
Thank you for tracking this down. I guess this should be fixed in the linux4.14 template then.

chneukirchen:
Fixed with 9d4c9dcbf3.

PR #9922, opened 2017-12-11
protobuf: update to 3.5.1.

maxice8:


pullmoll:
Merging this will, again, need attention because of qt5 which stresses our builders.

Issue #6386, opened 2017-05-13
wpa_supplicant should depend on crda package


amak79:
I recently installed Void on a Raspberry Pi Zero W. Everything went fine except that wpa_supplicant (and `iw`) couldn't set the WiFi regulatory domain. Eventually I discovered that the crda package wasn't installed. Shouldn't the wpa_supplicant package have a runtime dependency on crda? Or is it expected that users install the crda package manually?

the-maldridge:
This is an interesting issue you bring up.  I think most of our devs assert that the regulatory domain is already set.  The raspberry pi of course does not seem to come with the domain already set as laptop wifi cards seem to.  I think the easiest way to solve this is to add a note to the wiki that setting the domain will require the `crda` package as you describe.

As a side note the `depends` field in the template is meant for things that absolutely must be present or a package won't work (interpreters for scripts, data packages, etc).

cr6git:
Hello @the-maldridge , during my holidays I took my HP laptop to continue to study Void (it is difficult to stop.) 
I use the `wpa_supplicant` hook with `dhcpcd` to connect to Wi-Fi network. 
This morning, after reading this issue, I noticed this thing: 

````
$ sudo dmesg | grep CRDA
[    6.843451] cfg80211: Calling CRDA to update world regulatory domain
[    9.988054] cfg80211: Calling CRDA to update world regulatory domain
[   13.139053] cfg80211: Calling CRDA to update world regulatory domain
[   16.290060] cfg80211: Calling CRDA to update world regulatory domain
[   19.441095] cfg80211: Calling CRDA to update world regulatory domain
[   22.592118] cfg80211: Calling CRDA to update world regulatory domain
[   25.743146] cfg80211: Calling CRDA to update world regulatory domain
[   28.894160] cfg80211: Calling CRDA to update world regulatory domain
[   32.045188] cfg80211: Calling CRDA to update world regulatory domain
[   35.196216] cfg80211: Calling CRDA to update world regulatory domain
[   38.347241] cfg80211: Calling CRDA to update world regulatory domain
[   41.498273] cfg80211: Exceeded CRDA call max attempts. Not calling CRDA
````

Well, as you can see, the regulatory domain **is not already set.** 

To fix this problem, I installed the `crda` package, then I modified the config file to have: 

````
$ cat /etc/wireless-regdom.conf
WIRELESS_REGDOM="FR"
````

And now the problem is fixed: 

````
$ sudo dmesg | grep CRDA
[...]

PR #9930, opened 2017-12-12
linux4.1: update to 4.1.47.

cr6git:
`crda` is required [#6386](https://github.com/voidlinux/void-packages/issues/6386).


cr6git:
4.1.48 released today.

PR #9933, opened 2017-12-12
evolution: update to 3.26.3.

maxice8:


pullmoll:
This didn't work. Evolution-3.26.3 can't be installed due to missing solibs. I'm currently trying to revert to the previous version, which seems to be troublesome.

pullmoll:
@Gottox it seems that merging evolution-ews was a bad idea, or more specificially adding the common/shlibs lines. I'm going to try to build evolution 3.26.3 locally again (after having removed the shlibs lines) and upate again if it works out.

PR #9942, opened 2017-12-12
dracut: add grep to the network module dependencies

jnbr:
since version 046 dracut's network module makes use of grep but does not install it.

Same patch is accepted upstream: https://github.com/dracutdevs/dracut/commit/69acd20d4e067e51cf60e509bea96a680c7bfff7


Gottox:
Is this already reported upstream?

jnbr:
Yes, they merged the same patch: https://github.com/dracutdevs/dracut/commit/69acd20d4e067e51cf60e509bea96a680c7bfff7
Or do you mean something else?

PR #9962, opened 2017-12-13
New package: nvme-cli-1.4

phf:
It's my first attempt at putting a package in, please feel free to tell me all the things that are wrong with this.


srcpkgs/nvme-cli/template
```diff
@@ -0,0 +1,12 @@
...
+pkgname=nvme-cli
+version=1.4
+revision=1
+build_style=gnu-makefile
+make_install_args="SBINDIR=/usr/bin"
+short_desc="NVMe management command line interface"
+maintainer="Peter Froehlich <peter.hans.froehlich@gmail.com>"
+license="GPL-2"
+homepage="https://github.com/linux-nvme/nvme-cli"
+distfiles="https://github.com/linux-nvme/nvme-cli/archive/v1.4.tar.gz"
```
pullmoll: Use `.../v${version}.tar.gz` here.

cr6git: `homepage="http://nvmexpress.org"`

phf:
Thanks for the feedback! (I have to admit that the "push -f" instructions freaked me out a little, but then again the detailed history of these is maybe not very important.)

phf:
It does seem that the "push -f" breaks how github tracks this PR. @pullmoll I wonder if you need to confirm that the changes I made are what you wanted? (Sorry, not too familiar with this process.)

cr6git:
@phf No problem with your `git push -f` , and the changes you made are correct.

srcpkgs/nvme-cli/template
phf: Don't want to go on your nerves, but those are fixed. Any other reason holding back the merge? I am happy to work on it!

PR #9968, opened 2017-12-13
fcitx: fix build with musl

yshui:
These changes are also submitted upstream for inclusion.


cr6git:
Version 4.2.9.1 is **outdated**, see [#9681](https://github.com/voidlinux/void-packages/issues/9681).

yshui:
@cr6git I will rebase after that is merged

PR #9971, opened 2017-12-13
godot: fix musl build and set nocross

Johnnynator:


pullmoll:
Already done in https://github.com/voidlinux/void-packages/commit/ddacb53abad4a53f126fb0dc31587adfea90c6a2

PR #9977, opened 2017-12-13
New package: sudoku-1.0.5

codingHahn:


codingHahn:
How can I prevent the package from automatically detecting glibc as an dependency>

cr6git:
Sudoku solver? lol That's cheating!! 
Looking at your build errors on Travis, it seems like you need a patch for musl.

newbluemoon:
I think it would be OK to set musl broken for now, e.g.

```
        *-musl) broken="https://travis-ci.org/voidlinux/void-packages/jobs/316059557"
esac
```

PR #9979, opened 2017-12-13
gifski: update to 0.7.0.

maxice8:


cr6git:
@maxice8 You are too fast, you git-pushed before me! Thank you. 😉

Issue #7185, opened 2017-07-29
grantlee5-devel: conflicts with grantlee-devel


ebfe:
```
grantlee5-devel-5.1.0_1.x86_64 <-> grantlee-devel-0.5.1_1.x86_64:
	/usr/include/grantlee/abstractlocalizer.h
	/usr/include/grantlee/abstractmarkupbuilder.h
	/usr/include/grantlee/bbcodebuilder.h
	/usr/include/grantlee/cachingloaderdecorator.h
	/usr/include/grantlee/context.h
	/usr/include/grantlee/engine.h
	/usr/include/grantlee/exception.h
	/usr/include/grantlee/filter.h
	/usr/include/grantlee/filterexpression.h
	/usr/include/grantlee/grantlee_version.h
	/usr/include/grantlee/markupdirector.h
	/usr/include/grantlee/mediawikimarkupbuilder.h
	/usr/include/grantlee/metatype.h
	/usr/include/grantlee/node.h
	/usr/include/grantlee/outputstream.h
	/usr/include/grantlee/parser.h
	/usr/include/grantlee/plaintextmarkupbuilder.h
	/usr/include/grantlee/qtlocalizer.h
	/usr/include/grantlee/rendercontext.h
	/usr/include/grantlee/safestring.h
	/usr/include/grantlee/taglibraryinterface.h
	/usr/include/grantlee/template.h
	/usr/include/grantlee/templateloader.h
	/usr/include/grantlee/texthtmlbuilder.h
	/usr/include/grantlee/token.h
	/usr/include/grantlee/typeaccessor.h
	/usr/include/grantlee/util.h
	/usr/include/grantlee/variable.h
	/usr/include/grantlee_templates.h
	/usr/include/grantlee_textdocument.h
```

PR #9987, opened 2017-12-14
grantlee-devel: avoid conflict with grantlee5-devel

cr6git:
resolves [#7185](https://github.com/voidlinux/void-packages/issues/7185), I found this trick on the AUR [PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=grantlee-qt4). 
However the license is `LGPL-2.1`, see [COPYING.LIB](https://github.com/steveire/grantlee/blob/master/COPYING.LIB).


cr6git:
`aarch64` and `armv7hf` failed? oooh what a surprise! (just kidding)

Issue #6990, opened 2017-07-07
libedit-devel: conflicts with readline-devel


ebfe:
```
/usr/share/man/man3/history.3
```

PR #9990, opened 2017-12-14
libedit-devel: avoid conflict with readline-devel

cr6git:
resolves [#6990](https://github.com/voidlinux/void-packages/issues/6990) by removing the file `man3/history.3`. 
This is what Arch is doing, see [here](https://git.archlinux.org/svntogit/packages.git/commit/trunk?h=packages/libedit&id=2434c24dfed41a25a13f3cc5dfea1aea03d7fc9b).


Duncaen:
This is not right.

cr6git:
@Duncaen ok, I will look for another solution.

cr6git:
@Duncaen I just forgot to ask you **why** this change is not right? 

I ask the question because `man editline` works, even without `man3/history.3`, so I'm a bit lost...

Duncaen:
Because they are different man pages, but I guess its ok to just delete it, `history.3` in `libedit-devel` is just a symlink `/usr/share/man/man3/history.3 -> /usr/share/man/man3/editline.3`.

cr6git:
oh yes it's just a symlink (I hadn't noticed at first). I guess the change is ok, as it doesn't seem to break anything.

pullmoll:
I take @Duncaen 's comment for an okay to merge.

PR #9992, opened 2017-12-14
keepassxc: update to 2.2.4.

maxice8:


cr6git:
@maxice8 you can remove `qt5-devel` from `hostmakedepends` for cross build.

PR #10006, opened 2017-12-14
jack: update to 1.9.12.

maxice8:


srcpkgs/jack/template
```diff
@@ -1,6 +1,6 @@
 # Template file for 'jack'
 pkgname=jack
-version=1.9.10
+version=1.9.12
 revision=11
```
asergi: `revision` hasn't been reset to 1.

PR #10027, opened 2017-12-15
New package: aseprite-1.2.4

yshui:


yshui:
Just realized aseprite has a EULA :(

PR #10031, opened 2017-12-15
New package: cross-arm-none-eabi 7.2

teajay-fr:
This package replaces the arm-none-eabi-binutils and arm-none-eabi-newlib packages with a complete bare metal toolchain for arm processors.
Cross compilation for this package is not possible.


pullmoll:
I'm trying locally. The changes to `common/xbps-src/shutils/common.sh` however are unrelated as it seems (not used, not mentioned) and need to go in a separate PR.

teajay-fr:
The file common/xbps-src/shutils/common.sh was a left over I overlooked and wasn't meant to be kept. Thank you !

PR #10042, opened 2017-12-16
mediainfo: add build options for building only CLI/GUI

SolitudeSF:


srcpkgs/mediainfo/template
```diff
@@ -5,40 +5,47 @@ revision=2
...
 	done
 	vlicense ${wrksrc}/MediaInfo/License.html
 
-	vmkdir usr/share/icons/hicolor/scalable/apps
-	vinstall ${wrksrc}/MediaInfo/Source/Resource/Image/MediaInfo.svg 644 \
-		usr/share/icons/hicolor/scalable/apps mediainfo.svg
-	vinstall ${wrksrc}/MediaInfo/Source/Resource/Image/MediaInfo.png 644 \
-		usr/share/pixmaps mediainfo-gui.png
-	vinstall ${wrksrc}/$_gnudir/GUI/mediainfo-gui.desktop 644 usr/share/applications
+	if [ $(vopt_if GUI GUI false) = "GUI" ]; then
```
pullmoll: I think you can use `if [ "$build_option_GUI" ]; then` here.

srcpkgs/mediainfo/template
```diff
@@ -5,40 +5,47 @@ revision=2
 short_desc="Display technical and tag data for video and audio files"
 homepage="http://mediaarea.net/MediaInfo"
 maintainer="Georg Schabel <gescha@posteo.de>"
+build_options="CLI GUI"
+build_options_default="CLI GUI"
```
pullmoll: I think you should add
```
desc_option_CLI="Build CLI version"
desc_option_GUI="Build GUI version"
```

srcpkgs/mediainfo/template
SolitudeSF: i suspected, that there was sane alternative.

SolitudeSF:
done and done

srcpkgs/mediainfo/template
```diff
@@ -5,40 +5,49 @@ revision=2
...
+desc_option_CLI="Build CLI version"
+desc_option_GUI="Build GUI version"
 hostmakedepends="automake libtool pkg-config"
-makedepends="libmediainfo-devel zlib-devel wxWidgets-devel"
-depends="desktop-file-utils hicolor-icon-theme"
+makedepends="libmediainfo-devel zlib-devel $(vopt_if GUI wxWidgets-devel)"
+depends="$(vopt_if GUI 'desktop-file-utils hicolor-icon-theme')"
 create_wrksrc=yes
 _gnudir="MediaInfo/Project/GNU"
+_targets="$(vopt_if CLI) $(vopt_if GUI)"
```
Johnnynator: Did you ever check if this works properly? `$(vopt_if CLI)` does never return anything and therefore `$_targets` is always empty. #11062

SolitudeSF: well, im pretty sure i tested bunch of combinations of these, since CLI only version is installed now on my machine. idk

SolitudeSF: hmmm, i guess i messed up somewhere, dont know how i missed this

Johnnynator: No Problem, at least it didn't end in replacing any existing packages. I have fixed it in my PR.

SolitudeSF: yeah, thank you for noticing

PR #10047, opened 2017-12-16
New package: crystal-0.24.0

### LINKED ISSUE #1781: Package request: crystal-lang

LVMBDV:
This is the source package (fixes #1781) for Crystal, compiled with the precompiled binary from `crystal-bin`. Travis job for xlint will fail due to xlint misinterpreting backticks in the sed replacement string as "command" backticks.


Johnnynator:
You dependencies do look messed up. Your depends looks like a mix between makedepends and depends and makedepends looks like mostly hostmakedepends. Also if you pull in -devel you don't need to pull in the corresponding non devel package in the same dep type.

LVMBDV:
Alright, thanks. Will sort those out.

srcpkgs/crystal/template
```diff
@@ -0,0 +1,54 @@
+# Template file for 'crystal'
+pkgname=crystal
+version=0.24.0
+revision=1
+makedepends="crystal-bin libxml2 llvm git"
```
Vaelatern: `hostmakedepends` for these, I think.

srcpkgs/crystal/template
```diff
@@ -0,0 +1,54 @@
...
+short_desc="The Crystal Programming Language"
+maintainer="lvmbdv <ata.kuyumcu@protonmail.com>"
+license="Apache-2.0"
+homepage="https://crystal-lang.org/"
+distfiles="https://github.com/crystal-lang/crystal/archive/${version}.tar.gz"
+checksum=c34f4326ee576f60b8960e331aae8d5265a12b30707582e1f5c6c6032d364a80
+conflicts="crystal-bin"
+
+do_extract() {
+	tar xfz ${XBPS_SRCDISTDIR}/${pkgname}-${version}/${version}.tar.gz --strip-components=1 -C ${wrksrc}
```
Vaelatern: Is this really necessary?

LVMBDV: Apparently not. Sorry, I copied it directly from `crystal-bin` which I wrote looking at `yarn-bin`.

LVMBDV:
The build started out right without `do_extract` but I didn't see it through because it takes forever. Also I would like a bit of hand-holding on seperating the dependencies if they are still wrong :(

pullmoll:
Rule of thumb for the depends:
* `hostmakedepends` tools, compilers which run on the host architecture (usually x86_64* or i686*) and sometimes a `*-devel` package which is also in makedepends, if it otherwise isn't found by the compiler or also contains a tool which is for development.
* `makedepends` packages required by the package, usually only `*-devel` packages, except when there is none for a library.
* `checkdepends` tools or libraries required for running a package's tests (e.g. make check).
* `depends` required only, if the package needs tools or dynamically loads libraries which were not linked to it. The solibs which are found by inspecting the package's own solibs and binaries do not need to be listed here.

It would probably be inspiring for you to take a look at other package's templates to get the idea.

AFAICT without testing in this case they should probably be
```
hostmakedepends="crystal-bin git llvm"
makedepends="libatomic_ops-devel libxml2-devel gc-devel pcre-devel libevent-devel"
depends="llvm"
checkdepends="libyaml-devel libxml2-devel gmp-devel git"
```

LVMBDV:
Some of those `*-devel` packages are in depend because crystal links against those in a regular `*.cr` compilation. And the others are still used in Crystal API (like YAML). To test this, I compiled a simple program in Crystal and checked what libraries it was linked against.

```
atak@jet crystal-0.24.0-2>echo 'puts "Hello World!"' > hello.cr
atak@jet crystal-0.24.0-2>ldd hello
	linux-vdso.so.1 (0x00007fff4efa7000)
	libpcre.so.1 => /usr/lib/libpcre.so.1 (0x00007fbe418ec000)
	libgc.so.1 => /usr/lib/libgc.so.1 (0x00007fbe41679000)
	libpthread.so.0 => /usr/lib/libpthread.so.0 (0x00007fbe4145a000)
	libevent-2.1.so.6 => /usr/lib/libevent-2.1.so.6 (0x00007fbe41204000)
	librt.so.1 => /usr/lib/librt.so.1 (0x00007fbe40ffc000)
	libdl.so.2 => /usr/lib/libdl.so.2 (0x00007fbe40df8000)
	libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x00007fbe40be1000)
	libc.so.6 => /usr/lib/libc.so.6 (0x00007fbe4082c000)
	/lib/ld-linux-x86-64.so.2 (0x00007fbe41de9000)
	libatomic_ops.so.1 => /usr/lib/libatomic_ops.so.1 (0x00007fbe40629000)
```
Shall I move all `*-devel` packages to `makedepends` or keep the ones Crystal **always** uses?
Also, Crystal uses `cc` to compile things, should we add `gcc` or `clang` to `depends`?

LVMBDV:
[Here](https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/crystal) is how Arch dealt with it:

```
depends=(gc libatomic_ops pcre libevent llvm-libs)
makedepends=(libxml2 llvm)
checkdepends=(libyaml libxml2 gmp inetutils git)
optdepends=('shards: crystal language package manager'
            'libyaml: For YAML support'
            'gmp: For BigInt support'
            'libxml2: For XML support')
```

pullmoll:
Oh, I now understand more of the reason why you did include devel packages in depends.
As long as no configure options are there then all (optional) devel pkgs should probably all be in depends.
You can add gcc to depends, sure, even though most people using it will already have `base-devel` installed.

LVMBDV:
I'm trying to get the check phase working now. 2 failures so far. First one is about being able to touch `/bin/ls`. Is it because of the chroot?

```
it "raises if file cannot be accessed" do
	expect_raises Errno, "Operation not permitted" do
		File.touch("/bin/ls")
	end
end
```

And the second one, I have no idea why it would work in the first place.

```
it %(changes timezone with ENV["TZ"]) do
	old_tz = ENV["TZ"]?

	begin
		ENV["TZ"] = "America/New_York"
		offset1 = Time.local_offset_in_minutes

		ENV["TZ"] = "Europe/Berlin"
		offset2 = Time.local_offset_in_minutes

		offset1.should_not eq(offset2)
	ensure
		ENV["TZ"] = old_tz
	end
end
```

pullmoll:
I think do_check() can be fixed at any time. The chroot environment is probably different in several aspects from a *real* root, so no worries.

It does take fscking bloody ages to compile crystal, no? Sitting here and waiting for a result. The build seems to run single threaded, no parallelism whatsoever. I'll wait and see and eventually merge this PR.

pullmoll:
665b5231ca0bf3b622e98315fa7692c46667c46a

LVMBDV:
> I think do_check() can be fixed at any time.

Thanks, up to my neck in schoolwork at the moment.

> It does take fscking bloody ages to compile crystal, no? Sitting here and waiting for a result. The build seems to run single threaded, no parallelism whatsoever.

Yeah the crystal part of the build has no parallelism, takes long to build each time :(

Thanks for the `nocross` on the final commit, I tried my hand at armv7 and aarch64 builds a few days ago with no success.

pullmoll:
I think the problem is asking the host `llvm-config --cxxflags` for the flags to use. Then you get e.g. `-mtune=generic` for x86_64 which of course doesn't work for e.g. cross aarch64.

You'd have to replace this call with the (fixed?) flags which the llvm-config for the target architecture would return, or perhaps just ignore them and see if the default CXXFLAGS work alone.

PR #10048, opened 2017-12-16
sudo: fix flags to enable insults, add license

cr6git:
The flag `-with-all-insults` was obviously not correctly set to enable insults. 
````
  --with-insults          insult the user for entering an incorrect password
  --with-all-insults      include all the sudo insult sets
````


cr6git:
Travis is complaining about licenses: 
````
srcpkgs/sudo/template:17: license 'BSD', but no use of vlicense
srcpkgs/sudo/template:17: license 'ISC', but no use of vlicense
````

Coding-Doctors:
Have you tested it to need ``--with-insults``? do the insults work without that flag?

cr6git:
Yes. 
Without the flag `--with-insults` during the `do_configure` stage you will see: 
````
checking whether to include insults... no
````

Coding-Doctors:
Thanks for confirming! 👍

PR #10056, opened 2017-12-17
python-irc: update to 16.2.

maxice8:


Vaelatern:
Ah, can you rework this and let me know in IRC when you have? Your changes are good.

PR #10068, opened 2017-12-17
tcc: update to 0.9.27.

copyleftie:
update tcc to v0.9.27

the git snapshot is release tagged, so this should be compliant with Void's packaging policy


copyleftie:
please ignore this request, clearly I don't know how to GitHub

PR #10085, opened 2017-12-18
Add fcitx rime and its dependencies

yshui:


yshui:
x86_64-musl target depend on #9701

PR #10086, opened 2017-12-18
python-PyQt4: fix cross compilation (proposition)

newbluemoon:
The problem seems to be to use the right flags in the right place.

First I changed the template to use `QMAKE_CC=...` instead of `CC=...`, etc.,
because else the wrong compiler is used when `qmake` creates the makefiles.

The next problem is that when generating the makefiles for the applications
`pylupdate` and `pyrcc` above variables aren’t respected and the host compiler
is used instead. Because the `.pro` files are generated by `configure-ng.py` and
the flags cannot just be added via a `sed` line I modified `configure-ng.py` to
include a config file in the `.pro` file. This config file is created in the
template to set up the variables with respect to the target.

Then there is the `qtdetail` binary which is built and executed to gather
information about the installed Qt version. To compile it with the host compiler
I used the same ‘trick’ as above.

To find the right includes it was necessary to use the
`--sysroot=$XBPS_CROSS_BASE` option with `configure-ng.py`. But then the
`$XBPS_CROSS_TRIPLET` has to be removed from the installation path and in
`pyuic4`. `grep`ing for the triplet didn’t show any other locations where it
might be hiding.

I compared an armv7l-musl cross build with a native one: the file list is
identical and the package sizes are about the same. So I don’t think something’s
missing. Tested python-PyQt4 with puddletag and git-cola on a Raspberry Pi,
everything worked.

I don’t know if this is too much of a hack, tried to keep it as simple as
possible. :)


pullmoll:
I assigned this to @asergi as the template owner. To me it looks good enough and having python-PyQt4 will enable a number of packages depending on it to become cross compilable.

asergi:
#13203

PR #10087, opened 2017-12-18
python-PyQt5: fix cross compilation (proposition)

newbluemoon:
Similar to python-PyQt4 #10086.

For arm* and aarch64* (and probably mips*, too) some things are missing in Qt5:
  - There is no libQt5Designer.so.* in qt5-tools.
  - ‘QOpenGLFunctions’ isn’t present, build system test aborts with
    `cfgtest__QOpenGLFunctions_2_0.cpp:(.text.startup+0x10): undefined reference to 
    QOpenGLFunctions_2_0::QOpenGLFunctions_2_0()'`
    (this is also the reason why ‘shotcut’ fails to cross compile)

So I added a check to only package `_QOpenGLFunctions_2_0.so` (in python{,3}-PyQt5-opengl),
`QtDesigner.so` (in python{,3}-PyQt5-tools), and `libpyqt5.so` (python3-PyQt5-tools) if they were built.

But I couldn’t find out why those libs are missing in Qt5 and cannot find any hint in the template.
Could it be that an additional flag needs to be set?


pullmoll:
The problem with libQt5Designer.so.* is due to how qt5 checks for the availability of a module (widgets) to tell if it should build Qt5Designer or not. In a cross environment this detection fails.

It should be patched out and the module unconditionally built, because we know we have widgets (and png).

I tried to achieve that in [qt5-5.9.3 here](https://github.com/pullmoll/void-packages/blob/8210cb28d263e74f5fc488c721af36dca632d183/srcpkgs/qt5/patches/fix-cross-assitant-pixeltool-designer.patch)

newbluemoon:
Thanks for the info on Qt5Designer!

I think I figured out the issue with QOpenGLFunctions_2_0: because `qtdetail` is built for the host it checks the host’s opengl features! So now I’m trying to manually set it to use QOpenGLFunctions_ES2 when building for arm*/aarch64* ...

Can you think of any other features that might differ between platforms and would get detected wrong?

newbluemoon:
I manually disabled ‘Desktop OpenGL’ when cross building for arm*/aarch64*. However, detection of QOpenGLFunctions_ES2 fails:

```
Checking to see if the _QOpenGLFunctions_ES2 module should be built...
/usr/bin/qmake -o cfgtest__QOpenGLFunctions_ES2.mk cfgtest__QOpenGLFunctions_ES2.pro
make -f cfgtest__QOpenGLFunctions_ES2.mk
armv7l-linux-musleabihf-c++ -c -D_FORTIFY_SOURCE=2 -O2 -pipe -march=armv7-a -mfpu=vfpv3 -mfloat-abi=hard -I/usr/armv7l-linux-musleabihf/usr/include -O2 -Wall -W -D_REENTRANT -fPIC -DQT_NO_DEBUG -DQT_G
UI_LIB -DQT_CORE_LIB -I. -I /usr/include/qt5 -I /usr/include/qt5/QtGui -I /usr/include/qt5/QtCore -I. -I /usr/include/libdrm -I/usr/lib/qt5/mkspecs/linux-g++ -o cfgtest__QOpenGLFunctions_ES2.o cfgtest
__QOpenGLFunctions_ES2.cpp
cfgtest__QOpenGLFunctions_ES2.cpp: In function 'int main(int, char**)':
cfgtest__QOpenGLFunctions_ES2.cpp:5:9: error: expected type-specifier before 'QOpenGLFunctions_ES2'
     new QOpenGLFunctions_ES2();
         ^~~~~~~~~~~~~~~~~~~~
make: *** [cfgtest__QOpenGLFunctions_ES2.mk:524: cfgtest__QOpenGLFunctions_ES2.o] Error 1
```

pullmoll:
> Can you think of any other features that might differ between platforms and would get detected wrong?

Not from the top of my head, no. It would require a file-by-file comparison of the 50+ qt5 packages' contents to see if anything else is missing or extra.

newbluemoon:
I tried to build python-PyQt5 natively to compare the file list with the one of the cross build, but I get:

```
=> python-PyQt5-5.8.2_2: running do_build ...
Querying qmake about your Qt installation...
Error: Failed to determine the detail of your Qt installation. Try again using
the --verbose flag to see more detail about the problem.
Determining the details of your Qt installation...
/usr/bin/qmake -o qtdetail.mk qtdetail.pro
Could not find qmake configuration file devices/void-armv7l-linux-musleabihf-g++.
Error processing project file: qtdetail.pro
```

`qmake -query` returns
```
QT_SYSROOT:/usr/armv7l-linux-musleabihf
QT_INSTALL_PREFIX:/usr
QT_INSTALL_ARCHDATA:/usr/lib/qt5
QT_INSTALL_DATA:/usr/share/qt5
QT_INSTALL_DOCS:/usr/share/doc/qt5
QT_INSTALL_HEADERS:/usr/include/qt5
QT_INSTALL_LIBS:/usr/lib
QT_INSTALL_LIBEXECS:/usr/lib/qt5/libexec
QT_INSTALL_BINS:/usr/lib/qt5/bin
QT_INSTALL_TESTS:/usr/tests
QT_INSTALL_PLUGINS:/usr/lib/qt5/plugins
QT_INSTALL_IMPORTS:/usr/lib/qt5/imports
QT_INSTALL_QML:/usr/lib/qt5/qml
QT_INSTALL_TRANSLATIONS:/usr/share/qt5/translations
QT_INSTALL_CONFIGURATION:/etc/xdg
QT_INSTALL_EXAMPLES:/usr/share/qt5/examples
QT_INSTALL_DEMOS:/usr/share/qt5/examples
QT_HOST_PREFIX:/builddir/qt-everywhere-opensource-src-5.8.0/qtbase/host
QT_HOST_DATA:/builddir/qt-everywhere-opensource-src-5.8.0/qtbase/host
QT_HOST_BINS:/builddir/qt-everywhere-opensource-src-5.8.0/qtbase/host/bin
QT_HOST_LIBS:/builddir/qt-everywhere-opensource-src-5.8.0/qtbase/host/lib
QMAKE_SPEC:linux-g++
QMAKE_XSPEC:devices/void-armv7l-linux-musleabihf-g++
QMAKE_VERSION:3.1
QT_VERSION:5.8.0
```

And `/usr/lib/qt5/mkspecs/qconfig.pri` contains

```
!host_build {
    QMAKE_CFLAGS    += --sysroot=$$[QT_SYSROOT]
    QMAKE_CXXFLAGS  += --sysroot=$$[QT_SYSROOT]
    QMAKE_LFLAGS    += --sysroot=$$[QT_SYSROOT]
}
host_build {
    QT_ARCH = x86_64
    QT_TARGET_ARCH = arm
} else {
    QT_ARCH = arm
}
QT.global.enabled_features = shared rpath accessibility c++11 c++14 c++1z concurrent pkg-config
QT.global.disabled_features = framework debug_and_release simulator_and_device build_all force_asserts separate_debug_info static
PKG_CONFIG_SYSROOT_DIR = /usr/armv7l-linux-musleabihf
PKG_CONFIG_LIBDIR = /usr/armv7l-linux-musleabihf/usr/lib/pkgconfig:/usr/armv7l-linux-musleabihf/usr/share/pkgconfig:/usr/armv7l-linux-musleabihf/usr/lib/armv7l-linux-musleabihf/pkgconfig
QT_CONFIG += shared rpath accessibility release c++11 c++14 c++1z concurrent dbus mremap reduce_exports release_tools stl
CONFIG += shared cross_compile release
QT_VERSION = 5.8.0
QT_MAJOR_VERSION = 5
QT_MINOR_VERSION = 8
QT_PATCH_VERSION = 0
QT_EDITION = OpenSource
```

This seems to be the configuration for the host when cross compiling, but when trying to compile something natively using qmake this will apparently fail. So are native builds using a cross compiled Qt5 broken per default? I tried to manually remove all ‘cross stuff’ from the configuration files in `/usr/lib/qt5/mkspecs` but I can’t find a way to get rid of `QT_SYSROOT` or `QMAKE_XSPEC`.

newbluemoon:
> It would require a file-by-file comparison of the 50+ qt5 packages' contents to see if anything else is missing or extra.

Maybe it is enough to just check the features for which `qtdetail` is looking?

```
    // Defined in QtCore.
#if defined(QT_SHARED) || defined(QT_DLL)
    out << "shared\n";
#else
    out << "static\n";
#endif

    // Determine which features should be disabled.

    // Defined in ???.
#if defined(QT_NO_ACCESSIBILITY)
    out << "PyQt_Accessibility\n";
#endif

    // Defined in ???.
#if defined(QT_NO_SESSIONMANAGER)
    out << "PyQt_SessionManager\n";
#endif

    // Defined in ???.
#if defined(QT_NO_SSL)
    out << "PyQt_SSL\n";
#endif

    // Defined in QtPrintSupport.
#if defined(QT_NO_PRINTDIALOG)
    out << "PyQt_PrintDialog\n";
#endif

    // Defined in QtPrintSupport.
#if defined(QT_NO_PRINTER)
    out << "PyQt_Printer\n";
#endif

    // Defined in QtPrintSupport.
#if defined(QT_NO_PRINTPREVIEWDIALOG)
    out << "PyQt_PrintPreviewDialog\n";
#endif

    // Defined in QtPrintSupport.
#if defined(QT_NO_PRINTPREVIEWWIDGET)
    out << "PyQt_PrintPreviewWidget\n";
#endif

    // Defined in ???.
#if defined(QT_NO_RAWFONT)
    out << "PyQt_RawFont\n";
#endif

    // Defined in QtGui.
#if defined(QT_NO_OPENGL)
    out << "PyQt_OpenGL\n";
    out << "PyQt_Desktop_OpenGL\n";
#elif defined(QT_OPENGL_ES) || defined(QT_OPENGL_ES_2) || defined(QT_OPENGL_ES_3)
    out << "PyQt_Desktop_OpenGL\n";
#endif

#if QT_VERSION < 0x050200
// This is the test used in qglobal.h in Qt prior to v5.2.  In v5.2 and later
// qreal is double unless QT_COORD_TYPE is defined.
#if defined(QT_NO_FPU) || defined(Q_PROCESSOR_ARM) || defined(Q_OS_WINCE)
    out << "PyQt_qreal_double\n";
#endif
#else
    if (sizeof (qreal) != sizeof (double))
        out << "PyQt_qreal_double\n";
#endif

    // Defined in QtCore.
#if defined(QT_NO_PROCESS)
    out << "PyQt_Process\n";
#endif
```

pullmoll:
@newbluemoon To be honest I never really tested, if the cross compiled qmake works on native hardware. It needs to be built separately after building the project because the cross build process of course produces and uses a native (e.g. x86_64) qmake.

It would be senseless to leave this in the tarball and of course trying to do so will make xbps-src error out anyway. So qmake (and some other host binaries) have to be rebuilt for the target.

Since not many (to my knowledge) distributions do cross build qt5 like Void does, i.e. with a separate toolchain and no multiarch gcc/libs setup, all this stuff was deduced by trial and error.

It may well be that there is some way to create a *proper* qmake for the target, i.e. one which has no `QMAKE_XSPECS` set, which is indeed only used on the host. It's just that I do not know and could not find hints for that issue.

Since you seem to be the "cross hacker" around here, perhaps you want to delve into the depths of the qt5 template and find a way to build an actually usable qmake in the `do_install()` stage?

newbluemoon:
@pullmoll I just like to figure stuff out and while doing so fix things ;) This seems to be a bit bigger though, but I’ll give it a try. Can’t promise anything... :)

pullmoll:
/me hums "With A Little Help From My Friends" and tries to imitate Joe Cocker :)

I'm sure we can fix things in a joint effort. It's just that qt5 is probably the biggest single package and my usual trial-and-error approach is _very_ time consuming, especially if you need - or want - to test builds for 16 combinations of arch and libc...

newbluemoon:
qmake’s builtin paths can be overridden by creating a `qt.conf` in the same directory in which the qmake binary resides. `qmake -query` shows the wanted configuration but for some reason this is ignored when creating PyQt5’s makefiles. :(

```
% cat /usr/lib/qt5/bin/qt.conf 
[Paths]
Sysroot =
Prefix = /usr
ArchData = /usr/lib/qt5
Data = /usr/share/qt5
Documentation = /usr/share/doc/qt5
Headers = /usr/include/qt5
Libraries = /usr/lib
LibraryExecutables = /usr/lib/qt5/libexec
Binaries = /usr/lib/qt5/bin
Tests = /usr/tests
Plugins = /usr/lib/qt5/plugins
Imports = /usr/lib/qt5/imports
Qml2Imports = /usr/lib/qt5/qml
Translations = /usr/share/qt5/translations
Settings = /etc/xdg
Examples = /usr/share/qt5/examples
HostPrefix = /usr
HostData = /usr/lib/qt5
HostBinaries = /usr/lib/qt5/bin
HostHeaders = /usr/include/qt5
HostLibraries = /usr/lib
Spec = linux-g++
TargetSpec = linux-g++
```

newbluemoon:
> I'm sure we can fix things in a joint effort.

Sure we can! :) Maybe it’s possible without heavy recompilation of Qt5.

The `qt.conf` from above does work. It’s just that in the cross-built qmake package there are two module directories: `/usr/lib/qt5/mkspecs/modules` and `.../modules-inst`. qmake uses the former. The include files there setup the paths and then include the corresponding file from `modules-inst`. This  makes `qt.conf` useless.

For example:

`/usr/lib/qt5/mkspecs/modules/qt_lib_core.pri`:
```
QT_MODULE_BIN_BASE = /usr/lib/qt5/bin
QT_MODULE_INCLUDE_BASE = /usr/lib/qt5/include
QT_MODULE_IMPORT_BASE = /usr/lib/qt5/imports
QT_MODULE_QML_BASE = /usr/lib/qt5/qml
QT_MODULE_LIB_BASE = /usr/lib/qt5/lib
QT_MODULE_HOST_LIB_BASE = /usr/lib/qt5/lib
QT_MODULE_LIBEXEC_BASE = /usr/lib/qt5/libexec
QT_MODULE_PLUGIN_BASE = /usr/lib/qt5/plugins
include(/usr/lib/qt5/mkspecs/modules-inst/qt_lib_core.pri)
QT.core.priority = 1
include(/usr/lib/qt5/mkspecs/modules-inst/qt_lib_core_private.pri)
QT.core_private.priority = 1
```

`/usr/lib/qt5/mkspecs/modules-inst/qt_lib_core.pri` respects variables specified in `qt.conf`:
```
QT.core.VERSION = 5.8.0
QT.core.MAJOR_VERSION = 5
QT.core.MINOR_VERSION = 8
QT.core.PATCH_VERSION = 0
QT.core.name = QtCore
QT.core.module = Qt5Core
QT.core.libs = $$QT_MODULE_LIB_BASE
QT.core.includes = $$QT_MODULE_INCLUDE_BASE $$QT_MODULE_INCLUDE_BASE/QtCore
QT.core.frameworks =
QT.core.bins = $$QT_MODULE_BIN_BASE
QT.core.libexecs = $$QT_MODULE_LIBEXEC_BASE
QT.core.plugins = $$QT_MODULE_PLUGIN_BASE
QT.core.imports = $$QT_MODULE_IMPORT_BASE
QT.core.qml = $$QT_MODULE_QML_BASE
QT.core.depends =
QT.core.uses =
QT.core.module_config = v2
QT.core.CONFIG = moc resources
QT.core.DEFINES = QT_CORE_LIB
QT.core.enabled_features = properties animation textcodec big_codecs codecs commandlineparser textdate datestring filesystemiterator filesystemwatcher gestures itemmodel proxymodel identityproxymodel
library mimetype process statemachine qeventtransition regularexpression settings sharedmemory sortfilterproxymodel stringlistmodel systemsemaphore temporaryfile timezone topleveldomain translation xm
lstream xmlstreamreader xmlstreamwriter
QT.core.disabled_features =
QT_CONFIG += properties animation textcodec big_codecs clock-monotonic codecs textdate datestring doubleconversion eventfd filesystemiterator filesystemwatcher gestures glib itemmodel proxymodel ident
ityproxymodel inotify library mimetype process statemachine regularexpression settings sharedmemory sortfilterproxymodel stringlistmodel systemsemaphore temporaryfile threadsafe-cloexec translation xm
lstream xmlstreamreader xmlstreamwriter
QT_MODULES += core
```

As a workaround I just replaced `modules` with `modules-inst`.

Then the next problem arises: in all cross-built Qt5 sub-module devel-packages there is no `/usr/lib/qt5/mkspecs/modules/` with the sub-module project include files needed by qmake.
I think we have to figure out if these includes are there in a cross build in the first place. Because I don’t see `_install_subpkg()` or `_install_devel()` in the qt5 template do anything differently when cross compiling.

After removing some ‘cross stuff’ (like $XBPS_CROSS_BASE; still trying to find the minimum of changes needed) from `/usr/lib/qt5/mkspecs/{qconfig,qmodule}.pri` I was looking for a small package using qmake build style and no Qt sub-modules. Found one with ykpers-gui and, guess what, the package was successfully built natively! :)

So I think if we can figure out where the sub-module project includes are in a cross build it’s possible to use the cross built qmake natively with ‘just’ a couple of configuration file modifications.

pullmoll:
Thanks for the analysis. I now see in a build log of qt5-5.9.3 when cross compiling for aarch64 the *.pri files being installed to the `host` subdirectory instead of the `/usr/lib/qt5/mkspecs` directory. Here's an example:
```
/builddir/qt-everywhere-opensource-src-5.9.3/qtbase/bin/qmake -install qinstall /builddir/qt-everywhere-opensource-src-5.9.3/qtbase/mkspecs/modules-inst/qt_lib_core.pri /destdir/aarch64-linux-gnu/qt5-5.9.3/builddir/qt-everywhere-opensource-src-5.9.3/host/mkspecs/modules/qt_lib_core.pri
/builddir/qt-everywhere-opensource-src-5.9.3/qtbase/bin/qmake -install qinstall /builddir/qt-everywhere-opensource-src-5.9.3/qtbase/mkspecs/modules-inst/qt_lib_core_private.pri /destdir/aarch64-linux-gnu/qt5-5.9.3/builddir/qt-everywhere-opensource-src-5.9.3/host/mkspecs/modules/qt_lib_core_private.pri
```

This means that in the cross build case the `_install_devel()` function needs to move them from there to their final destination. Or as an alternative we would need to check, if setting a `-hostprefix host` in the configure options is required at all.

For qt5-5.9.3 I'll first try to move the files to the place where they belong and inspect their contents.

pullmoll:
Without setting `-hostdir host` the cross build fails for qt5-5.8.0 for armv7l-musl after having built qgltf from the qt3d submodule. This means I need to modify the `_install_devel()` function to move the `*.pri` files.

pullmoll:
For `qt5-5.9.2` leaving `-hostprefix` unset seems to work, i.e. `*.pri` files are installed where they belong. There also setting `-proprietary-codecs` works for cross builds (currently building for `aarch64`) so I fear we just can't add this to 5.8.0 already.

newbluemoon:
> I fear we just can't add this to 5.8.0 already.

If only that nasty issue with the builders was solved to finally be able to move to Qt 5.9!

newbluemoon:
I think I found the set of changes needed for `/usr/lib/qt5/mkspecs/{qconfig,qmodule}.pri`:

```
for f in qconfig qmodule; do
	sed -i -e "s|$XBPS_CROSS_BASE||g" \
	  -e "s/cross_compile//" \
	  -e "/^PKG_CONFIG_*/d" \
	  -e "s/^!\?host_build/0/" \
	  /usr/lib/qt5/mkspecs/$f.pri.test
done
```
- 1  remove cross base from path names
- 2  remove ‘cross_compile’ option from CONFIG
- 3  remove PKG_CONFIG_* as we use standard paths in a native build
- 4 what was the target when cross building is now the host;
  modify the condition to select the correct scope

I still have to test this for all platforms, though.

pullmoll:
I accidentially merged it before all tests for cross platforms were complete :)
I doubt it will cause problems but rather solve some of them...
Meanwhile trying to create a qt5-5.10.0 modularized build is giving me headaches.. but that's another story.

newbluemoon:
> I doubt it will cause problems

I think so, too. The problems above were due to trying to build it natively with the cross-built qt5 and are related to qt5 itself.
There are only a handful of packages that have python-PyQt5 in `*depends`. I’ll check if there are (new) problems and report back.

> trying to create a qt5-5.10.0 modularized build

If there’s anything I can do to help just let me know. :)

pullmoll:
qt5-5.10.0 very WIP: https://github.com/pullmoll/void-packages/tree/qt5.10
Only qt5-base is building to some extent. I currently get strange error messages from xbps-src after qt5-base is built. Something tries to chdir to a non-existing directory /builddir/_install/...

newbluemoon:
Cross-compiled and tested on a RPi3 (as far as I could make use of the application):
- python-jupyter_qtconsole
- hplip (built with -o gui)
- scudcloud (needs some packages added to `depends`; will do a PR shortly)

python-matplotlib depends on python-numpy which currently cannot be cross-compiled.

Calibre (which is `nocross`) needs some work done (if it’s possible at all; haven’t looked into it, yet).

`grep -r python-PyQt5` in `srcpkgs` didn’t reveal any other packages depending on PyQt5. So I’d be so keen and say it looks good. :)

newbluemoon:
> Something tries to chdir to a non-existing directory /builddir/_install/

In `qt5-base/template` you source in line 6 `functions.sh` and in that file (line 32) you define `_install_package()`, but xbps-src understands this as ‘create a subpackage named _install’ and looks for `srcpkgs/_install`.
When renamed e.g. to `_install_pkg()` it works. :)

asergi:
Update: 922e8adb3b3db27526453fca8a1b2800b9bfcb1a.

PR #10093, opened 2017-12-18
lxappearance: add build option for GTK3 version and dbus

SolitudeSF:


cr6git:
To be honest `--enable-dbus` should be an option too, because `dbus` has never been necessary to run `lxappearance`, just to say...

SolitudeSF:
added option for dbus, but since i dont understand the implications of switching it off i left it as default

PR #10097, opened 2017-12-18
qt5: add configure option -proprietary-codecs

### LINKED ISSUE #9985: HTML 5 video (H.264) does not work with qutebrowser (webengine)
### System

* xuname:  `Void 4.14.5_2 x86_64-musl GenuineIntel uptodate rFF`
* package:  `qutebrowser-1.0.4_1`

### Expected behavior

After installing gstreamer plugins, HTML 5 video should work.

### Actual behavior

Visiting https://html5test.com/ indicates H.264 is not working.

 1. Install `qutebrowser`, `gstreamer1`, `gst-plugins-{bad,good,ugly,base}1`, `gst-libav`. 
 2. Set `allow-plugins` to `true` in qutebrowser.
 3. Visit https://html5test.com/ to show H.264 video is not working.

```
xbps-query -s gst
[*] gst-libav-1.12.3_1         GStreamer libav plugin (1.x)
[*] gst-plugins-bad1-1.12.3_3  GStreamer plugins from the bad set (v1.x)
[*] gst-plugins-base1-1.12.3_1 GStreamer Base Plug-ins (v1.x)
[*] gst-plugins-good1-1.12.3_1 GStreamer set of well-maintained good plug-ins (1.x)
[*] gst-plugins-ugly1-1.12.3_1 GStreamer plugins from the ugly set (1.x)
[*] gstreamer1-1.12.3_1        Core GStreamer libraries and elements (1.x)
```

Note that this seems similiar to #4355 but instead of webkit, webengine is being used as the backend.

wesbarnett:
Fixes #9985.

### DIFF
--- a/srcpkgs/qt5/template
+++ b/srcpkgs/qt5/template
@@ -1,7 +1,7 @@
 # Template file for 'qt5'
 pkgname=qt5
 version=5.8.0
-revision=10
+revision=11
 wrksrc="qt-everywhere-opensource-src-${version}"
 build_style=gnu-configure
 homepage="http://qt.io/"
@@ -222,6 +222,7 @@ do_configure() {
 	opts+=" -system-zlib"
 	opts+=" -system-sqlite"
 	opts+=" -system-harfbuzz"
+	opts+=" -proprietary-codecs"
 	# opts+=" -silent"
 	opts+=" -v"
 


newbluemoon:
But https://doc.qt.io/qt-5/qtwebengine-features.html#audio-and-video-codecs says:

> Warning: When distributing proprietary codec libraries, you must acquire licenses for them.

So unless I’m getting it wrong, I don’t think Void can distribute binaries built with that option.

wesbarnett:
I'm assuming the Chromium package is built with proprietary codecs since H.264 videos work. Firefox can play them as well. So Void is probably already distributing them.

On December 19, 2017 3:06:58 AM EST, newbluemoon <notifications@github.com> wrote:
>But
>https://doc.qt.io/qt-5/qtwebengine-features.html#audio-and-video-codecs
>says:
>
>> Warning: When distributing proprietary codec libraries, you must
>acquire licenses for them.
>
>So unless I’m getting it wrong, I don’t think Void can distribute
>binaries built with that option.
>
>-- 
>You are receiving this because you authored the thread.
>https://github.com/voidlinux/void-packages/pull/10097#issuecomment-352668962

-- 
Sent from mobile.

pullmoll:
The problem with this PR and our build infrastructure is this: building qt5 is a real beast! There are 2-3 more things pending (enabling cross Qt5Designer which is currently missing) and if at all, qt5-5.8.0 will perhaps be built once more when all these things are *in*. So if anything, I would include your change in the *last* build of 5.8.0.

qt5-5.9.3 would enable the proprietary stuff but this PR is waiting for action by the infrastructure owners.

And yes, chromium (which is the base of qt5-webengine) also enables proprietary codecs. Enabling them does not change the license AFAICT, but then I dunno what other team members think about this.

PR #10105, opened 2017-12-19
harfbuzz: update to 1.7.4.

maxice8:


maxice8:
Chromium seems to be working

maxice8:
Firefox seems to be working

PR #10120, opened 2017-12-19
sudo: disable insults by default

mvf:
#10048 enabled `sudo` insults by default. This is a regression because the messages aren't localized. Example with `LANG=ja_JP.UTF-8`:
```
$ sudo
パスワード: 
Speak English you fool --- there are no subtitles in this scene.
```
Expected:
```
$ sudo
パスワード: 
残念、また試してください。
```
This change makes it so users have to explicitly enable insults by putting a `Defaults insults` into `/etc/sudoers`.


cr6git:
@mvf Why you consider the English to be a regression? I do not agree with you.

cr6git:
@mvf Have you noticed: **musl** is not localized, is it a regression?

PR #10137, opened 2017-12-20
livestreamer: removed, unmaintained upstream

asergi:
streamlink (fork of livestreamer) is the successor.

@lemmi


lemmi:
I'm fine with this. I more or less dropped `livestreamer` in favour of `youtube-dl` years ago. Since `streamlink` is already packaged, I think it's save to merge.

PR #10142, opened 2017-12-20
gwenview: revbump to trigger build

Hoshpak:
The package was merged in cdb708578f233859ddb9ecc358712cc27ef791ff but didn't show up in either the buildbot interface or the repositories. It seems to me that the builders just somehow missed the template or commit. This PR is intended to bump the revision to trigger a build of the package of the builders. If you think this is the wrong approach or there is something else amiss, please let me know.


Hoshpak:
obsoleted by f1fb2d52149e119dba69c1f0cd221b6850148822

PR #10165, opened 2017-12-21
tesseract-ocr: fix module-dependency

Piraty:
`osd` and `equ` were handled as they were language data file, but they actually are not.
At least `osd` is relevant for simple text documents.

I installed `tesseract-ocr` along with only one specific language subpackage and it complained about missing osd.trainingdata, so that seems to be mandatory, therefore at least `osd` should be added to depends (or not be subpackaged at all).


Gottox:
The fix introduced a dependency-loop that broke other packages. The real fix is to remove the tesseract-ocr-osd tesseract-ocr-equ subpackages and ship the files with tesseract-ocr.

Piraty:
obsolete due to 
https://github.com/voidlinux/void-packages/commit/0367e0e6dda558ba96f0b450d5f5d811b281d3e7
https://github.com/voidlinux/void-packages/commit/a0a407a81be3ab8ea3b76a33d4076d2cb2fd72f4
e384fa24047ca9fc7fa67f03ba63cb99e1299419

PR #6495, opened 2017-05-24
Cinnamon 3.4

Gottox:
With this update the following has changed:

* switched from upower0 to upower
* switched from ConsoleKit2 to elogind

PR #10178, opened 2017-12-22
[WIP] Update Cinnamon to 3.6

Johnnynator:
Menu is currently crashing on musl for me so have to check that first #6495


pullmoll:
Then instead of `mozjs38` probably `mozjs24` can be removed?
It seems `cjs` was the last package referencing it.

pullmoll:
Oops, this merge was by accident. I'll take a look at why menu fails with musl libc.

Johnnynator:
xD, at least 3.6 isn't as buggy on musl as our current 3.2. So the merge only improved the situation.

pullmoll:
Glad to hear that sometimes accidents improve a situation :smile_cat:

PR #10193, opened 2017-12-25
crystal: update to 0.24.1

### LINKED ISSUE #1781: Package request: crystal-lang

LVMBDV:
`crystal` now comes with `shards`, its package manager. I also tried to use Crystal's `--threads` option to allow multithreaded builds but I don't know if it actually works, it was still pretty slow.

PS `libyaml-devel` added to `makedepends` because shards needs it.

PPS This fixes #1781 maybe, probably?


aeosynth:
also update to 0.24.1? https://github.com/crystal-lang/crystal/releases/tag/v0.24.1

LVMBDV:
Oh cool, sure.

LVMBDV:
Done, it built fine. I said "upgrade" instead of "update" in the commit message but I amended that, those are not 2 seperate commits.

LVMBDV:
Also I tried to list itself as a hostmakedepend instead of `crystal-bin` but it gets stuck in a loop trying to build itself to build itself ad infinitum. How do we approach this?

pullmoll:
If you make it depend on the host crystal, the cross builders will build the host version first and (sometimes or always) fail at doing so. It's better to keep crystal-bin to bootstrap. All IMO.

LVMBDV:
Done.

PS Should I update `crystal-bin` too then?

pullmoll:
Hmm.. as long as crystal-0.24.x or later can be built with crystal-bin-0.24.0 I think an update is not necessary. Who wants to use crystal will certainly use the Void built package.

It may make sense, however, to simply integrate crystal-bin into crystal itself instead of having a separate package. There is no real need for crystal-bin other than bootstrapping, or is it?

LVMBDV:
Although it is functional for regular use, `crystal-bin` is just for bootstrapping. And I think we should keep the bootstrap compiler up to date as well.

I can integrate the binary compiler into `crystal` if you really want `crystal-bin` gone.

Anything else related to this PR?

pullmoll:
I don't *really want*, it's just a suggestion. I think similar packages which require some bootstrap binary also download it as part of the templates' distfiles, but there is no written policy AFAICT.

Other than that: you're installing a license for shards but list crystal, now including shards, as being Apache-2.0? For Apache-2.0 there is a license file in /usr/share/licenses and would not need to be installed. If shards has a different license, your should mention it in the combined package.

LVMBDV:
They are both under Apache-2.0 so I just left the license variable as it was. And I installed their own `LICENSE` files as they contain a copyright notice which we are obligated to include when we distribute these software, I think?

pullmoll:
I guess that's okay then.

LVMBDV:
Anything I can do that is keeping this from being merged?

pullmoll:
Bear with my laziness? It's just that I always try to build the PRs locally and if they seem ok, merge them. But this christmas I've been drunk most of the time and didn't want to merge too much problems into master, until I was sober enough to clean up after me....

LVMBDV:
Heh, glad you enjoyed yourself :)

da99:
@LVMBDV Should `pkg-config` be put as in `depends=`?
(I needed to manually install `pkg-config` to get Crystal to compile to use SSL features. For some reason it was not installed by default on my Void distro.)

LVMBDV:
I wouldn't mind adding it since we already have a few that could be considered optional from crystal's perspective. About SSL, I sent a [PR](https://github.com/crystal-lang/crystal/pull/5676) to crystal for better LibreSSL support so we'll have a smoother experience once it is merged into the next release :)

da99:
@LVMBDV Thanks for your service to Void and The World :)

PR #10203, opened 2017-12-25
radare2: update to 2.2.0.

maxice8:


jnbr:
as radare2 provides shared libraries, I think they should be added to `common/shlibs`.
For version `2.2.0`, the list should look like this:

```
libr2.so.2.2.0 radare2-2.2.0_1
libr_core.so radare2-2.2.0_1
libr_config.so radare2-2.2.0_1
libr_cons.so radare2-2.2.0_1
libr_util.so radare2-2.2.0_1
libr_flag.so radare2-2.2.0_1
libr_asm.so radare2-2.2.0_1
libr_bin.so radare2-2.2.0_1
libr_anal.so radare2-2.2.0_1
libr_lang.so radare2-2.2.0_1
libr_debug.so radare2-2.2.0_1
libr_egg.so radare2-2.2.0_1
libr_bp.so radare2-2.2.0_1
libr_search.so radare2-2.2.0_1
libr_io.so radare2-2.2.0_1
libr_hash.so radare2-2.2.0_1
libr_socket.so radare2-2.2.0_1
libr_parse.so radare2-2.2.0_1
libr_syscall.so radare2-2.2.0_1
libr_crypto.so radare2-2.2.0_1
libr_fs.so radare2-2.2.0_1
libr_magic.so radare2-2.2.0_1
libr_reg.so radare2-2.2.0_1
```

pullmoll:
> as radare2 provides shared libraries, I think they should be added to common/shlibs.

There is no radare2-devel package. Listing `*.so` files without soversion does not make much sense and radare2 has no major version only `*.so.2` files, which means that common/shlibs would need to be changed with every update.

I think the best we can do is just revbump the packages depending on radare2:  r2-bindings, bokken, and cutter, with every update of radare2.

PR #10245, opened 2017-12-28
konsole: update to 17.12.0.

Chocimier:


jambonmcyeah:
I see, so it's not only me that's having it use the wrong desktoptojson

jambonmcyeah:
Any idea how to fix it?

Chocimier:
Look at ark template.

jambonmcyeah:
thanks

PR #10246, opened 2017-12-28
kde5-baseapps: update to 17.12.0.

Chocimier:


Chocimier:
This won't crossbuild, because it transitively depends on broken-crossbuild kdelibs4support-devel.

PR #10259, opened 2017-12-28
cutter: update to 1.1.

jnbr:


jnbr:
package creation fails due to missing entries in `common/shlibs`.

PR #10260, opened 2017-12-28
radare2: update shlibs

jnbr:
they are outdated / incomplete
as stated by @pullmoll in #10203, it makes no sense to have them here


jnbr:
this breaks packaging of cutter (#10259). So maybe keeping them up to date makes sense.

PR #10266, opened 2017-12-29
pywal: update to 1.0.1.

maxice8:


Johnnynator:
shouldn't it depend on python3-setuptools? ATM I have to install it manually to get it working.

PR #10277, opened 2017-12-29
New package: ykpivmgr

Aloz1:


lemmi:
hm.. the naming seems odd. is this how other distributions name it as well?

Aloz1:
Other distros keep the name identical to upstream, which is yubico-piv-tool, with the gui being yubikey-piv-manager (the GUI isn't in the repo's yet). I've renamed it to ykpivmgr in the intrest of keeping it consistent with the other yubikey software currently in the repos. e.g. ykneomgr, ykpersonalize, ykpers-gui, ykparse, ykinfo, ykinfo, etc...

I am happy to change it back to yubico-piv-tool if you think that is more appropriate.

lemmi:
It's probably more yubico's fault for having weird names. I don't really care that much. But maybe add yubico-piv-tool to the description, so we are able to find it with a package search.

Aloz1:
Done

PR #10283, opened 2017-12-30
New package: dislocker-0.7.1

samsky72:


newbluemoon:
With `build_style=cmake` cmake is pulled in automatically, you don’t need to add it to hostmakedepends.

Checking `dislocker-0.7.1/src/CMakeLists.txt` you can specify the library directory with 

    configure_args="-DLIB_INSTALL_DIR=lib"

so you can remove `post_install()` and all should be fine. :)

samsky72:
Thank you for your help, newbluemoon!

newbluemoon:
@samsky72 No problem :)

But when making changes you should use `git commit --amend` and then force push so that there is just one commit per new/changed package in the PR. Or you can squash multiple commits into one via
`git rebase -i ...`.

srcpkgs/dislocker/template
```diff
@@ -0,0 +1,14 @@
+# Template file for 'dislocker'
+pkgname=dislocker
+version=0.7.1
+revision=1
+build_style=cmake
+configure_args="-DLIB_INSTALL_DIR=lib"
+makedepends="mbedtls-devel fuse-devel"
+depends="fuse mbedtls"
```
Vaelatern: This line should be deleted.
https://travis-ci.org/voidlinux/void-packages/jobs/323174884#L1053-L1058

samsky72: Ok, the line has been removed.

PR #10286, opened 2017-12-30
pngcrush: fix warning about png{.c,.h} mismatch

copyleftie:


Vaelatern:
How does a rebuild fix the problem?
What is the problem to begin with?

copyleftie:
The package in the repo throws a warning that png.c and png.h versions are different on every invocation:

```
$ pngcrush
Warning: versions are different between png.h and png.c
  png.h version: 1.6.32
  png.c version: 1.6.34

CPU time decode 0.000000, encode 0.000000, other 0.000001, total 0.000001 sec
```

Building pngcrush in void-packages as is prevents this error from ocurring, for some reason. 

```
$ ./xbps-src pkg -f pngcrush && xi -yf pngcrush
# package builds and installs here...
0 downloaded, 1 installed, 0 updated, 1 configured, 0 removed.
$ pngcrush
CPU time decode 0.000000, encode 0.000000, other 0.000001, total 0.000002 sec
```

I'm not sure why this works really --- my best guess libpng might have updated after the last time pngcrush was built?

Duncaen:
> I'm not sure why this works really --- my best guess libpng might have updated after the last time pngcrush was built?

Yes, has a versioned shared library and a stable api, if a rebuild would be required they would increase the library version.
No idea why pngcrush has this arbitrary check.

PR #10289, opened 2017-12-30
boswars: fix cross compilation

newbluemoon:
The cross-compiler wasn’t respected.


Vaelatern:
You added `glu-devel` and `python`, did that alter the build?

Vaelatern:
`python` at least I can see being used.

newbluemoon:
I don’t think so: `SDL` is built for i686*/x86_64* with option ‘opengl’ per default which pulls in `glu-devel`. On all other architectures `SDL` is built without that option (per default). When cross-compiling `boswars` an SDL include which includes `GL/glu.h` is used (but I don’t know if boswars makes use of glu at all). Adding `glu-devel` solves this. In my opinion that doesn’t alter the build.

As for `python`: the build system is a python script. It is pulled in via a dependency, but cannot be used when cross-compiling, of course. :)

And I thought a `$CROSS_BUILD` branch isn’t necessary, because in a native build those packages are there anyway (via dependencies).

Vaelatern:
Excellent.

newbluemoon:
I’ll be more verbose next time :)

PR #10311, opened 2017-12-31
New package: nheko-0.1.0.

maxice8:


srcpkgs/nheko/template
```diff
@@ -0,0 +1,32 @@
+# Template file for 'nheko'
+pkgname=nheko
+version=0.1.0
+revision=1
+wrksrc="nheko"
+build_style=cmake
+hostmakedepends=""
+makedepends="lmdb-devel qt5-tools-devel qt5-multimedia-devel"
+depends=""
```
Johnnynator: You have missed some empty variables

maxice8:
Oh yes, I was in a hurry for new years party (I'm on GMT -3),will take care
of it next year.

Thanks for noticing

srcpkgs/nheko/template
```diff
@@ -0,0 +1,30 @@
+# Template file for 'nheko'
+pkgname=nheko
+version=0.1.0
+revision=1
+wrksrc="nheko"
+build_style=cmake
+makedepends="lmdb-devel qt5-tools-devel qt5-multimedia-devel"
```
Vaelatern: So I guess `qt5-host-tools` for `hostmakedepends`

PR #10312, opened 2018-01-01
New package: navit-0.5.1+rc1

Vaelatern:
See files installed. Advice on fixing the new root level directory would be welcome.

It also downloads a file during the build, can that be fixed easily? I couldn't see how.

### DIFF
--- a/srcpkgs/navit/template
+++ b/srcpkgs/navit/template
@@ -1,0 +1,24 @@
+# Template file for 'navit'
+pkgname=navit
+version=0.5.1+rc1
+revision=1
+_ver="${version/+/-}"
+wrksrc="${pkgname}-${_ver}"
+build_style=cmake
+configure_args="-DLIBDIR=lib/navit -Wno-dev"
+hostmakedepends="pkg-config qt5-host-tools"
+makedepends="SDL-devel qt5-devel gpsd-devel imlib2-devel dbus-glib-devel
+ fontconfig-devel zlib-devel libpng-devel qt5-declarative-devel qt5-svg-devel
+ qt5-location-devel SDL_ttf-devel SDL_image-devel libfreeglut-devel gtk+-devel
+ pango-devel atk-devel gdk-pixbuf-devel qt5-sensors-devel qt5-multimedia-devel
+ python3-devel"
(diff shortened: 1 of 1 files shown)


Johnnynator:
-DSAMPLE_MAP=FALSE is probably the argument you are searching

newbluemoon:
The issue with the `LIB_DIR` seems to be a bug in `CMakeLists.txt`: there is

    GET_FILENAME_COMPONENT (LIB_DIR ${LIBDIR} NAME)

but according to [cmake’s documentation](https://cmake.org/cmake/help/v3.10/command/get_filename_component.html) I think instead of `NAME` there should be `DIRECTORY`. So with

    configure_args="-DLIBDIR=lib/navit -Wno-dev"

and

```
pre_configure() {
        sed -i -e 's/\(GET_FILENAME_COMPONENT (LIB_DIR ${LIBDIR}\) NAME/\1 DIRECTORY/' CMakeLists.txt
}
```

the right directory is used.

newbluemoon:
Maybe a simpler way is to just use

```
pre_configure() {
        sed -i -e 's/IF (CMAKE_SIZEOF_VOID_P EQUAL 4)/IF (1)/' CMakeList.txt
}
```

without setting `-DLIBDIR=lib/navit` in `configure_args`, because if unset CMake tries to autodetect `LIB_DIR`. It just uses `lib64/navit` on 64bit systems. The sed command above makes sure to enter the 32bit branch which sets `LIB_DIR` to the desired `lib/navit`. :)

Vaelatern:
Thanks, poking at it again.

Vaelatern:
The problem really is as follows: many files of the form: `masterdir/destdir/navit-0.5.1+rc1/builddir/navit-0.5.1-rc1/build/lib/navit/binding/libbinding_python.so`, not leakage, but the `/builddir/navit-0.5.1-rc1/build` dir shouldn't be installed on user machines.

newbluemoon:
I had this problem when using `-DLIB_DIR=lib/navit`, but not with `-DLIBDIR=lib/navit`. And also not when using the second approach (without specifying `-DLIBDIR...` or `-DLIB_DIR...` at all).

PR #10314, opened 2018-01-01
New package: squirrel-3.1

wpbirney:
New package: squirrel-devel-3.1


Vaelatern:
Quote:  The current stable release is 3.0.7

wpbirney:
where do you see that?

The homepage and github repo both say 3.1 is stable

srcpkgs/squirrel/template
```diff
@@ -0,0 +1,22 @@
+# Template file for 'squirrel'
+pkgname=squirrel
+version=3.1
+revision=1
+build_style=cmake
+configure_args="-DCMAKE_INSTALL_PREFIX=/usr"
+hostmakedepends="gcc"
```
Vaelatern: Please remove this line. And also `configure_args`

Vaelatern:
```
 Development state

    The current stable release is 3.0.7
    The project has been compiled and run on Windows(x86 & x64), Linux(x86 & x64), Illumos(x86 & x64), Mac OS X, FreeBSD, iOS and Android.
    Has been tested with the following compilers:

        MS Visual C++ 6.0,7.0,7.1,8.0,9.0 and 10.0(x86 & x64)
        MinGW gcc 3.2 (mingw special 20020817-1)
        Cygwin gcc 3.2
        Linux gcc 3.x
        Linux gcc 4.x
        Illumos gcc 4.x
        XCode 4


    The documentation has to be improved.

    I'd like to have some feed back and maybe help to design/port/test it.


Work in Progress

In the next release(3.1.x stable): 
```

wpbirney:
Ahh, i did not notice it said 3.0.7 as the current stable

I saw in the top section of http://www.squirrel-lang.org/ that 3.1 stable has been released 
As well as here https://github.com/albertodemichelis/squirrel/blob/master/README

I can change it to 3.0.7 if you'd like

pullmoll:
IMO since the main page announces `Squirrel 3.1 stable has been released, check it out.` and v3.1 is the only available release, this can certainly be merged.

I took the freedom to clean up the template to meet our standards

@wpbirney please take a close look at the changes once I pushed it. You should install `xtools` and run `xlint srcpkgs/<yourpkg>/template` to get hints what's missing or wrong in your new templates.

The `<pkgname>-devel` packages generally contain headers, pkgconfig or cmake files, static libraries and the symbolic links `*.so` files to version solibs.

The xbps-src log output tells you which libraries are missing from `common/shlibs` so that a package depending on `<yourpkg>-devel` can find the library files.

In this case, a compiler, there should be a separate `<pkgname>-libs` package which includes just the shared libraries, not the compiler itself. The main package then depends on `<pkgname->-libs` for the dynamically linked executable.

Last not least perhaps the documentation should be integrated into the template, i.e. download the PDF or similar files and add a `squirrel-doc` subpackage with `noarch=yes` which installs things to `usr/share/doc/squirrel`.

wpbirney:
@pullmoll thank you for your help!

I've taken a look at your changes and seen what all i missed, will help alot for next time!
Ill handle the doc and examples packages and create a PR for the update

Thanks again

PR #10318, opened 2018-01-01
pywal: update to 1.0.4.

maxice8:


pullmoll:
...and give us our daily pywal :grimacing:

PR #10325, opened 2018-01-01
fatrace: fix not working on non-x86 kernels

ElementW:
Removes the use of a kernel-side flag (O_LARGEFILE) that is different from the hardcoded value for basically all non-x86 kernels. On ARM, this caused the 2nd argument of `fanotify_init` to be `O_RDONLY|O_NOFOLLOW`, where the kernel returns `EINVAL` because of `O_NOFOLLOW`.

Since the kernel bug `fatrace`'s developers intended to fix [has been patched for a while now](http://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/?id=1e2ee49f7), I think it's only fair to drop the constant to 0 to avoid side effects.


chneukirchen:
Good job!

PR #10332, opened 2018-01-01
atom: update to 1.23.1

pbui:


wpbirney:
the build/install can be greatly simplified 

https://github.com/wpbirney/void-packages/blob/7f9c280c0e2dfcf6de0368c306eb375191a4e982/srcpkgs/atom/template

pbui:
@wpbirney I'll close this PR and you can send your own.

Issue #10345, opened 2018-01-02
./xbps-src bootstrap is broken


cr6git:
### System

* xuname:  `Void 4.14.10_1 x86_64-musl GenuineIntel uptodate rF`

* package:  `mpfr-3.1.6_1`

### Expected behavior
Build all bootstrap packages from scratch.
### Actual behavior
It fails, because `mpfr` needs to be updated to the [latest release](http://www.mpfr.org/mpfr-current/).
### Steps to reproduce the behavior
````
$ ./xbps-src bootstrap
[...]
=> mpfr-3.1.6_1: running do-fetch hook: 00-distfiles ...
=> mpfr-3.1.6_1: fetching distfile 'mpfr-3.1.6.tar.xz'...
looking up www.mpfr.org
connecting to www.mpfr.org:80
requesting http://www.mpfr.org/mpfr-current/mpfr-3.1.6.tar.xz
http://www.mpfr.org/mpfr-current/mpfr-3.1.6.tar.xz: Not Found
=> ERROR: mpfr-3.1.6_1: failed to fetch mpfr-3.1.6.tar.xz.
````

pullmoll:
Recently mpfr was updated to 4.0.0 and it seems they moved the old tarball elsewhere. Perhaps you can fix that yourself and create a PR. Meanwhile I'm looking at what changes are required to update mpfr to 4.0.0.

pullmoll:
Should be fixed with a5f54347240eaffee1c939398aaf806e164bb922

PR #10346, opened 2018-01-02
mpfr: update to 4.0.0.

cr6git:
resolves [#10345](https://github.com/voidlinux/void-packages/issues/10345).


pullmoll:
That's by far not all that is required: a number of revbumps for other packages and a fix for libmpc at least.
I'm working on it.

PR #10350, opened 2018-01-02
fpc: update to 3.0.4

mporrato:
This version is required to build Lazarus 1.8.0.
Changed distfiles URLs to use $SOURCEFORGE_SITE.


pullmoll:
fpc on x86_64-musl needs testing and possibly modification of the patch to call the code supposed to be called by glibc's __lib_csu_init/fini functions.

Issue #10353, opened 2018-01-02
Packages with SHA256 mismatches / wrong distfiles


pullmoll:
- [x] Cutegram
      (9189b96f9b07775311c30f89c6ab702581325fbe9e5bf0d44633ada2b84dc5c6)
- [x] Fritzing
      (cd9f5709b8481639ff8649d516d34cb18a096efa3981e3d20f3cc9d6c9953c97)
- [x] ImageMagick
      https://www.imagemagick.org/download/ImageMagick-6.9.9-30.tar.xz: Not Found
- [x] Pyrex
      http://code.enthought.com/src/Pyrex-0.9.9.tar.gz: Forbidden
- [x] R-cran
      https://cran.r-project.org/src/contrib/data.table_1.10.4.tar.gz: Not Found
- [x] R-cran-nloptr
      (84225b993cb1ef7854edda9629858662cc8592b0d1344baadea4177486ece1eb)
- [ ] R-cran-tibble
      https://cran.r-project.org/src/contrib/tibble_1.3.4.tar.gz: Not found
- [x] TSC
      ftp://ftp.secretchronicles.org/releases/TSC-2.0.0.tar.xz: Connection refused
- [ ] adom
      http://www.ancardia.com/download/adom_linux_debian_64_r60_pub.tar.gz: Not Found
- [x] allegro5
      http://download.gna.org/allegro/allegro/5.2.2/allegro-5.2.2.tar.gz: Connection timed out
- [x] alpine
      http://alpine.freeiz.com/alpine/patches/alpine-2.21/all.patch.gz: Gone
- [x] ampache
      (5e284c91c7d52abdaef43a0fd5b7d252e8d06e18dafad93c2361c96b014bb966)
- [x] apache-directory-studio
      http://wwwftp.ciril.fr/pub/apache/directory/studio/2.0.0.v20151221-M10/ApacheDirectoryStudio-2.0.0.v20151221-M10-linux.gtk.x86_64.tar.gz: Not Found
- [x] apache-maven
      http://apache.arvixe.com/maven/maven-3/3.3.9/binaries/apache-maven-3.3.9-bin.tar.gz: No route to host
- [x] arandr
      http://christian.amsuess.com/tools/arandr/files/arandr_0.1.9.orig.tar.gz: Not Found
- [x] awoken-icons
      https://dl.dropbox.com/u/8029324/AwOken-2.5.zip: Not Found
- [x] banner
[...]

PR #10358, opened 2018-01-02
awoken-icons: fix wrong distfiles

cr6git:


cr6git:
Too large for Travis ?

PR #10369, opened 2018-01-03
New package: gloox-1.0.20

wpbirney:


wpbirney:
I can look into patching this for musl in a future update

newbluemoon:
Why patch it for musl only in a future update? ;)
All that’s needed is to include `string.h` in `dns.cpp` and `connectiontcpserver.cpp`:

```
--- src/connectiontcpserver.cpp.orig	2017-02-21 13:32:58.000000000 +0100
+++ src/connectiontcpserver.cpp	2018-01-03 08:38:57.121764510 +0100
@@ -51,7 +51,7 @@
 #endif
 
 #include <cstdlib>
-#include <string>
+#include <string.h>
 
 #ifndef _WIN32_WCE
 # include <sys/types.h>
--- src/dns.cpp.orig	2017-02-21 13:32:58.000000000 +0100
+++ src/dns.cpp	2018-01-03 08:36:12.832722565 +0100
@@ -22,6 +22,7 @@
 #endif
 
 #include <stdio.h>
+#include <string.h>
 
 #if ( !defined( _WIN32 ) && !defined( _WIN32_WCE ) ) || defined( __SYMBIAN32__ )
 # include <netinet/in.h>
```

wpbirney:
@newbluemoon thanks!
I did not realize it was such a simple musl patch, i didn't have time to look into it yesterday..
I will push the patch shortly

wpbirney:
patch applied and builds with musl now

PR #10401, opened 2018-01-03
New package: mirage-0.9.5.2-1

sakhmatd:
Adding new package: mirage-0.9.5.2-1

### DIFF
--- a/srcpkgs/mirage/template
+++ b/srcpkgs/mirage/template
@@ -1,0 +1,15 @@
+# Template file for 'mirage'
+pkgname=mirage
+version=0.9.5.2
+revision=1
+build_style="python2-module"
+hostmakedepends="python-devel pygtk-devel"
+makedepends="python-devel pygtk-devel"
+depends="pygtk desktop-file-utils"
+short_desc="A simple GTK+ image viewer"
+maintainer="Sergei Akhmatdinov <sakhmatd@riseup.net>"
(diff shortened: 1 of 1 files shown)


sakhmatd:
Not sure what to do about aarch64 and armv7hf. I explicitly added python to makedepends, but build job on Travis still claims that python is not found.

cr6git:
Move `python` to the `hostmakedepends` instead!

sakhmatd:
Thanks! That helped, but it looks like the installer only supports x86_64 and i686 anyway?
Unless I am doing something wrong again. Using python-module as a build_style kinda gets me nowhere.

cr6git:
`hostmakedepends="python-devel pygtk-devel"`

No, not the `-devel` packages, only `python` ...

sakhmatd:
Will that make a difference in setup.py detecting arch?

Package manual advices to use `python-devel` in both `hostmakedepends` and `makedepends`, so that's what I did and it does seem to launch `setup.py` now.

I'll try just `python` though. I'm pretty certain `pygtk-devel` is needed to build it too.

sakhmatd:
Using `python2-module` worked out. Thanks for the help!

Apologies for lots of amended commits.

Vaelatern:
That's exactly as it should be!

Vaelatern:
I changed the commit message and merged.

PR #10418, opened 2018-01-04
fpc: fix unit search path in textmode IDE

mporrato:


pullmoll:
BTW did you try if my musl hack creates a package that works for x86_64-musl?
Or did anyone else try? I have so many things to do ... ;)

mporrato:
I don't have a musl system at hand. Perhaps I'll try in a container or vm but it will take some time.

PR #10423, opened 2018-01-04
openbabel: update to 2.4.1 / unbreak arm* and aarch64*

newbluemoon:
The patch is taken from upstream where I reported the [issue](https://github.com/openbabel/openbabel/issues/1689).


newbluemoon:
This is strange: locally avogadro is built without any problem, I even zapped the masterdir and all is OK.

cr6git:
Travis complains about `eigen2`
````
/usr/include/eigen3/Eigen/Core:321:2: error: #error Eigen2-support is only available up to version 3.2.
````

The AUR [package](https://aur.archlinux.org/packages/avogadro) is using a patch related to eigen2 something... Perhaps this is the solution?

ebfe:
We have eigen2 and eigen3.2 for this.

ebfe:
The reason it works sometimes is that avogadro depends on eigen2 and openbabel on eigen... which should conflict but doesn't. So which version you get depends on package install order...

newbluemoon:
So maybe I should do a separate PR for avogadro then?

ebfe:
Hmm, the problem is not the conflicts. It's that avogadro prefers eigen 3.x if available (which is pulled in by openbabel). Since it relies on the 2.x compatibility, the only solution i see is to make them both to depend on eigen3.2. No need for a new PR.

newbluemoon:
Ok, made both use eigen3.2. Let’s wait and see... ;)

newbluemoon:
That did it. Thank you!

PR #10425, opened 2018-01-04
gtk-engine-equinox: fix wrong distfiles & license

cr6git:


pullmoll:
This distfile has a different checksum on each download it seems. Doesn't make sens to try to fix that, unless we find a download URL which delivers a constant archive.

cr6git:
I found this [PPA](https://launchpad.net/~tiheum/+archive/ubuntu/equinox).

pullmoll:
Looks good to me. Testing.

PR #10427, opened 2018-01-04
skype: fix wrong checksum

cr6git:


cr6git:
Updated to a more recent version [#10457](https://github.com/voidlinux/void-packages/pull/10457).

PR #8631, opened 2017-10-28
since: update homepage,distfiles,maintainer

cr6git:
Project dead [#8626](https://github.com/voidlinux/void-packages/issues/8626). 
The original maintainer @beefcurtains is no longer active [on github](https://github.com/beefcurtains). 

I guess a revision bump is not needed however.

PR #10430, opened 2018-01-04
since: fix homepage, distfiles, maintainer

cr6git:

Switch to [Debian](https://packages.debian.org/sid/since).


cr6git:
Unfortunately, it fails in the same way as my previous attempt [#8631](https://github.com/voidlinux/void-packages/pull/8631). 😠

PR #10431, opened 2018-01-04
epstopdf: update to 2.27.

cr6git:


cr6git:
`http://mirrors.ctan.org/support/epstopdf.zip: No route to host`

Strange, this error appears only on the `aarch64` build log.

PR #10429, opened 2018-01-04
ii: add runit service

jmi2k:
It's very useful as a user service. It _needs_ a conf file to set `II_DIR`.

PR #10434, opened 2018-01-04
ii: add runit service

jmi2k:
The same as #10429, but I can't reopen it, lesson learned.

It's very useful as a user service. It _needs_ a conf file to set `II_DIR`.


Duncaen:
I don't think its smart to include a service like this, this might result in users activating it as system service and therefor running it as root.

jmi2k:
I've suffixed it with _generic_ as in `agetty-generic` It's not intended for direct activation, but to make a derived service (`agetty-tty1`, `agetty-ttyS0`, etc.).

Furthermore, I said it's very useful as a user service because that's the most common use case, but I can imagine it being used for IRC bots or something alike. The behavior of `ii` (opening channels as FIFO files) makes it great for that purpose. It would make sense running it system-wide.

Duncaen:
Yes but this would be the first "user service" in `/etc/sv`, we don't have user services by default.
The `-generic` construct from `agetty` is to avoid duplicating run and/or finish scripts, but your generic run script **must** be edited if enabled system wide, running `ii` as root is really not a good idea.

I personally think including this service just leads to confusion, if someone want's a service for `ii` its simple enough to create one.

jmi2k:
Ok I understand your point. Is sad, providing user services using the package manager itself is very convenient (I come from GuixSD where this is being done). In fact, I was preparing another PR for bundling a `runsvdir-generic` system service for enabling user services (`runsvdir-jmi2k` for example).

I'll close the PR (again!) because it seems like it needs more thinking and discussing beforehand. Maybe it can be reconsidered in the future if packaged user services become a thing.

What do you think? Should I open a follow-up issue to discuss it?

Duncaen:
I think this was the last time there was a discussion about it https://github.com/voidlinux/void-runit/pull/14.
No change except a FAQ entry on how to create user services.

PR #10436, opened 2018-01-04
New package: Kvantum-0.10.5

gspe:


newbluemoon:
You need to add `qt5-host-tools` to `hostmakedepends` when cross-compiling.

newbluemoon:
There is a space missing:

    hostmakedepends+=" qt5-host-tools qt5-devel"

gspe:
Thank for you suggestion and time.
Now should build correctly

pullmoll:
Do you really want the uppercase first letter? Other distributions seem to use `kvantum` and users generally complain about Uppercase or even CamelCase package names :)

gspe:
Not really, I've just followed the indication in the manual

> For those packages the upstream packages name should be used. Remember that in contrast to many other distributions, void doesn't lowercase package names. As a rule of thumb, if the tar.gz of a package contains uppercase letter, then the package name should contain them too; if it doesn't, the package name is lowercase.

but I agree that Lowercase is just nice, so I'm going to rename it Lowercase

pullmoll:
It's a matter of taste actually and Void does not enforce one or the other AFAIR. In fact there are quite well known packages with camel case like `ModemManager` or `NetworkManager`, first character uppercase like `Thunar`, or some packages with an all-uppercase name like `SDL`.

KDE5 packages seem to prefer all lowercase...

PR #10443, opened 2018-01-04
New package: davmail-4.8.2

Anachron:
DavMail is a POP/IMAP/SMTP/Caldav/Carddav/LDAP exchange gateway


Anachron:
No idea why the `x86_64` package gets generated empty ... will look into it.

newbluemoon:
You forgot to use `$DESTDIR` in `do_install()`. And it’s probably better to use `vmkdir` instead of `mkdir`. So when changing the last three lines to

        vmkdir usr/share/$pkgname
        mv install/$pkgname-linux-${_arch}-${version}/* ${DESTDIR}/usr/share/$pkgname
        vmkdir usr/bin
        ln -s /usr/share/$pkgname/davmail.sh ${DESTDIR}/usr/bin/$pkgname

the package is built just fine. :)

Anachron:
@newbluemoon thank you for your reply! Then why does it work on `i686`?
I'll update the template accordingly.

Johnnynator:
@Anachron There is only an empty use folder and no files. But don't ask me why this folder was created only for i686

```
Files of davmail:
/home/travis/build/voidlinux/void-packages/masterdir/destdir//davmail-4.8.2
/home/travis/build/voidlinux/void-packages/masterdir/destdir//davmail-4.8.2/usr
```

newbluemoon:
I think because when building a 32bit package xbps automatically creates `$DESTDIR/lib` and links `lib32` to it for compatibility reasons. So even if nothing gets installed in `$DESTDIR` it’s not empty and a package is built. The template installed davmail in `masterdir/usr/share/davmail`.

Anachron:
The thing is the package `dist/$pkgname-linux-${_arch}-${version}-trunk.tgz` is empty on `x86_64` but the `x86` (`i686`) is correct. 

I am not talking about the package that `xbps-src` creates.

newbluemoon:
Hm, just checked and both `*-trunk.tgz` versions are not empty here.

PR #10455, opened 2018-01-04
entr: fix wrong checksum, add license

cr6git:
````
$ xlint srcpkgs/entr/template
srcpkgs/entr/template:8: license 'ISC', but no use of vlicense
````


cr6git:
Oups! I forgot the licence.

PR #10472, opened 2018-01-04
discord-ptb: carry over desktop patch from discord

Coding-Doctors:
Noticed after I built and installed the package that the desktop file patch wasn't carried from discord, so this fixes that so the desktop file can actually launch the program.


Coding-Doctors:
Of course the icon wasn't changed for the different name either, fixing that now.

srcpkgs/discord-ptb/template
```diff
@@ -18,7 +18,7 @@ checksum="64f7a5d7d4e6116297a8979cb4723d1a9e576c5e4c564915c8e0309b9c98ba04"
 do_install() {
 	local package_location="usr/lib/$pkgname" item
 	vmkdir usr/share/pixmaps
-	vcopy discord.png /usr/share/pixmaps/
+	vinstall discord.png 644 /usr/share/pixmap/ discord-ptb.png
```
Coding-Doctors: Used ``vinstall`` since it can rename a file. Unless there's a better/preffered way to do this, this pr is good for merge.

PR #10473, opened 2018-01-05
linux4.14: follows kspp advised configs

thypon:


srcpkgs/linux4.14/files/arm64-dotconfig
```diff
@@ -7787,7 +7791,9 @@ CONFIG_DEBUG_KERNEL=y
 #
 # CONFIG_PAGE_EXTENSION is not set
 # CONFIG_DEBUG_PAGEALLOC is not set
-# CONFIG_PAGE_POISONING is not set
+CONFIG_PAGE_POISONING=y
+CONFIG_PAGE_POISONING_NO_SANITY=y
```
Duncaen: Does this not conflict with hibernation anymore?

Duncaen:
Not sure if I like all the `DEBUG_*` configs, but i guess running them is the only way to find out.

srcpkgs/linux4.14/files/arm64-dotconfig
```diff
@@ -310,7 +311,7 @@ CONFIG_MODULES=y
 # CONFIG_MODULE_FORCE_LOAD is not set
 CONFIG_MODULE_UNLOAD=y
 # CONFIG_MODULE_FORCE_UNLOAD is not set
-# CONFIG_MODVERSIONS is not set
+CONFIG_MODVERSIONS=y
```
Duncaen: Someone checked if this does not break dkms?

Duncaen:
I really think this shouldn't be merged that fast, I see a lot of problems in #voidlinux, some problems with dkms some problems with hibernation.

Gottox:
@Duncaen disabling `CONFIG_MODVERSIONS` and `CONFIG_PAGE_POISONING{,_NO_SANITY}` again?

Duncaen:
I don't think `CONFIG_MODVERSIONS` is the actual problem, other distributions seem to have it enabled.
Currently looking into `PAGE_POISONING`.

Gottox:
Preparing a new PR.

Gottox:
Following the IRC discussion I reverted the whole PR in 7194df2f715cd25314f10d9b362019e0290fc43c.

thypon:
@Gottox Which IRC discussion?

Could we test the Idea patches before merging and list which problem we had on these configs before continuing changing these configs, without testing them a bit?

I had no problems with these configurations but could be otherwise for other users.

PR #7637, opened 2017-09-12
synfig: update to 1.2.1.

diogoleal:

PR #10482, opened 2018-01-05
New package: synfigstudio-1.2.1 / synfig: update to 1.2.1

newbluemoon:
closes #7637 (has been inactive for months)

Synfigstudio is the GUI for synfig (which is the command line renderer).


newbluemoon:
Cross-compilation fails because synfigstudio needs synfig to create some icons, but Travis cannot build synfig for the host first.

PR #10490, opened 2018-01-05
New package: connman-ncurses-1.0

sakhmatd:
Adding new package: connman-ncurses-1.0


pullmoll:
I removed the configure_args `--disable-optimization` and `--enable-debug`. If you want to debug the code, you'd have to add them to your local template. The regular package for a user should be optimized and doesn't need debug info.

PR #10492, opened 2018-01-05
New package: papirus-icon-theme-20171223

gspe:


gspe:
It compile just fine, the problem is that the log is too long to be handled by travis

PR #10494, opened 2018-01-05
gopass: update to 1.6.9.

SevereOverfl0w:


newbluemoon:
Creating the cross-built packages has to fail because you use the target built `gopass` binary to create the completions. Maybe add `gopass` to `hostmakedepends` and make `post_build()` use the host `gopass`?

newbluemoon:
The created completion scripts all have as a first line

    Warning: GPG not found: no gpg binary found

So I guess `gnupg2` should also be in `hostmakedepends`.

Just tested adding `gopass` to `hostmakedepends`: if it already exists for the hosts everything works fine, but if not xbps-src seems to be trapped in a circle when checking the dependencies.

SevereOverfl0w:
The other option is to mark this as `nocross`.

newbluemoon:
But wouldn’t `nocross` be a bit too much, because `gopass` itself can be cross-built without any problem; just generating the shell completion scripts poses a problem?

SevereOverfl0w:
Perhaps.

I'm reluctant to perform this in INSTALL because we lose some benefits of being a packaged file to xbps.

The gopass team have decided that completion files are part of the distribution from their perspective, and that running gopass is the way to complete the distribution. In that sense, we are unable to create a full distribution in a cross-compatible way.

newbluemoon:
> I'm reluctant to perform this in INSTALL because we lose some benefits of being a packaged file to xbps.

That’s what I was thinking, too; just mentioned it as an option.

But what about using a `$CROSS_BUILD` branch to only generate the completion scripts when compiling natively?

PR #10495, opened 2018-01-05
freetype: downgrade to 2.8, fixes fonts in electron apps

Coding-Doctors:
Freetype 2.8.1 causes artifacting on electron apps, reverting the package until this is fixed in electron or a newer freetype version. Here's the tracking issue for the problem https://github.com/electron/libchromiumcontent/issues/384


Duncaen:
No

PR #10497, opened 2018-01-05
vimb: add PREFIX=/usr, fixes broken motion comands

kazufukurou:


Vaelatern:
Can you rework this to use the build style?

PR #10500, opened 2018-01-05
remove dependency on lxc

emdete:
as the webpages [relationship-with-lxc](https://linuxcontainers.org/lxd/introduction/#relationship-with-lxc) states lxd depends on liblxc but not lxc. in fact installing both is a bit ugly as lxd doesn't read the data from lxc (containers done with lxc aren't seen in lxd)

### DIFF
--- a/srcpkgs/lxd/template
+++ b/srcpkgs/lxd/template
@@ -1,13 +1,13 @@
 # Template file for 'lxd'
 pkgname=lxd
 version=2.21
-revision=1
+revision=2
 build_style=go
 go_import_path="github.com/lxc/lxd"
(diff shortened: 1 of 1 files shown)


Vaelatern:
check the travis builds for redundant automatically detected dependencies

iaroki:
Initially it was only liblxc dependency in the template and lxd was complaining about missing lxc while starting container.

iaroki:
Archlinux `lxd` package depends on `lxc` too. Ubuntu `lxd` package with `lxcfs`, `liblxc` also depends on some `lxc-common`. So I think it's the way how `lxc` is packaged.

PR #10503, opened 2018-01-05
New package: gammu 1.39.0

skrzyp:
I just wanted to sync contacts and send ringtones on my Nokia 6310i.

SMSD is separated because its an enterprise-class service to send SMSes on headless servers.
Docummentation is a bit buffy, contains many C examples, so i moved it to `-devel`

It's all okay, pls merg ^_^


srcpkgs/gammu/template
```diff
@@ -0,0 +1,35 @@
...
+build_style=cmake
+hostmakedepends="pkg-config intltool bash-completion"
+makedepends="libbluetooth-devel libusb-devel"
+short_desc="Cell phone connectivity and management tool"
+maintainer="Jakub Skrzypnik <jakub.skrzypnik@interia.pl>"
+license="GPL-2"
+homepage="https://wammu.eu"
+distfiles="https://dl.cihar.com/gammu/releases/gammu-1.39.0.tar.bz2"
+checksum="f20ca4e2367cd0beb33e49027e7e8e4b884ee9ed15d4a5353eb08e8bc269f8e2"
+#subpackages="gammu-devel gammu-smsd"
```
Johnnynator: If this is not needed anymore then remove it

Johnnynator:
You should use tabs for indentations instead of spaces

skrzyp:
Should be OK now.

PR #10517, opened 2018-01-06
alpine: update to 2.21.99.

cr6git:

New version 2.21.99 released 5 Dec 2017, see [alpine.git](http://repo.or.cz/alpine.git/commit/04737118aa1d9ad3db63fb1064267187ae871856).

As the patch cannot be found, I removed it.
The AUR [package](https://aur.archlinux.org/packages/alpine) did the same.


chneukirchen:
The patch is not needed because it's applied to the repo.

PR #10534, opened 2018-01-06
atom: update to 1.23.2

wpbirney:


wpbirney:
odd, x86_64 builds fine locally for me

jarbowski:
Hi,
It seems that the script builds a wrong /usr/share/applications/atom.desktop
I just installed it and the executable and icon are in a wrong path (buiddir)

wpbirney:
@jarbowski 
thanks for the update i must have overlooked that... (I don't personally use .desktop icons), ill add the patch back for the .desktop

jarbowski:
@wpbirney 
%F was good but %U would work too
thank you, anyway

PR #10542, opened 2018-01-07
linux4.14: enable gcc plugins

thypon:
- Enable PAGE_POISONING in AARCH64 (it is already enabled in the other architecture that supports it)
- Enable the GCC plugins support and the following plugins:
  * LATENT_ENTROPY: mixes the return value of the latent_entropy marked functions in order to increase the initial entropy in case of low entropy, for generating keys. It should slow down the boot a bit.
  * STRUCTLEAK: this plugin zeroes the pages of the struct passed by ref and not initialized in order to avoid potential leaks in case these structs are printed without being initialized. It actually increase the kernel size on disk by 10%.
  * ~~RANDSTRUCT: this randomize the structure fields in order to made difficult for a run everywhere exploit to know where the structures fields are placed. Since I don't want to hit the performance too much I enabled the RANDSTRUCT_PERFORMANCE option that tries to RANDSTRUCT but only if it is possible to do that without messing the alignment and the size of these structures in actual memory or the access time. It shouldn't be a problem with dkms modules but it is possible that poorly wrote external modules that use private copy of internal headers (.h) could not work anymore with the randstruct plugin.
https://forum.voidlinux.eu/t/nvidia-driver-fails-to-build-for-4-14-12-1-kernel/4289~~


thypon:
Since RANDSTRUCT is incompatible with some of our supported dkms modules we should test all of them before deploying RANDSTRUCT.

pullmoll:
We definitely need to verify if/what impact these flags (namely RANDSTRUCT) has on each and every dkms package. I failed to verify that when going from 4.14.8 to 4.14.10 and learned my lesson by breaking nvidia.

The referenced nvidia patch can't hurt whether or not we actually employ RANDSTRUCT. Positional initialization is so 1990s ;-P All in all it's beginning to look as if we should write our own Void test procedures for kernel updates, or should we?

chneukirchen:
Imo RANDSTRUCT is useless for us as we provide the same kernel to all users. Not worth it.

thypon:
I disabled RANDSTRUCT. It will need more testing.

PR #10544, opened 2018-01-07
New package: fcitx-qt5-1.2.1

yshui:


CasperVector:
Sorry for the disturbance, but is this not a duplicate of libfcitx-qt5?
(Of course, the latter needs a version bump...)

PR #10552, opened 2018-01-07
xbps-src: make cmake find/use the wx-config wrapper

newbluemoon:
Cmake’s `FindwxWidgets.cmake`  script doesn’t find the `wx-config` wrapper when cross-compiling. This adds the wrapper explicitly to cmake’s cross setup.

With that addition the workarounds in `openbabel` and `passwordsafe` are obsolete. `hugin` uses the same mechanism.

In the `openbabel` template I moved `eigen3.2` from `hostmakedepends` to `makedepends`, because it wasn’t detected and used in cross-builds. Therefore the revbump.


newbluemoon:
armv6hf-musl:

```
[host] cmake-3.10.1_1: installed.
...
/void-packages/common/build-style/cmake.sh: line 38: cmake: command not found
=> ERROR: hugin-2017.0.0_6: failed to run do_configure() at line 38.
```
?

pullmoll:
building hugin for armv6l-musl worked here, though. Perhaps you environment wasn't clean?

newbluemoon:
It works for me locally, too, this is just what Travis complains about. Sorry if it was unclear. :)

PR #10554, opened 2018-01-07
New package: matio-1.5.11

eichf:


newbluemoon:
You can check `config.log` from a native build. There `va_copy` gets detected and the variable `ac_av_va_copy=C99` is set. When the build style is `gnu-configure` you can trick the cross-configure script by setting this variable manually and use it as a cached variable with

```
        configure_args+=" ac_cv_va_copy=C99"
fi
```

and so the check is skipped. Other templates do this in a similar way. It works because we know the target environment. But sadly it doesn’t matter anyway, because `hdf5` cannot be cross-compiled.
But I think it’s better style to add the cross branch and remove `nocross`, because it’s not `matio` which cannot be cross-compiled, but a dependency and this is tracked by xbps-src.

And you should make use of `$pkgname` and `$version` in `distfiles`, e.g.

`distfiles="https://github.com/tbeu/matio/releases/download/v${version}/${pkgname}-${version}.tar.gz"`

This makes updates easier. :)

eichf:
i would like to try this, but as you say it fails because hdf5-devel cannot be cross compiled. however, as far as my understanding goes, `va_copy` is not the only compile-and-execute check in the configure step of matio that would fail the cross build. i may of course be wrong about this.

using `$pkgname` and `$version` as per your suggestion now, don't know why i forgot to do this.

newbluemoon:
Hm, did you try to cross compile before adding `hdf5-devel` to `makedepends`? Because xbps-src stops right when installing the dependencies and you wouldn’t even have reached the configure stage. So I don’t really understand the issue with va_copy (my first impression was that cross compilation only fails because of it). I tried to fix cross compilation of hdf5, but in the build stage it executes target built binaries and I stopped there.

> don't know why i forgot to do this

That just happens ;) Been there, too :)

newbluemoon:
Just built a native hdf5 package on my Raspberry Pi and used it to cross-compile matio: va_copy is the only compile-and-execute check. With the above workaround everything is fine (except for hdf5, of course ;)).

eichf:
you were right, it builds now. adding

```
        configure_args+=" ac_cv_va_copy=C99"
fi
```

to template and removing `nocross`.

This is rather surprising. I thought the various `printf()` calls would also fail the cross build.

(comment written but unsent before seeing your RPi comment - thanks for putting so much effort into this! :))

newbluemoon:
No problem! :)

PR #10559, opened 2018-01-07
New package: Haru-2.3.0

eichf:


newbluemoon:
Why not use the name ‘libharu’ as it’s the name of the upstream repository?
You can also make use of `$pkgname` and `$version` in `distfiles` and `wrksrc`:

```
distfiles="https://github.com/libharu/libharu/archive/RELEASE_${version//./_}.tar.gz"
wrksrc="${pkgname}-RELEASE_${version//./_}"
```

or with `pkgname=libharu`:

`distfiles="https://github.com/${pkgname}/${pkgname}/archive/RELEASE_${version//./_}.tar.gz"`

eichf:
As per [the website of the project](http://libharu.org/), the name is actually _Haru_ (or so I interpret the intent of the varying forms to be found on website and Github), which is why I used that. Would you like me to change it and issue a new pull request?

Thanks for the hint with `$version`. i didn't know this syntax :)

My thinking was that if the tarball was missing the license file it would be missing it also in the future, so i specified the README as license since it also contains the license notice and can therefore act as that. Would you prefer I add it separately?

newbluemoon:
Regarding the name: It’s not always clear which name to chose so I was just stating my observation that on the website I can only see ‘Haku’ beeing used in just one place and also on the release download page it says ‘libhaku’. ;) (But if you do want to change the name, you don’t have to open a new PR, you can modify this one.)

As for the license: You’re right, sorry. It’s embedded in the `README`. I just noticed a commit on the [github project page](https://github.com/libharu/libharu) with the message ‘Missing LICENSE file was created’ and assumed there was none. But maybe it’s better to change the name from `README` to `LICENSE` to prevent misunderstandings? You can add a second argument to `vlicense` to do this.

eichf:
I'm going to stick with _Haru_ for the name as it's either that or _libHaru_ (with capital H) on the website and I think it's more consistent with the Void package naming convention (ie not adding the _lib_ if it's not a library belonging to an application with the same name).

Renaming README to LICENSE now, you're right, that's a better idea.

PR #10570, opened 2018-01-08
drbd-utils: update to 8.9.11.

cr6git:


cr6git:
I can build the package without error, see my [build log](https://gist.github.com/cr6git/b2d3bc04565af5f17dbe8431e6833f8e).

I don't know why Travis failed.

newbluemoon:
This might be a parallelism issue: Your [log](https://gist.github.com/cr6git/b2d3bc04565af5f17dbe8431e6833f8e#file-gistfile1-txt-L284-L287) shows that `drbdsetup.xml` is processed before `drbd.conf.xml` and in the [Travis log](https://travis-ci.org/voidlinux/void-packages/jobs/326246572#L1253-L1256) it’s the other way around. But `drbd.conf` references `drbdsetup` and so I think the processing order matters. Sometimes it’s in the correct order and works, sometimes not.
You can try to add `disable_parallel_build=yes` to the template.

cr6git:
Thank you.
Added `disable_parallel_build=yes` to the template.
Also added `--without-83support` (reference: [Arch](https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/drbd-utils#n34) and [Alpine](https://git.alpinelinux.org/cgit/aports/tree/main/drbd-utils/APKBUILD?h=3.6-stable#n46))

cr6git:
It changes nothing. 😞

newbluemoon:
Sorry, but I thought it was worth a shot. It’s working locally for me, too.

However, I think the initial error is easier to spot now:
```
/usr/bin/xsltproc \
--xinclude --stringparam variablelist.term.break.after 1 \
http://docbook.sourceforge.net/release/xsl/current/manpages/docbook.xsl drbdsetup.xml
http://docbook.sourceforge.net/release/xsl/current/html/its.xsl:1: parser error : Document is empty
```

So it seems to me Travis can’t get the xls stylesheets. Perhaps some network issue?

PR #10578, opened 2018-01-08
feh: update to 2.23.

Chocimier:


cr6git:
[#10524](https://github.com/voidlinux/void-packages/pull/10524)

PR #10590, opened 2018-01-08
sc-controller: update to version 0.4.0.1_1

Vintodrimmer:
Also add `python-evdev` to dependencies(thanks to @maxice8) and `python-devel` to `makedepends`.


maxice8:
squash all the commits.

`git rebase -i master` should provide enough direction.

Vintodrimmer:
I think I screwed up the rebasing because of the conflicts.
WIll thy again.

maxice8:
> What's the proper way to see how it will look before pushing?

`git show`, `git log --oneline --decorate --graph`

PR #10595, opened 2018-01-08
pysolfc: update to 2.1.2.

maxice8:


ebfe:
This seems to be a beta version and misses dependencies `python-six` and `python-random2` (not packaged).

wpbirney:
#10720 should add the python-random2 module required

PR #10604, opened 2018-01-09
python-msgpack: update to 0.5.0.

maxice8:


chneukirchen:
Please revert, this breaks borg.

PR #10608, opened 2018-01-09
Cutegram: update aseman-qt-tools to 1.0.0, fix cross compilation

cr6git:


cr6git:
Failed?
It's mysterious...
I was able to cross build these packages without error on my machine.

newbluemoon:
That’s because the updated libqtelegram-ae-devel is also in `hostmakedepends`. It has to be built twice: for the host and for the target and this doesn’t seem to be possible with Travis. So I think all’s well :)

PR #10612, opened 2018-01-09
xen: update to 4.10.0

pullmoll:
Works for `x86_64{,-musl}`


pullmoll:
I still don't speak ocaml. Anyone who can help with this?
I think I remember something about a parameter for ocaml strings vs. bytes somewhere but can't remember the details.

pullmoll:
Thank you for the helping hand. I'm trying to incorporate your changes.

Why did you add `ghostscript` to the `fig2dev` makedepends? It should perhaps be a hostmakedepends and, if it is really required, be in a separate PR.

I'm going to try to build xen locally to see how it works out.

lodvaer:
It broke with acpica-utils-20180427. Patch: https://gist.github.com/lodvaer/9e8c158c0d438b3bd4014e078b06655c

I think fig2dev does some configure-time detection of the ghostscript command. Without it the build fails with `sh: -q: command not found` because GSEXE in fig2dev is an empty string. You are right that it should be hostmakedepends/separate PR.

By the way, I don't need the efi binary with [this][grub-patch] patch to GRUB, so, as far as I'm concerned, feel free to remove it if changing binutils is undesired.

[grub-patch]: https://lists.xen.org/archives/html/xen-devel/2017-03/txtCeHTNmz1hZ.txt

pullmoll:
Thank you @lodvaer ! I added the fig2dev and binutils patches to master and could build xen-4.10.0 for x86_64. Now testing the other targets.

pullmoll:
Oops! The binutils patch broke `mips*` builds at least. I changed it so that `--enable-target=x86_64-pep` is added to configure for `x86_64*` only.

pullmoll:
Now works for x86_64 and x86_64-musl. I dunno if or what to do about aarch64, but we didn't have xen for that arch before anyways.

PR #10613, opened 2018-01-09
New package: rustup-1.9.0

dleeram:
Outdates #5943.
In this approach, the user mantains the toolchains.


Vaelatern:
Is the `INSTALL.msg` superfluous when you consider other upstream docs?

dleeram:
I agree that upstream documentation is better. Maybe a wiki page explaining the different methods of using Rust on Void.

srcpkgs/rustup/template
```diff
@@ -0,0 +1,29 @@
+# Template file for 'rustup'
+pkgname=rustup
+version=1.9.0
+revision=1
+wrksrc="${pkgname}.rs-${version}"
+conflicts="cargo rust"
```
Vaelatern: Can you explain the `conflicts`?

Vaelatern: (Travis didn't build just now so I can't just look at the files there)

dleeram: The rustup binary is a [proxy](https://github.com/rust-lang-nursery/rustup.rs/blob/38e1d3a6e5049b510f371c7c55972b35eb8a3132/src/rustup-cli/main.rs#L3-L9) depending on arg0 when we run it. We provide `/usr/bin/cargo` and `/usr/bin/rustc` as symlinks to the rustup binary to use this functionality.

PR #10614, opened 2018-01-09
touchegg: fix distfile

Piraty:


Piraty:
https://github.com/voidlinux/void-packages/issues/10353

cr6git:
In the Travis [log](https://travis-ci.org/voidlinux/void-packages/jobs/326943075#L921): there was clearly a problem with geis.

Piraty:
yes of course, it seems it currently cannot be crosscompiled, but i didn't investigate any further on it

PR #10648, opened 2018-01-10
New package: verifast-17.06

raedwulf:
Verifast is a C program verification tool that uses Z3 and ocaml-lablgtk2.

z3: adds ocaml bindings as a subpackage
ocaml-lablgtk2: packages up and split into subpackages
verifast: built and separated into the cli package, verifast (low number of runtime dependencies), and the gui package (runtime depends on some gtk libraries)


pullmoll:
Please rebase for the meanwhile changed common/shlibs. Hint: not adding to the end but somewhere near *related packages* will many times avoid merge conflicts.

srcpkgs/z3/template
```diff
@@ -34,5 +35,24 @@ do_install() {
 	for f in build/python/z3/*; do
 		vinstall $f 0644 usr/lib/python2.7/site-packages/z3
 	done
+	vmkdir usr/lib/ocaml/Z3
+	vinstall src/api/ml/z3.mli 0644 usr/lib/ocaml/Z3
+	for f in build/api/ml/*; do
+		if [ ${f: -2} == ".o" ]; then
```
chneukirchen: use here a `case` with `*.o)` and `*.so)`.

srcpkgs/z3/template
```diff
@@ -1,10 +1,10 @@
 # Template file for 'z3'
 pkgname=z3
 version=4.6.0
-revision=1
+revision=2
 wrksrc="${pkgname}-${pkgname}-${version}"
-hostmakedepends="python"
-makedepends="libgomp-devel gmp-devel"
+hostmakedepends="python ocaml ocaml-findlib"
+makedepends="libgomp-devel gmp-devel ocaml-num ncurses-devel"
```
chneukirchen: perhaps the ocaml support should be an option that is disabled when cross compiling

srcpkgs/z3/template
```diff
@@ -13,9 +13,10 @@ license="MIT"
 homepage="https://github.com/Z3Prover/z3"
 distfiles="https://github.com/Z3Prover/z3/archive/${pkgname}-${version}.tar.gz"
 checksum=511da31d1f985cf0c79b2de05bda4e057371ba519769d1546ff71e1304fe53c9
+shlib_provides="libz3.so"
```
chneukirchen: sooner or later this template will switch to cmake, which uses a soname for the lib (as is strongly preferred by Void),
see https://github.com/Z3Prover/z3/pull/739/commits/657b0de2fcb1c25009139e296532c6b8a5dc52d8

raedwulf: this sounds good however their cmake build system does not support OCaml bindings yet (and hence why they still have the old build system): https://github.com/Z3Prover/z3/issues/986

srcpkgs/z3/template
```diff
@@ -34,5 +39,24 @@ do_install() {
 	for f in build/python/z3/*; do
 		vinstall $f 0644 usr/lib/python2.7/site-packages/z3
 	done
+	if [ $(vopt_if ocaml 1) != "" ]; then
```
chneukirchen: Use `if [ "$build_option_ocaml" ]`

chneukirchen:
One more tiny fix, then it can be merged.  Thanks!

PR #10652, opened 2018-01-10
ngspice: add shared library

faulesocke:
This is useful for building various other software and the upcoming kicad will also require it.


faulesocke:
Don't know, why building fails on this single architecture and works on all the others...

newbluemoon:
I think it’s a musl problem. The other musl architectures are cross-compiled, but that’s not possible:

`ERROR: ngspice-27_2: cannot be cross compiled, exiting...`

cr6git:
Oddly enough, on the Arch [PKGBUILD](https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/ngspice) the `./configure` step is executed in two passes.

newbluemoon:
Configure option `--with-ngshared` conflicts with `--with-x`, see `INSTALL` file:

```
1.4.2 Options Specific to Enable Ngspice as a shared library

 --with-ngshared
        This option let you compile ngspice as a shared
        library or dll, allowing an application controlling
        ngspice. This option excludes using --with-x or
        --with-wingui. Useful additional options are
        --enable-xspice --enable-cider --enable-openmp.
        No graphics inetrface is provided, this has to be
        handled by the controlling application.
```

I think this is why Arch uses a two-step build.

faulesocke:
Okay, I've modified the template once more, now it does a two-step-like build.

Vaelatern:
The `--with-x` never worked quite right. Rather than complicating the template, I'd like to see as follows: shared library, perhaps `--enable-xspice`, and a new template for a new Xorg controller for ngspice.

newbluemoon:
I think I found a way to fix cross compilation. Maybe you can make use of it. It’s based on the template before your two-step modification.

The target built binary `ngmakeidx` is run to create an index file. This just compiles it with the host compiler under a different name (I chose `makeidx`), because the real `ngmakeidx` is part of the package.

```
	pre_build () {
		cd src
		gcc -Iinclude -I${XBPS_CROSS_BASE}/include -o makeidx makeidx.c
		sed -i -e 's|\./ngmakeidx$(EXEEXT)|\./makeidx|' Makefile
	}
fi
```

Just to be sure I compared the generated file with a native build and they’re identical.

Vaelatern:
I'd be happy to see `--with-x` removed.

Vaelatern:
I'm handling this

PR #10656, opened 2018-01-10
gwenhywfar: change prefix detection for the cmake config file

Hoshpak:
The cmake configuration file for gwenhywfar always uses /usr as prefix
which causes issues while cross-compiling. This adds a patch to change
the detection to do it the same way most KDE/QT software does it.

CC @Gottox


Gottox:
👍

Issue #9759, opened 2017-12-03
package request: qMasterPassword


tadeokondrak:
https://github.com/bkueng/qMasterPassword

cr6git:
````
In file included from src/crypto.cpp:26:0:
include/crypto_functions.h:32:10: fatal error: libscrypt.h: No such file or directory
 #include <libscrypt.h>
          ^~~~~~~~~~~~~
compilation terminated.
make[1]: *** [Makefile.Release:418: release/crypto.o] Error 1
make[1]: *** Waiting for unfinished jobs....
make[1]: Leaving directory '/builddir/qMasterPassword-1.2.2'
make: *** [Makefile:38: release] Error 2
=> ERROR: qmasterpassword-1.2.2_1: failed to run do_build() at line 29.
````
Unfortunately, it fails.

Johnnynator:
I think it expects this scrypt library and not the one we have in the repository. https://github.com/technion/libscrypt

cr6git:
If one wants to try, here's my [template](https://gist.github.com/cr6git/5a26f8e82b981fdf9a206d5ac8c71887).

PR #10666, opened 2018-01-10
New package: libscrypt-1.21

wpbirney:


wpbirney:
i just noticed this #9759,

I made this template for other purposes but it would allow qMasterPassword to be built as well

PR #10684, opened 2018-01-11
New package: parity-1.8.6

wpbirney:


srcpkgs/parity/template
```diff
@@ -0,0 +1,27 @@
...
+hostmakedepends="cargo pkg-config"
+makedepends="eudev-libudev-devel libressl-devel"
+short_desc="Fast, light, robust Ethereum implementation"
+maintainer="Wilson Birney <wpb@360scada.com>"
+license="GPL-3"
+homepage="https://github.com/paritytech/parity/"
+distfiles="https://github.com/paritytech/parity/archive/v${version}.tar.gz"
+checksum=0923b18ff2f52d3a6144d587cef0e489abb1308844c525b8fb86d4150134c320
+nocross=yes
+only_for_archs="i686 x86_64"
```
Vaelatern: no

srcpkgs/parity/template
```diff
@@ -0,0 +1,27 @@
...
+
+system_accounts="parity"
```
Vaelatern: Start system name with an underscore.

srcpkgs/parity/INSTALL.msg
```diff
@@ -0,0 +1,3 @@
+After enabling the parity service
+
+visit web3.site in your browser to access parity
```
Vaelatern: remove file

srcpkgs/parity/template
wpbirney: @Vaelatern it fails to build on any other arch including musl

wpbirney:
i previously had `only_for_archs="i686 x86_64"` because it fails to build on musl..

newbluemoon:
> it fails to build on musl

If you can’t fix it you can use something like:

```
        *-musl) broken="<url to build log>";;
esac

```

wpbirney:
@newbluemoon thank you, i was not sure of the proper approach when a package fails to build on musl

PR #10687, opened 2018-01-11
python3-yarl: update to 0.18.0.

maxice8:


asergi:
Since 6934a3683c36416a01f4d2d114cfdf03bf1e0849 this package is broken. As of `0.17.0`, it requires `idna` too.

asergi:
There is no need to rage quit. Thank you for your collaboration.

PR #10692, opened 2018-01-11
opera: update to 50.0.2762.45. Fix /usr/share

gspe:
/usr/share was incorrectly copied to /usr/share/share
I've changed the shell script in /usr/bin that execute /usr/lib/opera/opera with a simlink


pullmoll:
Meanwhile there is http://get.geo.opera.com/pub/opera/desktop/50.0.2762.58/
Other than that looks ok to me. I would probably have used relative symlinks, but that's a matter of taste :)

gspe:
Corrected both of yours observation

pullmoll:
Please squash commits to just one, it makes the live of us maintainers easier.
I did this after pulling your PR in this case.
Thanks for the update!

PR #10708, opened 2018-01-11
New package: cackey-0.7.5

abenson:
I don't believe this should affect shlibs, since it's meant to be dynamically loaded by applications such as Firefox.


srcpkgs/cackey/template
```diff
@@ -0,0 +1,17 @@
...
+version=0.7.5
+revision=1
+build_style=gnu-configure
+makedepends="pcsclite-devel"
+depends="pcsc-tools"
+short_desc="DOD CAC PKCS#11 provider library"
+maintainer="Andrew Benson <abenson+void@gmail.com>"
+license="MIT"
+homepage="http://www.rkeene.org/projects/info/wiki/161"
+distfiles="http://cackey.rkeene.org/download/0.7.5/cackey-0.7.5.tar.gz"
```
the-maldridge: please parameterize this version number, then ping me when its done.

srcpkgs/cackey/template
```diff
@@ -0,0 +1,17 @@
+# Template file for 'cackey'
+pkgname=cackey
+version=0.7.5
+revision=1
+build_style=gnu-configure
+makedepends="pcsclite-devel"
+depends="pcsc-tools"
+short_desc="DOD CAC PKCS#11 provider library"
```
the-maldridge: There's more than one DOD, maybe clarify that this is the US one in case other governments would like to implement overly complex standards of their own?

abenson:
I just removed the reference to DoD since the library implements a standard not specific to DoD.

PR #10720, opened 2018-01-12
New package: python-random2-1.0.1

wpbirney:
Adds missing python-random2 module needed for 
https://github.com/voidlinux/void-packages/pull/10595#issuecomment-357221541


pullmoll:
Now we can fix pysolfc adding python-six and python-random2 to depends, if I got this right?

wpbirney:
@pullmoll yes this allowed pysolfc to start for me

pullmoll:
Ok, I'll update the template.

pullmoll:
Running pysol now tells me
```
No cardsets were found!!!
Main data directory is:
/usr/share/PySolFC

Please check your PySol installation.
```

/usr/share/PySolFC is there and contains data. What may be wrong here?

wpbirney:
@pullmoll iirc the cardsets are a seperate download, perhaps a cardsets template should be made, or we can add the cardsets to the pysolfc template

http://sourceforge.net/projects/pysolfc/files/PySolFC-Cardsets/PySolFC-Cardsets-2.0

wpbirney:
looks like i could be wrong, appears the default cardset is included

https://github.com/shlomif/PySolFC in the extras part of the readme

wpbirney:
@pullmoll i just confirmed downloading the cardsets and putting them under /usr/share/PySolFC/cardsets/* works

ebfe:
Or we could just revert the update and wait for an official (non-beta) release that might actually include the correct cardset.

wpbirney:
@ebfe afaik the original non beta from 2009 version did not include the cardsets either

pullmoll:
The package `aisleriot` downloads and installs the PySolFC cardset as `/usr/share/aisleriot/cards`.
I think that pysolfc could do the same just installing to the default location. If pysolfc-cardset is a separate package, I could change aisleriot to depend on that instead, patching the source to use the shared resource.

newbluemoon:
I have a template ready and could do a PR if that’s OK with everybody.

wpbirney:
@newbluemoon sounds good to me!

newbluemoon:
Ok, done. #10726 :)

PR #10723, opened 2018-01-12
xbps-src: use proxy if XBPS_PROXY is set

pulux:


Duncaen:
https://github.com/voidlinux/void-packages/pull/8117

PR #10731, opened 2018-01-12
mpfr: update to 4.0.0, libmpc: update to 1.1.0, revbumps

pullmoll:
This updates the GNU MPFR and MPC libraries to their latest releases.
The new `libmpfr.so.6` requires revbumps for packages now depending on `libmpfr.so.4`.
Also updated the `cross-*` packages that were still on binutils-2.29 to binutils-2.29.1

I am currently testing by doing yet another zap+bootstrap with these libraries.

Please review but give me some time to check there are no problems with the update before merging.
This may be faster than usual with my shiny new Ryzen 1950X :racehorse: 
I'll remove the [NOMERGE] once I'm confident it works.


pullmoll:
Hmm.. the gcc revbump has most probably to go to a seperate PR.
Or if things don't work that way, we need another bootstrap, because of the mpfr <-> gcc interdependency.

pullmoll:
Update: there don't seem to be any problems. Built ~14500 packages for all targets without new problems.
Still, this can't be merged but needs a bootstrap AFAICT, or we'll be trapped in a dependency loop.

pullmoll:
To avoid the interdependencies (dependency loop) between the libraries `gmp`, `mpfr`, `libmpc` and `isl15` with `gpc` I now integrated these libs into the `gcc` template.

The packages versions are the same as the external ones and are simply copied to `${wrksrc}/<libname>/` so that configure finds and uses them.

This way `gcc` can be built first and separately from the other dependencies.
Now waiting for @Gottox or someone else(?) to add the patches to enable `RETPOLINE` support in gcc and integrate them here as well.

Gottox:
@pullmoll Currently my change management has a few flaws (cough timemanagement cough), so don't expect them from me in the next few days.

If anyone wants to patch our gcc, these patches are very very welcome!

pullmoll:
It seems there is no patchset for RETPOLINE for 7.2.0 yet, so perhaps we should wait for gcc-7.3.0 and then at the same time update mpfr and libmpc and dependencies.

pullmoll:
It'd be fine if someone who has a *homegrown* bootstrap could test if this PR merges without causing a dependency loop. I'm currently some time ahead of the point where I could try myself and a second test can't hurt anyway.

It'd be best to backup one's `hostdir/binpkgs` before corrupting it with a failing test ;-P

PR #10732, opened 2018-01-12
New package: blind-1.1

copyleftie:
I'm wondering if imagemagick should be a runtime dependency as well? You "only" need it to support other image formats than [farbfeld](https://tools.suckless.org/farbfeld/) and PAM. Maybe add a post-install message?


Vaelatern:
this template sucks.

PR #10743, opened 2018-01-13
appstream-glib: update to 0.7.4.

maxice8:
- Adapt to new meson build system, use configure options to remove gtk-doc
- Expand on libstemmer packaging, add URL pointing to issue blocking libstemmer packaging
- Change license to LGPL-2.1
- Add patch to fix musl builds by adding missing stdint.h include for archive_entry.h include
- add gir build option to disable gobject-introspection on cross-compilation
- Remove static archive (.a) vmove that doesn't exist anymore


maxice8:
Seems to be fixed by adding 

```
#include <stdint.h>
```

before the 

```
#include <archive_entry.h>
```

PR #10748, opened 2018-01-13
python-matplotlib: update to 2.1.1.

pbui:
Update patch for new version (maintaining previous behavior).


asergi:
`s/python-functools32/python-backports.functools_lru_cache/`

PR #10750, opened 2018-01-13
lz4: update to 1.8.1.

maxice8:


cr6git:
````
post_install() {
	vlicense lib/LICENSE
}
````

maxice8:
https://github.com/voidlinux/void-packages/pull/8115

I just leave as is because maintainer disapproved last motion.

PR #10772, opened 2018-01-13
New package: otfcc-0.10.2alpha

renatoaguiar:


srcpkgs/otfcc/template
```diff
@@ -0,0 +1,32 @@
+# Template file for 'otfcc'
+pkgname=otfcc
+version=0.10.2alpha
+revision=1
+_version=${version/alpha/-alpha}
+wrksrc=${pkgname}-${_version}
+only_for_archs="i686 x86_64"
```
Johnnynator: Pls also add i686-musl and x86_64-musl, it should work there fine.

pullmoll:
I took the freedom to hack the generated x86/x64 centric `*.make` files to support cross builds in favor of setting `only_for_archs=...`. We really ought to not abuse this variable for (sorry) braindead Makefiles or generators :)

PR #10775, opened 2018-01-13
lz4: update to 1.8.1.

pbui:


cr6git:
[#10750](https://github.com/voidlinux/void-packages/pull/10750)

PR #10779, opened 2018-01-13
sxiv: add support for compile-time config file

jmi2k:


jmi2k:
Oops, mistapped the close button. Sorry

Duncaen:
Hm is it necessary to keep the default file in the repository? I think its easier if the template checks if the config.h file is in FILESDIR and then copies, while still using the default if it doesn't exist.

jmi2k:
I made it according to how it's currently managed in `st`, `dwm`, etc. It's a good idea but I think it should be addressed in another PR for all those packages.

Issue #10780, opened 2018-01-13
Broken templates: FTBFS


chneukirchen:
The following templates were detected to try to compile, but they fail to build from source.

Please help by adding patches!

- [ ] libmemcached
- [ ] mDNSResponder
- [ ] mac
- [ ] mariadb
- [ ] msp430-toolchain
- [ ] oath-toolkit
- [ ] rapidjson
- [ ] reptyr
- [ ] tacacs

wpbirney:
It appears mariadb needs updated to 10.2.12 as well, i can tackle this later today if no one else jumps on it

wpbirney:
it appears mariadb needs updated to 10.2.12
mariadb fails to build 10.2.12 due to pcre reporting/having an incorrect recursion stack size.
I was able to build mariadb by using a pcre configured with --disable-stack-for-recursion, but im not sure that is the right approach as it might impact performance

chneukirchen:
I propose to use the bundled pcre instead.

newbluemoon:
I get no error when building tacacs.

pullmoll:
OTOH stack recursion is something which musl with its limited default stack size may not like too much and we can't be sure every package depending on pcre does increase the stack size when it should. According to the [wikipedia entry](https://en.wikipedia.org/wiki/Perl_Compatible_Regular_Expressions) pcretest needs 8MB of stack space... IMO we could compare the real, practical impact on speed of disabling stack recursion altogether.

pullmoll:
FWIW tacacs seems to have a parallel build issue. `disable_parallel_build=yes` fixes it.
Realized this while modifying the template to enable musl and all target architectures.

wpbirney:
mariadb has build issues still, appears the new version creates different libs...
I wont be able to get to it for about an hour or so

pullmoll:
I reverted the mariadb update for now until it is fixed as it blocks updating other things.

pullmoll:
Surpising measurements of running `time make test` inside the `./xbps-src chroot` after building `./xbps-src -Q build pcre` with and without stack recursion.


without `--disable-stack-for-recursion` | with `--disable-stack-for-recursion`
----------|----------
real	0m4.186s | real 0m4.154s
user	0m3.527s | user 0m3.506s
sys	0m1.124s | sys 0m1.131s

[...]

PR #10783, opened 2018-01-13
mDNSResponder: update to 878.1.1 and fix build #10780

pbui:
- Use byacc instead of bison
- Explicitly set OPTIONALTARG to extra targets and set BISON to byacc


pbui:
musl does not have `nss.h`.  Should we disable this package on those platforms or should we try to patch around it?

maxice8:
This patch should help https://github.com/GregorR/musl-pkgsrc-patches/blob/master/net-mDNSResponder.diff

pbui:
Thanks @maxice8, I just pushed an update that includes a patch which should fix building on musl.

Issue #6981, opened 2017-07-07
tevent-devel: conflicts with samba-devel


ebfe:
```
/usr/include/tevent.h
```

PR #10792, opened 2018-01-13
tevent: update to 0.9.35.

maxice8:
New `libtalloc.so` library that is on `/usr/lib/tevent/libtalloc.so` seems to be for exclusive use, original `libtalloc.so` library belongs to `talloc` package


cr6git:
@maxice8 Do you think it would be possible to find a solution to fix [#6981](https://github.com/voidlinux/void-packages/issues/6981) ?

maxice8:
On Sun, Jan 14, 2018 at 04:33:19AM -0800, cr6git wrote:
> @maxice8 Do you think it would be possible to find a solution to fix [#6981](https://github.com/voidlinux/void-packages/issues/6981) ?
> 
> -- 
> You are receiving this because you were mentioned.
> https://github.com/voidlinux/void-packages/pull/10792#issuecomment-357508749

samba template removes /usr/include/samba/tevent, shouldn't it also remove /usr/include/tevent.h

makes sense that tevent package controls /usr/include/tevent.h

maxice8:
On Sun, Jan 14, 2018 at 04:45:31AM -0800, cr6git wrote:
> ````
> $ xbps-query -Rf samba-devel |grep tevent.h
> /usr/include/tevent.h
> 
> $ xbps-query -Rf tevent-devel |grep tevent.h
> /usr/include/tevent.h
> ````
> Still here, however.
> 
> -- 
> You are receiving this because you were mentioned.
> https://github.com/voidlinux/void-packages/pull/10792#issuecomment-357509436

i am aware of that, if it wasn't the case then there wouldn't be any problem to talk about would it ?

cr6git:
sorry, misinterpreted! 😉 
I spent some time trying to modify the samba template but was not able to find the solution, unfortunately.

maxice8:
On Sun, Jan 14, 2018 at 12:52:47PM +0000, cr6git wrote:
> sorry, misinterpreted! 😉 
> I spent some time trying to modify the samba template but was not able to find the solution, unfortunately.
> 
> -- 
> You are receiving this because you were mentioned.
> https://github.com/voidlinux/void-packages/pull/10792#issuecomment-357509843

precursory looks tells that adding

rm -f "${DESTDIR}/usr/include/tevent.h" at line 76, which is inside post_install() , should solve it, but i didn't test, taking shower.

PR #10800, opened 2018-01-14
retroarch: update to 1.7.0

SolitudeSF:


pullmoll:
Usually a new version also has a new checksum, except for 1 in 1·10<sup>253</sup> cases ...just kidding and fixed it.

pullmoll:
You may have seen that arm* is broken with this update?
It seems either the flags for linking against the libEGL stuff in `/opt/vc/lib` is wrong or they forgot to add (or removed) a library specific for arm* builds.

SolitudeSF:
yeah, i noticed, but i have no clue what to do about that, i just wanted a version bump :^)

SolitudeSF:
actually, if the build is failing anyway, should the CFLAGS lines even be there? without them it fails at compiling stage, but im not comptent enought to know if this was even the right approach.

pullmoll:
I left the CFLAGS for the case the fix is elsewhere.
I can try to fix the problem but thought you may have read about it upstream already.
It's probably visible in the difference between CMakeLists.txt of previous and 1.7.0.

pullmoll:
The [best I could make it](https://gist.github.com/pullmoll/c7fb58c446bc70be800b871e3f3f6442) but still missing symbol(s) when linking.

PR #10822, opened 2018-01-14
New package: waitforfile-0.1.7

cr6git:
[waitforfile](https://github.com/nroi/waitforfile)
Very useful !


Vaelatern:
Well that is a new approach to cross compiling rust.

pullmoll:
Hmm.... **Bahnhof** @Vaelatern ***nod and agree*** Is this for real? Reinterpreted:
```
if [ -z "$CROSS_BUILD" ]; then ...
    # Compile from source for all other archs
    ...
fi
```

I guess I should have been drinking a lot less than I did. Or is it that just my alternate reality bit flipped again? *lol* Perhaps I should get sober and take another look at that template again :)

Vaelatern:
I don't know: is it better to have coverage at the expense of actually building the binaries ourselves?

pullmoll:
I don't know either. It's just that the template looks very confusing to me.
And as per coverage I prefer binaries built by us, but in case that's not possible I think it is ok to
use binary distributions in cases like this.

srcpkgs/waitforfile/template
```diff
@@ -0,0 +1,60 @@
...
+do_build() {
+	if [ -z "$CROSS_BUILD" ]; then
+		cargo build --release
+	else
+		printf "\033[1;32m=> precompiled binaries, nothing to do.\033[m\n"
+	fi
+}
+
+do_install() {
+	if [ -z "$CROSS_BUILD" ]; then
```
Vaelatern: Maybe you can move this logic out of the functions? To try and simplify the template.

cr6git:
@pullmoll Changes were made to make it more clear!

pullmoll:
This is actually close to what I can grok. I @Vaelatern is also happy he could merge?

PR #10823, opened 2018-01-14
linux4.14: add missing support for non-PC partition tables in x86 kernels

skrzyp:
I've just realized that I can't see an RDB disklabel from my Amiga hard disk on ThinkPad running Void. Looks like someone forgot to enable appropriate partition table support…

…surprisingly, it's enabled in ARM kernels without any complains :)

Could you merge this? Should I edit previous kernel version's configs as well?

Additional code isn't large or problematic, IIRC these are just some offset tables, but it releases many people from pain of using two disklabels (MBR and RDB) on the same device at once.


pullmoll:
Without revbump this doesn't change anything. So are you happy enough while waiting for 4.14.14?

pullmoll:
There you go :)

skrzyp:
Thanks for it, I was actually asleep when you asked for revbump :)

pullmoll:
*lol* I **am** usually asleep while merging PRs. That's why it takes 2-3 commits per change until the builders can finish.

PR #10840, opened 2018-01-15
adom: update to 3.0.6

newbluemoon:
#10353 

Reworked the platform checks to prevent building empty packages on unsupportet architectures.


srcpkgs/adom/template
```diff
@@ -1,6 +1,6 @@
 # Template file for 'adom'
 pkgname=adom
-version=60
+version=3.0.6
```
pullmoll: I fear this version scheme change will not work because 3.0.6 < 60.
Perhaps it works with a reverts line, i.e.:
```
pkgname=adom
reverts=60_1
version=3.0.6
…
```

newbluemoon: Didn’t think of that ;)

The homepage says:

> Latest free ADOM download: Version 3.0.6 (release 88)

and the distfiles filename now contains the version. So I think it’s better to change it to the (real) version number and not stick with the release.

PR #10848, opened 2018-01-15
Fix various builds

Gottox:
Addresses https://github.com/voidlinux/void-packages/issues/10780


srcpkgs/acmetool/patches/fix-uuid-build.patch
```diff
@@ -0,0 +1,12 @@
+--- storage/types.go.orig	2018-01-11 08:57:49.339089353 +0100
++++ storage/types.go	2018-01-11 08:58:12.707027749 +0100
+@@ -194,7 +194,8 @@ func (tgt *Target) ensureFilename() {
+ 		nprefix = tgt.Satisfy.Names[0] + "-"
+ 	}
+ 
+-	b := uuid.NewV4().Bytes()
++	u, _ := uuid.NewV4()
```
ebfe: this should probably panic in case of error, else it continues with an all-zero uuid.

Gottox: fixed, thx.

PR #10866, opened 2018-01-16
elogind: update to 235.2.

maxice8:
=== GLIBC RELATED

- [x] - Test-login fails, see this [bug report](https://github.com/elogind/elogind/issues/37)

=== MUSL RELATED

- [x] - Fix getenv definitions on musl-missing.h
- [x] - Fix HAVE_UTMP definitions on musl-missing.h
- [x] - Replace __register_atfork with pthread_atfork and guard the original definition
- [x] - Fix src/test/test-hexdecot.c not getting strndupa but adding \#include "missing.h" causing other compilations errors.
- [x] - Fix test-sizeof.c usage of __syscall_{u,s]long_t and __time_t


Gottox:
To fix the musl issues: We're switching to getenv in scenarios where secure_getenv isn't available.

maxice8:
On Tue, Jan 16, 2018 at 09:18:38AM +0000, Enno Boland wrote:
> To fix the musl issues: We're switching to getenv in scenarios where secure_getenv isn't available.
> 
> -- 
> https://github.com/voidlinux/void-packages/pull/10866#issuecomment-357899320

i traced the problem, musl_missing.h checks if HAVE_SECURE_[__]GETENV is defined, but now with config.h
from meson it is always defined, it should instead check if it's true ( non-zero ), or false ( 0 ), it also
should #undef HAVE_SECURE_GETENV before redefining it (i believe, not sure)

maxice8:
There is also the issue of __register_atfork which is a glibc-ism, but that is handled [here](http://www.openwall.com/lists/musl/2013/11/19/2) apparently

maxice8:
Well i'm hitting a bit of an wall here with strndupa.

```
In file included from ../src/basic/alloc-util.h:27:0,
                 from ../src/test/test-hexdecoct.c:20:
../src/test/test-hexdecoct.c: In function ‘test_unbase32hexmem’:
../src/test/test-hexdecoct.c:185:25: error: implicit declaration of function ‘strndupa’; did you mean ‘strndup’? [-Werror=implicit-function-declaration]
         assert_se(streq(strndupa(mem, len), ""));
                         ^
```

Problem is, if i \#include "missing.h" it will error with a redefinition of `len` for some reason.

maxice8:
Error if musl_missing.h is added

```
In file included from ../src/basic/alloc-util.h:27:0,
                 from ../src/test/test-hexdecoct.c:20:
../src/test/test-hexdecoct.c: In function ‘test_unbase32hexmem’:
../src/shared/musl_missing.h:60:24: error: declaration of ‘len’ shadows a previous local [-Werror=shadow]
                 size_t len = strnlen(in, (n)) + 1; \
                        ^
```

srcpkgs/elogind/template
```diff
@@ -1,24 +1,42 @@
...
+	# since /etc/machine-id is expected to be a file.
+	# test-id128 tries to get a id generated by dbus-uuidgen from
+	# /etc/machine-id, since we don't want to add a checkdependency
+	# on dbus we just provide a string generated from dbus-uuidgen
+	# for use in all cases.
+	echo 'dcb30309cd6c8b7cc20383d85a5c7012' > /etc/machine-id
+
+	# test-copy searches for /etc/os-release but this file is provided
+	# by runit-void, create a dummy file to avoid a dependency on runit-void
+	echo 'dummy-file' > /etc/os-release
```
Gottox: We need to change this. this will polute the masterdir.

maxice8:
On Tue, Jan 16, 2018 at 12:18:25PM +0000, Enno Boland wrote:
> Gottox commented on this pull request.
> 
> 
> 
> > +		*-musl) cp build/config.h src/shared;;
> +	esac
> +}
> +
> +pre_check() {
> +	# test-fs-util checks for /etc/machine-id/foo and expects ENOTDIR
> +	# since /etc/machine-id is expected to be a file.
> +	# test-id128 tries to get a id generated by dbus-uuidgen from
> +	# /etc/machine-id, since we don't want to add a checkdependency
> +	# on dbus we just provide a string generated from dbus-uuidgen
> +	# for use in all cases.
> +	echo 'dcb30309cd6c8b7cc20383d85a5c7012' > /etc/machine-id
> +
> +	# test-copy searches for /etc/os-release but this file is provided
> +	# by runit-void, create a dummy file to avoid a dependency on runit-void
> +	echo 'dummy-file' > /etc/os-release
> 
> We need to change this. this will polute the masterdir.
> 
> -- 
> https://github.com/voidlinux/void-packages/pull/10866#pullrequestreview-89076214

#10879

PR #10869, opened 2018-01-16
log4cxx: fix compilation.

maxice8:
mentioned on #10780


Gottox:
Please increase the revision here too.

maxice8:
On Tue, Jan 16, 2018 at 08:44:07AM +0000, Enno Boland wrote:
> Please increase the revision here too.
> 
> -- 
> https://github.com/voidlinux/void-packages/pull/10869#issuecomment-357890615

Done

PR #10870, opened 2018-01-16
mupen64plus: fix build failure

fallenwizard:
As seen on #10780.


newbluemoon:
Patches should be in p0 format. Why not just remove `a/` and `b/` from the paths in your patch? Then you don’t need `patch_args`.

PR #10871, opened 2018-01-16
panda: remove, deprecated upstream in favour of zef

maxice8:
https://github.com/tadzik/panda/commit/a763bcbc20275cfdd9a34292442ff1d2c5529e4c

mentioned in #10780 

I am currently packaging zef


asergi:
This broke `rakudo-star`. You can either try to update it or, as recommended upstream, remove it. See [piojo commented on 2017-10-29 04:26](https://aur.archlinux.org/packages/rakudo-star/).

Gottox:
Then we should remove it, imho.

PR #10879, opened 2018-01-16
elogind: cleanup after doing tests.

maxice8:


Gottox:
Is there a way to adjust the tests so that they don't access the files?

maxice8:
On Tue, Jan 16, 2018 at 04:26:26AM -0800, Enno Boland wrote:
> Is there a way to adjust the tests so that they don't access the files?
> 
> -- 
> https://github.com/voidlinux/void-packages/pull/10879#issuecomment-357944709

problably but i won't mess up with C source code.

Gottox:
os-release can easily be overwritten, machine-id is harder to change.

maxice8:
sorry but i don't understand what's the big wop.

Gottox:
Let's get this merged, and do the rest later 👍

PR #10895, opened 2018-01-16
sudo: update to 1.8.22.

maxice8:


Duncaen:
> When timestamp_type option is set to tty and a terminal is present, the time stamp record will now include the start time of the session leader. When the timestamp_type option is set to ppid or when no terminal is available, the start time of the parent process is used instead. This significantly reduces the likelihood of a time stamp record being re-used when a user logs out and back in again. Bug [bugzilla.sudo.ws#818](https://bugzilla.sudo.ws/show_bug.cgi?id=818). 

👍

PR #10897, opened 2018-01-16
st: split terminfo files in subpkg

xYuusha:


Vaelatern:
Why?

xYuusha:
To be able to only install the terminfo in a headless machine.
rxvt-unicode is doing the same.

wpbirney:
It's definitely great for headless machines

xtraeme:
revision must be increased

xYuusha:
I always forget to increment the revision. Done now.

PR #10907, opened 2018-01-16
nvidia340: update to 340.106

fallenwizard:


srcpkgs/nvidia340/template
```diff
@@ -41,7 +41,6 @@ do_extract() {
 do_configure() {
```
Hoshpak: kernel-4.11.patch can also be removed, the changes it makes are now in the driver itself. Since this leaves the do_configure function empty, it can be removed. Please don't just remove the patch commands, also remove the patch files from srcpkgs/nvidia340/files . There's no need to keep them around.

PR #10908, opened 2018-01-16
crystal: merge with bootstrap binary package

LVMBDV:
Merged `crystal-bin` into `crystal` as it was only meant to be used to bootstrap the build for `crystal`.

- Bumped the revision number up as the bootstrap compiler version is newer than `crystal-bin`.
- Added makedepends that were carried over from `crystal-bin`.
- Had to write an explicit `do_extract` due to crystal source release and binary release being extracted into the same directory with overlapping files.


Vaelatern:
I don't understand the benefit. Why did you complicate the crystal package? Why not keep crystal-bin around for when a new bootstrap is necessary?

LVMBDV:
No solid benefit besides disallowing users from installing `crystal-bin` for some odd reason. It doesn't have docs as it's meant to be only for building `crystal`. The plan is to depend on a prior version of itself (the source package) if/when xbps supports such a thing.

> Why not keep crystal-bin around for when a new bootstrap is necessary?

I don't understand this phrase, like a new version for crystal or something else that relies on crystal?
If it's the former case, we just change the `_bootstrapversion`.
And if it's the latter, it should depend on the source package to begin with.

Vaelatern:
Say that tomorrow we lose all packages. Or that you want to bootstrap without relying on our packages. You need the `-bin` package to bootstrap crystal again. This isn't a "Well, now it works, never need to make it work again" sort of deal.

LVMBDV:
I still don't follow. This still fetches the precompiled binary in order to bootstrap. `crystal-0.24.1-2` used `crystal-bin-0.24.0` to build, now the same thing happens except the precompiled binary isn't kept on the host after the build. Because it is not needed in any other case, just the build. It is not our job to keep a copy of the precompiled binary just in case.

Vaelatern:
Can we then have a `bootstrap` build option then, which only pulls in `crystal-bin` on that option being true?

LVMBDV:
That is a great idea. I still want to keep the bootstrap binary fetching in `crystal` tho.

Also, to make sure we are on the same page. Right now the `bootstrap` option is what this build already does. I am just speculating about the `bootstrap-with-the-source-package` option, I am not sure if it's actually doable. I will look further into it tomorrow.

pullmoll:
I guess it was my *fault* to, at some point, suggest to @LVMBDV to incorporate the bootstrap binary download into the crystal package itself.

I generally don't like to see that we are forced to rely on third party binaries to be able to bootstrap *any* compiling or scripting languages we provide. But then I don't want to manually enter the primary boot loader to load the first version of FORTRAN, written for PDP-1, available as FORTRAN.ASM and to be compiled by an assembler, which someone (who? entered manually through 8, 12 or 16 bit switches and a big, red STORE switch at a PDP-1 console, which, I hope, had no backdoor, either, except for perhaps squashing bugs which lived in that machine... (/me has an extra comma left `*shrug*` Germans and their sentence structure :)

The reason for this dislike is perhaps not-so-obvious and may seem to be irrational, yet it is very well [described in this document](https://www.ece.cmu.edu/~ganger/712.fall02/papers/p761-thompson.pdf). The question is not whether you're paranoid or not, but whether you're paranoid **enough**! :rofl: 

As long as any language offers some way to bootstrap itself using another language (preferably the *simple* or *primary* ones like Assembler, C or C++.. or even a script) I *see* benefit in incorporating the bootstrap process as an (default disabled?) option into the template. And if a precompiled binary is required, we should offer a trusted (trusted by whom?) source for that as an option.

I agree with @LVMBDV that keeping the crystal-bin as a separate package does not make sense beyond possibly confusing Void Linux users.

Vaelatern:
I'm okay with a one template form. See the `openjdk` package for how to seamlessly do that, @LVMBDV, in a way that works well for us so far.

LVMBDV:
Checked out the `openjdk` package, modified crystal accordingly. Now it has a `binary_bootstrap` option, enabled by default. Sorry @pullmoll, this is due to `hostmakedepends` looking for the most recent version to build a package. I think this is how `openjdk` deals with this too.

A repo host would need to build the recent version with `binary_bootstrap` to build it a second time without it. This is because repos don't keep previous versions of a package.

In another case, if a user has a `crystal` binary in their `PATH` (like from a previous version), they could build `crystal` with the `binary_bootstrap` option disabled.

I see there is a `bootstrap` variable to add the package to the list of packages considered to be part of the bootstrap process. Does it apply here? If not, I think this is pretty much done :)

LVMBDV:
I guess I should mention this. Ignore Travis negatives as they are caused by xlint false positives and compilations being too long for Travis.

pullmoll:
Yes, but not ignore errors for i686 :)

I cleaned up and fixed the template to properly add to the distfiles and checksum for either of x86_64 or i686. It's easier to do this over multiple lines than using `$(vopt_if ...)`.

LVMBDV:
Oh damn, I blame alarm fatigue. Thanks for the fix :)

PR #10910, opened 2018-01-16
mariadb: update to 10.1.30

wpbirney:
this will at least fix the build issue with the 10.1.28 template for the first iteration of #10780

some care will need to be taken when we upgrade to mariadb-10.2.12 because they changed a few sonamed shlibs to symlinks to the libmariadb.so.3. Im in the process of working that out now.

This version 10.1.30 is supported until 2020 and provides the libmysqld.so.18 that our other packages are expecting


wpbirney:
Did this ever cross compile properly? I only up'd it to 10.1.30 instead of 10.1.28

wpbirney:
aarch64 failed to build 
```
/builddir/mariadb-10.1.30/storage/mroonga/vendor/groonga/lib/ts/ts_expr_node.c:3824:18: internal compiler error: in gen_vec_cmpv2dfv2di, at config/aarch64/aarch64-simd.md:2495
       out_ptr[i] = grn_ts_op_ ## type ## _ ## kind(buf_ptrs[0][i],\
       ~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                                    buf_ptrs[1][i]);\
                                                    ~~~~~~~~~~~~~~~
```
building locally for aarch64 now to check, but it seems the others are building

chneukirchen:
triggers on aarch64 this bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81228

will be fixed in 7.2.1

pullmoll:
> triggers on aarch64 this bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81228
> 
> will be fixed in 7.2.1

I guess we can enable mariadb for aarch64 again now that we have gcc-7.3.0?

wpbirney:
Yes and the `disable_parallel_build=yes` should be able to be removed as well,
testing now

pullmoll:
Nope, gcc-7.3.0 still fails for me with the same internal compiler error :-(

pullmoll:
Adding the following lines to the bottom of `pre_build()` works around the bug:
```
        aarch64*) # Work around GCC bug https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81228
                sed -i build/storage/mroonga/vendor/groonga/lib/CMakeFiles/libgroonga.dir/flags.make \
                        -e "s; -O3;;g"
                ;;
        esac
```
If noone complains, I'll add that to mariadb/template and enable aarch64* again.

Issue #10899, opened 2018-01-16
Broken template, FTBFS - the finals


Gottox:
supersedes #10780 and #10881. Those are the remainders, false positives are possible though.

* [ ] R-cran-tibble
* [ ] R-cran-zoo
* [ ] android-tools
* [ ] apache-maven-bin
* [ ] brother-dcp197c-lpr
* [ ] btsync
* [ ] ccextractor
* [ ] criu
* [ ] darcs
* [ ] dbus-sharp
* [ ] debug-me
* [ ] dnscrypt-proxy
* [ ] ejabberd
* [ ] foomatic-db
* [ ] foomatic-db-nonfree
* [ ] fsbext
* [ ] gcc-multilib
* [ ] gnustep-base
* [ ] include-what-you-use
* [ ] ispc
* [ ] kiwix-lib
* [ ] libdbus-c++
* [ ] luaposix
* [ ] mfoc
* [ ] perl-Text-Bidi
* [ ] soundmodem
* [ ] suricata
* [ ] taglib-sharp
* [ ] tintin
* [ ] xnec2c
* [ ] yabasic

ebfe:
suricata is a never ending story since one of the distfiles is updated daily and no versions alternative exists.

Gottox:
same for foomatic-db{,-nonfree}

newbluemoon:
<i>xnec2c</i> can still be found in the old location, it’s just been updated:
http://www.qsl.net/5b4az/pkg/nec2/xnec2c/xnec2c-3.6.1-beta.tar.bz2

It says beta though.

Hoshpak:
taglib-sharp is a .Net package that hasn't seen a release since back in 2012. It currently fails in configure stage while trying to find gmcs (which isn't present any more in current mono versions). Since no package in the repository depends on it, I wonder if it's worth trying to fix the package or if it shouldn't just be dropped instead (same goes for dbus-sharp which is just there for dbus-sharp-glib and no package depends on that).

Gottox:
@Hoshpak Sounds reasonable.

@newbluemoon Not worth having a beta of that 👎

Johnnynator:
What is wrong with kiwix-lib? It compiles fine for me for all checked architectures (armv7*,aarch64*,x86_64*)

Gottox:
@Johnnynator true, it's a false positive.

wpbirney:
fsbext has a wrong checksum, running `xgensum` on it solves it and it builds...
[...]

PR #10927, opened 2018-01-17
ulatencyd: mark as broken

maxice8:
mentioned in #10899


Gottox:
But we have an .a-archive:
```
xbps-query -Rf procps-ng-devel
/usr/lib/libprocps.a
...
```

maxice8:
weird, didn't appear on xlocate

PR #10944, opened 2018-01-17
linux4.9: update to 4.9.77.

pbui:
This enables RETPOLINE on x86_64 and i686.


cr6git:
Why isn't it already merged? I wonder what the Void devs are waiting for...

PR #10959, opened 2018-01-18
git: update to 2.16.0.

maxice8:


pullmoll:
You could also update chroot-git with this PR then.

Issue #3545, opened 2016-02-05
fail2ban run script is incorrect


jsumners:
fail2ban does things in a silly way. The shipped `fail2ban-server` script isn't meant to actually start the server. In fact, it hard codes some configuration values (socket and pidfile). The right way to start fail2ban is with the `fail2ban-client` script. Thus, the run script should be:

```
#!/bin/sh
fail2ban-client start
exec pause
```

xtraeme:
Care to submit a proper pull request?

jsumners:
Turns out even that script isn't sufficient. I've submitted a query to upstream.

cr6git:
> I've submitted a query to upstream.

What was their response?

jsumners:
I don't know if they ever resolved the issue. This got lost in my stack and I forgot about it.

cr6git:
Do you still use `fail2ban` ? 
Today our goal is to determine whether the problem you decribed is resolved.

jsumners:
Actually, I just disabled it a couple days ago since this never got resolved (to my knowledge).

The issue looks to be https://github.com/fail2ban/fail2ban/issues/1139 and the fix merged in https://github.com/fail2ban/fail2ban/pull/1413 . Given that Void is shipping version "0.10.1" this might actually be resolved.

Indeed, it looks like the startup script is very close to what they suggest in https://github.com/fail2ban/fail2ban/blob/309a1cb337604e03f764bf50839bdd3cb8280757/files/fail2ban.service.in . Void is doing `fail2ban-server -f` and may benefit from `fail2ban-server -xf`. A stop script might also be a good thing to supply since it looks like they use `fail2ban-client stop` to shut down the server.

cr6git:
Did you try the command mentioned by sebres? 
````
$ fail2ban-testcases -fn
Fail2ban 0.10.1 test suite. Python 2.7.14 (default, Sep 18 2017, 08:48:18) [GCC 6.3.0]. Please wait...
......................s.s.............................................F......................................................................s...s.......ss.......................................s..............sss.............s......sssssssssssssss..............................................................................................EE........
.....................EE........ssssss..................s..
======================================================================
ERROR: testSampleRegexsZZZ-GENERIC-EXAMPLE (fail2ban.tests.samplestestcase.FilterSamplesRegex)
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

PR #10962, opened 2018-01-18
fail2ban: update to 0.10.2.

maxice8:


cr6git:
While you're at it, what do you think about [#3545](https://github.com/voidlinux/void-packages/issues/3545) ? Thanks.

PR #10975, opened 2018-01-18
cmake: update to 3.10.2.

maxice8:


concatime:
I think we should change to https.

maxice8:
Please do, I left void Linux completely and won't touch any package I am not a maintainer

concatime:
Understood.

Anachron:
What happened maxice8?

cr6git:
@maxice8 what's going on?

PR #10976, opened 2018-01-18
New package: master-pdf-editor-4.3.61

gspe:
This is a proprietary software free for non commercial use, I've asked if I can repack the .deb for Void and this is their answer:

> Hello,
>
>Yes of course, you can!
>Please, pack updates too when you can. 
>
>Sincerely, 
>Code Industry Ltd 
>http://code-industry.net


cr6git:
It looks interesting.

`ar x ${pkgname}-${version}_qt5.amd64.deb data.tar.xz`

I didn't know it was possible to extract only `data.tar.xz` from a `.deb` package, thank you for that!

PR #10978, opened 2018-01-18
New package: Samsung unified driver (Non free)

teajay-fr:
This is port from the arch package to enable printing with samsung printers.


cr6git:
The Travis build failed only for i686:
````
=> samsung-unified-driver-1.00.37_1: running do_install ...
cp: cannot stat 'uld/i686/libscmssc.so': No such file or directory
````

Here in your template:
````
vcopy uld/$XBPS_TARGET_MACHINE/libscmssc.so /usr/lib
````
For i686 architecture, $XBPS_TARGET_MACHINE returns i686, this is normal...
But the correct path in the tarball is `uld/i386/` , not i686!

**Comment by teajay-fr** [comment]
I will fix this. Thank you for the analysis.

**Comment by cr6git** [comment]
````
$ common/travis/xlint.sh
srcpkgs/samsung-unified-driver/template:16: do not use a newline before function opening brace
````
You made a mistake.
Rewrite the function:
````
post_extract() {
	mv uld/i386 uld/i686
}
````

**Comment by teajay-fr** [comment]
This last version shoud pass now !

File: srcpkgs/samsung-unified-driver/template [hunk 1]
```diff
@@ -0,0 +1,79 @@
...
+short_desc="Unified Linux Driver for Samsung printers and scanners"
+maintainer="Thomas Bernard <thomas@famillebernardgouriou.fr>"
+license="Propietary"
+homepage="http://www.samsung.com"
+distfiles="http://downloadcenter.samsung.com/content/DR/201512/20151210091120064/uld_v${version}_00.99.tar.gz"
+checksum="6b85253ea0bb51d241f6fd665ff0d39464cdad87084802a77a385c707fa2c664"
+nostrip=true
+
+post_extract() {
+	mv uld/i386 uld/i686
```
pullmoll: You need to change it like this:
```
[ "$XBPS_TARGET_MACHINE" = "i686" ] && mv uld/i386 uld/i686
```
or the template will fail for x86_64, no?

teajay-fr: I thought of something along these lines but as you saw, we can keep it simple.

teajay-fr:
Seems the build failed with an error in the  binary bootstrap. Is there anything I should do about this ?

PR #10980, opened 2018-01-19
New package: stegsnow-20130616

cr6git:
The patch comes from [here](https://aur.archlinux.org/cgit/aur.git/tree/stegsnow.patch?h=stegsnow).


srcpkgs/stegsnow/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'stegsnow'
+pkgname=stegsnow
+version=20130616
+revision=1
+wrksrc="snow-${version}"
+patch_args="-Np1"
+noarch=yes
```
pullmoll: This is wrong. stegsnow is compiled from C source.

cr6git: I understand. The AUR [PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=stegsnow) tricked me!
````
arch=('any')
````

PR #10994, opened 2018-01-20
busybox: update to 1.28.0.

xYuusha:


pullmoll:
cc @Duncaen please review the busybox.dotconfig changes. I won't merge this without a :+1:

Duncaen:
Many of the enabled features seem not really useful, do we really need `init` (we have `runit` already), package managers like `dpkg` and `rpm` are packaged too.
Some of the programs provided by `util-linux` might make sense if you want to replace both `util-linux` and `coreutils`.
And I don't think we need `dnsd`, `ftpd` etc. looks like the new config has every possible feature enabled.

pullmoll:
So this is more or less :-1: from you... Then let's perhaps request the full blown busybox.config as an option selectable by a `build_options="everything"` or similar and stay with the set we had enabled previously otherwise?

xYuusha:
I mostly enabled all that was by default as with the old config file, the build failed.

xYuusha:
I did a cleanup for the config file to be as close as possible to the current one, adding some new options and removing some old ones. I hope I'm not missing anything important.

PR #10996, opened 2018-01-20
choqok: update to 1.6.0.

Chocimier:


srcpkgs/choqok/template
```diff
@@ -1,13 +1,18 @@
...
 pkgname=choqok
-version=1.5
+version=1.6.0
 revision=1
 build_style=cmake
-hostmakedepends="automoc4"
-makedepends="kdelibs-devel kde-runtime-devel phonon-devel qjson-devel qoauth-devel telepathy-qt-devel"
+hostmakedepends="extra-cmake-modules pkg-config"
+makedepends="kcmutils-devel kdewebkit-devel kemoticons-devel knotifyconfig-devel
+ kparts-devel qca-qt5-devel qoauth-devel qt5-devel telepathy-qt5-devel"
```
pullmoll: Your are mixing Qt4 and Qt5 here: qoauth-devel is Qt4. There is no qoauth5-devel (yet) so either this needs to be removed or a qoauth5 package needs to be added.

pullmoll:
Currently won't build because kdewebkit-devel is waiting for qt5>5.8.0

PR #10998, opened 2018-01-20
cmake: update to 3.10.2

concatime:
Do you think we should change the link of `distfiles` to https?
PS: I’m new in Void Linux.


concatime:
Oh, there is already a pull request. https://github.com/voidlinux/void-packages/pull/10975

PR #11037, opened 2018-01-23
[test] http ~> https (common/environment/setup/misc.sh)

concatime:
Except:
- Ubuntu because no HTTPS available right now
- Debian because ERR_CERT_AUTHORITY_INVALID


maxice8:
ping

concatime:
Well, I have nothing to add, or maybe… we can change the link of GNU to the mirror which would redirect to _supposedly_ the nearest server (`https://ftpmirror.gnu.org/gnu`) and therefore, reduce the latency.

PR #11056, opened 2018-01-24
seafile-client-qt: add build option: shibboleth

Piraty:
if you don't like this option enabled by default, just chime in


yopito:
@Piraty : following issue #11067: if you want to trigger Travis CI builds, please force repush your commits (commits have to change).

For instance:
```
% git commit --amend --date $(date)
[don't change anything in commit]

% git push -f -u origin <your-branch>
```

Piraty:
@yopito 
it is a known fact that some arm* crosscompilation packages were broken in the last days, and i know my package works.

But since you are the maintainer of that package, are you ok with the added build option? because i need it

yopito:
@Piraty : I'm fine with it, let's go ahead.  
Quick test shows no regression on usage as before (without sibboleth use).

PR #11061, opened 2018-01-24
New package: kitty-0.7.0

SolitudeSF:
this package is a bit problematic. i added snippet from python3-module build style but it refuses to crosscompile.


srcpkgs/kitty/template
```diff
@@ -0,0 +1,59 @@
...
+if [ -n "$CROSS_BUILD" ]; then
+	hostmakedepends+=" python3"
+fi
+
+pre_build() {
+	case "$XBPS_TARGET_MACHINE" in
+		*-musl) sed -e 's;pow10;power10;g' -i kitty/parser.c
+			sed -e "s;<sys/sysctl\.h>;<linux/sysctl.h>;" \
+			-i kitty/shaders.c
+			;;
```
Vaelatern: why not make this an unconditional patch?

srcpkgs/kitty/template
```diff
@@ -0,0 +1,59 @@
...
+short_desc="A modern, hackable, featureful, OpenGL based terminal emulator"
+maintainer="SolitudeSF <solitudesf@protonmail.com>"
+license="GPL-3"
+homepage="https://github.com/kovidgoyal/kitty"
+distfiles="${homepage}/archive/v${version}.tar.gz"
+checksum=fef41783d20815cdab8f0b8a5d1c588e23589394472d25cc0a123069b8c13e44
+
+if [ -n "$CROSS_BUILD" ]; then
+	hostmakedepends+=" python3"
+fi
```
Vaelatern: Just keep it in `hostmakedepends` at all times.

srcpkgs/kitty/template
```diff
@@ -0,0 +1,59 @@
...
+pkgname=kitty
+version=0.7.0
+revision=1
+build_style=python3-module
+python_version=3
+hostmakedepends="pkg-config ncurses"
+makedepends="harfbuzz-devel zlib-devel ncurses-devel libXi-devel
+		python3-devel fontconfig-devel freetype-devel
+		startup-notification-devel libpng-devel libXrandr-devel
+		libXinerama-devel libXcursor-devel libxkbcommon-devel"
```
Vaelatern: Indentation is done with 1 space for these variables.

srcpkgs/kitty/template
```diff
@@ -0,0 +1,59 @@
+# Template file for 'kitty'
+pkgname=kitty
+version=0.7.0
+revision=1
+build_style=python3-module
+python_version=3
```
Vaelatern: Remove variable.

srcpkgs/kitty/template
```diff
@@ -0,0 +1,59 @@
...
+		LDFLAGS+=" -L${XBPS_CROSS_BASE}/${py3_lib} -L${XBPS_CROSS_BASE}/usr/lib"
+		CC="${XBPS_CROSS_TRIPLET}-gcc -pthread $CFLAGS $LDFLAGS"
+		LDSHARED="${CC} -shared $LDFLAGS"
+		env CC="$CC" CFLAGS="$CFLAGS" LDFLAGS="$LDFLAGS" \
+		LDSHARED="$LDSHARED" PYPREFIX="$PYPREFIX" \
+		python3 setup.py linux-package ${make_build_args}
+	else
+		python3 setup.py linux-package
+	fi
+}
```
Vaelatern: Just remove the `do_build` entirely.

srcpkgs/kitty/template
```diff
@@ -0,0 +1,59 @@
...
+
+do_install() {
+	vmkdir usr
+	vcopy linux-package/* usr
+}
```
Vaelatern: Does it not provide a way to install?

SolitudeSF: well, its `setup.py linux-package` is build + install combined, but i intentionally divided it in two steps

srcpkgs/kitty/template
SolitudeSF: well, its needed for musl, so i dont know what would be the implications of these on glibc

srcpkgs/kitty/template
SolitudeSF: if its not specified xbps-src converts shebangs to python2. is this how it supposed to work?

srcpkgs/kitty/template
SolitudeSF: then it fails to compile with ```/usr/bin/ld: cannot find libasan_preinit.o: No such file or directory
/usr/bin/ld: cannot find -lasan```

SolitudeSF:
also, for some reason travis cant even build x86_64 build even tho it shoud be able

srcpkgs/kitty/template
Vaelatern: Well, does it do `python3 setup.py build` then `python3 setup.py install`?

srcpkgs/kitty/template
Vaelatern: `libsanitizer-devel-6.3.0_3      /usr/lib/libasan_preinit.o`

srcpkgs/kitty/template
Vaelatern: Even for a `python3-module`? That would be a bug.

srcpkgs/kitty/template
Vaelatern: Musl is not a special case. Make the changes into a patch and they will be applied evenly.

srcpkgs/kitty/template
SolitudeSF: uhh, i dont know anything about python, but from what i understand, this setup.py isnt the standard setup.py that utilizes setuptools

Vaelatern: So provide a `do_install()` if you must.

SolitudeSF:
ok, i need some consulting. since this applies bunch of flags that i dont know meaning or importance of , how do i translate this to a void-intended way of invoking compiler? 

```python
def build_linux_launcher(args, launcher_dir='.', for_bundle=False, sh_launcher=False):
    cflags = '-Wall -Werror -fpie'.split()
    libs = []
    if args.profile:
        cflags.append('-DWITH_PROFILER'), cflags.append('-g')
        libs.append('-lprofiler')
    else:
        cflags.append('-O3')
    if for_bundle:
        cflags.append('-DFOR_BUNDLE')
        cflags.append('-DPYVER="{}"'.format(sysconfig.get_python_version()))
    elif sh_launcher:
        cflags.append('-DFOR_LAUNCHER')
    pylib = get_python_flags(cflags)
    exe = 'kitty-profile' if args.profile else 'kitty'
    cmd = [env.cc] + cflags + [
        'linux-launcher.c', '-o',
        os.path.join(launcher_dir, exe)
    ] + libs + pylib
    run_tool(cmd)```

**Comment by Vaelatern** [comment]
please post the latest version of the template so I can see the travis result from the above code.

**Comment by SolitudeSF** [comment]
this is almost finished except it doesnt build actual final binary, since i dont know how to properly

**Comment by Vaelatern** [comment]
Ideally the build would respect the CFLAGS and LDFLAGS environmental variables.

**Comment by SolitudeSF** [comment]
nah, cant figure it out. compiling with this command `	gcc ${CFLAGS} ${LDFLAGS} -I/usr/include/python3.6m -I/usr/include/python2.7 -lpython2.7 -lpython3.6m linux-launcher.c -o build/kitty
` fails with
```linux-launcher.c:(.text.startup+0x14f): undefined reference to `Py_DecodeLocale'
linux-launcher.c:(.text.startup+0x171): undefined reference to `Py_Main'
linux-launcher.c:(.text.startup+0x195): undefined reference to `PyMem_RawFree'```
i honestly have no clue whats going on here

File: srcpkgs/kitty/template (hunk 4 above)
Comment (Vaelatern) [reply srcpkgs/kitty/template]: @asergi any input?

File: srcpkgs/kitty/template (hunk 1 above)
Comment (Vaelatern) [reply srcpkgs/kitty/template]: Please make this into a patch file.

File: srcpkgs/kitty/template [hunk 7]
```diff
@@ -0,0 +1,74 @@
...
+			sed -e "s;<sys/sysctl\.h>;<linux/sysctl.h>;" \
+			-i kitty/shaders.c
+			;;
+	esac
+}
+
+post_build() {
+	python3 -m compileall -d build
+	cc -Wall -Werror -fpie -O3 -I/usr/include/python3.6m -I/usr/include/python3.6m linux-launcher.c -o build/kitty -lpthread -ldl -lutil -lm -L/usr/lib -lpython3.6m -Xlinker -export-dynamic
+}
```
Vaelatern: Just remove this.

srcpkgs/kitty/template
```diff
@@ -0,0 +1,74 @@
...
+	vmkdir usr/share/terminfo
+	vmkdir usr/share/icons/hicolor/256x256/apps
+
+	vcopy logo/kitty.rgba usr/lib/kitty/logo
+	vcopy logo/kitty.png usr/share/icons/hicolor/256x256/apps
+
+	tic -o ${DESTDIR}/usr/share/terminfo terminfo/kitty.terminfo
+	tic -o ${DESTDIR}/usr/lib/kitty/terminfo terminfo/kitty.terminfo
+
+	cat >${DESTDIR}/usr/share/applications/kitty.desktop <<EOF
```
Vaelatern: Do they not provide such a file?

SolitudeSF: no, they too write desktop file from script.

srcpkgs/kitty/template
SolitudeSF: this is part of the program, what should i substitute it for?

Vaelatern: Why is it not done by the build step?

Vaelatern: Their build script, that is.

SolitudeSF: the build step in this script is used for running from source so it doesnt perform every building step, for building everything as full package there is command `linux-package`. i guess the author didnt really care about packaging standarts, but for something like AUR or just compiling for personal use this certainly works. again, i dont really understand what is going on.
if this seems too out of void's standards i can close pr until something changes in building process upstream.

Vaelatern: can you at least not custom write compilation lines?

Vaelatern: That breaks on new python, on cross build, etc. etc. etc.

SolitudeSF: this is just placeholder, since i dont know how to do it other way
this package breaks on crossbuild with the developer intended way too

PR #11065, opened 2018-01-25
wget: update to 1.19.4

concatime:
\+ `tar.gz` ~> `tar.lz`


cr6git:
To extract `tar.lz` you need to add `hostmakedepends="lzip"`

You should use the `tar.xz` tarball instead.

concatime:
There is no xz in https://ftp.gnu.org/gnu/wget/

cr6git:
You're right, since version 1.19.1 they dropped `tar.xz`

concatime:
Damn, I was updating my comment :)

Vaelatern:
Confused. They provide https://ftp.gnu.org/gnu/wget/wget-1.19.4.tar.gz and why is that not good enough?

srcpkgs/wget/template
```diff
@@ -14,8 +14,8 @@ short_desc="The GNU wget download utility"
 maintainer="Juan RP <xtraeme@voidlinux.eu>"
 homepage="http://www.gnu.org/software/wget/wget.html"
 license="GPL-3"
-distfiles="${GNU_SITE}/wget/wget-$version.tar.gz"
-checksum=4f4a673b6d466efa50fbfba796bd84a46ae24e370fa562ede5b21ab53c11a920
+distfiles="${GNU_SITE}/$pkgname/$pkgname-$version.tar.lz"
```
Vaelatern: Don't change this. Variable substitution where it makes a line even longer is just ugly.

Vaelatern:
I'll leave this open so you can learn to squash commits and force push to the PR.

Vaelatern:
d6787daad4 but seriously, why the switch to lz?

concatime:
Because `.xz` is not available anymore. Tell this to GNU Softwares, not us.

Vaelatern:
We never were using `.xz`. We were using `.gz`.

concatime:
Take a look at https://github.com/voidlinux/void-packages/pull/8583/commits/6e5523df716c86b6873f87f286df49501e8d541c

cr6git:
@Vaelatern note: the `.gz` is double the size.
````
wget-1.19.4.tar.gz 4.1M
wget-1.19.4.tar.lz 2.0M
````

Vaelatern:
fair, we used to then.
Double the size, but at a small scale.

PR #11072, opened 2018-01-25
movit: update to 1.6.0

newbluemoon:


newbluemoon:
There’s a little conflict now with #11073. Perhaps it’s best if I removed the revbumps and then this PR is merged before the <i>mlt</i> update and the <i>shotcut</i> (which also depends on <i>mlt</i>) revbump is added to #11073?

jnbr:
I do not see the conflict. The mlt update should not require revbumps. Keep it at it is now and no matter what will be merged first, nothing should break.

newbluemoon:
But <i>movit</i> is a dependency of <i>mlt</i> and <i>shotcut</i>; <i>shotcut</i> itself depends also on <i>mlt</i>. With this update there comes a new .so version of libmovit and I do think this would break things and the order does matter. ;)

jnbr:
You're right. If #11073 gets merged first, mlt needs a revbump to `6.6.0_2` instead of `6.4.1_2`

newbluemoon:
Removed the revbumps. <i>shotcut</i> can be made a separate PR afterwards.
The order should be <i>movit</i> --> <i>mlt</i> --> <i>shotcut</i> to minimize the workload for the builders.

PR #11073, opened 2018-01-25
mlt: update to 6.6.0.

jnbr:


newbluemoon:
I don’t think you need to update `common/shlibs` because the .so version doesn’t change. As I understand it the second column in `shlibs` is the minimal required version of the package which provides the library in the first column.

jnbr:
thank you, fixed it

jnbr:
Thank you. I fixed it.

PR #11074, opened 2018-01-25
New package: bro-2.5.2

abenson:


abenson:
I can fix the musl issues. That's easy.

The cross compilation is going to be a bigger project later sometime.

PR #11084, opened 2018-01-26
New package: sterm-20171215

cr6git:


srcpkgs/sterm/template
```diff
@@ -0,0 +1,15 @@
+# Template file for 'sterm'
+pkgname=sterm
+version=20171215
+revision=1
+_githash=6d1729e4fb40b97752e79b79186f6f7059422e33
```
pullmoll: The last commit title is `Update changelog for new release` so there should be release tarballs, no?

cr6git: Inspired by [ttysterm PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=ttysterm), they use `git` but we don't do that, of course...
If what you mean is tarballs in the form `$pkgname-$version.tar.gz`, I don't know where to find them, I'm sorry.

cr6git: @pullmoll if you prefer, I can close this pull request. it's not a problem for me ;)

pullmoll: If there are no release tarballs at all then using a specific git hash and a date version is ok.

cr6git: cool 🙆

Neffistopheles:
I'm not sure on the proper etiquette for _deleting_ a package, but there's several problems here,

- sterm does provide releases. The latest is 0.7, so there's no need for the -git version. see https://dl.suckless.org/st/st-0.7.tar.gz
- There's no -git suffix here
- sterm is already packaged by its proper name, st, so this is a duplicate template. See template for "st"
- The existing package is already the latest version, and already handles terminfo and config.h correctly, which this template does not

cr6git:
@Neffistopheles You came out of nowhere, in the middle of the night...

Before opening a pull request to remove the package, you should have wait our answer.

As I pointed out, the two projects look different.

cr6git:
@Neffistopheles Perhaps you should advise the maintainer of the AUR [package](https://aur.archlinux.org/packages/ttysterm/) that he made the same mistake as me?

PR #11089, opened 2018-01-26
New package: perl-Svsh-1.002000.

maxice8:


pullmoll:
None of these packages are noarch=yes? I can't imagine that's true.

maxice8:
That is true, big oversight by me.

I'll fix this when I get home at midnight GMT -3.

My notebook HDD gave up so I can't work on Voidlinux from my uni anymore.

pullmoll:
NP. I just wanted to work on the pile of PRs again, from older to newer, and this was the next one.

maxice8:
On Tue, Feb 06, 2018 at 08:02:54PM +0000, Jürgen Buchmüller wrote:
> None of these packages are noarch=yes? I can't imagine that's true.
> 
> -- 
> https://github.com/voidlinux/void-packages/pull/11089#issuecomment-363547530

it is done

Issue #7184, opened 2017-07-29
appdata-tools: conflicts with appstream-glib


ebfe:
```
appdata-tools-0.1.7_1.x86_64 <-> appstream-glib-0.6.13_1.x86_64:
	/usr/share/aclocal/appdata-xml.m4
```

PR #11091, opened 2018-01-26
appstream-glib: update to 0.7.5.

maxice8:


cr6git:
And while you're at it, what about [#7184](https://github.com/voidlinux/void-packages/issues/7184) ?

PR #11095, opened 2018-01-26
New packages: arcan, durden and tools

Piraty:
https://arcan-fe.com/
https://github.com/letoram/arcan

This contribution is accepted and guided by arcan's author @letoram.

`aclip` and `aloadimage` are currently contained in the source of the arcan project, and will evolve into their own projects once they mature enough (according to the author). Therefore i made separate templates, but based on the same distfile as arcan itself.


srcpkgs/arcan/template
```diff
@@ -0,0 +1,55 @@
...
+ -DSTATIC_OPENAL=Off -DSTATIC_FREETYPE=Off
+ -DVIDEO_PLATFORM=egl-dri -DENABLE_SIMD=On -DENABLE_LTO=Off"
+
+makedepends=" libgbm libGL libdrm-devel MesaLib-devel libEGL LuaJIT-devel
+ freetype-devel sqlite-devel libusb-devel libxkbcommon-devel harfbuzz-devel
+ libopenal-devel vlc-devel ffmpeg-devel bzip2-devel liblzma-devel
+ libvncserver-devel  $(vopt_if tesseract tesseract-ocr-devel)"
+
+build_options="tesseract"
+desc_option_udev="Enable support for OCR via tesseract"
```
Johnnynator: You probably want to write tesseract instead of udev.

Piraty: right. copy-paste issue. thanks

PR #11111, opened 2018-01-27
gnome-builder: update to 3.26.3.

maxice8:


pullmoll:
`flatpak-builder` is broken for musl. This needs to be fixed or marked as `broken=<logfile>`

maxice8:
On Wed, Feb 07, 2018 at 09:21:56AM -0800, Jürgen Buchmüller wrote:
> `flatpak-builder` is broken for musl. This needs to be fixed or marked as `broken=<logfile>`
>

Noted, the failure is the same that flatpak has, i will just rebase the patch.

maxice8:
also noted lots of makedepends should be in hostmakedepends.

maxice8:
ok i updated, let's wait travis.

PR #11124, opened 2018-01-27
papirus-icon-theme: update to 20180116.

maxice8:


cr6git:
They changed the LICENSE to GPL-3, see [#1049](https://github.com/PapirusDevelopmentTeam/papirus-icon-theme/issues/1049#issuecomment-356685734).

PR #11129, opened 2018-01-27
libbsd: update to 0.8.7.

maxice8:


pullmoll:
This required a patch for musl. It'd be nice if besides just updating to a newer version you could also verify that the update builds for the default targets, i.e. our builders :)

PR #11130, opened 2018-01-27
crypto++: update to 600.

maxice8:


pullmoll:
crypto++610 annot be merged because this breaks MEGAsdk, also the latest v3.3.1

PR #11135, opened 2018-01-27
perl-Mail-Transport: update to 3.002.

maxice8:


pullmoll:
I see no libs or binaries being compiled - can't this be noarch=yes?

maxice8:
On Sat, Feb 24, 2018 at 03:41:54AM -0800, Jürgen Buchmüller wrote:
> I see no libs or binaries being compiled - can't this be noarch=yes?
> 
> —
> 

yes, added, this comes from before i added noarch=yes as a default on my `xnewperl` script

PR #11168, opened 2018-01-27
dvtm: add support for compile-time config file

copyleftie:


Duncaen:
Whats the point of putting the default config file in the repository.
This just makes updates harder, you have to check every time if something is changed.
Please just check if there is a config.h in the files dir and copy it.
Just because some teamplates do it doesn't mean its good.

PR #11180, opened 2018-01-28
micro: update to 1.4.0.

maxice8:


cr6git:
`homepage="https://micro-editor.github.io"`

PR #11181, opened 2018-01-28
calc: update to 2.12.6.4.

maxice8:


Chocimier:
2.12.6.4 is [unstable](http://www.isthe.com/chongo/src/calc/).  What about updating to stable 2.12.6.5?

PR #11182, opened 2018-01-28
ncurses: update to 6.1.

maxice8:


chneukirchen:
**Fallout**

Rebuild required:
- [x] bash
- [x] yash
- [x] rc (with static)

Other issues:
- tmux requires server restart
- potential slang fallout (couldn't reproduce yet: https://bugs.archlinux.org/task/57300)

PR #11188, opened 2018-01-28
intellij-idea-community-edition: update to 2017.3.3.

asiekierka:


srcpkgs/intellij-idea-community-edition/template
```diff
@@ -37,13 +37,15 @@ do_install() {
 		rm ${DESTDIR}/usr/share/intellij-idea/bin/fsnotifier
 		rm ${DESTDIR}/usr/share/intellij-idea/bin/idea.vmoptions
 		rm ${DESTDIR}/usr/share/intellij-idea/lib/libpty/linux/x86/libpty.so
+		rm -d ${DESTDIR}/usr/share/intellij-idea/lib/libpty/linux/x86
```
asiekierka: This line (and the matching one) was added to prevent a warning from the 03-remove-empty-dirs hook.

pullmoll: You can safely ignore ths warnings from 03-remove-empty-dirs. They are purely informational and many packages leave empty directories in the tree, especially after e.g. having moved files to subpackages.

asiekierka: Alright! In that case, I have corrected my PR.

PR #11192, opened 2018-01-28
New package: gnucobol-2.2_1

dertuxmalwieder:


pullmoll:
Since this is a compiler I think it would make sense to split the binary and the libs and development files into subpackages like `gnucobol-libs` and `gnucobol-devel`.

+ `gnucobol` depends on `gnucobol-libs`, i.e. just the `libcob.so.4.0.0`
+ `gnucobol-libs` does just  `vmove usr/lib/*.so.*` in its `pkg_install()`
+ `gnucobol-devel` depends on `gnucobol-libs` as well and `vmove usr/include`, `vmove usr/lib/*.so` and `vmove usr/lib/*.a` files, perhaps `vmove usr/bin/cob-config` and also  `vmove usr/share/gnucobol` files?
+ add a common/shlibs entry `libcob.so.4.0.0 gnucobol-libs-2.2_1`

I haven't tried, so do binaries compiled with `cobc` need `cobrun` to be executed, or just the `libcob.so.4.0.0`, if they are shared and nothing at all for a static link?

dertuxmalwieder:
@pullmoll Good points - this was my first try though and I'll still have a lot to learn about xbps packaging, obviously. :-(

It would make more sense like that ^ indeed... mind to help? (I'll be "out of office" until Monday -- sorry.)

pullmoll:
First never use `git merge …` to merge `voidlinux/void-packages/master` (a.k.a. `upstream`) into a branch but always use `git pull --rebase`. After doing your changes, perhaps rebasing using `git rebase -i origin/master` and committing the changes, then use `git push -f` to force update the HEAD of your branch.

Now that I think about it, everything but the shared library should probably be kept in the main package. Programs which link against the shared library should not install the compiler, headers, etc., though, so just a `gnucobol-libs` subpackage should suffice:
```
# Template file for 'gnucobol'
pkgname=gnucobol
version=2.2
revision=1
build_style=gnu-configure
makedepends="gmp-devel db-devel"
short_desc="Free COBOL compiler (formerly OpenCOBOL)"
maintainer="Cthulhux <git@tuxproject.de>"
license="GPL-3"
homepage="https://sourceforge.net/projects/open-cobol/"
distfiles="${SOURCEFORGE_SITE}/open-cobol/gnu-cobol/${version}/${pkgname}-${version}.tar.xz"
checksum=dc18fc45c269debfe86a4bbe20a7250983cba6238ea1917e135df5926cd024a0
depends="gnucobol-libs>=${version}_${revision}"

gnucobol-libs_package() {
        short_desc+=" - runtime libraries"
        pkg_install() {
                vmove usr/lib/*.so.*
                vmove usr/lib/gnucobol
        }
}
```

And somewhere in `common/shlibs` the following line must be added:
```
libcob.so.4.0.0 gnucobol-libs-2.2_1
```
If you don't do it at the end but somewhere else chances are higher that your PR won't see a merge conflict too soon.

dertuxmalwieder:
Ah, git is not really my favorite VCS. I always lose myself in merge/pull/commit conflicts. :-(

PR #11206, opened 2018-01-29
sterm: Remove duplicate package/template.

Neffistopheles:
sterm is already packaged under its proper name "st", which is already up to
date with a well-formed template.

Furthermore, the duplicate template has a few issues as outlined https://github.com/voidlinux/void-packages/pull/11084#issuecomment-361118485


maxice8:
Installing both here, they seem very different

st opens st

sterm returns this

```
No device specified
Usage: sterm [options] <device>
Options:
  -b <duration> send break signal
  -c        enter command mode
  -d[PULSE] make pulse on DTR
  -e        ignore '~.' escape sequence
  -n        do not switch the device to raw mode
  -r[PULSE] make pulse on RTS
  -s <baudrate>
  -v        verbose mode

PULSE is a number specifying the pulse. Absolute value defines the
length of the pulse in milliseconds, sign determines the polarity of
the pulse. Alternatively, PULSE can be either '+' or '-', which
corresponds to +1 or -1.
```

cr6git:
@Neffistopheles Are you sure these two projects are the same?

https://git.suckless.org/st

https://rtime.felk.cvut.cz/gitweb/sojka/sterm.git

They look different.

Vaelatern:
@Neffistopheles seems the burden of proof is on you, can you confirm they are the same project?

cr6git:
@Vaelatern Seems this guy is just a **troll**. Please close this.

pullmoll:
st and sterm are different packages.

PR #11218, opened 2018-01-29
hugo: fix dependencies and license

dleeram:
Chroma is used by [default](https://gohugo.io/content-management/syntax-highlighting), but can still use optionally Pygments.
Their license changed to Apache 2.0.


maxice8:
0.37 was released, including the update with the fixes would be nice.

pullmoll:
Can you please rebase @dleeram ?

PR #11229, opened 2018-01-29
New package: wordgrinder-0.7.1

abenson:


pullmoll:
This does not build (anymore?) for me:
```
=> wordgrinder-0.7.1_1: running do_build ...
Bootstrapping build
/bin/sh: o: command not found
make: [Makefile:149: /tmp/wg-build/lua] Error 127 (ignored)
/bin/sh: /tmp/wg-build/lua: No such file or directory
make: *** [Makefile:116: /tmp/wg-build/build.ninja] Error 127
=> ERROR: wordgrinder-0.7.1_1: failed to run do_build() at line 9.
```

abenson:
This isn't building for me either, wth.

pullmoll:
This doesn't cross build because the scripts try to run tests with the cross built binary.
Is there perhaps a `--disable-tests` like switch which could be defined when cross compiling?

pullmoll:
Remove the `pre_configure()` step. It is not necessary because the builders are clean when a package is built.

abenson:
It's not an issue with the builders, it's a required step for the package.

It might not be required, author got back to me about building without testing. Going to try that and see if it a) fixes the cross compile b) negates the need for the extra clean.

pullmoll:
Ok, I'll wait until you're done with the template.

abenson:
Split out the X11 version to a sub-package and was able to reduce the number of required patches.

pullmoll:
I don't know why it worked for you, it does not for me:
```
=> wordgrinder-0.7.1_1: running do_build ...
make: '/tmp/wg-build/build.ninja' is up to date.
ninja: error: unknown target 'bin/xwordgrinder-builtin-x11-release'
=> ERROR: wordgrinder-0.7.1_1: failed to run do_build() at line 19.
```

abenson:
Try this now...

It uses system Lua, doesn't build in /tmp...

PR #11248, opened 2018-01-30
New package: fortune-mod-zh-2.6

cr6git:


cr6git:
Why Travis doesn't start?

cr6git:
Weird.

PR #11250, opened 2018-01-30
ranger: update to 1.9.0

NoctuaNivalis:


maxice8:
https://github.com/voidlinux/void-packages/pull/11212/files

NoctuaNivalis:
Wow, I'm slow. I though checking the last page would be enough.

PR #11255, opened 2018-01-30
flatpak: update to 0.11.1

maxice8:


maxice8:
xdg-desktop-portal needs to be updated in lockstep according to upstream, also new xdg-desktop-portal needs libpipewire, please see #11123

maxice8:
do not merge until commit with pipewire update to 0.1.8 (required by xdg-desktop-portal) is added.

Vaelatern:
odd, it seems to build.

maxice8:
it has --disable-pipewire, if pipewire is enabled then it isn't enough because pipewire 0.1.8 is required

maxice8:
i'm currently working on bumping pipewire to 0.1.8 but it is a glibc fest of `error.h` and `error()`

maxice8:
There we go, the patches to fix musl compat amount to `269` lines.

PR #11270, opened 2018-01-31
New package: fgit-1.0.2

cr6git:
Based on [PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=fgit).


srcpkgs/fgit/template
```diff
@@ -0,0 +1,28 @@
+# Template file for 'fgit'
+pkgname=fgit
+version=1.0.2
+revision=1
+noarch=yes
+create_wrksrc=yes
```
pullmoll: Remove this...

srcpkgs/fgit/template
```diff
@@ -0,0 +1,28 @@
...
+build_wrksrc="${pkgname}"
```
pullmoll: ...and this...

srcpkgs/fgit/template
```diff
@@ -0,0 +1,28 @@
...
+checkdepends="git"
+depends="bash git ncurses"
+short_desc="Run a Git command in several repositories"
+maintainer="cr6git <quark6@protonmail.com>"
+homepage="https://github.com/l0b0/fgit"
+license="GPL-3"
+distfiles="https://github.com/l0b0/fgit/archive/${version}.tar.gz"
+checksum=c14cb8b076ccdac7f05f883cc48af48bebacba5460d0420ed06580bb13e04784
+
+post_extract() {
```
pullmoll: ...and this function.
What did you want to achieve? The tarball contains a `$pkgname-$version` directory and there is no reason to fiddle with that AFAICS.

cr6git: Because of a [Makefile](https://github.com/l0b0/fgit/blob/master/Makefile) weirdness...
If you don't change `$pkgname-$version` to -> `$pkgname` , the program will be installed like that:
````
/usr/share/fgit-1.0.2/warning.sh
/usr/share/fgit-1.0.2/error.sh
/usr/share/fgit-1.0.2/usage.sh
/usr/share/fgit-1.0.2/variables.sh
/usr/bin/fgit-1.0.2
````
But this is not what we want.
And I used the variable `build_wrksrc` just because `wrksrc` cannot be changed during the build!

newbluemoon: Just have a look at the `Makefile`: You can also set the `name` variable in `do_install()` like you do with `PREFIX` and `DESTDIR`: 

    make name=${pkgname} PREFIX=${DESTDIR}/usr DESTDIR=/usr install

Then you won’t need `create_wrksrc`, `build_wrksrc`, and `post_extract()`.

cr6git: Thanks @newbluemoon ! 😃

cr6git: @pullmoll I think it's good know.

PR #11275, opened 2018-01-31
Openscad revbump

pienjo:
The recent upgrade of mpfr broke the openscad package, as this still depends on the old library (libmpfr.so.4). A rebuild fixes this problem.


pienjo:
Argh, I messed up my pull requests, I'll retry.

PR #11277, opened 2018-01-31
linux4.9: update to 4.9.79.

pbui:


cr6git:
See [3e4e64a](https://github.com/voidlinux/void-packages/commit/3e4e64a4bb119b7757965998c293b2917369a38c).

PR #11296, opened 2018-02-01
New package: just-0.3.7

TeddyDD:
[Just](https://github.com/casey/just) is command runner with syntax inspired by make.


TeddyDD:
Looks like @wpbirney has already added it ¯\_ツ_/¯

Issue #8242, opened 2017-10-17
gnome-documents fails to start


maxice8:
```
$ gnome-documents
Unsatisfied dependency: GdPrivate
```

How to reproduce:

1. Install gnome-documents ( xbps-install -S gnome-documents )
2. Try to run it

Gottox:
maybe related to #8182.

maxice8:
Thanks for the pointer, i'll try again when #8121 effects hit the repo

maxice8:
No, the addition of tracker-miners doesn't seem to fix the problem

greywolf01:
Over here no problems opening up Documents.

maxice8:
On Tue, Oct 17, 2017 at 08:19:19PM +0000, greywolf01 wrote:
> Over here no problems opening up Documents.
> 
> -- 
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/voidlinux/void-packages/issues/8242#issuecomment-337357713

Alright, i'll assume it's due to me not running a full gnome-environment ( gnome-shell )

Gottox:
Actually, this is still an issue. gnome-documents should have all dependencies that are needed to run it.

PR #11300, opened 2018-02-01
gnome-documents: update to 3.26.2.

maxice8:


cr6git:
@maxice8 And what about [#8242](https://github.com/voidlinux/void-packages/issues/8242) ?

PR #11304, opened 2018-02-01
gifski: update to 0.7.4.

maxice8:


cr6git:
[#11411](https://github.com/voidlinux/void-packages/pull/11411)

PR #11318, opened 2018-02-02
glibc: update to 2.27.

maxice8:


chneukirchen:
I think `--enable-static-pie` would be good.

chneukirchen:
Also we should use `--enable-kernel=3.2` which is the lowest kernel actually supported by this glibc.

maxice8:
On Fri, Feb 02, 2018 at 01:07:38PM +0000, Leаh Neukirchen wrote:
> I think `--enable-static-pie` would be good.
> 
> -- 
> https://github.com/voidlinux/void-packages/pull/11318#issuecomment-362581461

it also needs gcc 8

chneukirchen:
@pullmoll can you do a quick test run with this glibc? changelog sounds rather boring.

pullmoll:
I can try a bootstrap and the usual base system builds for x86_64 and i686, yes. Will probably take a couple of hours, though.

chneukirchen:
No hurry, I just don't want to bump glibc blindly. :)

pullmoll:
Of course :) Should we perhaps update to binutils-2.30 at the same time? We'll have to update the cross compilers anyway.

pullmoll:
Nope, won't work without additional patching:
```
=> gcc-7.3.0_1: running do_configure ...
mkdir: relocation error: /tmp/void/glibc-1/usr/lib/libc.so.6: symbol _dl_exception_create, version GLIBC_PRIVATE not defined in file ld-linux-x86-64.so.2 with link time reference
=> ERROR: gcc-7.3.0_1: failed to run do_configure() at line 204.
```

pullmoll:
I tried with `export CFLAGS+=" -DNO_RTLD_HIDDEN"` in `do_configure()` but then building glibc fails.

The missing symbol in the listing of `nm /usr/lib/libc.so.6` is:
```
         U _dl_exception_create@@GLIBC_PRIVATE
```
This line is not present in the output of `nm /usr/lib/libc.so.6` of `glibc-2.26`.
So it seems as if this symbol should have been resolved internally?

ebfe:
Misses common/shlibs update.

maxice8:
ebfe: added.

PR #11336, opened 2018-02-02
New package: pipes.sh-1.3.0

abenson:


maxice8:
`build_style=meta` are for packages that don't install anything but just pull other packages, just remove it, you provide a do_install() function so it will work.

abenson:
Declined via IRC.

PR #11348, opened 2018-02-02
New Package: arc-kde

jambonmcyeah:


cr6git:
[#7634](https://github.com/voidlinux/void-packages/pull/7634)

PR #11352, opened 2018-02-03
New package: arc-theme-20170302

jambonmcyeah:


cr6git:
[#7634](https://github.com/voidlinux/void-packages/pull/7634)

jambonmcyeah:
@cr6git I know but other theme packages like #10491 #10492 got merged so I'm going to give it a try

cr6git:
@Vaelatern I think you have to clarify the situation about the themes. Thank you.

jambonmcyeah:
@Vaelatern Please allow the packaging of themes because it's so much more convenient for the users compared to manually downloading themes. It's allowed in every other distribution and why not void?

PR #11358, opened 2018-02-03
New package: xgalaga++-0.9

abenson:
xgalaga++ seems better maintained than xgalaga.


srcpkgs/xgalaga++/template
```diff
@@ -0,0 +1,26 @@
+# Template file for 'xgalaga++'
+pkgname=xgalaga++
+version=0.9
+revision=1
+build_style=gnu-makefile
+make_build_target="all"
+high_scores_dir="/var/games/xgalaga++"
```
maxice8: custom variables should be prefixed with _

srcpkgs/xgalaga++/template
```diff
@@ -0,0 +1,26 @@
...
+short_desc="A classic single screen vertical shoot ’em up"
+maintainer="Andrew Benson <abenson+void@gmail.com>"
+license="GPL-3"
+homepage="http://marc.mongenet.ch/OSS/XGalaga/"
+distfiles="http://marc.mongenet.ch/OSS/XGalaga/xgalaga++_${version}.tar.gz"
+checksum=c8db4bb9808c1cbbd001533b08318f6ab76b877ab6ed313d977d6a8546cc7fc9
+
+post_extract() {
+	sed -i 's|$(installprefix)/games|$(installprefix)/bin|' Makefile
+	sed -i 's|EXE_GROUP=games|EXE_GROUP=users|' Makefile
```
maxice8: Optional: replace all sed calls with just 3 `-e` expressions and 1 sed call

abenson:
Mistakenly deleted wrong branch.

PR #11359, opened 2018-02-03
liborcus: update to 0.13.2.

maxice8:


pullmoll:
Nope. 0.13.x won't work (yet) with libreoffice, which is the only consumer.

PR #11361, opened 2018-02-03
wine: update to 3.1.

cr6git:
**Note:**
`libX11-devel` was redundant, already pulled by the other `libX...-devel` packages.
`dbus-devel` was duplicate in `makedepends`.


cr6git:
[a8b8ce3](https://github.com/voidlinux/void-packages/commit/a8b8ce386dff0cbfe7a1e2a3c34be9093273bbb7)

PR #11367, opened 2018-02-03
widelands: multiple packaging fixes

Hoshpak:
- widelands looks for it's data relative to the path were the binary resides,
  use an absolute path instead so the data will be found
- wl_render_richtext was installed in /usr, move it to /usr/bin
- COPYING, CREDITS, Changelog and VERSION are installed in /usr, remove them
  before packaging

CC @pullmoll


pullmoll:
Thank you!

PR #11368, opened 2018-02-03
New package: iasl-20180105

tim241:


srcpkgs/iasl/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'iasl'
+pkgname=iasl
+version=20180105
+revision=1
+wrksrc="acpica-unix-$version"
+only_for_archs="i686 x86_64"
+build_style=gnu-configure
+make_build_args="iasl"
+hostmakedepends="gcc flex bison"
+makedepends="gcc flex bison"
```
maxice8: There is no need to mark GCC as a dependency it is automatically in

srcpkgs/iasl/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'iasl'
+pkgname=iasl
+version=20180105
+revision=1
+wrksrc="acpica-unix-$version"
+only_for_archs="i686 x86_64"
```
maxice8: Please add a comment with the reason why making this exclusive to those arches ( something brief like,  `only arches supported by upstream` )

srcpkgs/iasl/template
```diff
@@ -0,0 +1,19 @@
...
+hostmakedepends="gcc flex bison"
+makedepends="gcc flex bison"
+short_desc="An ACPI tool"
+maintainer="Tim Wanders <timwanders241@gmail.com>"
+license="GPL-2,Intel"
+homepage="https://www.acpica.org"
+distfiles="https://acpica.org/sites/acpica/files/acpica-unix-$version.tar.gz"
+checksum=414f843ac6c7c53bbd2a830b092a2a27c49172b0878fd27fe306dd526b78b921
+do_configure() {
+	make clean
```
maxice8: This looks weird, why override the configure phase ?

srcpkgs/iasl/template
```diff
@@ -0,0 +1,19 @@
...
+revision=1
+wrksrc="acpica-unix-$version"
+only_for_archs="i686 x86_64"
+build_style=gnu-configure
+make_build_args="iasl"
+hostmakedepends="gcc flex bison"
+makedepends="gcc flex bison"
+short_desc="An ACPI tool"
+maintainer="Tim Wanders <timwanders241@gmail.com>"
+license="GPL-2,Intel"
```
maxice8: Style: add a space between licenses, if the license isn't found in /usr/share/licenses please add a vlicense.

srcpkgs/iasl/template
tim241: There is no `configure` file, only a `Makefile`

maxice8: Then the build_style should be gnu-makefile

srcpkgs/iasl/template
tim241: There is no `LICENSE` file, only on their site, there is license information about the source weirdly enough.
More info here: https://acpica.org/Licensing

srcpkgs/iasl/template
cr6git: `makedepends="gcc flex bison"`
What is that??
These are `hostmakedepends`, not `makedepends` !

tim241: okay, did not know that, thanks!

srcpkgs/iasl/template
maxice8: I'd just pick the GPL-2 license, the licensing fluff gives us the option to pick any of the three, the BSD ones require we also ship a LICENSE File.

tim241: okay

srcpkgs/iasl/template
cr6git: It's the same on the [PKGBUILD](https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/iasl#n26), however.

tim241: okay

maxice8: @cr6git yeah but that should be either pre_build() or post_extract()

cr6git: @maxice8 oh sorry, I should have reload the page in my browser, in order to see your last comment. 😉

srcpkgs/iasl/template
```diff
@@ -0,0 +1,19 @@
...
+maintainer="Tim Wanders <timwanders241@gmail.com>"
+license="GPL-2"
+homepage="https://www.acpica.org"
+distfiles="https://acpica.org/sites/acpica/files/acpica-unix-$version.tar.gz"
+checksum=414f843ac6c7c53bbd2a830b092a2a27c49172b0878fd27fe306dd526b78b921
+pre_build() {
+	make clean
+}
+do_build() {
+	make
```
maxice8: If your build_style is gnu-makefile then do_build already does make for you.

tim241: if I try to remove that, the build breaks, this resolves the build issue

maxice8: log ? it shouldn't break unless some rogue variable is set.

tim241: here's the build without the `do_build` part: https://gist.github.com/3fee6425e36fcb139cc3c0aa6144e4eb

maxice8:
On Sat, Feb 03, 2018 at 03:36:37AM -0800, Tim Wanders wrote:
> 
> here's the build without the `do_build` part: https://gist.github.com/3fee6425e36fcb139cc3c0aa6144e4eb

Taking a look into it

srcpkgs/iasl/template
tim241: and here's the build with the `do_build` part: https://gist.github.com/10384511ef35d0f6aa5acb73fed48b48

maxice8: yeah, looks fine to me.

cr6git:
@maxice8 Probably needs `pciutils-devel` ?

tim241:
I didn't see acpica-utils, so this package is unrequired, closing!

Issue #11006, opened 2018-01-20
java sym links not updated


chriscamacho:
### System

* xuname:  
 Void 4.14.14_1 x86_64 GenuineIntel uptodate rrrmF
* package:  
  oracle-jdk 8u162_1

### Expected behavior
java issued from the command line should execute java

### Actual behavior
 instead it says java not found
### Steps to reproduce the behavior
update oracle-jdk
run java

### workaround
xbps-alternatives -g java -s oracle-jdk

additionally JAVA_HOME is not set properly the JAVA_HOME=${JAVA_HOME.... trick doesn't seem to work, removing ${JAVA_HOME and closing } fixes this

cr6git:
> JAVA_HOME is not set properly

This was discussed [here](https://forum.voidlinux.eu/t/java-home-incorrect/290) a long time ago.

chriscamacho:
> This was discussed here a long time ago.
yeah I started the thread, nothing resolved, and JAVA_HOME still not properly set until I fix the profile script every time java updates...

cr6git:
> until I fix the profile script

Please could you describe (precisely) how you fixed it?

chriscamacho:
> removing ${JAVA_HOME and closing } fixes this

so

`JAVA_HOME=${JAVA_HOME......}`

becomes

`JAVA_HOME=......`

please don't let this side issue distract from the main issue of xbps-alternatives symlinks not being updated properly....

cr6git:
I think the real culprit is `openjdk-jre` , see [openjdk/template#L203](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/openjdk/template#L203).

chriscamacho:
i *might* have had the alternatives set to openjdk before the update but can't be sure to be honest

cr6git:
I think we should modify [template#L203](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/openjdk/template#L203) to **force** the creation of a new symlink:
````
ln -sf $_openjdk_version ${PKGDESTDIR}/usr/lib/jvm/openjdk
````

chriscamacho:
is there no way to do

`xbps-alternatives -g java -s <users last choice>`

I have oracle and open versions installed for testing, it would be polite to update the alternatives links to provide the same selection as before the update...

chriscamacho:
just to be clear, it was all the links to each java tool that were broken

[...]

PR #11371, opened 2018-02-03
oracle-jdk: fix symlinks & JAVA_HOME

cr6git:
Trying to resolves [#11006](https://github.com/voidlinux/void-packages/issues/11006), using exactly the same correction as [#11047](https://github.com/voidlinux/void-packages/pull/11047).


ebfe:
Runnin xbps-alternatives in the INSTALL script changes the alternative back to oracle-jdk on every update.

cr6git:
If you have a better idea, no problem.

ebfe:
Actually fixing the bug in xbps-alternatives instead of layering more workarounds that create problems elsewhere might be an idea. This is probably already fixed in https://github.com/voidlinux/xbps/pull/193.

cr6git:
I didn't know that there was a bug in `xbps-alternatives`, but I trust you.

So, maybe you should close this pull request, revert [#11047](https://github.com/voidlinux/void-packages/pull/11047), and fix this bug once and for all.

Me I have no idea what more I can do, I'm sorry.

ebfe:
So the above patch is already in srcpkgs/xbps/patches so it obvisouly doesn't fix the issue at hand. However my point still stands that this should be fixed in xbps-alternatives (not updating changed symlink targets at an update). No need to be sorry. 

Since you moved stuff to an unversioned directory things should work on the next update without the extra xbps-alternatives invocation.

cr6git:
Yes, moved stuff to an unversioned directory, as it the case for [openjdk](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/openjdk/template#L74) and [openjdk-jre](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/openjdk/template#L187) for example.
I thought it was correct.

OK, I will `--amend` this commit to remove the extra `xbps-alternatives` invocation.

ebfe:
> Yes, moved stuff to an unversioned directory, as it the case for [openjdk](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/openjdk/template#L74) and [openjdk-jre](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/openjdk/template#L187) for example.
> I thought it was correct.
> 
> OK, I will `--amend` this commit to remove the extra `xbps-alternatives` invocation.

Only running it when oracle-jdk is actually the currently set
alternative might be an option too. But i'm not sure if the alternatives
are setup correctly at the time the INSTALL script runs.

cr6git:
No, no, I was wrong, it doesn't work, now the update is broken:
````
Name       Action    Version           New version            Download size
oracle-jdk update    8u161_1           8u162_2                - 

Free space on disk:           4091MB

Do you want to continue? [Y/n] Y

[*] Downloading binary packages

[*] Verifying package integrity
oracle-jdk-8u162_2: verifying SHA256 hash...

[*] Running transaction tasks
oracle-jdk-8u161_1: updating to 8u162_2 ...
 ...
ERROR: oracle-jdk-8u162_2: [unpack] failed to extract file `./usr/lib/jvm/oracle-jdk': Le dossier n'est pas vide
ERROR: oracle-jdk-8u162_2: [unpack] failed to extract files: Le dossier n'est pas vide
ERROR: oracle-jdk-8u162_2: [unpack] failed to unpack files from archive: Le dossier n'est pas vide
````
😭

cr6git:
ok, fixed what I wrecked yesterday... by modifying INSTALL, to add an `elif`:
````
if [ -d usr/lib/jvm/jdk$ORACLE_JDK_LONG_VERSION ]; then
	rm -rf usr/lib/jvm/jdk$ORACLE_JDK_LONG_VERSION
elif [ -d usr/lib/jvm/oracle-jdk ]; then
	rm -rf usr/lib/jvm/oracle-jdk
fi
````

cr6git:
@newbluemoon of course! stupid me!
````
mv jdk$ORACLE_JDK_LONG_VERSION usr/lib/jvm/oracle-jdk || return 1
````
obviously completely wrong!
I reverted this to the original:
````
mv jdk$ORACLE_JDK_LONG_VERSION usr/lib/jvm/jdk$ORACLE_JDK_LONG_VERSION || return 1
````

and now it's ok:
````
$ ls /usr/lib/jvm -al
total 64
drwxr-xr-x  3 root root  4096  4 févr. 13:59 .
drwxr-xr-x 67 root root 49152  4 févr. 13:57 ..
drwxr-xr-x  8   10  143  4096 20 déc.  05:52 jdk1.8.0_162
lrwxrwxrwx  1 root root    12  4 févr. 10:10 oracle-jdk -> jdk1.8.0_162
-rw-r--r--  1 root root   262  4 févr. 10:10 oracle-jdk-vars.sh
````

pullmoll:
Hmm.. I didn't see this, just had to update because 8u162 was no longer available.
Is it ok now?

PR #11373, opened 2018-02-03
New package: xnp2-0.86

hipperson0:


cr6git:
````
wrksrc=${pkgname}-${version}/x11
````
Travis complains about that: `template:5: wrksrc should be a top-level directory`
Maybe what you mean is:
````
build_wrksrc=x11
````

**Comment by hipperson0** [comment]
I tried that but then it fails to apply the patch (Maybe something is wrong with my patch?).

```
=> xnp2-0.86_1: patching: define_cplusplus.patch.
The text leading up to this was:
--------------------------
|--- compiler.h.old	2018-02-02 17:00:29.174895866 +0000
|+++ compiler.h	2018-02-02 17:00:41.199160781 +0000
--------------------------
File to patch: 
Skip this patch? [y] 
1 out of 1 hunk ignored
=> ERROR: xnp2-0.86_1: failed to run post-extract_00-patches() at line 34.

```

PR #11376, opened 2018-02-03
skype: update to 8.17.76.3

lonwillett:


cr6git:
@lonwillett update to 8.17.76.3

PR #11382, opened 2018-02-04
mu/mu4e: update to 1.0

emacsomancer:
also bump xapian-core to rebuild with latest gcc in order to avoid errors building mu/mu4e


chneukirchen:
Why is nopie needed? Builds fine here.

emacsomancer:
@chneukirchen For some reason my local build failed without it.

Johnnynator:
@emacsomancer can you test it after running `./xbps-src bootstrap-update`? Your problem seems to to be an outdated masterdir.

PR #11390, opened 2018-02-04
alacritty: update to latest-commit: commit (bot)

Arzte:


Vaelatern:
Can they please tag versions?

mad-s:
I think the version should be 20180204, as it is not April yet.

Coding-Doctors:
Got the date switched sorry about that, fixed. As far as versions go, from one the maintainers, it'll likely happen once scrollback is implemented. There is a pr open on that repo, but they don't like how some of it is done, and therefore are working on other changes they want to land to make scrollback more performant.

Arzte:
I've updated the template with the current date and commit.

Arzte:
I've updated to the latest commit. Additionally added man pages, bash & zsh completions. (https://github.com/jwilm/alacritty/pull/1133) I'm unsure how long it'll be till this gets merged, but if its going to be a bit longer, it might be worth waiting for [scrollback](https://github.com/jwilm/alacritty/labels/Project%3A%20Scrollback) to land.

cr6git:
[57146b2](https://github.com/voidlinux/void-packages/commit/57146b2a7d3d9ac042de2c16a79cd36af2dc45e1)

PR #11393, opened 2018-02-04
New package: SIR-3.2

cr6git:


Vaelatern:
Should this pkgname be in all caps? @voidlinux/pkg-committers

cr6git:
It's done.

PR #11398, opened 2018-02-05
Update --enable-kernel parameter for glibc 2.26

winks:
glibc 2.26's README says: 
```
When working with Linux kernels, this version of the GNU C Library requires Linux kernel version 3.2 or later.
```
and running it on a 3.0 kernel indeed gives the "FATAL: kernel too old" message. The 2.6.32 value was true for glibc 2.25 (and certain architectures).

So I don't see any reason that anyone should be misled (like I was) by this version statement.


cr6git:
related to [#11318](https://github.com/voidlinux/void-packages/pull/11318#issuecomment-362592247).

PR #11405, opened 2018-02-05
New package: oomox-1.4.5.2

tim241:


srcpkgs/oomox/template
```diff
@@ -0,0 +1,22 @@
+# Template file for 'oomox'
+pkgname=oomox
+version=1.4.99
+revision=1
+# python-gobject failes to install on other platforms with this error:
+# ERROR: python-gobject-3.26.1_1: cannot be cross compiled, exiting...
+only_for_archs="i686 x86_64-musl x86_64"
```
Vaelatern: No. Just no. Please search the documentation for a variable much more suited to not being able to cross compile.

srcpkgs/oomox/template
```diff
@@ -0,0 +1,22 @@
...
+build_style=meta
```
Vaelatern: Please read the description of this build stye again and tell me if it really matches.

srcpkgs/oomox/template
```diff
@@ -0,0 +1,22 @@
...
+maintainer="Tim Wanders <timwanders241@gmail.com>"
+license="GPL-3"
+homepage="https://github.com/actionless/oomox"
+do_fetch() {
+	if [ ! -d oomox-$version ]; then
+		git clone --recursive --depth=1 --branch $version https://github.com/actionless/oomox/ oomox-$version
+	fi
+}
+do_install() {
+	bash ./packaging/install.sh . /destdir/oomox-$version/
```
Vaelatern: I think you meant `${DESTDIR}`

srcpkgs/oomox/template
```diff
@@ -0,0 +1,22 @@
...
+depends="python python-gobject gtk+3 gdk-pixbuf glib-devel glib sassc bc sed librsvg librsvg-utils optipng parallel inkscape"
+short_desc="Application for generating different color themes"
+maintainer="Tim Wanders <timwanders241@gmail.com>"
+license="GPL-3"
+homepage="https://github.com/actionless/oomox"
+do_fetch() {
+	if [ ! -d oomox-$version ]; then
+		git clone --recursive --depth=1 --branch $version https://github.com/actionless/oomox/ oomox-$version
+	fi
+}
```
Vaelatern: Strongly dislike this. We can't cach anything in this format!

srcpkgs/oomox/template
```diff
@@ -0,0 +1,22 @@
+# Template file for 'oomox'
+pkgname=oomox
+version=1.4.99
+revision=1
+# python-gobject failes to install on other platforms with this error:
+# ERROR: python-gobject-3.26.1_1: cannot be cross compiled, exiting...
+only_for_archs="i686 x86_64-musl x86_64"
+build_style=meta
+hostmakedepends="git"
+depends="python python-gobject gtk+3 gdk-pixbuf glib-devel glib sassc bc sed librsvg librsvg-utils optipng parallel inkscape"
```
Vaelatern: Please limit column widths. 80 is a reasonable number. Indent subsequent lines with a single space.

srcpkgs/oomox/template
```diff
@@ -0,0 +1,22 @@
...
+short_desc="Application for generating different color themes"
```
Vaelatern: `Graphical application for generating color variations of some themes`

srcpkgs/oomox/template
tim241: Okay

srcpkgs/oomox/template
tim241: I have no idea which other one to pick

srcpkgs/oomox/template
tim241: So, should I remove the `if` statement completely?

srcpkgs/oomox/template
tim241: okay

srcpkgs/oomox/template
tim241: thank you

srcpkgs/oomox/template
tim241: @Vaelatern I searched around, couldn't find anything about it, it seems like it should work on other archs [here](https://packages.debian.org/sid/python-gobject) is an example of debian having it for all archs

Vaelatern: You missed my point. That doesn't reflect on this package at all.

srcpkgs/oomox/template
Vaelatern: It's almost a `fetch` build style. Almost. I wish it weren't.

srcpkgs/oomox/template
Vaelatern: Well, I'd prefer if the do_fetch() got removed completely. Search for submodule updating elsewhere in the package tree for better examples.

srcpkgs/oomox/template
```diff
@@ -0,0 +1,25 @@
...
+revision=1
+# python-gobject failes to install on other platforms with this error:
+# ERROR: python-gobject-3.26.1_1: cannot be cross compiled, exiting...
+only_for_archs="i686 x86_64-musl x86_64"
+build_style=meta
+hostmakedepends="git"
+depends="python python-gobject gtk+3
+	gdk-pixbuf glib-devel glib sassc
+	bc sed librsvg librsvg-utils
+	optipng parallel inkscape"
```
Vaelatern: Space, not tabs please. I think this makes sense as two lines.

srcpkgs/oomox/template
Vaelatern: I don't know if debian cross compiles.

srcpkgs/oomox/template
tim241: well, would you rather have a `fetch` build style or `meta` build style?

srcpkgs/oomox/template
tim241: when I look at how other [packages](https://github.com/voidlinux/void-packages/search?utf8=%E2%9C%93&q=submodule+file%3Atemplate) do this, it's basically the longer version of this without the `if` statement

srcpkgs/oomox/template
tim241: would this be better as comment instead of that one ^ `python-gobject is a dependency from oomox, which doesn't support cross compiling to arm devices, that's why we only allow 32 and 64 bit`

srcpkgs/oomox/template
tim241:
```
> ~/dev/void-packages xlint srcpkgs/oomox/template 
srcpkgs/oomox/template:11: indent with tabs
```

srcpkgs/oomox/template
tim241: This also allows you to build the package without having a clean env, so without executing `./xbps-src clean`

srcpkgs/oomox/template
Vaelatern: Meta is simply wrong. Read the description.

srcpkgs/oomox/template
Vaelatern: You don't though. It's not about the comment, it's about you not checking the manual for the highly related `nocross` variable.

srcpkgs/oomox/template
Vaelatern: @voidlinux/pkg-committers ?

pullmoll:
Perhaps I can help to clarify some requests which @Vaelatern made or rather the policies regarding Void templates and how we believe they should look like.

If at all possible, avoid `git clone` and try to use releases - is [1.4.5.2](https://github.com/actionless/oomox/archive/1.4.5.2.tar.gz) not sufficiently new? And if it isn't, you're cloning 1.4.99 if I understood your template?

We generally don't like packages which are not in some definend and reproducible state for anyone who tries to build them locally and `git clone …` has a tendency to break every now and then. I think we even have a (silent?) agreement to not accept new `*-git` packages, which means packages which don't download distfiles but use some RCS to fetch files.

If no releases exist, you should use the git revision of the commit you think is usable and download the tarball from e.g. the latest commit [fa9eb21f69d7ac76e9c673d91b9a7b2ee0b72118](https://github.com/actionless/oomox/archive/fa9eb21f69d7ac76e9c673d91b9a7b2ee0b72118.tar.gz).

For going this way our common practise is to define a `version=` including a date stamp, e.g. 1.4.99.20180128 and also define a variable like `_git_hash=fa9eb21f69d7ac76e9c673d91b9a7b2ee0b72118` to be used in the distfiles line like this:
```
distfiles="https://github.com/actionless/oomox/archive/${_git_hash}.tar.gz>${pkgname}-${version}.tar.gz"
```

Now if a github (or other git repository's) package contains references to other packages with a specific git hash of their own, download them along with the main package using the same scheme. In this case you could define `_oomoxify_hash=abcbadbadbad...` (find it in the tree) and use that hash in a second line in `distfiles="…`.

Finally in a `post_extract()` or perhaps `pre_configure()` function in your template move the referenced packages where they are expected to reside.

For an example of a template doing just that [take a look at tracebox](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/tracebox/template).

Finally a template should never try to optimize for possibly already existing `git clone …` directories. You can think of the builders as always being in a clean state when a package is to be built. We do everything to make sure they are clean, because otherwise no build would be reproducible.

Collecting the necessary info from a package and its submodules is a burden which upstream puts on us by not using released versions of their integrated packages. They all should be slapped around with a large trout – or rather be kindly asked to use a downloadable set of subpackages instead of recursive submodules. Most of them probably would never change their mind, though, and this is where a packager's job is beginning to become tedious... you asked for it :grimacing: 

HTH + Cheers
Jürgen

tim241:
Alright, thanks for the explanation @pullmoll, I have 1 question though, with this `install` script, which build style should I choose? I cannot find one that fits the needs of this packages(meta is not allowed, configure doesn't fit either because I don't have a `do_build` part and fetch doesn't extract from the start, so that means I'd need to extract all the files myself{possible, but not really clean})?

Johnnynator:
@tim241 not specifying any build style should be the correct way

pullmoll:
As @Johnnynator said: if no build style matches, remove that line from the template and write your own `do_build()` and `do_install()` functions. You can find quite a few templates w/o `build_style=…` in the repository.

tim241:
Alright, thanks!

pullmoll:
I merged it after applying some changes to your template:
+ It's a (shell) script only package, so I set `noarch=yes`
+ It doesn't expect python, but python3, so I modified the `depends=…`
+ If a package doesn't cross compile solely because of a dependency, there shouldn't be a `nocross=yes`, except for cases where this dependency will never be cross compilable or is hidden deep in a tree. `python3-gobject` may some day be available for cross builds and then `oomox` as well.
+ Changed to install to `/usr/share/oomox` rather than `/opt/oomox` because Void doesn't like `/opt` (exceptions exist, though)

Issue #10347, opened 2018-01-02
firefox: adobe-flash-plugin crashes when enabled


7185:
Since firefox 57, the flash plugin crashes as soon as I enable it on a webpage.
I tried to update adobe-flash-plugin to the latest version but the issue is still here.
It works fine with firefox binary builds from Mozilla (tested with Nightly on my machine).
I'm using Void glibc on an x86_64 arch.

I saw that there is already a thread on the [Void forum](https://forum.voidlinux.eu/t/flash-player-crashes-instantly/4221) about this issue with interesting information, but I think it's a good idea to create an issue here as well for better visibility.

cr6git:
[Updated](https://github.com/voidlinux/void-packages/pull/10624).

ncmprhnsbl:
updated flash (28.0.0.137) crashes also (firefox-57.0.4_1)

s-p-k-m:
I can also reproduce the crash of the flashplugin and fiddled around with the build options. It seems that https://github.com/voidlinux/void-packages/commit/fc0a6682a9897b7c731245460bd52c0b24f26809 broke it.
When I build the package without `ac_add_options --enable-release` the flashplugin no longer crashes.
I don't know what the flag does in detail, but the commit log says
> pass '--enable-release' to disable gold linker on musl

So I assume linke incompatibilities with the proprietary flashplugin.
I adjusted the template to only add the flag for musl builds.

cr6git:
Updated flash [#11410](https://github.com/voidlinux/void-packages/pull/11410)

7185:
Still waiting for someone to review/merge #11409

PR #11409, opened 2018-02-06
fixes issue #10347 - firefox: adobe-flash-plugin crashes when enabled

s-p-k-m:


pullmoll:
Firefox fails to build. I can't tell if it is due to this change or due to updated dependencies. For now I reverted this PR.

s-p-k-m:
I just pulled the latest version of the firefox template and it fails to build for me too, with and without my patch, same error as the buildbot: `error: Could not compile "encoding_c".`
So the failed build must be due to something else.

asergi:
Merged in 06e2b1301b026ed97169d0e1f4bb58ac9a5a5d19.

PR #11411, opened 2018-02-06
gifski: update to 0.8.0.

cr6git:


maxice8:
0.8.2 available

cr6git:
ah thank you! will update soon, I'm currently busy on something else in my "real life".

PR #11412, opened 2018-02-06
clojure: update to 1.9.0.326.

SevereOverfl0w:


Vaelatern:
Please update more like we do vim. Less often just when it seems big things have happened.

Vaelatern:
If you provide a short summary of changes, I also work with clojure and will happily merge. Just so long as it's not "Oh, the version bumped"

PR #11419, opened 2018-02-06
New package: imagescan-3.35.0

gspe:


pullmoll:
Modifying the host environment to make the Makefile find host directories is a *Bad Idea™*
I changed the Makefile.in to always use `/etc/sane.d/dll.d` instead.

PR #11420, opened 2018-02-06
python-networkx: update to 2.1.

pbui:


maxice8:
python-setuptools is required at runtime, see #11127

PR #11428, opened 2018-02-06
Kde5: update to 5.12.0

Sir-Boops:
* Tested and working on x86_64 👍 
* Alot of these are starting to need `>=qt5.9` meaning a few have been marked broken 😢


gspe:
Please in systemsettings add dependency for setxkbmap and xrdb:
setxkbmap is needed to change keyboard layout
xrdb is needed to change fonts dpi, you need this for 4k screen

Sir-Boops:
Since they are not compile time dependencies shouldn't they be optional 
dependencies ( aka just add a message stating that those are there for 
the purpose )? As most people don't have 4k so it won't effect most users.

gspe:
They are not compile dependency they are runtime dependency that should be mapped into: "depends="
without them systemsettings doesn't work correctly under some circumstance

pullmoll:
Both `setxkbmap` and `xrdb` are tiny packages compared to KDE5 itself, so I'm all for adding this kind of optional dependencies... not just because I **have** a 4K display :-)

Sir-Boops:
I'll add them to depends then since it seems that they will be used!

pullmoll:
I thought you had marked the packages that are broken for qt5... ok, now I will add `broken=` where required.

Sir-Boops:
I did add `broken=` to the packages that i did find broken when bumping?
It looks like the builder is hitting other broken libs and then it errors out.

pullmoll:
Yep, it also breaks when a dependency is broken. I have to mark all pkgs which fail and once qt5 is updated we'll need to remove these again :P

Sir-Boops:
Ahh, ok I'll watch for those errors when test building now! :)

PR #11439, opened 2018-02-07
bash: update to 4.4.019.

maxice8:


pullmoll:
++ `chroot-bash`

maxice8:
On Wed, Feb 07, 2018 at 12:17:29PM +0000, Jürgen Buchmüller wrote:
> ++ `chroot-bash`
> 

Should be now.

PR #11447, opened 2018-02-07
gitea: add missing absolute paths, use sqlite

damadmai:
The paths

- LOCAL_COPY_PATH
- APP_DATA_PATH
- REPO_INDEXER_PATH

were relative so it did not work, now they are prepended by `/var/lib/gitea/` like the others.

`DB_TYPE` is changed to `sqlite3` so it can work out of the box after installing and activating the service without setting up a database server.


damadmai:
Added version bump. If I should squash the two commits, just tell me.

pullmoll:
Yes, please always squash to one commit per package.

damadmai:
Okay, done.

damadmai:
Found much better solution, added GITEA_WORK_DIR to runit command line.

damadmai:
Solution worked only because gitea read the already set values from the database, so the paths are added again now, so that it also works if there was no config before.

PR #11450, opened 2018-02-07
mcabber: add mcabberrc.example

damadmai:
Install `mcabberrc.example` to `/usr/share/doc/mcabber/` for reference.

### DIFF
--- a/srcpkgs/mcabber/template
+++ b/srcpkgs/mcabber/template
@@ -1,7 +1,7 @@
 # Template file for 'mcabber'
 pkgname=mcabber
 version=1.1.0
-revision=1
+revision=2
 build_style=gnu-configure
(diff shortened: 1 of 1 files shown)


srcpkgs/mcabber/template
```diff
@@ -13,3 +13,7 @@ license="GPL-2"
 homepage="http://mcabber.com"
 distfiles="${homepage}/files/mcabber-${version}.tar.bz2"
 checksum=04fc2c22c36da75cf4b761b5deccd074a19836368f38ab9d03c1e5708b41f0bd
+
+post_install() {
+    vdoc mcabberrc.example
```
ebfe: Alternatively `vsconf`

damadmai:
Changed vdoc to vsconf as suggested by @ebfe

PR #11454, opened 2018-02-07
New package: blackbox-1.20170611

damadmai:
No dependency on gnupg or gnupg2 to let users decide which to install.
Needs `GPG=gpg2` if used with GnuPG 2


Piraty:
maybe use `INSTALL.msg` for this info about gpg/gpg2?

damadmai:
Thanks, I added it.

PR #11457, opened 2018-02-07
acme-client: update to 0.1.16

freshprince:
ATTENTION: acme-client changed default directories from "letsencrypt" to "acme" so this might break running setups.

Should I add a `install.msg` with a note?


Vaelatern:
Huh, will check what upstream expected as a migration mechanism

Vaelatern:
Yes please, a 1-line `INSTALL.msg`, max 80 characters.

freshprince:
I hope the wording is alright.

freshprince:
Crossbuilds are now fixed too.

PR #11191, opened 2018-01-28
qt5: update to 5.10.1

pullmoll:
Build succeeded here for:
- [x] x86_64
- [x] i686
- [x] x86_64-musl
- [x] i686-musl
- [x] aarch64
- [x] aarch64-musl
- [x] armv7l
- [x] armv7l-musl
- [x] armv6l
- [x] armv6l-musl
- [ ] armv5tel qtwebengine fails with `Project ERROR: armv5te architecture is not supported`
- [ ] armv5tel-musl (same)
- [ ] mips-musl
- [ ] mipshf-musl
- [ ] mipsel-musl
- [ ] mipselhf-musl

It seems arm* can be cross compiled from an x86_64 environment with `-no-webengine-v8-snapshot`. Testing this and a bunch of fixes for qtwayland.

[ci skip]

P.S.: For the adventurous *early adopter* there is [a repository to play with](http://repo.voidlinux.de/current/qt5). I tested the x86_64 build with some of my qt5 projects and saw no issues so far.

PR #11481, opened 2018-02-08
fcitx-qt5: update to 1.2.2.

maxice8:


mvf:
This should be merged ASAP. `fcitx-qt5` is a platform plugin and needs to be rebuilt against Qt 5.10.1 merged with #11191. Currently it doesn't load anymore:

```
22147:	/usr/lib/qt5/plugins/platforminputcontexts/libfcitxplatforminputcontextplugin.so: error: relocation error: symbol _ZTI21QPlatformInputContext, version Qt_5 not defined in file libQt5Gui.so.5 with link time reference (fatal)
```

PR #11484, opened 2018-02-08
New package: android-file-transfer-linux-3.2

Vaelatern:
Weird musl problems


chneukirchen:
```
00:33:11      <leah2> Vaelatern: the musl issue on #11484 is the usual
         strerror_r thing
00:33:24      <leah2> it has a different prototype on musl, and doesnt match 
         glibc
00:33:44      <leah2> (that requires patching)

       int strerror_r(int errnum, char *buf, size_t buflen);
                   /* XSI-compliant */

       char *strerror_r(int errnum, char *buf, size_t buflen);
                   /* GNU-specific */
```

PR #11500, opened 2018-02-09
dbus: update to 1.12.4.

maxice8:


maxice8:
thanks to @cr6git for mentioning the cmake files on the normal package, a `vmove usr/lib/cmake` was added.

PR #11524, opened 2018-02-10
wavpack: fix configure flags

cr6git:
````
configure: WARNING: unrecognized options: --enable-mmx
````

### CODE REVIEW

File: srcpkgs/wavpack/template [hunk 1]
```diff
@@ -12,8 +12,8 @@ checksum=1939627d5358d1da62bc6158d63f7ed12905552f3a799c799ee90296a7612944
 
 pre_configure() {
 	case "$XBPS_TARGET_MACHINE" in
-		i686|*x86_64*) configure_args+=" --enable-mmx";;
 		armv5*) configure_args+=" --disable-asm";;
+		*) ;;
```
pullmoll: This kind no-op is not necessary IMO.

cr6git: ok, I will remove this line.

PR #11534, opened 2018-02-10
wine: add x86_64 target with i686 support

tim241:


thypon:
@voidlinux/pkg-committers comments?

Duncaen:
Does this even work?
There are bootstrap issues, iirc you cant depend on `-32bit` packages.

tim241:
it builds correctly on my machine

thypon:
why removed and deleted?

PR #11535, opened 2018-02-10
New package: Qwt-1.6.3

gspe:
This is a dependency package for build QGIS


srcpkgs/qwt/template
```diff
@@ -0,0 +1,44 @@
+# Template file for 'qwt'
+pkgname=qwt
+version=6.1.3
+revision=1
+build_style=qmake
+hostmakedepends="qt5-qmake"
+makedepends="qt5-tools-devel qt5-svg-devel"
+depends=""
```
jnbr: remove empty variable

gspe: Done, thanks.

srcpkgs/qwt/template
```diff
@@ -0,0 +1,47 @@
+# Template file for 'qwt'
+pkgname=qwt
+version=6.1.3
+revision=1
+build_style=qmake
+hostmakedepends="qt5-qmake"
+makedepends="qt5-tools-devel qt5-svg-devel"
+short_desc="Qt Widgets for Technical Applications"
+maintainer="Giuseppe Fierro <gspe@ae-design.ws>"
+license="Qwt License,LGPL-2.1"
```
pullmoll: `license="Qwt-License LGPL-2.1"`, i.e. a blank separated list of license names

jnbr:
the problem with cross builds is that qt5-tools doesn't include the QT5Designer on these platforms.
I think it would be ok to build qwt without the designer components when cross building or better, when the architecture isn't x86 or i686.

pullmoll:
Yes, the missing Qt5Designer is a problem of current qt5 cross builds. This will be fixed once the qt5-5.10.0 update is working for cross builds.

srcpkgs/qwt/template
```diff
@@ -0,0 +1,47 @@
...
+	hostmakedepends+=" qt5-host-tools qt5-tools-devel qt5-svg-devel"
+fi
+
+pre_configure() {
+	sed -e '/^\s*QWT_INSTALL_PREFIX/ s|=.*|= /usr|' \
+		-e '/^QWT_INSTALL_DOCS/ s|/doc|/share/doc/qwt|' \
+		-e '/^QWT_INSTALL_HEADERS/ s|include|&/qwt|' \
+		-e '/^QWT_INSTALL_PLUGINS/ s|plugins/designer|lib/qt5/&|' \
+		-e '/^QWT_INSTALL_FEATURES/ s|features|lib/qt5/mkspecs/&|' \
+		-i qwtconfig.pri
```
jnbr: disable designer plugin on anything than x86 or i686:
```
        i686*|x86_64*)
            ;;
        *)
            sed -e 's/^\s*QWT_CONFIG\s*+=\s*QwtDesigner//g' -i qwtconfig.pri
            ;;
    esac
```

gspe: I've made this change.

jnbr: With the updated qt5 version, this is not necessary anymore.

gspe:
Made changes 
Now everything should be ok

srcpkgs/qwt/template
```diff
@@ -0,0 +1,47 @@
+# Template file for 'qwt'
+pkgname=qwt
+version=6.1.3
+revision=1
+build_style=qmake
+hostmakedepends="qt5-qmake"
+makedepends="qt5-tools-devel qt5-svg-devel"
+short_desc="Qt Widgets for Technical Applications"
+maintainer="Giuseppe Fierro <gspe@ae-design.ws>"
+license="Qwt-License LGPL-2.1"
```
jnbr: nit-pick: these should be separated by a comma

gspe: nice catch!
Corrected

pullmoll:
FYI The common/shlibs entry was missing.

PR #11545, opened 2018-02-10
New package: darch-0.18.3

pauldotknopf:
This is to use Void Linux with Darch.

https://godarch.com/
https://godarch.com/concepts/recipes/distributions/

This is my first pull request here, so be nice ;)


maxice8:
Can't the GO packages be built using build_style=go ?

pauldotknopf:
I tried, but I couldn't get the GOPATH and GOBIN working right.

I tried to refer to existing templates, but they all didn't work. I opted for a less magic, more explicit, approach.

maxice8:
fair

srcpkgs/runc/template
```diff
@@ -0,0 +1,48 @@
...
+pkgname=runc
+version=1.0.0
+revision=1
+_subversion="rc4"
+_version="$version-$_subversion"
+wrksrc="$pkgname-$_version"
+create_wrksrc=yes
+hostmakedepends="git go pkg-config"
+makedepends="libseccomp-devel"
+depends="libseccomp-devel"
```
maxice8: Is it really necessary to set seccomp-devel to runtime dependencies ? most of the time the package manager will detect that a library is used and add it itself on the proper library.

pauldotknopf: Nope, I don't need to depend on the devel version.

Are you saying that I don't even have to specify a ```depends```? That ```xbs-src``` will ```ldd``` to see the dependencies, and auto establish them for me?

mvf: Exactly. [Documentation Link](https://github.com/voidlinux/void-packages/blob/master/Manual.md#runtime-dependencies)

maxice8: > Are you saying that I don't even have to specify a depends? That xbs-src will ldd to see the dependencies, and auto establish them for me?

This is the hook in question that does the rdeps generating.

https://github.com/voidlinux/void-packages/blob/master/common/hooks/pre-pkg/04-generate-runtime-deps.sh

It will try it's best but will fail if the program is not cooperative, we recently had to rebuild some packages due to ncurses update from `6.0` -> `6.1`, some packages broke because they didn't have dependencies on the proper ncurses libraries.

srcpkgs/containerd/template
```diff
@@ -0,0 +1,50 @@
...
+pkgname=containerd
+version=1.0.2
+revision=1
+_subversion="rc.1"
+_version="$version-$_subversion"
+wrksrc="$pkgname-$_version"
+create_wrksrc=yes
+hostmakedepends="git go"
+makedepends="libbtrfs-devel"
+depends="runc libbtrfs"
```
maxice8: libbrtfs should also be generated from rdeps

srcpkgs/runc/template
```diff
@@ -0,0 +1,48 @@
...
+short_desc="CLI tool for spawning and running OCI containers"
+maintainer="Paul Knopf <pauldotknopf@gmail.com>"
+license="Apache-2.0"
+homepage="https://github.com/opencontainers/runc"
+
+do_fetch() {
+	mkdir -p ${wrksrc}/tmp/src/github.com/opencontainers
+	mkdir -p ${wrksrc}/tmp/bin
+	git clone git://github.com/opencontainers/runc ${wrksrc}/tmp/src/github.com/opencontainers/runc
```
maxice8: I'm not sure this is the best way to fetch the package, i packaged runc pretty quickly and using distfiles= to download the release tarball .xz and it worked (apparently)

pauldotknopf:
Guys, I am having troubles with the cross builds. How do I cross compile locally?

```
export XBPS_TARGET_MACHINE=aarch64
export CROSS_BUILD="yes"
./xbps-src pkg runc
```

This doesn't seem to do a cross-compilation.

srcpkgs/runc/template
pauldotknopf: Yeah. I have been looking at it more, looking into cross compilation.

I think I see my hangup using the ```go``` build style. It also has some cross-build environment variables that needs to be set.

I am going to revamp this templates, will update later today.

Thanks for your help though. I appreciate the time.

pauldotknopf:
Alright guys, all updated/fixed. Waiting for your review!

srcpkgs/containerd/template
```diff
@@ -0,0 +1,30 @@
...
+go_import_path=github.com/containerd/containerd
+wrksrc="$pkgname-$_version"
+hostmakedepends="git go unzip"
+makedepends="libbtrfs-devel"
+depends="runc"
+short_desc="An open and reliable container runtime"
+maintainer="Paul Knopf <pauldotknopf@gmail.com>"
+license="Apache-2.0"
+homepage="https://github.com/containerd/containerd"
+distfiles="https://github.com/containerd/containerd/archive/v${_version}.zip"
```
maxice8: There is a preference for .tar.gz files which don't need the extra unzip dep, if possible change it.

pauldotknopf: No problem.

Making changes to previous commits are difficult, since I can't just do ```git commit --amend```. Since ```containerd``` isn't the tip, can you make sure there isn't anything else that needs to be changed before I make this change?

newbluemoon: You can use `git rebase --interactive` to edit previous commits. See https://help.github.com/articles/using-git-rebase-on-the-command-line/

pauldotknopf:
Updated again, ready for review!

Vaelatern:
Anywhere with `version=` that isn't `version`, e.g. `subversion`, please make it something like `_subver=` to not confuse certain scripts.

Vaelatern:
The `darch` template makes me think it could have been done better.

srcpkgs/darch/template
```diff
@@ -0,0 +1,40 @@
...
+homepage="https://godarch.com/"
+make_dirs="/var/lib/darch 0755 root root"
+conf_files="/etc/darch/hooks/ssh/hook /etc/darch/hooks/hostname/hook /etc/darch/hooks/fstab/hook"
+
+do_fetch() {
+	mkdir -p "$(dirname ${GOSRCPATH})"
+	git clone --branch "v${version}" \
+		git://github.com/godarch/darch \
+		${GOSRCPATH}
+}
```
Vaelatern: using distfiles

srcpkgs/darch/template
```diff
@@ -0,0 +1,40 @@
...
+
+do_configure() {
+	export GOBIN=${GOPATH}/bin
+	export PATH=$PATH:$GOBIN
+	go get github.com/LK4D4/vndr
+	cd ${GOSRCPATH}
+	vndr
+}
```
Vaelatern: this sounds like something for the go build style.

srcpkgs/darch/template
```diff
@@ -0,0 +1,40 @@
...
+
+do_build() {
+	cd ${GOSRCPATH}
+	make build TRAVIS_TAG="v${version}" GO_BUILD_FLAGS="-buildmode=pie"
+}
```
Vaelatern: Not sure why the makefile is necessary.

srcpkgs/darch/template
```diff
@@ -0,0 +1,40 @@
...
+
+do_install() {
+	cd ${GOSRCPATH}
+	make install DESTDIR=${DESTDIR}
+}
```
Vaelatern: This I can understand. Except the `cd` at the top with a good build style will not need this.

srcpkgs/runc/template
```diff
@@ -0,0 +1,27 @@
...
+
+do_build() {
+	cd ${GOSRCPATH}
+	make
+}
+
+do_install() {
+	cd ${GOSRCPATH}
+	make install BINDIR=${DESTDIR}/usr/bin
+}
```
Vaelatern: I think you want to learn about `build_wrksrc`.
Don't know why you use a `build_style=go` then overwrite it.

srcpkgs/darch/template
pauldotknopf: to makefile depends on using git.

A new darch version can remove the need, but it is needed now

srcpkgs/darch/template
pauldotknopf: Add vndr support go he go build style?

srcpkgs/darch/template
pauldotknopf: Building darch requires the use of make.

The built in build stuff for the go build template is too rudimentary.

srcpkgs/darch/template
pauldotknopf: The Makefile exists in the repository, under src/github.com/godarch/darch.

Again, this had to be done because I had to manually setup go path because I can't use distfiles at the moment.

srcpkgs/runc/template
pauldotknopf: The Makefile with darch has built in deploy logic.

pauldotknopf: I mainly use the go template to set the GOSRCPATH environment variable.

I can remove build style go and setup GOSRCPATH manually.

Vaelatern: The do build sets up the GOSRCPATH for you as well

srcpkgs/darch/template
```diff
@@ -0,0 +1,40 @@
+# Template file for 'darch'
+pkgname=darch
+version=0.18.3
+revision=1
+build_style=go
+go_import_path=github.com/godarch/darch
+hostmakedepends="git go"
+depends="containerd"
+only_for_archs="x86_64"
```
Vaelatern: This is my biggest complaint.

pauldotknopf: Darch only supports 64 bit.

The executable itself could run cross plat, but to use it, you run docker images. The images only support 64 bit.

Even if someone installs it on arm, it would be useless.

Containers and runc are xplat

srcpkgs/darch/template
pauldotknopf: I forgot to mention.

I had to do it this way because vndr doesn't work with symlink dirs.

srcpkgs/darch/template
Vaelatern: and musl?

pauldotknopf: I can add musl. Forgot about that

Vaelatern: I dislike the use of `only_for_archs` because it doesn't describe the situation at all. At least with a conditional then `broken="reason"` on matching cause, you get a clear explanation that can be checked against upstream's current reality.

Vaelatern: `only_for_archs` seems to say "Yeah, this thing is a binary linked against this exact libc and you will get nowhere if you try to work around this."

srcpkgs/runc/template
pauldotknopf: I tried again removing ```cd ${GOSRCPATH}```, but Go can't find the vendored libraries. The cd has to be there.

pauldotknopf: I am still using build_style=go because it sets up environment variables for cross compilation.

pauldotknopf:
@Vaelatern, I updated what I could.

* Replaced "verison" with "ver".
* Added musl as a target for Darch. Added a ```broken``` reason for the failure.

Vaelatern:
still don't like the use of `only_for_archs`, but it's not so bad it can't be merged. Just so bad that I hope you will maintain it.

pauldotknopf:
I use Darch daily. Consider it my hypervisor for Void Linux. I guarantee it will be maintained. :)

PR #11578, opened 2018-02-12
New package: amp-0.3.3

wpbirney:


wpbirney:
I believe with some simple scripting and rustup we should be able to make all these rust/cargo packages (that are all nocross=yes) cross compile capable, ill mess with it and see if i can get it working

pullmoll:
Unfortunately amp is broken with the most recent libressl update.

wpbirney:
@pullmoll amp is now at 0.3.4, i can make a PR for the update, that version builds locally for me

PR #11586, opened 2018-02-12
perl-Params-ValidationCompiler: update to 0.27.

maxice8:


newbluemoon:
I’m curious why you changed `makedepends`. I did it the way it was because the dependencies aren’t needed for creating the package, just at runtime and I thought it’s better not to pull them in unnecessarily.

maxice8:
On Mon, Feb 12, 2018 at 11:48:28AM +0000, newbluemoon wrote:
> I’m curious why you changed makedepends. I did it the way it was because the
> dependencies aren’t needed for creating the package, just at runtime and I
> thought it’s better not to pull them in unnecessarily.
> 
> —
> 

they are required during testing (which i hope to make it default on Void Linux some time in the future )
most perl packages take this approach.

```
hostmakedepends="perl" # perl-Module-Build if build_style is perl-ModuleBuild
makedepends="${hostmakedepends} dependencies"
depends="${makedepends}"
checkdepends="dependenciesfortesting"
```

newbluemoon:
OK. Thanks for clarifying. :)

PR #11597, opened 2018-02-12
mpc: import upstream patch for manpages

maxice8:


newbluemoon:
No offense, but why not wait for a new release? The package itself won’t change and to me that one line in `post_install()` looks better than adding a patch which does the same and will be included in the next release, anyway.

PR #11542, opened 2018-02-10
[wip] plasma-workspace: fix CVE-2018-6791 

Piraty:
As mentioned by  @thypon in IRC.
https://www.kde.org/info/security/advisory-20180208-2.txt

Fix itself is small, but some bumps need to be reverted due to this: https://github.com/voidlinux/void-packages/pull/11428

PR #11611, opened 2018-02-13
kde5-baseapps: update to 17.12.2

mstraube:


mstraube:
Cross building dolphin fails:
```
   [runtime] kio-extras>=0: not found.
=> ERROR: kdesignerplugin-devel-5.42.0_1: cannot be built, it's currently broken; see the build log:
=> ERROR: kdesignerplugin-devel-5.42.0_1: kdewebkit requires qt5 >= 5.9.0
=> ERROR: kio-extras-17.12.1_1: failed to transform dependency kdesignerplugin-devel
```
I'm not sure how to handle this.

Hoshpak:
My suggestion would be to wait with the update until either #11542 or #11191 are successfully merged. Marking the package as broken would also work but wouldn't be of any use for anyone imho. I'm also waiting for the next qt5 update or revert of broken packages since the new version of kmymoney also depends on kdewebkit.

mstraube:
Thanks for the links. I also think waiting is better than marking broken.

PR #11622, opened 2018-02-13
dropbear service add conf options

xYuusha:


Duncaen:
The revision needs to be increased to make this change effective.
And change the commit subject to something like `dropbear: add conf to service`.

xYuusha:
Updated.

PR #11625, opened 2018-02-14
New package: helm-0.9.0

laserbat:


srcpkgs/helm/template
```diff
@@ -0,0 +1,13 @@
+# Template file for 'helm'.
+pkgname=helm
+version=0.9.0
+revision=1
+build_style=gnu-configure
```
jnbr: there is no `configure` file, so `gnu-makefile` is the better choice here

jnbr:
Crossbuilds fail because the Makefile uses `uname -m` to select `SIMDFLAGS`. These should be set depending on the target architecture, not the host architecture.

PR #11630, opened 2018-02-14
clipgrab: update to 3.6.7.

maxice8:


cr6git:
`homepage="https://clipgrab.org"`

PR #11672, opened 2018-02-16
strace: update to 4.21.

maxice8:


cr6git:
`aarch64` and `aarch64-musl` failed with the same error:
````
configure: error: Cannot enable m32 personality support
````

I think these errors are related to:
````
$ ./configure --help

  --enable-mpers=yes|no|check|m32|mx32
                          whether to enable multiple personalities support
                          required for proper decoding of structures used by
                          tracees with personalities that differ from the
                          personality of strace, default is yes.
````

PR #11673, opened 2018-02-16
New build-style: cargo

zethra:
Cargo is the build system used for rust packages.  While building with cargo is already easy and requires few commands I felt that having a cargo build-style would be useful so that packaging rust applications would be even easier.


Vaelatern:
Why rustup instead of just the cargo commands currently used by rust packages?

zethra:
rustup allows the package to use an alternate toolchain ei. the nightly build of the compiler.

Anachron:
Think of rustup for Rust as virtualenv for python or nvm for node.

In my opinion rustup is required since there are a lot of different crates with different Rust versions required.

zethra:
Yes.  rustup is required to build certain packages because they use specific compiler versions.  I do think using straight cargo by default would be beneficial as it would decrease build time and complexity.

Vaelatern:
I am conflicted on this because on the one hand, having it work is good. On the other hand, we don't like "Hey, now you have this random source from wherever on the system which you can't cache at all".

zethra:
I'm sorry. I don't follow.

Anachron:
Vaelatern in the Rust community there are a few different toolchains, for example stable & nightly.

Some packages require nightly, others build fine on stable.

Rustup managers different versions of Rust transparently,- meaning: You just define which channel your crate (package) needs to be build and rustup does the rest.

chneukirchen:
NACK, this should use the rust and cargo we have in repos.

common/build-style/cargo.sh
```diff
@@ -0,0 +1,39 @@
...
+	cargo build --release ${cargo_build_args}
+}
+
+do_check() {
+	cargo check --release
+	cargo test --release
+}
+
+do_install() {
+	vbin target/release/${pkgname}
```
DirectorX: This part can be better, we should iterate over all the binaries in that directory.

The a look at golang build style https://github.com/voidlinux/void-packages/blob/master/common/build-style/go.sh#L20

common/build-style/cargo.sh
```diff
@@ -0,0 +1,39 @@
+#
+# This helper is for templates using cargo
+#
+
+cargo_build_args=""
+
+do_configure() {
+	# local cargo_cmd
+	# local cargo_build_args
+	if [ "$rust_toolchain" ]; then
```
DirectorX: Well, this is not a **must** but we need to look into cross compiling, `rustup` is good help to achieve this.

A good reference: https://github.com/japaric/rust-cross#rust-cross

Anachron:
@chneukirchen cargo cannot handle different Rust versions easily,- libraries requiring `nightly` Rust will not be able to be build.

Vaelatern:
@Anachron and I'm not sure why that's a packaging problem. If it requires nightly, then it's too unstable to be packaged, imo.

zethra:
@DirectorX Thanks for the suggestions.

zethra:
@Vaelatern I'll deiffer to your opinion on this.  I'll have it just use cargo.

wpbirney:
using rustup would make cross compiling a hell of a lot easier

chneukirchen:
afaics, rustup does not support aarch64 nor musl, so I don't see how anything gets easier here.

Also the builds are not reproducible anymore if always the latest nightly compiler is used.

Vaelatern:
unless rustup is capable of just using our rust and cargo?

zethra:
rustup does support aarch64 and musl. You can lock your build on a specific version of nightly so it's reproducible(not that you should) and rustup installs rust and cargo for the toolchain you've specified. It conflicts with the rust and cargo packages.

zethra:
I don't see the cross compiling use case being needed. Using the rust and cargo packages rust packages already build for all of void's toolchains. The only reason I can see one needing it for packaging is for grabbing specific compiler versions. And as @Vaelantern said. I'm not sure we should be doing that anyway.

Anachron:
For cross compilation:
https://github.com/rust-lang-nursery/rustup.rs/blob/master/README.md#cross-compilation

Rustup supports all platforms that Rust supports. And Rust supports a lot:

https://forge.rust-lang.org/platform-support.html

Frankly I don‘t understand the issue at hand here. 

If a Rust library required a specific release of Rust (lets say 1.10) and Rustup can handle it easily but the default build system (just latest Rust and Cargo) cant ... where is the issue?

newbluemoon:
@Anachron I think the point is to not depend on third party binaries of which one cannot be sure (trust) how they’re built.

zethra:
I'm new to this community so for give my questions, but how do we come to a final decision on this?

Anachron:
Rustup is not a third party library... it‘s from the Rust team: https://github.com/rust-lang-nursery/rustup.rs

newbluemoon:
@Anachron Ok, sorry, not the right choice of words ;) But I think compilers are a very sensitive subject because everything is based on them. So building them from source whenever possible is always preferable than using a prebuilt binary. That’s how I feel and that’s how I understand the hesitation, just wanted to point it out clearly, because you asked :)

zethra:
I altered the script so that it will only use rustup if the template specified an alternate toolchain.  It'll also install all of the binaries generated.

Anachron:
@newbluemoon but then you would have to compile Rust itself if you don't trust the build? And that is a huge task. We don't do that, neither for cargo nor for rust. 

If it's all about the missing knowledge of Rustup: We're here to help and fill up the missing links.

newbluemoon:
@Anachron you said above you don’t understand the issue at hand and I just wanted to say how it presents itself from an outside perspective, but maybe I’m making it worse? ;)

Anyway, I’m under the impression that Void does build rust from source (else I don’t understand what the template does) and so the argument is to make the new build style use that. In the end it’s all a matter of trust, I guess.

Anachron:
@newbluemoon sorry I was looking wrong,- right,- we are building the Rust binary ourselves.
https://github.com/voidlinux/void-packages/blob/e421d5923aca137cdda76e4fe9e3543f8588f707/srcpkgs/rust/template

It's a pretty old Rust version though ... Rust is at `1.24.0` right now, the repo is still at `1.15.1`. That version is one year old. 

Rustup is already a package:
https://github.com/voidlinux/void-packages/blob/332c34ddcb22bda97a6922538869c972d852075b/srcpkgs/rustup/template

So since we are already building Rustup and Rust.

chneukirchen:
Will it work with Void cross-chroots? Many Rust projects using bindings to native libraries.

zethra:
As far as I can tell, as long as you list the correct deps and pkg-config, they seem to link fine.  A lot of projects link to openssl.

PR #11679, opened 2018-02-16
libgusb: update to 0.3.0.

maxice8:


maxice8:
Hard requires gobject-introspection, 100% unfriendly to cross-compilation.

PR #11702, opened 2018-02-16
New package: mugshot

zethra:
mugshot is used for user profile managment for xfce.


jnbr:
it should be safe to set `noarch=yes`

PR #11708, opened 2018-02-17
New package: ufetch-0.0.0.20170506

cr6git:
Tiny little thing.


Vaelatern:
I don't think we can accept this.

Vaelatern:
It's just an unversioned 52 line shell script.

cr6git:
@Vaelatern OK!

* cr6git throws away ufetch

PR #11729, opened 2018-02-18
New Package: telegram-desktop and dependencies

Johnnynator:
Notes:
- Compilation for i686 and x86_64{,-musl} succeeded fine locally
- package tested successfully on x86_64-musl

### DIFF
--- a/srcpkgs/gyp/template
+++ b/srcpkgs/gyp/template
@@ -1,0 +1,20 @@
+# Template file for 'gyp'
+pkgname=gyp
+version=0.0.0.20180218
+revision=1
+_commit=4d467626b0b9f59a85fb81ca4d7ea9eca99b9d8f
+create_wrksrc=yes
+build_style=python-module
+hostmakedepends="python-setuptools git"
+depends="python-setuptools"
+short_desc="GYP (Generate Your Projects) meta-build system"
+maintainer="John <johnz@posteo.net>"
+license="BSD"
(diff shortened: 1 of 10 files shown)


Johnnynator:
Compiled fine locally for all supported archs. Travis errors because it exceeded the max time limit of ~50min.

skrzyp:
How's the status of that? From the last discussion on IRC I've thought it's generally okay-ish, but still no one accepted the PR.

pullmoll:
We have a serious lag now because of many huge packages hitting the builders, changes like the libressl and its dependencies updates and failing builders were holding up things. I'm sure everyone of @voidlinux/pkg-committers does her best to catch up.

srcpkgs/gyp/template
```diff
@@ -0,0 +1,25 @@
+# Template file for 'gyp'
+pkgname=gyp
+version=20180218
```
Vaelatern: Who decides this version?

Johnnynator: Yep, I changed it. The project it self doesn't have any versioning. At least no visible one.

pullmoll:
Provided #12055 is accepted, you could then make `gyp` download the tarball.
I really hate package templates fetching distfiles with `git clone` everytime I build them :)

Johnnynator:
Updated gyp template to use the tarball

PR #11731, opened 2018-02-18
ckbcomp: update to 1.180.

cr6git:


maxice8:
ey, add `pkgname=console-setup` to `srcpkgs/ckbcomp/update` so `./xbps-src update-check` can do proper update checking

cr6git:
sighs... if only I had a MERGE button... :unamused:

PR #11732, opened 2018-02-18
Update kdeframeworks to 5.43.0 and unbreak KDE5, necessary rebuild for qt5-5.10.1

Hoshpak:
This PR contains the following changes:
* <s>Revbump of qt5-webkit (needs rebuild for qt5-5.10.1)</s>
* Update of the KDE frameworks to 5.43.0
* Unbreaking of all KDE5 packages that were waiting for qt5>=5.9

The KDE frameworks packages need a rebuild for the qt5-5.10.1 update and since that requires touching all templates, I figured I might as well just update them at the same time.


pullmoll:
@voidlinux/pkg-committers should we merge this now or rather wait for the `*-musl` builders to catch up?

Vaelatern:
I vote wait for musl builders.

pullmoll:
The qt5-5.10.1_4 build for x86_64-musl seems to be broken. qt5-declarative-5.10.1_4.x86_64-musl.xbps cannot be extracted. It seems it was corrupted when packaging.

Yet another revbump?

pullmoll:
It seems kdesignerplugin cross building always fails.
I'll give disable_parallel_buil=yes a try or otherwise we need to make cross builds depend on the host's kdesignplugin and the contained /usr/bin/kgendesignerplugin.

pullmoll:
I can't solve this problem. I tried what I could imagine might help and failed.

pullmoll:
I found a hacky solution. I don't know if the results are usable, though.

I think we should create a separate subpackage `kgendesignerplugin` with just the binary to use for cross builds, because I think `kdesignerplugin` for the host pulls in a lot of unused stuff.

pullmoll:
The kde5 meta package should include the `plasma-*` stuff only for no-cross as it is now, because we have the unsolved problem of plasma-framework reporting:
```
CMake Error at src/declarativeimports/core/CMakeLists.txt:29 (add_library):
  Target "corebindingsplugin" links to target "Qt5::Gui_GLESv2" but the
  target was not found.  Perhaps a find_package() call is missing for an
  IMPORTED target, or an ALIAS target is missing?
```

PR #11738, opened 2018-02-19
New package: volume_key-0.3.9

abenson:


cr6git:
`short_desc="A library for manipulating storage volume encryption keys"`
Didn't anyone tell you that the "A" must be avoided?

abenson:
No. I just copied it from the webpage. I'll fix it.

abenson:
If that's a real rule, the Manual should be updated, also.

cr6git:
@Vaelatern does not like the "A", and I agree with him.
This rule is not in the Manual? Maybe we should add it?

abenson:
Yeah, it should be added. There are a few hundred packages with the "A", so I wasn't aware. It should be more visible.

cr6git:
I know there are a few hundred packages with the "A"...
It'll take a while to remove all of them! ;)

Vaelatern:
It's a thing we've unofficially borrowed from the OpenBSD porting guide, no beginning with an indefinite article

cr6git:
@abenson I can cross compile `volume_key` without error, using a slightly modified template:
````
# Template file for 'volume_key'
pkgname=volume_key
version=0.3.9
revision=1
build_style=gnu-configure
short_desc="Library for manipulating storage volume encryption keys"
hostmakedepends="pkg-config"
makedepends="libblkid-devel glib-devel cryptsetup-devel nss-devel gpgme-devel python-devel"
maintainer="Andrew Benson <abenson@gmail.com>"
license="GPL-3"
homepage="https://pagure.io/volume_key"
distfiles="http://releases.pagure.org/volume_key/volume_key-${version}.tar.xz"
checksum=450a54fe9bf56acec6850c1e71371d3e4913c9ca1ef0cdc3a517b4b6910412a6

	hostmakedepends+=" python-devel"
fi

volume_key-devel_package() {
	pkg_install() {
		vmove usr/include
		vmove usr/lib/*.so
	}
}
````

**Comment by cr6git** [comment]
@abenson oops... I only tested cross compilation for `aarch64` and it was ok. Looking at your last build, `armv7hf` and `armv6hf-musl` failed with the same error:
````
"LONG_BIT definition appears wrong for platform (bad gcc/glibc config?)."
````
I don't know what that is?
See [#L1011](https://travis-ci.org/voidlinux/void-packages/jobs/343424898#L1011) and [#L879](https://travis-ci.org/voidlinux/void-packages/jobs/343424900#L879) for reference.

**Comment by abenson** [comment]
That is an odd one.

**Comment by abenson** [comment]
It looks like it is an issue when cross compiling from 64bit to 32bit, at least based on discussions for other projects. I'll have to look into how to fix it for this one.

**Comment by pullmoll** [comment]
We have a similar issue with LuaJIT which fails for `arm*-musl` builds because the environment is `x86_64`, not `i686`. Fortunately I could solve another similar issue for `qt5` by disabling the qtwebengine (Chromium) V8 engine snapshot code generation which would also have required an `i686` build env.

**Comment by abenson** [comment]
So... there is no fix? That's unfortunate. That kind kills this whole project for the time being, unless blockdev can be modified to not use volume_key. This is part of the overall effort to upgrade udisks2

**Comment by pullmoll** [comment]
I did not say there is no fix. It may well be possible to use a 64 bit environment and `gcc-multilib` to support the `-m32` flag in CFLAGS for x86_64. This doesn't work with the current builders and cross `arm*-musl`, though, because there is no `musl-multilib`. So a workaround is more probably to find in modifying either configure.ac or configure, the resulting Makefile or whatever this package uses.

I would have to take a closer look at the configure or Makefile, while currently I'm busy with paid work, which every now and then I need to do to pay the :cheese: for my :bread: ... sorry.

**Comment by newbluemoon** [comment]
The problem are lines like `In file included from /usr/include/python2.7/Python.h` in the error log. Had this problem a while ago, too: python host includes aren’t compatible with the target so we need to find a way to use the right ones. 

In `Makefile.am` and `Makefile.in` there’s the line
`PYTHON_CPPFLAGS = -I/usr/include/python$(PYTHON_VERSION)`

Don’t add python-devel to `hostmakedepends`, only python, and pass the cross base prefixed include path to make:

```
	hostmakedepends+=" python"
	make_build_args+=" PYTHON_CPPFLAGS=-I${XBPS_CROSS_BASE}/usr/include/python2.7"
fi
```

cr6git:
@newbluemoon Splendid! that works, I can confirm.

abenson:
That's awesome! It worked!

Issue #2867, opened 2015-11-01
Package request: xmonad and xmonad-contrib


sramov:
http://xmonad.org/

chneukirchen:
First stab using `stack`:

```
# Template file for 'xmonad'
pkgname=xmonad
version=0.11.1
revision=1
build_style=haskell-stack
stackage="lts-3.11"
make_build_args="--flag xmonad:small_base"
nocross=yes
hostmakedepends="ghc stack"
makedepends="libXrandr-devel libXinerama-devel libXScrnSaver-devel"
short_desc="Universal converter between markup formats (with citeproc support)"
maintainer="Christian Neukirchen <chneukirchen@gmail.com>"
license="3-clause-BSD"
homepage="http://xmonad.org/"
distfiles="http://hackage.haskell.org/package/${pkgname}-${version}/${pkgname}-${version}.tar.gz"  
checksum=39e0bf227df782d7a5799c811ad0d8b70c9d6eaaa94e8cc395fcf25895d6d2dd

post_extract() {
        # wrong dependency, fixed upstream
        sed -i /utf8-string/s/0.4/1.1/ xmonad.cabal
}  
```

`stack` only can install binaries for now, perhaps someone clueful can see which files from
./masterdir/builddir/xmonad-0.11.1/.stack-work/install/x86_64-linux/lts-3.11/7.10.2/ need to go to /usr.

PR #11741, opened 2018-02-19
New Package: xmonad and xmonad-contrib

phikal:
This should solve #2867.

These are the first packages I created myself, so please tell me if I did something wrong. I could build them on my system, and it seems to work.

### DIFF
--- a/srcpkgs/xmonad-contrib/template
+++ b/srcpkgs/xmonad-contrib/template
@@ -1,0 +1,14 @@
+# Template file for 'xmonad'
+pkgname=xmonad
+version=0.13
+revision=1
+build_style=haskell-stack
+stackage="lts-10.5"
+makedepends="libX11-devel libXinerama-devel libXext-devel"
+depends="xmonad"
+short_desc="Third party tiling algorithms, configurations and scripts for xmonad"
+maintainer="Philip K. <philippija@gmail.com>"
+license="BSD-3"
+homepage="http://xmonad.org/"
+distfiles="http://hackage.haskell.org/package/${pkgname}-${version}/${pkgname}-${version}.tar.gz"
+checksum="a760827fe5b1f99d783f52ccbb72b272d02d53daa26757363cde3ceba014476e"
--- a/srcpkgs/xmonad/template
+++ b/srcpkgs/xmonad/template
@@ -1,0 +1,17 @@
+# Template file for 'xmonad'
+pkgname=xmonad
+version=0.13
+revision=1
+build_style=haskell-stack
+stackage="lts-10.5"
+makedepends="libX11-devel libXinerama-devel libXext-devel libXrandr-devel"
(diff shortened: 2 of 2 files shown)


srcpkgs/xmonad-contrib/template
```diff
@@ -0,0 +1,14 @@
+# Template file for 'xmonad'
+pkgname=xmonad
```
Duncaen: pkgname wrong

srcpkgs/xmonad/template
```diff
@@ -0,0 +1,17 @@
...
+makedepends="libX11-devel libXinerama-devel libXext-devel libXrandr-devel"
+short_desc="A dynamic tiling X11 window manager that is written and configured in Haskell"
+maintainer="Philip K. <philippija@gmail.com>"
+license="BSD-3"
+homepage="http://xmonad.org/"
+distfiles="http://hackage.haskell.org/package/${pkgname}-${version}/${pkgname}-${version}.tar.gz"
+checksum="f9f81b63569f18c777a939741024ec3ae34e4ec84015e5cc50f6622034a303ca"
+
+do_install() {
+	vman man/$pkgname.1
```
fallenwizard: add: **vlicense LICENSE** to do_install() for both of your packages

fallenwizard:
xmonad-contrib/template:
**libXrandr-devel** and **libXft-devel** are missing from makedepends=""
pkg-config is missing from hostmakedepends=""
use xlint (xtools package) to get rid of template issues.

Also you may want to squash your commits so the people with push privileges have an easier time.

phikal:
>Also you may want to squash your commits so the people with push privileges have an easier time.

I will do that, no problem there, and I've implemented most of the fixes - but my problem is that for some reason the `.xbps` file is generated without any binary (there's just a LICENSE, xmonad.1 and the standard xbps files).

gspe:
The funcion
 ```
do_install()
```
is wrong!
As is now it only install license and manpage, you would do this in function
```
post_install()
```

phikal:
@gspe While that was quite a stupid mistake, I'm afraid it didn't quite fix it... xbps-src tells me that it fails to install `ghc-8.2.2_1` as a dep., and that `less-530_1` is broken (because?) of an unresolvable shlib `libpcre.so.1`. The whole transaction fails because of the last message.

gspe:
Hi, I did some research and I've found that you can't use haskell-stack style to build apps like xmonad.
Stack tool lack of `--prefix=` option so it doesn't copy libraries and other things when you use `stack install` command.
See this bug report [Add stack build --prefix option](https://github.com/commercialhaskell/stack/issues/848)

I'm trying to make a template using `runhaskell` command maybe we can try to do this together if you want

phikal:
While I would like to help, I only have minimal experiances with building haskall packages, and I was just hoping that  `stack-install` would take care of it...

PR #11760, opened 2018-02-20
New package: chatty-9.0

beta-phenylethylamine:


srcpkgs/chatty/template
```diff
@@ -0,0 +1,39 @@
...
+  vmkdir /usr/share/${pkgname}/sounds
+  vinstall install/sounds/ding.wav 644 /usr/share/${pkgname}/sounds
+  vinstall install/sounds/dingdong.wav 644 /usr/share/${pkgname}/sounds
+  vinstall install/sounds/typing.wav 644 /usr/share/${pkgname}/sounds
+
+  vinstall $FILESDIR/${pkgname} 755 /usr/bin/
+  vinstall $FILESDIR/${pkgname}.png 644 /usr/share/pixmaps/
+  vinstall $FILESDIR/${pkgname}.desktop 755 /usr/share/applications/
+
+  vlicense install/LICENSE
```
Vaelatern: Remove all leading slashes (etc: replace `/usr/share/applications/` with `usr/share/applications/`)

Vaelatern: Fix xlint bug: indent with tabs!

srcpkgs/chatty/template
```diff
@@ -0,0 +1,39 @@
...
+depends="virtual?java-runtime"
+short_desc="Chatty is a Twitch Chat Client for Desktop"
+maintainer="Phenethylamine <beta.phenylethylamine@gmail.com>"
+license="GPL-3"
+homepage="http://chatty.github.io/"
+distfiles="https://github.com/chatty/chatty/archive/v${version}.tar.gz"
+checksum=856623df053fa5a4498df736825380834c4c5ef9407f81ba13cc947e3fed1a99
+
+do_build () {
+  gradle allPlatformsZip
```
Vaelatern: No better target than a zip file?

Vaelatern:
Please don't have a PNG file in this.
Please get all resources from other places. That's .desktop files and png files.

beta-phenylethylamine:
Alright. This is my first time doing this, sorry. I'll sort those out.

beta-phenylethylamine:
@Vaelatern is the `chatty` script that's put in /usr/bin to launch it also a problem?

beta-phenylethylamine:
I've fixed everything but the inclusion of `chatty.png` and `chatty.desktop`, but it doesn't seem like there's exactly an upstream source for either. There's an ICO file in the repository and a PNG on their website, but I would consider the icon's location on the chatty website to be a little too volatile for me to be comfortable using it in this, and I'm not sure if ICO files can be used.

Since there's no source for a .desktop file should there just be no desktop file?

(I'll squash the commits after I'm done fixing this all)

Vaelatern:
And the png from the website is fine: we sha256sum our dependencies.

beta-phenylethylamine:
I'm looking at the manual and I'm not actually seeing how I should exactly fetch the image, actually. I know I can't use the distfiles variable since `.png` isn't an approved suffix, but should I just use `curl`?

beta-phenylethylamine:
Got it sorted out. Let me know if my method for accessing chatty.png was incorrect. It was all I could figure out, honestly.

beta-phenylethylamine:
Any updates on this? @Vaelatern (don't really know who else to ping, so sorry if you're the wrong person)

PR #11765, opened 2018-02-20
caddy: update to 0.10.11.

cr6git:


cr6git:
[dec5456](https://github.com/voidlinux/void-packages/commit/dec54566857b92e3c93333a75e55fc83842b1750)

PR #11793, opened 2018-02-21
Xorgproto: Transition

maxice8:
Xorgproto is a new package by upstream that unifies all the separate xyzproto packages headers into one main package.

#### Work to be done

- [x] Create xorgproto package
- [x] transform xyzproto into dummy packages
- [x] replace dependencies in packages


maxice8:
for the interested: how xorgproto came to be https://lists.x.org/archives/xorg-announce/2018-February/002835.html

chneukirchen:
I think we can skip the whole transitional packages, and just use `provides=` in the main package.

maxice8:
On Thu, Feb 22, 2018 at 01:55:44PM +0000, Leаh Neukirchen wrote:
> I think we can skip the whole transitional packages, and just use provides= in
> the main package.
> 
> 

Sounds good, i'd honestly just remove them since i also remove all mentions of them and replaced with xorgproto.

chneukirchen:
Just removing them is not enough since everyone who has them installed (e.g. because they compile X11 stuff their own) won't get them updated then.

maxice8:
On Thu, Feb 22, 2018 at 02:00:10PM +0000, Leаh Neukirchen wrote:
> Just removing them is not enough since everyone who has them installed (e.g.
> because they compile X11 stuff their own) won't get them updated then.
> 
> 
Ok, how should i put the replaces= ? just xyzproto-${version}_${revision} ? is there the possibility that the xorgproto version is smaller than of the protocol package and it doesn't get picked up ?

chneukirchen:
2018 should be bigger than any number used (< 10 i think).

But you want `replaces="xyproto>=0 ..."`

maxice8:
On Thu, Feb 22, 2018 at 06:05:48AM -0800, Leаh Neukirchen wrote:
> 2018 should be bigger than any number used (< 10 i think).
> 
> But you want replaces="xyproto>=0 ..."
> 
> 

oh yeah sorry i typed wrongly, i did the replaces= part i mean the provides=

chneukirchen:
yes, provides with 2018 then.

maxice8:
ok i'll update the PR, mind if i notify for review ?

maxice8:
ok done, i made the removal of all proto packages into one commit.

maxice8:
xorgproto 2018.4 was released together with xorg-server-1.20-RC1, xorgproto was updated in this PR.

PR #11794, opened 2018-02-21
stellarium: update to 0.17.0.

Johnnynator:


Johnnynator:
#11529 I should check open PR's before making updates...

PR #11796, opened 2018-02-21
Sudo sssd

inzanez:
sssd support for sudo in LDAP environments. Current version doesn't support sssd.


Vaelatern:
@the-maldridge at least it doesn't seem to pull in a new dependency.
@inzanez please learn how to format commit messages

the-maldridge:
sudo is security critical code, sssd is low quality code.  I cannot approve of this change.  If you want to do this have sudo do a normal group check and derive your authority from that.

inzanez:
@the-maldridge I thought that change wouldn't pull in any other libraries, so the 'sssd' support activated by the change is only code from 'sudo' itself. It would only matter if somebody actually installed 'sssd' and configured it accordingly, wouldn't it?

the-maldridge:
Yes, it would turn on more code within sudo which would attempt to talk to sssd.  This would increase sudo's attack surface without actually increasing its functionality in an appreciable way.

pullmoll:
It could be added as a `build_option="sssd"`, default off, and configure_args containing `$(vopt_enable sssd)`.

inzanez:
@pullmoll I guess that would mean one needed to build it from source anyway, right? So it wouldn't exist as binary package in the main repo.

pullmoll:
@inzanez exactly. It would it make a tad easier for those who need the package to build their own sssd enabled xbps, that's all.

inzanez:
@pullmoll I see. Well, maybe I'll the advice of @the-maldridge and just work with groups. We'll see...

the-maldridge:
I would accept the build option route, that would be a reasonable substitute to turning on this option for all users of sudo.

As an aside, sudo is in the base system, the bar for modifying the base system components is decidedly quite high.

inzanez:
@the-maldridge I guess that makes totally sense. Do I have to close this PR or how does that work? We just decided to use normal sudoers files and work with groups. If we ever needed sssd inside sudo we would possibly host a local repository with a 'sudo-sssd' package,...

Vaelatern:
If you don't mind, go ahead and push a build option approach as @pullmoll suggested? That way at least you just use a local repo and `./xbps-src pkg -o sssd sudo`

PR #11797, opened 2018-02-21
New package: pdfjam-208

Vaelatern:
Doesn't depend on LaTeX, but might need to... it really needs to depend on a tlmgr package.


Vaelatern:
seems to be under tlmgr as well.
I presume one day we will have those. Probably after dxpb.

arete:
Hi was the status of dxpb ?

Il mer 21 feb 2018, 22:20 Toyam Cox <notifications@github.com> ha scritto:

Vaelatern:
@arete the latest code is on github. If you would like, see the latest commit on the master branch and give a shot at continuing the testing of the grapher.

srcpkgs/pdfjam/template
```diff
@@ -0,0 +1,23 @@
+# Template file for 'pdfjam'
+pkgname=pdfjam
+version=208
+revision=1
+wrksrc="${pkgname}"
+build_style=fetch
+short_desc="Collection of shell scripts as a simple interface to pdfLaTeX"
+maintainer="Toyam Cox <Vaelatern@Voidlinux.eu>"
+license="GPL-2"
```
maxice8: use spdx identifier

PR #11801, opened 2018-02-22
python-zeroconf: update to 0.20.0.

maxice8:


asergi:
https://github.com/jstasiak/python-zeroconf#0200

PR #11805, opened 2018-02-22
New package: qmidinet-0.5.0

laserbat:


Johnnynator:
Musl patch
```

--- src/qmidinetUdpDevice.cpp	2017-12-16 13:06:59.438768239 +0100
+++ -	2018-02-22 08:47:58.409130080 +0100
@@ -32,6 +32,7 @@
 #include <unistd.h>
 #include <fcntl.h>
 #include <sys/ioctl.h>
+#include <sys/select.h>
 inline void closesocket(int s) { ::close(s); }
 #endif
```

Johnnynator:
For fixing cross you can try adding qt5-devel to hostmakedepends additionally. It is not a nice solution but it works quite often.

laserbat:
Thank you! I'll add the patch asap, and thanks for the tip with qt5-devel, I'll try it.

newbluemoon:
The problem is that this is a mixture of gnu-configure and qmake. `configure` generates the first level `Makefile` which calls qmake to generate `src/Makefile`. So `QMAKE_CXX` etc. aren’t set and it uses the wrong compiler. But we can add the missing variables:

```
	hostmakedepends+=" qt5-devel"

	pre_configure() {
		sed -i -e "$ a QMAKE_CC=${CC}\nQMAKE_CXX=${CXX}\nQMAKE_LINK=${CXX}" src/src.pri.in
	}
fi
```

Also, you can remove qt5-qmake from `hostmakedepends` because it’s pulled in automatically by qt5-devel.

newbluemoon:
Sorry, what I meant was to remove qt5-qmake and not replace it with qt5-devel. ;)
qt5-devel is already in `makedepends` which is ok for a native build and when cross building we add it to `hostmakedepends` which also makes qmake available. :)

laserbat:
Thank you, fixed it!

PR #11806, opened 2018-02-22
rosegarden: update to 17.12.1

newbluemoon:


newbluemoon:
The log is too big again.

PR #11811, opened 2018-02-22
 kmymoney: update to 5.0.0

Hoshpak:


Hoshpak:
Travis fails to install cmake despite ```build_style=cmake```, builds fine locally.

PR #11816, opened 2018-02-22
postfix: update to 3.3.0.

maxice8:


chneukirchen:
> Postfix 3.3 will run in the foreground with "postfix start-fg".

This sounds like a good idea for the sv script perhaps.

newbluemoon:
There is a [forum thread](https://forum.voidlinux.eu/t/postfix-utf-8-support-unavailable/4690) about the cross-built postfix not using <i>icu</i>. Perhaps this could be fixed as well?

A solution I found was to change the `sed` command in `do_build()` with

```
sed -e "s,\(/usr/include/db.h\),${XBPS_CROSS_BASE}/\1,g" \
 -e 's,^\(\s\+\)CCARGS="$CCARGS -DNO_EAI",\1CCARGS="$CCARGS $icu_cppflags"\n\1SYSLIBS="$SYSLIBS $icu_ldflags",' \
 -i makedefs
```

which just sets the values used when the test for <i>icu</i> was successful in the branch when the test failed, but we know that <i>icu</i> is there, anyway.

maxice8:
On Fri, Feb 23, 2018 at 01:52:49PM +0000, Leаh Neukirchen wrote:
>     Postfix 3.3 will run in the foreground with "postfix start-fg".
> 
> This sounds like a good idea for the sv script perhaps.
> 
> —
> 

Made the change

chneukirchen:
Reverted it again, it breaks sv stop. Nvm.

PR #11823, opened 2018-02-23
New package: ssh-audit-1.7.0

abenson:


srcpkgs/ssh-audit/template
```diff
@@ -0,0 +1,23 @@
...
+short_desc="SSH server auditing"
+maintainer="Andrew Benson <abenson+void@gmail.com>"
+license="MIT"
+homepage="https://github.com/arthepsy/ssh-audit"
+distfiles="https://github.com/arthepsy/ssh-audit/archive/v${version}.tar.gz"
+checksum=cba29cc19ec2932e4f43c720b2c49a7d179219e23482476aeb472f7463713b68
+
+do_build() {
+	echo "Nothing to build."
+}
```
Vaelatern: `build_style=fetch`

srcpkgs/ssh-audit/template
```diff
@@ -0,0 +1,23 @@
...
+
+do_install() {
+	vbin ssh-audit.py ssh-audit
+
+	# License was missing in release
+	vlicense ${FILESDIR}/LICENSE
```
Vaelatern: Then find the license elsewhere. We will NOT vendor it.

abenson: Ah, fair enough. I can have it fetch it as well

jnbr:
since it's only a python script, there should be `noarch=yes`

Vaelatern:
Ok, finally I think I'd like to see it on python3.

abenson:
I've been wondering about the python/python2/python3 and if there's a
transition plan.

--
AndrewB
Knowing is Half the Battle.

Vaelatern:
The "transition plan" is to depend on python3 instead of python where possible. Then we wake up one morning and nothing depends on python2.7

Vaelatern:
As per: common/hooks/pre-pkg/03-rewrite-python-shebang.sh
set `python_version=3`

PR #11845, opened 2018-02-23
New package: argtable

Magister-Bergerac:


jnbr:
Since this package provides a shared library, an entry in `common/shlibs` is needed: 
`=> WARNING: argtable-2.13_1: libargtable2.so.0 not found in common/shlibs!`

Also, files only needed for development should go into a `argtable-devel` subpackage.

maxice8:
> Since this package provides a shared library, an entry in common/shlibs is needed:

Not really, lots of packages that provide libs have no shlib entry, it is most common to add only when a package makes use of it.

jnbr:
> Not really, lots of packages that provide libs have no shlib entry, it is most common to add only when a package makes use of it.
Thank you, wasn't aware of that.
Sorry @Magister-Bergerac for wrong information.

maxice8:
If you're going to add shlib then make it in the same commit as the one you introduced the package, also clustalo seems unrelated to argtable, make a separate PR for it if you plan on adding it.

Magister-Bergerac:
How do I add it in the same commit?

Vaelatern:
You amend commits. Or fix them up. Google it.

common/shlibs
```diff
@@ -16,6 +16,7 @@
 # PLEASE NOTE: when multiple packages provide the same SONAME, the first
 # one (order top->bottom) is preferred over the next ones.
 #
+libargtable2.so.0 glibc-2.25_1
```
Vaelatern: NO.
NO.
NO.

srcpkgs/argtable/template
```diff
@@ -0,0 +1,12 @@
+# Template file for 'argtable'
+pkgname=argtable
+version=2.13
+revision=1
+wrksrc="argtable2-13"
```
Vaelatern: `wrksrc="${pkgname}${version//./-}"`

common/shlibs
Vaelatern: No.
Remove the change completely.

Vaelatern:
Shlibs still broken. I'm not closing the PR, but I'm not merging it.

Vaelatern:
Please clean the commit history. You need to have one commit.
That or I can squash and merge, but I won't with the current `short_desc`. Understand why as follows:
```
[-] argtable-2.13_1                 Argtable is an ANSI C command line parser
```
So it should be `ANSI C command line parser`

PR #11847, opened 2018-02-23
openrazer-meta: update to 2.2.2

z3ntu:
I didn't test it because my void vm is broken for some reason - but the code works fine on other distros and the packaging didn't change so it should be fine


z3ntu:
@pullmoll done

PR #11851, opened 2018-02-24
linux4.1: update to 4.1.49.

cr6git:
Currently using it.


cr6git:
No one seems intesrested in merging this? ok, I close.

PR #11872, opened 2018-02-25
New package: VeraCrypt-1.21

gustavoheinz:
This closes #7035, there is no updates since October... So I created a new template for it.

### DIFF
--- a/srcpkgs/VeraCrypt/template
+++ b/srcpkgs/VeraCrypt/template
@@ -1,0 +1,33 @@
+# Template file for 'Veracrypt'
+
+pkgname=VeraCrypt
+version=1.21
+revision=1
+create_wrksrc=1
+build_wrksrc="src"
+build_style=gnu-makefile
+make_use_env=1
+hostmakedepends="yasm pkg-config"
+makedepends="wxWidgets-devel fuse-devel"
+short_desc="Disk encryption with strong security based on TrueCrypt"
+maintainer="Gustavo Heinz <gustavoheinz95@gmail.com>"
+license="Apache 2.0, TrueCrypt 3.0"
+homepage="https://www.veracrypt.fr"
+distfiles="https://launchpad.net/veracrypt/trunk/${version}/+download/${pkgname}_${version}_Source.tar.bz2"
+checksum="44717d24b8deff2db9dd0365572063e38bbe4a59438a8c3c07675b5eed370358"
+
+case "$XBPS_TARGET_MACHINE" in
+	i686*|x86_64*) ;;
+	*) make_build_args+=' NOASM=1' ;;
+esac
+
+if [ -n "$CROSS_BUILD" ]; then
+	make_build_args+=' NOTEST=1'
+fi
+
+do_install() {
+	vbin Main/veracrypt
+	vlicense License.txt
+	vinstall Setup/Linux/veracrypt.desktop 644 usr/share/applications/
+	vinstall Resources/Icons/VeraCrypt-256x256.xpm 644 usr/share/pixmaps/ veracrypt.xpm
+}


cr6git:
````
do_install() {
	cd ${wrksrc}/src
````
As you have already defined `build_wrksrc="src"` , are you sure you need to `cd` into it?

Also, you can replace these 4 lines:
````
vmkdir /usr/share/applications
vmkdir /usr/share/pixmaps
vcopy Setup/Linux/${_pkgname}.desktop usr/share/applications/
vcopy Resources/Icons/VeraCrypt-256x256.xpm /usr/share/pixmaps/{_pkgname}.xpm
````
With something like:
````
vinstall Setup/Linux/${_pkgname}.desktop 644 usr/share/applications/
vinstall Resources/Icons/VeraCrypt-256x256.xpm 644 usr/share/pixmaps/ {_pkgname}.xpm
````

**Comment by gustavoheinz** [comment]
@cr6git I have no idea why I put that `cd` in there, but yes, it's not needed. Also, the `vinstall` is much cleaner, thanks.

@newbluemoon Thanks for the suggestions, I tried to look into the makefile and I did not find other archs in the code, so I concluded that no arch other than x86_64 and i686 was supported.

I'm interested in doing the `NOGUI=1` binary, but it's not clear to me how to do it in the template.

**Comment by newbluemoon** [comment]
To create a nogui-package simultaniously you have to use a custom `do_build()` stage, because there have to be two builds. And remove `build_wrksrc` because we’re changing that manually:

```
do_build() {
	cp -a src src-nogui
	cd src
	make $makejobs $make_build_args
	cd ../src-nogui
	make $makejobs $make_build_args NOGUI=1 NOTEST=1
}
```
Had to add `NOTEST` to the nogui-build else it complains: ‘Unable to initialize GTK+’. I’ll try to find out where this comes from.

And create a subpackage, e.g.:

```
VeraCrypt-nogui_package() {
	short_desc+=" (nogui)"
	conflicts="VeraCrypt"
	pkg_install() {
		vbin ${wrksrc}/src-nogui/Main/veracrypt
		vlicense ${wrksrc}/src-nogui/License.txt
	}
}
```

Lastly, make both packages conflict with each other (they provide the same binary name); similarly to the subpackage add `conflicts="VeraCrypt-nogui"` to the main template body.

The `_pkgname` variable is just used once in `do_install()`. I think it’s better to remove it and simply use
`vinstall Resources/Icons/VeraCrypt-256x256.xpm 644 usr/share/pixmaps/ veracrypt.xpm`
(The `$` was missing anyway. ;))

This should be it. :)

P.S.: Don’t forget to create a symbolic link in `srcpkgs`: `ln -s VeraCrypt VeraCrypt-nogui`

**Comment by newbluemoon** [comment]
In `do_install()` you have to do a `cd src` now.

**Comment by gustavoheinz** [comment]
> Had to add NOTEST to the nogui-build else it complains: ‘Unable to initialize GTK+’. I’ll try to find out where this comes from.
Is that because VeraCrypt is heavily dependent on wxWidgets?

@newbluemoon Thanks a lot for the help, you ended up doing all the work :)

**Comment by newbluemoon** [comment]
> Is that because VeraCrypt is heavily dependent on wxWidgets?

You know, I just realised that the nogui package has the same dependencies as the gui package! I guess the hard truth is that this makes it quite useless in respect to why we made it in the first place. So perhaps it’s better to just have one (the gui) version? Any opinions on this?

> you ended up doing all the work

Well, I was tinkering with this recently and the work was already (almost) done, so why not share it? :)

File: srcpkgs/VeraCrypt/template [hunk 1]
```diff
@@ -0,0 +1,52 @@
+# Template file for 'Veracrypt'
+
+pkgname=VeraCrypt
+conflicts=VeraCrypt-nogui
```
Vaelatern: No, this variable does not belong here at all.

srcpkgs/VeraCrypt/template
```diff
@@ -0,0 +1,52 @@
...
+create_wrksrc=1
+build_style=gnu-makefile
+make_use_env=1
+hostmakedepends="yasm pkg-config"
+makedepends="wxWidgets-devel fuse-devel"
+short_desc="Disk encryption with strong security based on TrueCrypt"
+maintainer="Gustavo Heinz <gustavoheinz95@gmail.com>"
+license="Apache 2.0, TrueCrypt 3.0"
+homepage="https://www.veracrypt.fr"
+_filename="${pkgname}_${version}_Source.tar.bz2"
```
Vaelatern: Variable used in one place? Get rid of it.

gustavoheinz:
> So perhaps it’s better to just have one (the gui) version? Any opinions on this?

@newbluemoon Yeah, I didn't thought that through very well, maybe it's actually saner to just have one version of the package and the users can simply run `veracrypt -t`, until the veracrypt team actually makes it non dependant on wxWidgets...

srcpkgs/VeraCrypt/template
```diff
@@ -0,0 +1,52 @@
...
+do_build() {
+	cp -a src src-nogui
+	cd src
+	make $makejobs $make_build_args
+	cd ../src-nogui
+	make $makejobjs $make_build_args NOGUI=1 NOTEST=1
+}
+
+do_install() {
+	cd src
```
Vaelatern: Please don't. Work relative to whatever dir you are already in.

srcpkgs/VeraCrypt/template
```diff
@@ -0,0 +1,52 @@
...
+if [ -n "$CROSS_BUILD" ]; then
+	make_build_args+=' NOTEST=1'
+fi
+
+do_build() {
+	cp -a src src-nogui
+	cd src
+	make $makejobs $make_build_args
+	cd ../src-nogui
+	make $makejobjs $make_build_args NOGUI=1 NOTEST=1
```
Vaelatern: Wasn't it decided that nogui has the same output as gui?

gustavoheinz: I'll remove the nogui subpackage

gustavoheinz:
Forgot to comment, I have fixed all the comments from @Vaelatern, should be good now.

PR #11880, opened 2018-02-25
New package: frandom-1.2: fast kernel PRNG. 

bitshark:


maxice8:
everything as one commit, remove commented sections on template

bitshark:
@maxice8   Okay , updated

biopsin:
Nice to see this in repo, tnx!
What about the 10-frandom.rules not been copied to /usr/lib/udev/rules.d/ 
will only be availible to root without.

bitshark:
@biopsin   ah nice catch...  you're right -- in current form, only root can use /dev/frandom.   I will update it to install the rule into udev.

biopsin:
oh I forgot the 10-frandom.rules need unfortunately a tweak, remove the NAME="%k", or a warning will show on boot, while at it maybe adding to INSTALL.msg
```
:extra tip
adding a weekly cronjob to reinitialize after long uptimes
  /usr/bin/dd if=/dev/frandom of=/dev/null count=1 >/dev/null 2>&1

substitute random/urandom with frandom/erandom in rc.local
  (ln -sf /dev/frandom /dev/random && ln -sf /dev/erandom /dev/urandom) &
```
however the last tip requires another tweak to MODE in 10-frandom.rules from 444 to 666

chneukirchen:
IMO frandom is quite useless, it uses insecure crypto and was never meant to be secure. Also, since 4.9 urandom is pretty quick. Do we really need this?

bitshark:
@biopsin   okay I will make those changes

@chneukirchen  I've found frandom has useful applications (mainly in statistics and research, for use with languages such as R , etc).  While you probably would not want to use frandom for cryptographically secure entropy, I think it's useful for two reasons.  (1) with older 3.x and 4.x kernels it is much faster than urandom,  (2) it passes PRNG statistical tests like Diehard.

Additionally, most practical concerns with regard to using RC4 as a PRNG can be resolved by using RC4-drop[n], where n is the first n bytes of the keystream.  

``` 
The paper by Ilya Mironov says: Our most conservative recommendation is based on the experimental data on the tail probability of the strong uniform time T (Section 5.5). This means that discarding the initial 12 × 256 bytes most likely eliminates the possibility of a strong attack.
```
http://www.users.zetnet.co.uk/hopwood/crypto/scan/cs.html#RC4-drop 

I will add a patch to convert this module to RC4-Drop[8192] as a precaution , and I will add a warning that frandom should only be used for research and statistical applications, and not for any cryptographic ones.

bitshark:
@biopsin @maxice8 

Okay, so here's the changes I added:
(1) Updated description to clarify frandom is only faster on pre 4.9 series kernels.
(2) Included a patch to ensure frandom defaults to the more secure RC4-Drop[n] algorithm (with n=8192)
(3) Included a patch to remove the NAME=%k directive on the udev rule to avoid warning on boot
(4) Added a warning to the INSTALL message that frandom should be used with caution in production systems which require a high-security CS-PRNG, with a link to an article explaining the full crypto details.

chneukirchen:
If you need fast random numbers, why not generate them in user-space in first place? Saves lots of context-switches too.

YMMV.

PR #11888, opened 2018-02-25
[do not merge] rust: update to 1.24.0.

jnbr:
can someone please put `rustc-1.23.0-x86_64-unknown-linux-musl.tar.gz` and `rust-std-1.23.0-x86_64-unknown-linux-musl.tar.gz` up on https://repo.voidlinux.eu/distfiles?


jnbr:
local builds are good for all three (`x86_46`, `x86_46-musl` and `i686`).

Vaelatern:
I think changing the link to those distfiles is in error.

srcpkgs/rust/template
```diff
@@ -18,40 +18,40 @@ maintainer="Leah Neukirchen <leah@vuxu.org>"
...
-checksum=7464953871dcfdfa8afcc536916a686dd156a83339d8ec4d5cb4eb2fe146cb91
+checksum=bb8276f6044e877e447f29f566e4bbf820fa51fea2f912d59b73233ffd95639f
 
 case "$XBPS_MACHINE" in
 x86_64-musl)
 	hostmakedepends+=" libcurl libgit2"
 	distfiles+="
-	 https://repo.voidlinux.eu/distfiles/rustc-1.22.1-x86_64-unknown-linux-musl.tar.gz
-	 https://repo.voidlinux.eu/distfiles/rust-std-1.22.1-x86_64-unknown-linux-musl.tar.gz
+	 https://repo.voidlinux.eu/distfiles/rustc-1.23.0-x86_64-unknown-linux-musl.tar.gz
```
Arzte: Here.

srcpkgs/rust/template
```diff
@@ -18,40 +18,40 @@ maintainer="Leah Neukirchen <leah@vuxu.org>"
...
 	;;
 x86_64)
 	# extract from src/stage0.txt
 	distfiles+="
-	 https://dev-static.rust-lang.org/dist/2017-11-20/rustc-1.22.0-x86_64-unknown-linux-gnu.tar.gz
-	 https://dev-static.rust-lang.org/dist/2017-11-20/rust-std-1.22.0-x86_64-unknown-linux-gnu.tar.gz
-	 https://dev-static.rust-lang.org/dist/${_date}/cargo-0.24.0-x86_64-unknown-linux-gnu.tar.xz"
+	 https://static.rust-lang.org/dist/${_date}/rustc-1.23.0-x86_64-unknown-linux-gnu.tar.gz
+	 https://static.rust-lang.org/dist/${_date}/rust-std-1.23.0-x86_64-unknown-linux-gnu.tar.gz
+	 https://static.rust-lang.org/dist/${_date}/cargo-0.24.0-x86_64-unknown-linux-gnu.tar.xz"
```
Arzte: And here.

srcpkgs/rust/template
```diff
@@ -18,40 +18,40 @@ maintainer="Leah Neukirchen <leah@vuxu.org>"
...
 	;;
 i686)
 	# extract from src/stage0.txt
 	distfiles+="
-	 https://dev-static.rust-lang.org/dist/2017-11-20/rustc-1.22.0-i686-unknown-linux-gnu.tar.gz
-	 https://dev-static.rust-lang.org/dist/2017-11-20/rust-std-1.22.0-i686-unknown-linux-gnu.tar.gz
-	 https://dev-static.rust-lang.org/dist/${_date}/cargo-0.24.0-i686-unknown-linux-gnu.tar.xz"
+	 https://static.rust-lang.org/dist/${_date}/rustc-1.23.0-i686-unknown-linux-gnu.tar.gz
+	 https://static.rust-lang.org/dist/${_date}/rust-std-1.23.0-i686-unknown-linux-gnu.tar.gz
+	 https://static.rust-lang.org/dist/${_date}/cargo-0.24.0-i686-unknown-linux-gnu.tar.xz"
```
Arzte: And here.

jnbr:
> I think changing the link to those distfiles is in error.
Rust version 1.n is usually build with rust version 1.(n-1). I See two options here:
1) update the distfiles with each rust update (as it was done previous)
2) keep the distfiles and build all previous releases between the one in the distfiles and the pkgversion

option 2 doesn't really scale.
Do you see another option?

srcpkgs/rust/template
jnbr: good idea, thanks

Vaelatern:
@Arzte I think you meant `$((version-.01))`

Vaelatern:
Can it be built by choosing a version and just going from there? Messing with distfile mirrors is annoying. That or change those mirrors to be upstream.

jnbr:
> Can it be built by choosing a version and just going from there?

not by itself. We would need to run it in a loop ourselves.

srcpkgs/rust/template
```diff
@@ -1,8 +1,10 @@
 # Template file for 'rust'
 pkgname=rust
-version=1.23.0
+version=1.24.0
 revision=1
-_date=2018-01-02
+_date=2018-01-04
+_rust_dist_version=1.23.0
```
Arzte: This supposed to be 1.23?

jnbr: no. 1.23.0 is the version.

Arzte: Why is the rust package out of sync with the rust version?

jnbr: bootstrapping is done with previous version

Arzte:
@Vaelatern Why would it be ``$((version-.01))`` shouldn't it be ``${version}``? Since ``${version}`` would equal ``1.24.0`` (per this pr bumping the package version) which would be correct to how rust puts out their dist files?

jnbr:
@Arzte maybe this helps to understand: https://github.com/rust-lang/rust/blob/1.24.0/src/stage0.txt

Arzte:
@jnbr Seems like a somewhat silly way to do things... Might want to make a note incase someone else is confused by it. (very least that link)

Vaelatern:
alternately we have two pkgs. "rust-prev" and "rust", depending on "rust-prev" to build "rust"

jnbr:
than we would have the same problem with "rust-prev" as we have now with "rust". We would need the full chain, starting with the latest version of the musl-tarball on repo.voidlinux.eu/distfiles

chneukirchen:
Please. I'll do this. Thanks @jnbr for updating the template. I'll get around to it this week.

PR #11901, opened 2018-02-26
clipgrab: update to 3.6.8.

maxice8:


cr6git:
[#11831](https://github.com/voidlinux/void-packages/pull/11831)

Issue #11911, opened 2018-02-26
Libressl 2.6 Updates


Gottox:
* [ ] FreeRADIUS
* [ ] Ice
* [x] MEGAsdk
* [x] android-tools
* [ ] bacula-common
* [x] badvpn
* [x] balsa
* [x] bind
* [x] bitchx
* [x] bitcoin
* [ ] bitlbee
* [x] boinc
* [x] borg
* [x] botan
* [ ] burp-server
* [x] burp2-server
* [ ] ca-certificates
* [x] calibre
* [x] cargo
* [x] ccnet
* [x] cgit
* [x] clamav
* [ ] collectd
* [ ] coturn
* [x] cryptsetup
* [x] crystal
* [x] ctpp2
* [x] curl
* [x] cyrus-sasl
* [x] dillo
* [x] dma
* [x] dovecot
* [x] easyrsa
* [ ] ecryptfs-utils
* [x] efl
* [x] eiskaltdcpp-qt
* [ ] ejabberd
* [x] elinks
* [x] encfs
* [ ] erlang
* [x] ettercap
* [x] facter
* [x] fdm
* [x] fetchmail
* [x] fossil
* [x] freerdp
* [ ] freshplayerplugin
* [x] gambit
* [x] gatling
* [x] gftp
* [x] git
* [x] git-crypt
* [x] git-series
* [ ] gkrellm
* [ ] glusterfs
* [x] godot
* [x] grub-customizer
* [x] gsoap
* [ ] gst-plugins-bad
* [ ] gst-plugins-bad1
* [ ] gwenhywfar
* [ ] h2o
* [x] haproxy
* [ ] hexchat
* [x] hfsprogs
* [ ] hitch
* [x] hostapd
* [ ] hplip
* [ ] httperf
* [ ] httrack
* [x] i2pd
* [ ] icecast
* [ ] icinga2
* [x] imapfilter
* [ ] inadyn
* [ ] inspircd
* [ ] ipmitool
* [ ] ipmiutil
* [x] iputils
* [x] irssi
* [x] isync
* [x] john
* [x] kea
* [x] keepalived
* [ ] kicad
* [ ] kvirc
* [ ] lastpass-cli
* [ ] ldapvi
* [x] ldns
* [x] lftp
* [x] libarchive
* [x] libesmtp
* [x] libetpan
* [x] libevent
* [x] libfetch
* [ ] libgda
* [x] libgit2
* [ ] libimobiledevice
* [ ] libmega
[...]

PR #11918, opened 2018-02-26
newsbeuter: warn of deprecation

maxice8:
for #11911


Vaelatern:
Why is newsbeuter obsolete? Link?

maxice8:
Not maintained anymore, https://github.com/akrennmair/newsbeuter

Vaelatern:
Then no transitional package. They are different projects.
A one-line INSTALL.msg saying "This package has been abandoned upstream" and a revbump is probably best.

PR #11987, opened 2018-02-28
New package: alert-after-1.4.3

cr6git:
Very useful.
For this to work you will need a notification daemon, like `xfce4-notifyd` for example.


maxice8:
can't this really be a simple shellscript ?

cr6git:
@maxice8 no clue... I propose you send your suggestions to the author [Corey Farwell](https://github.com/frewsxcv).

PR #11993, opened 2018-02-28
New package: vtk-8.1.0

Piraty:
This is a dependency for the [soon awaited freecad](https://forum.freecadweb.org/viewtopic.php?f=10&t=26174) update.

The build usually hits travis' timecap.


faulesocke:
Is it really necessary to compile vtk with all these features (especially Qt5)? I did a successful FreeCAD git build today with just the standard vtk build options. It only required MesaLib-devel and libXt-devel as build dependencies.

Piraty:
I guess it doesn't have to be full-blown.
but since my testing of vtk's interaction with freecad so far was limited, i went to full-blown first instead of missing sth later on.
I guess that can be slimmed down, will approach that soon

Piraty:
ok, here go.

@faulesocke so i added qt5 build option which is disabled by default.
If you find anything else in particular that you don't like to see and that is not required for (currently `vtk`'s only dependency) freecad, please speak up.

Also, i will take a look in what specific `libvtk*` is required by freecad 0.17, so we can slim that down a bit (maybe split vtk into subpackages...?).

Piraty:
This is the slimmed-down version of it.

PR #11997, opened 2018-02-28
xfce4-terminal: update to 0.8.7.1.

maxice8:


cr6git:
[#11890](https://github.com/voidlinux/void-packages/pull/11890)

PR #11998, opened 2018-02-28
vice: switch to GTK3 UI by default, replace SDL with SDL2

skrzyp:
• GTK3 UI has more convenient monitor/debugger than SDL2 UI, and it's overall "better" for development purposes
• SDL UI has been upgraded to SDL2 as it's supported since 3.x. I'm still leaving it as it is because it's more suitable for ad-hoc emulation/gaming.
• I've merged GTK3 vopt with Pulseaudio and told the SDL UI to use SDL's sound subsystem instead (to reduce redundancy, there's no added latency)

**BUT**: I wanted to add `ReSID(fp)` which is the "real" SID emulation (with filters) instead of the "fast SID" emulation, the buildsystem says something weird:
```
configure: WARNING: g++ is not the intended cross-compiler, ReSID will not be configured
```

Surprisinglyy, when compiled out of the `xbps-src`, it builds fine with ReSID too. What could be wrong here?

### DIFF
--- a/srcpkgs/vice/template
+++ b/srcpkgs/vice/template
@@ -1,13 +1,26 @@
 # Template file for 'vice'
 pkgname=vice
 version=3.1
-revision=1
+revision=2
 build_style=gnu-configure
-configure_args="--enable-sdlui $(vopt_with pulseaudio pulse)
- --disable-option-checking"
+configure_args="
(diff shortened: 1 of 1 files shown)


skrzyp:
`*bump*`

(I would like to retry tests in-place, but looks like I can't)

PR #11999, opened 2018-02-28
mc: switch ncurses to slang, add mksh support patch

skrzyp:
* Switched to `slang` console library instead of `ncurses`:
  * It's more responsive when used with mouse
  * Does not produce garbage after exiting from programs acting unaccordingly with terminals, refreshes screen nicely
  * Allows to assign own characters for frames and other UI elements
  * Works with 256-16M color terminals
    * (both of these functionalities can be tested with bundled themes)
* Added a patch to support `mksh` shell as subshell, with default configuration it just dropped us to empty pty after pressing ^O

### DIFF
--- a/srcpkgs/mc/patches/mc-mksh-subshell-v2.patch
+++ b/srcpkgs/mc/patches/mc-mksh-subshell-v2.patch
@@ -1,0 +1,73 @@
+--- lib/shell.c	2016-09-22 18:24:12.000000000 +0000
++++ lib/shell.c	2017-02-16 13:47:34.933939000 +0000
+@@ -66,6 +66,10 @@
+     /* 3rd choice: look for existing shells supported as MC subshells.  */
+     if (access ("/bin/bash", X_OK) == 0)
+         mc_shell->path = g_strdup ("/bin/bash");
++    else if (access ("/usr/local/bin/mksh", X_OK) == 0)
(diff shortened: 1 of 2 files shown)


chneukirchen:
Why can't mc just use $SHELL?

skrzyp:
`mc` is just kinda weird and it handles subshell invocation **separately** for each shell.
The patch I've posted is coming from the bug opened on mc's tracker just to support `mksh` as subshell: https://midnight-commander.org/ticket/3748 

patch: https://midnight-commander.org/attachment/ticket/3748/mc-mksh-subshell-v2.patch

Back to slang, I can elaborate on some cases where ncurses fails (and slang works just fine) if you *really* want, but let's just assume that I somehow know what am I doing. Also, RHELs moved on slang backend a while ago, and it really helped us with running mc on a lot of servers.

skrzyp:
`*bump*`

pullmoll:
@voidlinux/pkg-committers to merge or not to merge? That is the question.

Gottox:
+1 for the mksh part.

slang is okay with me too. Are there any issues with slang?

pullmoll:
> Are there any issues with slang?

Probably not or [Arch](https://www.archlinux.org/packages/community/x86_64/mc/) would have a patch.

PR #12001, opened 2018-02-28
icecat: update to 52.6.0.

cr6git:


cr6git:
[f788af7](https://github.com/voidlinux/void-packages/commit/f788af7b6ade5484dfd9a3b896a0b178f17ff664)

Issue #6486, opened 2017-05-23
[Proposal] Remove artwork for grub


Gottox:
The artwork of grub in voidlinux has always been pretty ugly. Instead of improving the artwork itself I'm proposing to remove the background completely from grub.

IbnSaeed:
why ? just replace the background image yourself.

Vaelatern:
@IbnSaeed because we can set the policy.

The grub artwork has mostly made the boot options unreadable. I'd be happy with better artwork that didn't have this property

silvernode:
I'd like to see some sort of style in the grub menu, even if there is no image.

logenkain:
I usually don't even notice the artwork. I'm happy as is, but don't really care either way.

mrnhmath:
I agree.

Piraty:
Because the image is so uggly I always was thankful for that installer-option to use grub's textmode :-)

silvernode:
In my opinion the artwork is not actually bad, it's just low resolution. However I think there should be no image but instead, just match the grub theme (selection highlighting, text, ect) to the Void Linux website.

mrnhmath:
What about using a plain background with Void's green color #478061?

cr6git:
Just to say, I opened a pull request. 😉

cr6git:
@Gottox I opened a pull request for your proposal.

cr6git:
It would make sense to also remove the artwork in LXDM, which is the same as GRUB.
````
/etc/lxdm/lxdm.conf

## background of the greeter
bg=/usr/share/void-artwork/splash.png
````

PR #12003, opened 2018-02-28
lxdm: remove Void artwork

cr6git:
[#6486](https://github.com/voidlinux/void-packages/issues/6486)


cr6git:
I can build the package and install it, but `lxdm` is now broken.

PR #12055, opened 2018-03-02
hooks/do-fetch/00-distfiles.sh: contents checksum

pullmoll:
Add support for contents checksums. This is meant to be used for tarballs
like the ones coming from https://*.googlesource.com/ which on every download
change their outer checksum due to being packaged with different date/time
stamps or similar.

If such a distfile still contains the same files as the author of a template
found, then the checksum of its contents can be specified by prepending a
commercial at (@) to the corresponding checksum="..." line.

Depending on the format of the distfile its entire contents will then be
piped into sha256sum and the resulting digest is verified against what
follows after the @.

An example is the pending package addition of [gyp](https://github.com/voidlinux/void-packages/pull/11729). The following template uses the tarball:
```
# Template file for 'gyp'
pkgname=gyp
version=20180218
revision=1
_commit=4d467626b0b9f59a85fb81ca4d7ea9eca99b9d8f
create_wrksrc=yes
build_style=python-module
hostmakedepends="python-setuptools git"
depends="python-setuptools"
short_desc="GYP (Generate Your Projects) meta-build system"
maintainer="John <johnz@posteo.net>"
license="BSD"
homepage="https://gyp.gsrc.io"
distfiles="https://chromium.googlesource.com/external/gyp/+archive/${_commit}.tar.gz"
# The checksum changes every time it gets downloaded, thus specify the contents checksum
checksum=@26490d7015d997ba306f6b45cd430e40f6434a3e7215a5ab62c38b1f2a51bed1

post_install() {
        vlicense LICENSE
}
```

### DIFF
--- a/Manual.md
+++ b/Manual.md
@@ -419,6 +419,13 @@ Example:
 separated by blanks. Please note that the order must be the same than
 was used in `${distfiles}`. Example `checksum="kkas00xjkjas"`
 
+If a distfile changes its checksum for every download because it is packaged
+on the fly on the server, like e.g. snapshot tarballs from any of the
+`https://*.googlesource.com/` sites, the checksum of the `archive contents`
+can be specified by prepending a commercial at (@).
+For tarballs you can find the contents checksum by using the command
+`tar xf <tarball.ext> --to-stdout | sha256sum`.
+
 - `wrksrc` The directory name where the package sources are extracted, by default
 set to `${pkgname}-${version}`.
 
--- a/common/hooks/do-fetch/00-distfiles.sh
+++ b/common/hooks/do-fetch/00-distfiles.sh
@@ -20,23 +20,121 @@ get_cksum() {
 	echo "$cksum"
 }
 
+# Return the checksum of the contents of a tarball
+contents_cksum() {
+	local curfile="$1" cursufx cksum
+
+	case $curfile in
+	*.tar.lzma)   cursufx="txz";;
+	*.tar.lz)     cursufx="tlz";;
+	*.tlz)        cursufx="tlz";;
+	*.tar.xz)     cursufx="txz";;
+	*.txz)        cursufx="txz";;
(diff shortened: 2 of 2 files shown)


Vaelatern:
:shipit:

ebfe:
> If such a distfile still contains the same files as the author of a template
> found

Note that this (at least for tar files) doesn't cover filenames, permission or symlink targets.

pullmoll:
File permissions are disregarded when extracting a tarball, see `common/hooks/do-extract/00-distfiles.sh`. Symbolic links may impose a risk, that's true.

What consequences could a malicious symbolic link have?
I can't imagine what to attack inside the chroot build environment.

Perhaps we can improve the checksumming to also include the bare file list. With tar or bsdtar that would involve a second command to create the naked file list to prepend or append to the raw data.
For unzip there is an option `-c` which includes the filenames in the stream.

ebfe:
> I can't imagine what to attack inside the chroot build environment.

It has unrestricted access to the hostdir and the repo. File boundaries are lost too and filenames aren't hashed either, so you can choose an arbitrary slice of data from the content as e.g. Makefile/build script etc.

The build chroot doesn't really protect the host system from an attacker that can execute code in the chroot. He can just insert whatever command he wants into /void-packages/xbps-src and have it executed on the host when the next build is started.

But enough paranoia... this was mostly a comment on how the content checksum might actually check less than initially thought.

pullmoll:
Well, you are right, a content checksum is inherently less safe than an outer checksum of the archive. We will use it in only very few cases and there it is a real relief compared to checking out commits from a repo.

Perhaps we can do something to better protect the mounted host paths in the chroot. At least mounting `/void-packages` read-only to prevent anyone from doing evil things there should be easy enough. Also `/host/{binpkgs,sources,repocache}` could be mounted read-only as they don't have to be modified by the build process, but only by the common scripts or xbps-src itself. Just an idea which may not be feasible, though.

PR #12056, opened 2018-03-02
ssr: fix cross compile

abenson:


newbluemoon:
Perhaps it would be better to move the extra `configure_args` from the cross branch to something like

```
        i686*|x86_64*) ;;
        *) configure_args+=" -DENABLE_X86_ASM=no -DWITH_GLINJECT=no" ;;
esac
```
because a) e.g. i686 packages can be cross compiled and b) it would break if compiled natively on e.g. an arm device.

abenson:
Yeah, you're right.

PR #12064, opened 2018-03-03
 proj: update to 5.0.0.

Johnnynator:


Johnnynator:
I missed vtk and grass because they were merged after I created this PR.

PR #12067, opened 2018-03-03
nim: update to 0.18.0

SolitudeSF:


SolitudeSF:
temporarily added libssl43 to depends, since nim doesnt links against libbsl44

PR #12068, opened 2018-03-03
New package: clyrics-0.10

cr6git:
:microphone:


cr6git:
thanks @pullmoll 
question: is it mandatory to define `local f ` (as you have just done [here](https://github.com/voidlinux/void-packages/commit/6a3a9f3457884b10ba566f1258bfddc14d0d0608#diff-ac3c966d79fbb1d15bb77e110cd272e2R15)) before using this variable?

pullmoll:
No, it's not mandatory. The functions are executed as subshells AFAIR. This is more of a habit of mine to declare local variables.

PR #12084, opened 2018-03-04
mozc: fix aarch64-musl build

mvf:
This `#define`s `_GNU_SOURCE` in an attempt to fix the `aarch64-musl` build. I can't reproduce this problem locally.


pullmoll:
This PR doesn't fix the build for `aarch64-musl` here.
Linking some host tool `gen_emoticon_rewriter_data_main` seems to lack `pthread`, i.e. `-pthread` in the host `LDFLAGS`. I have no idea how to achieve that.

mvf:
Thanks for taking a look. It seems this broke even more in the meantime. When I opened this PR, local and Travis builds were working. What this fixes is that without `_GNU_SOURCE`, musl's `pthread.h` does not expose `pthread_setname_np` and compilation fails. `_GNU_SOURCE` seems to be defined by default on all build servers except `aarch64-musl`. I can't find the failing `mozc` build anymore, but here's a recent example of the same issue with `procps-ng`, just with a different header:

https://build.voidlinux.eu/builders/aarch64-musl_builder/builds/9232/steps/shell_3/logs/stdio

I can reproduce the new linking issue now, so I'll look into that.

PR #12100, opened 2018-03-04
id3lib: fix missing distfiles

cr6git:
tarballs have vanished from [sourceforge](https://sourceforge.net/projects/id3lib/files/).


cr6git:
tarballs have reappeared! :sweat_smile:

PR #12113, opened 2018-03-04
openvpn: update to 2.4.5

scadu:


scadu:
Seems it doesn't play that well with LibreSSL 2.6. For some reason OpenVPN 2.4.4 builds correctly.

PR #12125, opened 2018-03-05
New package: dmd-bootstrap-2.069

yshui:
Add a bootstrap compiler for D language.

dmd-2.069 is the last version of dmd that is written in C++, later dmd versions is written in D, thus requires a bootstrap compiler.


Vaelatern:
And whence do the patches come?

srcpkgs/dmd-bootstrap/files/musl.patch
```diff
@@ -0,0 +1,22 @@
...
++++ druntime/src/rt/backtrace/dwarf.d
+@@ -15,7 +15,7 @@ module rt.backtrace.dwarf;
+ version(CRuntime_Glibc) version = linux_or_freebsd;
+ else version(FreeBSD) version = linux_or_freebsd;
+ 
+-version(linux_or_freebsd):
++version(none):
+ 
+ import rt.util.container.array;
+ import rt.backtrace.elf;
```
Vaelatern: please work with upstream to patch this in

srcpkgs/dmd-bootstrap/template
```diff
@@ -0,0 +1,75 @@
...
+}
+
+pre_configure() {
+	case "$XBPS_TARGET_MACHINE" in
+		*-musl) patch -p0 < ${FILESDIR}/musl.patch
+	esac
+}
+
+do_build() {
+	cd $wrksrc/dmd/src
```
Vaelatern: `build_wrksrc`

srcpkgs/dmd-bootstrap/template
```diff
@@ -0,0 +1,75 @@
...
+	else
+		mkdir -p $wrksrc
+		cd $wrksrc
+		(mkdir dmd; cd dmd; git init; git remote add origin git://github.com/dlang/dmd)
+		(mkdir druntime; cd druntime; git init; git remote add origin git://github.com/dlang/druntime)
+		(mkdir phobos; cd phobos; git init; git remote add origin git://github.com/dlang/phobos)
+	fi
+	(cd dmd; git fetch --depth=1 --progress origin dmd-cxx; git reset --hard FETCH_HEAD)
+	(cd druntime; git fetch --depth=1 --progress origin dmd-cxx; git reset --hard FETCH_HEAD)
+	(cd phobos; git fetch --depth=1 --progress origin dmd-cxx; git reset --hard FETCH_HEAD)
```
Vaelatern: NO.

srcpkgs/dmd-bootstrap/template
```diff
@@ -0,0 +1,75 @@
...
+version=2.069
+revision=1
+build_style=gnu-makefile
+provides="d-compiler-${version}_${revision}"
+hostmakedepends="git"
+short_desc="DMD Compiler (last C++ version for bootstrap)"
+maintainer="Yuxuan Shui <yshuiv7@gmail.com>"
+license="custom"
+homepage="dlang.org"
+only_for_archs="i686 x86_64 i686-musl x86_64-musl"
```
Vaelatern: what? I mean....what?

srcpkgs/dmd-bootstrap/template
```diff
@@ -0,0 +1,75 @@
...
+
+	cd ../../druntime
+	make -f posix.mak MODEL=$bits DMD=../dmd/src/dmd RELEASE=1
+
+	cd ../phobos
+	make -f posix.mak MODEL=$bits DMD=../dmd/src/dmd RELEASE=1
+}
+
+do_install() {
+	cd $wrksrc
```
Vaelatern: No

srcpkgs/dmd-bootstrap/template
```diff
@@ -0,0 +1,75 @@
...
+	vbin dmd/src/dmd
+
+	vmkdir etc
+	echo -e "[Environment]\nDFLAGS=-I/usr/include/dlang/dmd -L-L/usr/lib -L-L/usr/lib32 -L--export-dynamic" > $DESTDIR/etc/dmd.conf
```
Vaelatern: printf not echo.

srcpkgs/dmd-bootstrap/template
```diff
@@ -0,0 +1,75 @@
...
+
+	vmkdir usr/share/man/man1
+	vmkdir usr/share/man/man5
+	vcopy dmd/docs/man/man1/* usr/share/man/man1/
+	vcopy dmd/docs/man/man5/* usr/share/man/man5/
+
+	vlicense dmd/src/backendlicense.txt
+	vlicense dmd/src/boostlicense.txt
+
+	vinstall phobos/generated/linux/release/$bits/libphobos2.a 0644 usr/lib
```
Vaelatern: hey lets put a random dollarsign in here... what?

srcpkgs/dmd-bootstrap/template
```diff
@@ -0,0 +1,75 @@
...
+hostmakedepends="git"
+short_desc="DMD Compiler (last C++ version for bootstrap)"
+maintainer="Yuxuan Shui <yshuiv7@gmail.com>"
+license="custom"
+homepage="dlang.org"
+only_for_archs="i686 x86_64 i686-musl x86_64-musl"
+
+case "$XBPS_TARGET_MACHINE" in
+	x86_64*) bits="64";;
+	i686*) bits="32";;
```
Vaelatern: no arm at all? really?

srcpkgs/dmd-bootstrap/files/musl.patch
yshui: This is a workaround for this version of dmd/druntime. Proper musl support is included in later version of druntime.

srcpkgs/dmd-bootstrap/template
yshui: No.

srcpkgs/dmd-bootstrap/template
yshui: Why?

srcpkgs/dmd-bootstrap/files/musl.patch
Vaelatern: Why not use that later version?

srcpkgs/dmd-bootstrap/template
Vaelatern: `echo -e` is not POSIX compatible, and I don't want random breakage. Also you will already have to change things and this is a few characters.

srcpkgs/dmd-bootstrap/template
Vaelatern: Why not?

srcpkgs/dmd-bootstrap/template
Vaelatern: Select on arch, mark broken if arm or aarc64... but `broken="Impressively good reason why a language can't be on a platform"`

srcpkgs/dmd-bootstrap/template
pullmoll:
```
distfiles="https://github.com/dlang/dmd/archive/v${version}.tar.gz
 https://github.com/dlang/druntime/archive/v${version}.tar.gz
 https://github.com/dlang/phobos/archive/v${version}.tar.gz"
checksum="<insert the ones for above here>"
```
In `post_fetch()` rename directories as they are expected to be.

srcpkgs/dmd-bootstrap/template
pullmoll: Or alternatively use
```
cat > $DESTDIR/etc/dmd.conf <<EOF
[Environment]
LDFLAGS=-I/usr/include/dlang/dmd -L-L/usr/lib -L-L/usr/lib32 -L--export-dynamic
EOF
```

srcpkgs/dmd-bootstrap/files/musl.patch
yshui: Because this version of dmd can't compile a later version of druntime

pullmoll:
[This template](https://gist.github.com/pullmoll/d58ab594b993749fc950c8a3d2fde9c6) almost works, but it seems I have the wrong dmdruntime or phobos? dmdruntime-2.069 does not exist, but only dmdruntime-2.069.0, or does it have to be an earlier version?

srcpkgs/dmd-bootstrap/template
yshui: The release archives are way older than the git 'dmd-cxx' branch. This branch is meant to be used for bootstrapping newest dmd.

srcpkgs/dmd-bootstrap/template
yshui: There's no reason why a language **cannot** be on a platform. But obviously support for those platforms are not implemented in this version.

yshui:
@pullmoll set TARGET_CPU=stub generates non-working compilers (no code gen support at all). I don't see why you would want to do that.

srcpkgs/dmd-bootstrap/template
yshui: No what?

yshui: No what?

srcpkgs/dmd-bootstrap/template
yshui: OK

pullmoll:
@yshui I haven't had time to look too deep into the documentation wrote what I could imagine from peeking at the posix.mk file. The template was meant as an example how to not use `git clone …` but download tarballs. I'll leave the details to you.

yshui:
@pullmoll The tarball you can find is too old. There's no tarball available for the `dmd-cxx` branch (which is being updated for bootstrapping new version of dmd). ARM support only comes in later version of dmd.

srcpkgs/dmd-bootstrap/template
pullmoll: You can download any git revision by their hash. See the first line in my example template.
I guess you need two more git hashes for druntime and phobos to download the latest revisions from the cxx branch then.

pullmoll:
@yshui You can download anything you can clone as a tarball from github, believe me.
We do this in several packages.

srcpkgs/dmd-bootstrap/template
pullmoll: Take a look at my example template again. I **am** downloading the latest dmd-cxx branch tarball there.

srcpkgs/dmd-bootstrap/template
pullmoll: So to rephrase: it is not possible to build a bootstrap dmd for e.g. arm, but once a bootstrap is available, a later version can be build which can then cross build for arm?

yshui:
@pullmoll OK, then I think it's best to pick the latest commit of branch `dmd-cxx`, and put commit has in $version?

srcpkgs/dmd-bootstrap/template
Vaelatern: Remove the `cd` command

srcpkgs/dmd-bootstrap/template
yshui: I never built a cross compiling dmd before, but I would assume that is the case.

pullmoll:
You can either keep the version number and only revbump on updates, or perhaps append a datestamp to the version, e.g. 2.069.20180220 or similar. Then if dmd-bootstrap needs an update to build newer dmd versions it should be easy enough to update it.

pullmoll:
Updated [the template](https://gist.github.com/pullmoll/d58ab594b993749fc950c8a3d2fde9c6) and it now builds for the 4 native targets.

pullmoll:
Merged the template. I guess next is a PR for `dmd`?

yshui:
@pullmoll Thanks a lot for fixing my template! Yes, the next will be `dmd`, but it might take a while.

Do you think it's better to package dmd and its standard library together, or separately?

yshui:
@pullmoll I want to correct one thing I got wrong. Actually dmd never had ARM support. However the other two D compilers, ldc and gdc, does support ARM.

PR #12148, opened 2018-03-05
build-style/cmake.sh: fix default do_check

Chocimier:


Vaelatern:
example of it being broken?

Chocimier:
evolution, any kde package

maxice8:
are packages that have working tests (like cmake itself) keep working ?

Chocimier:
Tests for cmake still run. Are there other packages that should be tested for regression?

ebfe:
Looks right to me

PR #12172, opened 2018-03-05
New Package: alure

Magister-Bergerac:


srcpkgs/alure/template
```diff
@@ -0,0 +1,12 @@
...
+pkgname=alure
+version=1.2
+revision=1
+build_style=cmake
+makedepends="libopenal-devel"
+short_desc="Utility library to help manage common tasks with OpenAL applications. "
+maintainer="Simon Whelan <si.whelan@gmail.com>"
+license="GPL-3"
+homepage="http://kcat.strangesoft.net/alure.html"
+distfiles="http://kcat.strangesoft.net/alure-releases/alure-1.2.tar.bz2"
```
maxice8: use ${version} instead of hardcoding

cr6git:
`license="GPL-3"`

Not sure about that...
The Arch [package](https://www.archlinux.org/packages/community/x86_64/alure/) is "MIT", while the Debian [copyright](http://metadata.ftp-master.debian.org/changelogs/main/a/alure/alure_1.2-6_copyright) says "Expat".

srcpkgs/alure/template
```diff
@@ -0,0 +1,12 @@
+# Template file for 'alure'
+pkgname=alure
+version=1.2
+revision=1
+build_style=cmake
+makedepends="libopenal-devel"
+short_desc="Utility library to help manage common tasks with OpenAL applications. "
+maintainer="Simon Whelan <si.whelan@gmail.com>"
+license="GPL-3"
```
maxice8: change to MIT and add

```
post_install() {
    vlicense COPYING
}
```

at the end

srcpkgs/alure/template
```diff
@@ -0,0 +1,12 @@
+# Template file for 'alure'
+pkgname=alure
+version=1.2
+revision=1
+build_style=cmake
+makedepends="libopenal-devel"
+short_desc="Utility library to help manage common tasks with OpenAL applications. "
```
jnbr: remove trailing white space and dot

srcpkgs/alure/template
```diff
@@ -0,0 +1,12 @@
...
+maintainer="Simon Whelan <si.whelan@gmail.com>"
+license="X11/MIT"
```
maxice8: we have no such concept of X11/MIT, we use MIT for all licenses.

Vaelatern:
single commit.

PR #12186, opened 2018-03-05
fcitx (and depending pkgs): fix cross compilation

newbluemoon:
The problem was that during the build some target built binaries were executed to generate data. However, these binaries are also packaged. So to make cross compilation work I just added <i>fcitx</i> to `hostmakedepends` and adjusted the cmake scripts to use the host binaries.

<s>To make cmake find `fcitx4-config` and not have to add it’s location to every template using it I added it to the toolchain file generated in `build-style/cmake.sh`.</s>

Also, I removed the ‘execinfo’ patch for musl and made use of libexecinfo instead. There was a little bug in libexecinfo-devel, too: `libexecinfo.so` was a symlink to `/usr/lib/libexecinfo.so.1` which fails when cross compiling; made it a relative link.

<i>fcitx-qt5</i> is a duplicate of <i>libfcitx-qt5</i>; so I removed the former (because it’s the one which was added last) and adjusted `common/shlibs`. I couldn’t find a package that depends on it.

@mvf I also enabled fcitx for cross compilation in mozc. Please check if you’re OK with the changes (I just dissolved the cross branches).

The other packages just needed some conditional `hostmakedepends`.


mvf:
Wow, thanks for tackling this beast, @newbluemoon! The `mozc` changes are fine but maybe the commit message could mention that it is `fcitx-mozc` being fixed.

newbluemoon:
@mvf You’re right, thanks. I adjusted the commit message. It’s more precise now. :)

newbluemoon:
Travis will most likely fail, because fcitx also being a `hostmakedepends` when cross compiling.

Vaelatern:
Resolve conflicts?

newbluemoon:
@Vaelatern Sorry, should have been more precise: fcitx has to be built for the host and the target and in the past Travis didn’t like situations like this.

newbluemoon:
Sorry again, misunderstood; now I see what you mean!

newbluemoon:
@Vaelatern Done. :)

common/build-style/cmake.sh
```diff
@@ -22,6 +22,7 @@ SET(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
 SET(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
 
 SET(wxWidgets_CONFIG_EXECUTABLE ${XBPS_WRAPPERDIR}/wx-config)
+SET(FCITX4_FCITXCONFIG_EXECUTABLE ${XBPS_CROSS_BASE}/usr/bin/fcitx4-config)
```
Vaelatern: Why should this be done here?

Vaelatern: @voidlinux/pkg-committers ?

newbluemoon: Just to add: packages using `fcitx4-config` are
* fcitx-configtool
* fcitx-rime
* libfcitx-qt5

I don’t know if three is a large number enough to put it in `cmake.sh`, but in my opinion it’s cleaner that way.

And to elaborate some more:
When cmake tries to locate fcitx the found cmake script calls `fcitx4-config --prefix` to include `FcitxMacro.cmake` (which resides in the same directory!). That’s what the sed in line 32 in fcitx’s template is for: to look for the file in the same directory. Else `fcitx4-config --prefix` has to be prefixed with $XBPS_CROSS_BASE. But this would cause e.g. fcitx-rime to install some files under $XBPS_CROSS_BASE which would then have to be removed again in the template ...

So I tried to find a way with the smallest amount of changes necessary.

pullmoll: The arguments seem reasonable to me while I also don't know if a package count of 3 for which this is useful is sufficient to justify a cmake build style change.

newbluemoon: > I guess it's not possible to modify the FcitxMacro.cmake file itself to become cross building aware

@pullmoll Sorry, just saw your edit: I did some digging and think it is possible by just changing one line in FcitxConfig.cmake and modifying the cmake build style won’t be necessary :) Need to test that everything else still works. Will report back later today.
Thanks for the nudge in the right direction :)

newbluemoon:
I think I found a better way to find `fcitx4-config` when cross compiling without modifying the cmake build style:

The problem is that the cmake cross toolchain file is setup to not look for programs under `$XBPS_CROSS_BASE`. This is the right thing to do in general, but in this case `fcitx4-config` (which is a shell script) cannot be found. So the simple solution is to add `$CMAKE_FIND_ROOT_PATH` to the list of paths with which `FcitxConfig.cmake` calls `FIND_PROGRAM` for `fcitx4-config`.
There already is a list of paths defined. One of them is `/opt/fcitx` which is used on MacOSX and to make the sed command less complex I just used this line and replaced it with `$CMAKE_FIND_ROOT_PATH`. Now everything works just fine for native and cross builds.

newbluemoon:
Did some further testing and found that `fcitx4-config` <i>is</i> found <i>without</i> the above adding `$CMAKE_FIND_ROOT_PATH` to `FIND_PROGRAM`; I guess the option `ONLY_CMAKE_FIND_ROOT_PATH` overrides `SET(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)` from the cmake cross toolchain file. So I removed the above change. I just have no idea why it didn’t work in the first place!

@pullmoll If you can spare a little time, could you please test if this works for you? (In case I messed something up.) :)

newbluemoon:
To clean this up: Apparently, I confused some things while trying out different approaches. `fcitx4-config` is found without any modification necessary. What really does the trick is just to fix how `FcitxMacro.cmake` is included in `FcitxConfig.cmake` (i.e. without invoking `fcitx4-config`). I triple checked and now everything should be ok. Sorry for the inconveniencies.

pullmoll:
Hmm.. libexecinfo now fails to build for cross `*-musl` targets for some reason and so trying to build fcitx for e.g. `aarch64-musl` fails here. I have idea what's going on and why `xbps-src` thinks it is not building for a `*-musl` arch when I use `-a aarch64-musl`.

newbluemoon:
Just tried for aarch64-musl and armv7l-musl and everything is just fine here.

pullmoll:
I'm using a glibc x86_64 environment to cross build for aarch64-musl, perhaps that makes a difference. But whatever it is, I think in case of libexecinfo we can resort to specifying the `only_for_archs` list and I would push the PR with that change if all looks ok here.

newbluemoon:
> I'm using a glibc x86_64 environment to cross build for aarch64-musl

Same here. Except that it works for me as is. Strange...

pullmoll:
Maybe it's related to #12324 but I would not know how. I don't feel like investigating this further, though :)

Issue #12152, opened 2018-03-05
poppler-qt4 no longer available with poppler-0.62 / removing kde4


ebfe:
Affected packages:

calligra
katarakt
kfilemetadata
qtikz

removing kfilemetadata (and all it's revdeps) is basically removing kde4.

Anyone still using kde4?

biopsin:
Thank you for the headsup, I use katarakt daily, will put qt on hold.

Johnnynator:
Katarakt can be updated to qt5 (at least Debian has it build with qt5).

I will check it this afternoon an make a PR if it works fine.

ebfe:
There are updates for calligra as well. kfilemetadata without poppler-qt4 seems to work too (no idea if it's still useful then), but this still might be a good opportunity to remove kde4 if it's no longer used.

biopsin:
Just tested and katarakt compiles just fine with qt5..

Johnnynator:
I'm atm also working on getting an updated Calligra template ready, so that we have only kde4 left as only dependency.

Johnnynator:
@ebfe is it ok if I prepare a PR that updates Calligra AND removes kde4.
atm I'm standing between the choice of adding a couple of ${name}5 package or updating the current ones that are for kde4 to kde5. e.g. marble, akonadi.

I could also add Calligra without these dependecies but imo they should be added.

ebfe:
> @ebfe is it ok if I prepare a PR that updates Calligra AND removes kde4.

That's part of the point of this issue. Figure out if we want to remove (parts
of) kde4. I don't use it and have no idea in what state our kde5 packages are.

cc @voidlinux/pkg-committers

pullmoll:
IMO we should strive for getting rid of kde4.

asergi:
+1 to remove KDE4.

PR #12212, opened 2018-03-06
qtikz: change qt version to qt5

Johnnynator:
@Piraty #12152


Piraty:
looks good, except broken theming...

Piraty:
you could bump to 0.12 while at it

ebfe:
@Piraty good to merge?

Piraty:
@ebfe sure.
Theming was my own issue only (need to use `QT_STYLE_OVERRIDE=gtk2` or sth similar for correct theming to apply)

PR #12213, opened 2018-03-06
xcursor-vanilla-dmz: add INSTALL script

cr6git:
This problem has been reported on the [forum](https://forum.voidlinux.eu/t/solved-updating-system-failed-update-due-to-directory-not-empty/4850).


maxice8:
please `echo pkgname=dmz-cursor-theme > srcpkgs/xcursor-vanilla-dmz/update` for the void-updates bot.

srcpkgs/xcursor-vanilla-dmz/INSTALL
```diff
@@ -0,0 +1,7 @@
+case "$ACTION" in
+pre)
+	# Actions to execute before the package files are unpacked.
+	rm -rf usr/share/icons/Vanilla-DMZ-AA/*
```
ebfe: These files belong to the xcursor-vanilla-dmz-aa package.

cr6git: of course! stupid me.

PR #12227, opened 2018-03-07
New package: kitty-0.8.3

laserbat:


srcpkgs/kitty/template
```diff
@@ -0,0 +1,27 @@
+# Template file for 'kitty'
+pkgname=kitty
+version=0.8.0
+revision=1
+makedepends="python pkg-config zlib-devel libpng-devel python3-devel harfbuzz-devel libXrandr-devel libXcursor-devel libXi-devel libXinerama-devel MesaLib-devel libxkbcommon-devel ncurses"
```
maxice8: i assume python is running something so it should be on hostmakedepends, pkg-config likewise, also no tsure why ncurses it there

srcpkgs/kitty/template
```diff
@@ -0,0 +1,27 @@
...
+depends="python harfbuzz freetype fontconfig"
```
maxice8: adding harfbuzz freetype and fontconfig shouldn't be necessary and it should be detected automatically

srcpkgs/kitty/template
```diff
@@ -0,0 +1,27 @@
...
+short_desc="A cross-platform, fast, feature full, GPU based terminal emulator"
+maintainer="Olga Ustuzhanina <me@laserbat.pw>"
+license="GPL-3"
+homepage="https://github.com/kovidgoyal/kitty"
+distfiles="https://github.com/kovidgoyal/kitty/archive/v${version}.tar.gz"
+checksum=de8e6ce097747246bc5a9b946cd9dc94702dc1ecd64a7601aefc6061406b06ea
+
+do_build() {
+	cd ${wrksrc}
+	python3 ./setup.py linux-package
```
maxice8: no need for cd $wrksrc and to use ./ , since you're running python3 on host it should be added to hostmakedepends

srcpkgs/kitty/template
laserbat: It uses ncurses to generate terminfo data. I'll move that, pkg-config and python to hostmakedepends

srcpkgs/kitty/template
laserbat: Noted.

srcpkgs/kitty/template
maxice8: speaking of terminfo it would be nice if a `kitty-terminfo` package could be created, see `rxvt-unicode` template for an example of how to do it.

laserbat: I'll look into this, thank you

laserbat:
Not sure why travis builds are broken, it compiles with no errors on my machine.

laserbat:
Build log from my machine
https://gist.github.com/laserbat/6e965871613f3ae379b49f1b53f485a1

mvf:
Most of the builds fail because the build process uses POSIX semaphores and there is no, or no writable `/dev/shm`. I had the same issue when packaging `mozc`. In my case the workaround was to disable parallel Makefile generation.

PR #12237, opened 2018-03-07
dmd: Update to 2.079.0

yshui:
Notes:

* Documents and sample code are now packaged in dmd-docs.
* Previous template installs some prebuilt binary, which won't work with musl. They are now removed. A `dtools` package will be added later to replace them.
* `6193.patch` is needed for `phobos` to compile against musl. It is upstreamed and should be included in next release.


yshui:
Since there's no way to put the package itself into the makedepends, guess we will have to settle with using dmd-bootstrap forever.

srcpkgs/dmd/template
```diff
@@ -1,88 +1,104 @@
...
 
-	vman man/man1/dmd.1
-	vman man/man1/rdmd.1
-	vman man/man5/dmd.conf.5
+	vman generated/docs/man/man1/dmd.1
+	vman generated/docs/man/man5/dmd.conf.5
+}
+
+dmd-docs_package() {
+	short_desc="Documentation and sample code for D programming language"
```
pullmoll: Please use the policy used in (almost?) all other packages:
```
short_desc+=" - documentation"
```

srcpkgs/dmd/template
```diff
@@ -1,88 +1,104 @@
...
+	license="Boost License"
```
pullmoll: Why is this different from the main package?

srcpkgs/dmd/template
```diff
@@ -1,88 +1,104 @@
...
+
+	pkg_install() {
+		cd dmd2
 
-	vmkdir usr/share/d/samples
-	cp -r samples/d/* ${DESTDIR}/usr/share/d/samples/
+		vmkdir usr/share/doc/d
```
pullmoll: I dunno if it is a good idea to put documentation for the package `dmd` under `/usr/share/doc/d`.
Is the documentation referring only to the language `d` and not the `dmd` compiler?

srcpkgs/dmd/template
```diff
@@ -1,88 +1,104 @@
...
-	for f in ddemangle dumpobj obj2asm rdmd; do
-		vbin linux/bin${_archbits}/${f}
-	done
+		vlicense license.txt
+	}
 }
 
 libphobos_package() {
 	short_desc="The phobos D standard library for DMD"
 	license="Boost License"
```
pullmoll: The same comment as for the `-docs` subpackage applies here.

yshui: Should be `short_desc+=" - standard library"`?

srcpkgs/dmd/template
yshui: @pullmoll this is documentation not just for dmd, but for the D language.

PR #12245, opened 2018-03-07
xkeyboard-config: fix pl layout

Chocimier:
Fix for upstream bug https://bugs.freedesktop.org/show_bug.cgi?id=105384


Chocimier:
Thanks!

PR #12249, opened 2018-03-07
spacefm: update to 1.0.6.

maxice8:


SolitudeSF:
huh, wonder if this means that spacefm-ng isn't needed anymore

srcpkgs/spacefm/template
```diff
@@ -11,7 +11,7 @@ configure_args="$(vopt_with gtk3)"
 depends="hicolor-icon-theme desktop-file-utils"
 short_desc="Multi-panel tabbed file manager"
 maintainer="Juan RP <xtraeme@voidlinux.eu>"
-license="GPL-3"
+license="GPL-3.0-or-later"
 homepage="http://ignorantguru.github.com/spacefm/"
```
cr6git: `homepage="http://ignorantguru.github.io/spacefm/"`

cr6git:
@SolitudeSF good question.

PR #12271, opened 2018-03-08
lightdm-gtk3-greeter: update to 2.0.4.

maxice8:


cr6git:
it closes [#11859](https://github.com/voidlinux/void-packages/pull/11859) then.

PR #12275, opened 2018-03-08
x264: update to 20180307

maxice8:


asergi:
No need to update x264 every day. Thanks.

PR #12284, opened 2018-03-08
grafana: update to 5.0.1

pulux:


the-maldridge:
Since we have a node-lts package now and a yarn that can use it, can you try building this with just the original source tarball?

pulux:
I'm not familar with node-js, you are the maintainer here.

Gottox:
Get this merged, move to a source build later.

PR #12289, opened 2018-03-08
crystal: semantic fixes and a new dependency

LVMBDV:
- Added `pkg-build` to depends because Crystal uses it to get fine-grained information on installed libraries.
- Fixed the build option binary_bootstrap checking the target architecture instead of host architecture, this would be a problem in the case of crossbuilding.

Speaking of crossbuilding, I tried to get it to work but I ran into a strange problem beyond `llvm-config --cxxflags`. I will try again when the project is a bit more mature.


maxice8:
there is a 0.24.2 version available to update, might want to take the opportunity to also bump it

LVMBDV:
I am aware of it but I want to wait a little bit more because it seems a bit [rough around the edges](https://github.com/crystal-lang/crystal/pull/5642) at the moment.

pullmoll:
Please squash to one commit.

pullmoll:
This wasn't based on upstream/master so I changed it to bump revision from 5 to 6.

LVMBDV:
Actually, let me start over later, with the 0.24.2 update, in a different PR. I'm too tired to fix this at the moment.

LVMBDV:
Oh wait you already fixed it, thanks :)

PR #12291, opened 2018-03-08
New package: shake.

CasperVector:
short_desc="Defragmenter that runs in userspace while the system is used"


CasperVector:
It seems that this package uses `error()` from `error.h`, and is therefore unsupported on musl.  However, I do not quite understand why help2man complains about `can't get '--help' info` on aarch64 and armv7hf.

Vaelatern:
Cross building means it can't run cross-built binaries.
It's a flaw unconsidered by those who rely on help2man

CasperVector:
Now I see...

CasperVector:
Now I see...

Vaelatern:
Straightforward fix is to rely on the pkg as a host pkg on cross build and have help2man do its magic on them.

newbluemoon:
Just had the same thought as @Vaelatern :) but you also have to make cmake use the host binary, e.g.:

```
	hostmakedepends+=" shake"

	post_extract() {
		sed -i -e 's;get_target_property (HELP2MAN_COMMAND ${command} LOCATION);set(HELP2MAN_COMMAND ${command});' \
		 CMakeLists.txt
	}
fi
```

CasperVector:
Thanks for the advice; template updated accordingly, cross-built successfully on my machine.

CasperVector:
Added `$broken` specifically for `-musl` architectures.

Vaelatern:
See how I amended the commit message before merging.

CasperVector:
Oops, I see.  All remaining templates updated now.

PR #12295, opened 2018-03-08
abook: update to 0.6.1.

CasperVector:
Optional support for .vcf files was introduced in this version, but it [requires](https://gitweb.gentoo.org/repo/gentoo.git/tree/app-misc/abook/abook-0.6.1-r1.ebuild) [libvformat](https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-libs/libvformat/libvformat-1.13-r2.ebuild), and is therefore not enabled here.


maxice8:
abook 0.6.1 is development version.

CasperVector:
Indeed, but the version was released in 2015, and has been marked "stable" in 2017 on gentoo for all architetures it supports.

maxice8:
On Thu, Mar 08, 2018 at 02:33:16PM +0000, Casper Ti. Vector wrote:
> Indeed, but the version was released in 2015, and has been marked "stable" in
> 2017 on gentoo for all architetures it supports.
> 
> —
> 

Gentoo is not upstream in this case, but this should pass by a void member.

CasperVector:
Yes, gentoo was not intended as an authoritative evidence, but as a (weakly?) supporting argument :)

Vaelatern:
Considering the stalled upstream, if an abook user is okay with this upgrade I can accept it.

CasperVector:
I can personally testify that this version always worked as expected on gentoo, and still works for me on void.  I will look into the CI failures on travis.

CasperVector:
`/usr/share/gettext/archive.dir.tar.xz: No such file or directory`?
Not sure what happened to gettext-devel on travis for the affected architectures...

Vaelatern:
Those are the cross builds.

Vaelatern:
Thus /usr isn't in the normal spot?

Vaelatern:
Or perhaps gettext-devel needs to be installed for the hostarch too?

CasperVector:
I finally understood what you mean by reading `common/xbps-src/shutils/common.sh`...
gettext-devel moved from `$makedepends` to `$hostmakedepends`; cross build succeeded on my machine.

Vaelatern:
so maybe `gettext` needs to be in `hostmakedepends` and not `gettext-devel`

CasperVector:
`/usr/share/gettext/archive.dir.tar.xz` is provided by gettext-devel, not gettext...
(Just tried, the build would fail with gettext-devel replaced by gettext.)

srcpkgs/abook/template
```diff
@@ -1,12 +1,21 @@
...
-checksum=0646f6311a94ad3341812a4de12a5a940a7a44d5cb6e9da5b0930aae9f44756e
+distfiles="http://abook.sourceforge.net/devel/${pkgname}-${version}.tar.gz"
+checksum=f0a90df8694fb34685ecdd45d97db28b88046c15c95e7b0700596028bd8bc0f9
+
+post_extract() {
+	autoreconf --install
+}
+
+post_install() {
+	vdoc BUGS ChangeLog FAQ README TODO sample.abookrc
```
Vaelatern: This is invalid.
```
for f in BUGS ChangeLog FAQ README TODO sample.abookrc; do
vdoc $f
done
```

CasperVector: Fixed, plus the same for #12293.

PR #12296, opened 2018-03-08
New package: virtualbox-addons (nonfree).

CasperVector:
short_desc="VirtualBox addition image and extension pack"
I did not find a way to let this package depend on `virtualbox-ose-${version}` regardless of the revision (`depends="virtualbox-ose-${version}"` does not work), so the `$revision` of this package has to be synchronised with that of virtualbox-ose for now.


CasperVector:
Sorry for neglecting to use `xlint`.  Template updated just now, which should hopefully pass the tests.

pullmoll:
Whenever a new version of virtualbox-ose is available, it will tell you about new guest drivers and I believe it's easy enough to download the extensions and ISO from within virtualbox to install the drivers in the guest.

This PR seems to create an additional maintenance burden (synchronizing revision) for what benefit? You still have to locate the files on your HD and manually install the drivers in the guest.

If any of @voidlinux/pkg-committers thinks different, please go and merge this PR - after it was updated to 5.2.8.

the-maldridge:
I agree that there is not much value in this PR as present, the built in fetch and install features are very nice, and don't require version tracking to work.  This can be reopened if anyone wants to argue this.

PR #12297, opened 2018-03-08
New packages: sunpinyin, sunpinyin-data, fcitx-sunpinyin.

CasperVector:
sunpinyin: A statistical language model based Chinese input method library.
My attempt to let sunpinyin `$depend` on sunpinyin-data resulted in `xbps-src` complaining about non-existent dependeny (even though the latter is only a runtime dependency), so sunpinyin-data needs to be installed manually in order for sunpinyin frontends (eg. fcitx-sunpinyin) to work correctly.  I guess this can be solved if `xbps-src` treats runtime dependencies like gentoo's `$PDEPEND`.


srcpkgs/sunpinyin-data/template
```diff
@@ -0,0 +1,30 @@
...
+	esac
+}
+
+do_build() {
+	make $makejobs ENDIANNESS="$(cat end)"
+}
+
+do_install() {
+	make $makejobs ENDIANNESS="$(cat end)" DESTDIR=$DESTDIR install
+}
```
Johnnynator: why not use the `gnu-makefile` build_style and set the ENDIANNESS as variable?
```
make_build_args="ENDIANNESS=$_end"
make_install_args="ENDIANNESS=$_end"

case $XBPS_TARGET_MACHINE in
 	i686*|x86_64*|arm*|mipsel*) _end=le;;
 	*) _end=be;;
 esac

```

srcpkgs/sunpinyin/template
```diff
@@ -0,0 +1,40 @@
...
+homepage="https://github.com/sunpinyin/sunpinyin"
+distfiles="https://github.com/sunpinyin/sunpinyin/archive/v$_ver.tar.gz"
+checksum=692455b645803bfc6c2e53151bb9ccbae1a4b813a18d87607d19d14ae9354780
+
+do_build() {
+	CXX="$CXX" CXXFLAGS="$CXXFLAGS" LDFLAGS="$LDFLAGS" scons ${makejobs} --prefix=/usr
+}
+
+do_install() {
+	scons --prefix=/usr --install-sandbox=$DESTDIR install
```
Johnnynator: scons build_style? and `--install-sandbox=$DESTDIR` into `make_install_args`

CasperVector: I think that for this particular template, `build_style=scons` greatly increases the complexity, and does not really quite improve the readability.

srcpkgs/sunpinyin-data/template
CasperVector: I did not find an official way to pass information between stages, and just followed the way in #12294.  Nevertheless, since the information in this particular case is independent of the package source and not too complex to generate, it is probably better to just generate it in the "preamble".  Many thanks for your suggestion, and I will update the template.

CasperVector: I updated the template, but without using `build_style=gnu-makefile`, for reasons not using `build_style=scons' in srcpkgs/sunpinyin/template.  BTW, I think your original snippet would not work properly, since it references `$_end' before assigning a value to it.

srcpkgs/sunpinyin-data/template
```diff
@@ -0,0 +1,30 @@
+# Template file for 'sunpinyin-data'
+pkgname=sunpinyin-data
+version=20140820
+revision=1
+makedepends="sunpinyin-utils"
```
Johnnynator: should be hostmakedepends

Johnnynator:
It seems like at least genpyt is segfaulting on musl, so sunpinyin-data can not be created.
You should either set sunpinyin as broken for musl or try to debug it.

srcpkgs/sunpinyin-data/template
CasperVector: Thanks; I see.

CasperVector:
@Johnnynator I see; I will attempt to reproduce the segfault on musl...

CasperVector:
The segfault was caused by genpyt(1) internally [attempting](https://github.com/sunpinyin/sunpinyin/blob/master/src/lexicon/pytrie_gen.cpp#L95) iconv(3) (GBK/GB2312 <-> UTF-8) in order to get the charset range of word strings; however, conversion to GBK/GB2312 is [not yet supported](https://wiki.musl-libc.org/functional-differences-from-glibc.html#iconv) on musl.
Luckily, sunpinyin does not attempt iconv(3) from/to GBK/GB2312 anywhere other than in genpyt(1), so one way to around this is to make separate big/little endian noarch packages and make meta packages that depend on them.
And the `*) echo _end=be;;` (should be `*) _end=be;;`) typo made aarch64 fail :(
Just renamed sunpinyin to sunpinyin-core, changed sunpinyin-data into two (mutually exclusive) noarch packages sunpinyin-data-{le,be}, and added a sunpinyin meta-package that drags in sunpinyin-core and sunpinyin-data-$_end; commit description updated accordingly.

pullmoll:
The `EDIT 2` assumption does not hold. The x86_64-musl repository is separate from the x86_64/i686 repository and thus noarch packages are built there, too.

PR #12300, opened 2018-03-08
busybox: fix `'/dev/nul'*` != `/dev/null` issue.

CasperVector:
cf. http://lists.busybox.net/pipermail/busybox/2018-March/086295.html


Vaelatern:
How does this fix the problem?

Vaelatern:
And what does `CONFIG_PIE` do?

CasperVector:
Since it seems that PIE is enabled for most packages in void, I adjusted the flag just by the way...
So you think I need to mention this in the commit message? (I would love to...)

Vaelatern:
Does it not otherwise respect our LDFLAGS?

CasperVector:
Just tried: yes, it does.  BTW, did I mention that the `LFLAGS="$LDFLAGS"` part in the old template means that LDFLAGS has always been passed through the environment?

Vaelatern:
it does not mean that.

CasperVector:
Did you realise `LFLAGS` is probably a typo?
```
$ grep -r '\<LFLAGS\>' busybox-1.28.1
$ 
```

Vaelatern:
Ah, did not realize.

chneukirchen:
I partially reverted this in 77d3415 but it still works, and I cannot reproduce the bug outside of xbps-src, no idea what's going one there.

CasperVector:
I can successfully reproduce it outside of xbps-src by:
```
tar xpf busybox-1.28.1.tar.bz2
cd busybox-1.28.1
make defconfig
make prepare
make CFLAGS=-O2
./busybox ash -c "echo '/dev/nul'*"
```

chneukirchen:
I can also reproduce it this way on Arch x86_64 with gcc 7.3 and Debian x86_64 with gcc 6.3.

Ok, I tracked it down: it works with `-funsigned-char`, but just passing CFLAGS= overrides this.

CasperVector:
Nice.  Just [noticed](http://lists.busybox.net/pipermail/busybox/2018-March/086297.html) upstream of this.

PR #12314, opened 2018-03-08
xbps-src: use HEAD commit time for SOURCE_DATE_EPOCH

ebfe:
In order to make builds more reproducible SOURCE_DATE_EPOCH was set to
the time of the last commit that touched the template. Since trying to
reproduce a build from a different revision is futile and looking up the
commit in question can take several seconds, stop wasting time an just use
HEAD.

### DIFF
--- a/common/environment/setup/git.sh
+++ b/common/environment/setup/git.sh
@@ -14,6 +14,6 @@ if [ -z "${SOURCE_DATE_EPOCH}" -a -n "$IN_CHROOT" ]; then
 	if $GIT_CMD -C ${XBPS_SRCPKGDIR}/${basepkg} status -u normal --porcelain template | grep "^?? " &> /dev/null; then
 		export SOURCE_DATE_EPOCH="$(stat -c %Y ${XBPS_SRCPKGDIR}/${basepkg}/template)"
 	else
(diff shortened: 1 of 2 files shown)


common/environment/setup/git.sh
```diff
@@ -14,6 +14,6 @@ if [ -z "${SOURCE_DATE_EPOCH}" -a -n "$IN_CHROOT" ]; then
 	if $GIT_CMD -C ${XBPS_SRCPKGDIR}/${basepkg} status -u normal --porcelain template | grep "^?? " &> /dev/null; then
 		export SOURCE_DATE_EPOCH="$(stat -c %Y ${XBPS_SRCPKGDIR}/${basepkg}/template)"
 	else
-		export SOURCE_DATE_EPOCH="$($GIT_CMD -C ${XBPS_SRCPKGDIR}/${basepkg} log --pretty='%ct' -n1 .)"
+		export SOURCE_DATE_EPOCH="$($GIT_CMD log --pretty='%ct' -n1 HEAD)"
```
Gottox: We don't rely on a specific object in the tree anymore. The call can be shortened to:
```bash
$GIT_CMD rev-parse HEAD
```

ebfe: We want a timestamp not a git revision.

Gottox: Okay, my mistake.

PR #12322, opened 2018-03-09
New package: bitfighter-0.19f

cr6git:
:rocket:


cr6git:
@pullmoll ah yes, adding a `bitfighter-data` pkg in [8aa4e9e](https://github.com/voidlinux/void-packages/commit/8aa4e9e370436575a5eed64aceaf4f7653312042) was clever. :+1:

pullmoll:
Many games don't have -data subpackages while they could.
This increases multiarch repositories and I'm close to the limit of 200GiB with https://repo.voidlinux.de/ :-)

PR #12325, opened 2018-03-09
New package: falkon-3.0.0

Hoshpak:


maxice8:
falkon is qupzilla renamed under tutelage of KDE, should qupzilla be turned into a transitional package ?

Hoshpak:
I'm not sure. For now, they can be installed alongside since they don't share the profile, the difference between the two is minimal, though. We already have two branches of qupzilla in the repository (1.8.x as qupzilla1 and 2.2.x as qupzilla). On point against turning qupzilla into a transitional package would be that there's no automatic profile migration, users will be informed by the INSTALL.msg that this has to be done manually but if one package replaces another via a transitional package, I would expect that to happen automatically. I think it also depends on how much qupzilla and Falkon are going to diverge in the future with the qupzilla branch being put in maintenance mode or discontinued completely.

Jakecheslekov:
In order to get your bookmarks back you have to copy the file from ~/.config/qupzilla/profiles/ 
Its the bookmarks.json
copy it to ~/.config/falkon/profiles/ 
And then you'll have all your bookmarks back. For me that was the most important thing.

kenny-w7:
Has anyone managed to get this to compile? I get errors about shiboken, pyside & pythonlibs, but apparently those are only optional dependencies. It usually stops at about 50% compiled.

Hoshpak:
If it didn't compile, I wouldn't have created this PR. In fact, I just compiled falkon again with the template from my branch and it worked just fine. The warnings about the python stuff are indeed optional dependencies for python extensions for falkon. Since the dependencies are not in the repository and at least one of them hasn't put out a release yet, I left them out for now. Compiling falkon wasn't a big issue for me, just add ` -Wl,--no-fatal-warnings` to your LDFLAGS for otherwise the linking fails because of a LibreSSL warning.

kenny-w7:
Oh, I'm sorry but, where do I add LDFLAGS?

Hoshpak:
If you are compiling outside of xbps-src, just define a variable in your shell called LDFLAGS before running the cmake step. It should pick it up and write the Makefile accordingly.

kenny-w7:
I'm sorry but I don't know how to do that! I don't know what defining a variable in my shell even means... 
Could you please just give me the command to enter?

Johnnynator:
export LDFLAGS=

Hoshpak:
Assuming you are operating outside of the xbps-src environment and LDFLAGS ist not defined yet, run:
```
LDFLAGS=" -Wl,--no-fatal-warnings"
```
Then run the cmake and make commands as you would usually do.

kenny-w7:
Am I doing something wrong? 

/usr/lib/libQt5Network.so.5.10.1: warning: warning: EVP_CipherFinal is often misused, please use EVP_CipherFinal_ex and EVP_CIPHER_CTX_cleanup
make[2]: *** [src/lib/CMakeFiles/FalkonPrivate.dir/build.make:5401: bin/libFalkonPrivate.so.3.0.99] Error 1
make[1]: *** [CMakeFiles/Makefile2:178: src/lib/CMakeFiles/FalkonPrivate.dir/all] Error 2
make: *** [Makefile:141: all] Error 2

Hoshpak:
Most likely you didn't define LDFLAGS before configuring with cmake or still had files from earlier around, that's exactly the error I meant. Perhaps it would be best if you deleted the directory, extracted the tarball again, defined LDFLAGS and went through the whole process from the beginning.

kenny-w7:
-- Could not set up the appstream test. appstreamcli is missing.

Could this have something to do with it? 

I deleted the build folder & then recreated it, then did the LDFLAGS & then ran cmake, now its running make & is at 25%, it would usually stop at 50%. I'm running Void x86_64-MUSL, so I guess that MIGHT be why, but idk.

Hoshpak:
No that's harmless and will not cause the build to fail. But now I remember that there's one fix for building with musl in this PR. You could download this file and apply the patch manually but perhaps at this point it would be easier if you just pulled from my branch and used xbps-rc to build a binary package you can install. The process would be:

```
git clone https://github.com/Hoshpak/void-packages.git
cd void-packages
git checkout falkon
./xbps-src binary-bootstrap
./xbps-src pkg falkon
cd hostdir/binpkgs/falkon
xdowngrade falkon-3.0.0_1.x86_64-musl.xbps
```

xdowngrade is part of the xtools package in case you haven't installed it yet.

kenny-w7:
Thank you very much Hoshpak. I am very greatful for your help. I will let you know how it goes, I've used xbps-src only a few times before, but from what I remember its not too hard :P

kenny-w7:
QT5 got an update last night / today, Qupzilla v2 is no longer flickering like mad. 
I'm compiling Falkon at the moment, does the Falkon package have to be updated to work with the new QT5 package? Its currently compiling libavcodec for some reason, I didn't remember it doing that last time... Could be wrong, maybe it did & I missed it... Not sure.

Johnnynator:
@kenny-w7 It does compile libavcodec because it got updated but Hoshpak's branch is still referencing the old one. You should rebase it against the current master branch because otherwise you will not be able to install the resulting package.
```
git remote add upstream https://github.com/voidlinux/void-packages.git
git pull upstream master --rebase
```

kenny-w7:
Johnnynator but is Falkon in there? I looked just a little while ago & didn't see it.

Hoshpak:
The command just adds a reference to the upstream repository, it does not check it out. The second command is to rebase the branch, applying my changes on to of the current upstream master. This way, your local branch will be up-to-date and xbps-src won't compile an outdated libavcodec version before compiling falkon.

kenny-w7:
First time I attempted it, it failed... I'm going to give it one more shot.
What's this "All checks have failed" ? Does that mean it can't be built right now?

PR #12340, opened 2018-03-09
libgit2: update to 0.26.2.

maxice8:


Vaelatern:
Finding their changelog.
I'm particularly cautious about libgit2 since I need it.

maxice8:
On Fri, Mar 09, 2018 at 07:56:45PM +0000, Toyam Cox wrote:
> Finding their changelog.
> I'm particularly cautious about libgit2 since I need it.
> 
> —
> 

v0.26.2
-------

This is a security release fixing memory handling issues when reading crafted
repository index files. The issues allow for possible denial of service due to
allocation of large memory and out-of-bound reads.

As the index is never transferred via the network, exploitation requires an
attacker to have access to the local repository.

PR #12349, opened 2018-03-09
Upower0: drop in favour of upower

Piraty:
All packages depending on `upower0`or `libupower-glib1` on compile time were migrated to `upower` and it's `libupower-glib3`.
All compiles ok (except `budgie-desktop`, which succesfully recognizes `upower-glib 0.99.7` but suffers from other issues so it won't build )

`xfce4`'s packages were tested by me, they do fine.

Using `replaces="upower0>=0"` in `upower` is obsolete i guess?

Inspired by @Gottox


Gottox:
Looks great! We should make sure the dependencies still work as expected. So far were sure for cinnamon, mate, and xfce?

Piraty:
"were" == "we're"?
`lightdm` works according to @abenson , who switched manually to `upower` because `upower0` dindn't work.

`mixxx`' configure picked it up, i guess it works (but i didn't see what it is supposed to do...)

Johnnynator:
Upower should have made a meta package before removing or the line that it provides upower removed and revbumbped.
Now we have a lot of systems that need to have manually upower, because upower0 claims to have upower.

Gottox:
@Johnnynator transitional package added, thanks!

PR #12351, opened 2018-03-09
octave: add missing dependencies

Piraty:
+ enable octave's editor via qscintilla-qt5 (it was missing!)
+ provide support for solving linear programs via glpk
+ provide support for audio file handling
+ provide support for openGL->ghostscript printing via gl2ps
+ provide option for large arrays (but needs 64bit BLAS)

+ add gui launcher script, see: https://savannah.gnu.org/bugs/?53034
(essentially, you can't start octave from dmenu etc. by simply running `octave`, but rather provide ` 
--force-gui` to it, which isn't that intuitive)

Also in the pipe is another big dependency `suitesparse`, which i will add later.


Piraty:
@diogoleal ok with you?

CoolOhm:
I've been meaning to ask this question to the dev community: why are we still building Octave on Qt 4.X? Does Octave simply not run well when built on Qt5?


This line in the template pulls both Qt4.X and Qt5.X stuff:
    $(vopt_if gui 'qt-devel qscintilla-qt5-devel') 

**_qt-devel_** is Qt 4.X.   Why are we mixing 4.X and 5.X? Is that acceptable practice? If so great, I would just be surprised if you could harmlessly mix major versions of a big framework like that. 

But if not, the line should just be:
    $(vopt_if gui 'qt5-devel qscintilla-qt5-devel')

CoolOhm:
Cool and thanks! Thank you  as well for being a reasonable and responsive for the quick reply. 

I hope this gets merged soon. :)

CoolOhm:
So I just built and complied Octave with the patches you (and I) have suggested, and there is a serious problem. Plotting intermittently fails (about 70% failure).

This is apparently a Qt4 v Qt5 thing. The solution is to compile with Qt4!

See this [post ](https://bbs.archlinux.org/viewtopic.php?pid=1723164#p1723164) and this [bugreport](http://savannah.gnu.org/bugs/index.php?51442).

SO, now I know why we compile with Qt4!

TLDR: there is a bug with Octave compiled with Qt5 that makes plotting useless, the solution is to go back to Qt4 for now.

Piraty:
But i can't reproduce what you mean by "plots breaking 70%". I run the version presented here in the template. I don't see failing plots or artefacts.

Can you clarify?
`4.4` will have qt>=5.1 support, as shown in their [(pre-)release announcement](https://hg.savannah.gnu.org/hgweb/octave/file/5f4550b5d31b/NEWS)

screenshot_octave_plot

Piraty:
Afterall, since the editor wasnt't included in the package before, i still consider this PR a big improvement.
4.4 is not far away they say, so that might correct things on this end.

CoolOhm:
Agreed.

Let's hope this PR gets merged soon.

Issue #6359, opened 2017-05-10
Package request: dkimproxy


thypon:
dkimproxy is the default opensmtp dkim helper.

http://dkimproxy.sourceforge.net/

jnbr:
dkimproxy is in the repo now. This can be closed

PR #12358, opened 2018-03-10
New package: dkimproxy-1.4.1.

jnbr:
#6359


srcpkgs/dkimproxy/template
```diff
@@ -0,0 +1,26 @@
+# Template file for 'dkimproxy'
+pkgname=dkimproxy
+version=1.4.1
+revision=1
+build_style=gnu-configure
+noarch=yes
```
the-maldridge: This is almost certainly wrong for something that uses gnu-configure build style.

jnbr: The configure script doesn't do anything architecture specific. It is mainly used for installing files into the correct location. The only changes it does to the scripts, is setting the include path for perl.
For example:
```
--- scripts/dkim_responder.pl   2010-01-22 15:53:24.000000000 +0100
+++ /usr/bin/dkim_responder.pl  2018-03-10 03:58:23.000000000 +0100
@@ -1,4 +1,4 @@
-#!/usr/bin/perl -I../lib
+#!/usr/bin/perl -I/usr/lib
 #
```
The same happens to `/usr/bin/dkimproxy.in` and `/usr/bin/dkimproxy.out`.
Do you still want the `noarch=yes` to be removed?

cr6git: @the-maldridge the Debian [package](https://packages.debian.org/sid/dkimproxy) and the Arch [package](https://www.archlinux.org/packages/community/any/dkimproxy/) are `noarch`.

the-maldridge: @jnbr if this is actually perl and not C then perhaps the perl-module style is more apt, if not put a comment above that that this is perl and not C.

jnbr: @the-maldridge yes, it is pure perl. I added a comment.

jnbr:
@the-maldridge is the comment about noarch clear enough or do you want it to be changed?

PR #12377, opened 2018-03-10
Add config option for RICOH to kernels

Follpvosten:
The RICOH card reader built into some Thinkpads doesn't work without this kernel config option. I have turned it on for the 4.x LTS kernels and the latest one (because i think most people will be using one of those).


chneukirchen:
Can this be built as a module?

Follpvosten:
I don't think so - it looks like it's some kind of subcomponent of "SDHCI support on PCI bus" in `make menuconfig`. Maybe it will only be loaded when that module is.

PR #12388, opened 2018-03-11
update flite to 2.1

Ophidiophobia:


Johnnynator:
Commit message should be `flite: update to 2.1`.
I would recommend using xbump for creating update commits, it is part of the xtools package.

PR #12401, opened 2018-03-11
New package: python-mysqlclient-1.3.12

alindobre:
This package is necessary so a Void Linux host with mariadb/mysql installed can be managed with Ansible using any of its own internal mysql_* module (check [1] for full list of modules, but `mysql_db` and `mysql_user` is what i would need).

Without this module, I wouldn't be able to use Ansible unless I would write my own recipes using the `mysql` command line client or some other workaround, which is a bit out of hand.

[1] http://docs.ansible.com/ansible/latest/list_of_database_modules.html

### DIFF
--- a/srcpkgs/python-mysqlclient/template
+++ b/srcpkgs/python-mysqlclient/template
@@ -1,0 +1,24 @@
+pkgname=python-mysqlclient
+version=1.3.12
+revision=1
+wrksrc="mysqlclient-${version}"
+build_style=python-module
+pycompile_module="mysqlclient"
+hostmakedepends="python-setuptools python3-setuptools"
+makedepends="libmysqlclient-devel python-devel python3-devel zlib-devel libressl-devel"
+depends="python"
+short_desc="Python2 interface to MySQL/MariaDB"
+maintainer="Alin Dobre <alin.dobre@outlook.com>"
+license="GPL-2"
+homepage="https://github.com/PyMySQL/mysqlclient-python"
+distfiles="${PYPI_SITE}/m/mysqlclient/mysqlclient-${version}.tar.gz"
+checksum=2d9ec33de39f4d9c64ad7322ede0521d85829ce36a76f9dd3d6ab76a9c8648e5
+
+python3-mysqlclient_package() {
+	pycompile_module="mysqlclient"
+	depends="python3"
+	short_desc="${short_desc/Python2/Python3}"
+	pkg_install() {
+		vmove usr/lib/python3*
+	}
+}
--- a/srcpkgs/python3-mysqlclient
+++ b/srcpkgs/python3-mysqlclient
@@ -1,0 +1,1 @@
+python-mysqlclient


srcpkgs/python-mysqlclient/template
```diff
@@ -0,0 +1,13 @@
+pkgname=python-mysqlclient
+version=1.3.12
+revision=1
+wrksrc="mysqlclient-${version}"
+build_style=python-module
+pycompile_module="mysqlclient"
+hostmakedepends="python-setuptools libmysqlclient-devel python-devel zlib-devel libressl-devel"
```
jnbr: To make cross-compiling work, `*-devel` packages must be in `makedepends`, not in `hostmakedepends`.

alindobre: Thanks, will update in a minute.

jnbr:
Please add a sub-package for python3 and squash the commits.

alindobre:
Done.

srcpkgs/python-mysqlclient/template
```diff
@@ -0,0 +1,27 @@
...
+short_desc="Python2 interface to MySQL/MariaDB"
+maintainer="Alin Dobre <alin.dobre@outlook.com>"
+license="GPL-2"
+homepage="https://github.com/PyMySQL/mysqlclient-python"
+distfiles="${PYPI_SITE}/m/mysqlclient/mysqlclient-${version}.tar.gz"
+checksum=2d9ec33de39f4d9c64ad7322ede0521d85829ce36a76f9dd3d6ab76a9c8648e5
+
+python3-mysqlclient_package() {
+  noarch=yes
+	build_style=python-module
```
jnbr: not needed, already covered in line 6

srcpkgs/python-mysqlclient/template
```diff
@@ -0,0 +1,27 @@
...
+depends="python"
+short_desc="Python2 interface to MySQL/MariaDB"
+maintainer="Alin Dobre <alin.dobre@outlook.com>"
+license="GPL-2"
+homepage="https://github.com/PyMySQL/mysqlclient-python"
+distfiles="${PYPI_SITE}/m/mysqlclient/mysqlclient-${version}.tar.gz"
+checksum=2d9ec33de39f4d9c64ad7322ede0521d85829ce36a76f9dd3d6ab76a9c8648e5
+
+python3-mysqlclient_package() {
+  noarch=yes
```
jnbr: use tab, not spaces

alindobre:
Learning curve :) Amended again.

srcpkgs/python-mysqlclient/template
```diff
@@ -0,0 +1,26 @@
+pkgname=python-mysqlclient
+version=1.3.12
+revision=1
+noarch=yes
```
jnbr: no. remove it. Same for python3-mysqlclient.
The package contains some architecture specific .so files.

alindobre: That's correct, removed noarch.

jnbr:
looks good to me

PR #12412, opened 2018-03-12
libgdiplus: upgrade to 5.6

nonchip:
this fixes an issue with mono not supporting `System.Drawing.Imaging.ImageFormat.Gif`


maxice8:
also rename the commit.

nonchip:
might be a stupid question, but: how? I'm not very used to git history rewriting, usually I just commit, branch and merge stuff /o\

Issue #12310, opened 2018-03-08
Package request: pragha


TalviT:
"Pragha is a Lightweight Music Player for GNU/Linux."

Website: https://pragha-music-player.github.io/
Code repository: https://github.com/pragha-music-player/pragha

VargMon:
This seems straightforward enough. BTW, cool media player

VargMon:
:-)

TalviT:
Awesome, thank you!

maxice8:
@ebfe this can be closed

VargMon:
You’re welcome my friend :)

PR #12418, opened 2018-03-12
New package: pragha-1.3.3

VargMon:
Tested on `x86_64`. Cross compiling to `i686` works. This is to fulfill package request: #12310 Please let me know if there's any improvements I make to the template. I didn't have to use(or apply) any patches to get this to compile properly. However, I ran into one a ./configure script bug, and did have to export one environment variable. :-)


common/shlibs
```diff
@@ -3110,3 +3110,4 @@ libwayland-egl++.so.0.2 libwaylandpp-2.2.0_1
 libwayland-client++.so.0.2 libwaylandpp-2.2.0_1
 libwayland-cursor++.so.0.2 libwaylandpp-2.2.0_1
 libwayland-client-extra++.so.0.2 libwaylandpp-2.2.0_1
+libdeviceclient.so.0 pragha-1.3.3_1
```
pullmoll: Remove this line. Entries in common/shlibs are required for `*-devel` subpackages only.

srcpkgs/pragha/template
```diff
@@ -0,0 +1,25 @@
...
+license="GPL-3"
+homepage="https://pragha-music-player.github.io/"
+distfiles="https://github.com/pragha-music-player/pragha/archive/v1.3.3.tar.gz"
+checksum=71babb217115f362fc8565072cd70d94f02d58b346f00d2d06e40cecaaf282c8
+
+pre_configure() {
+	# Bug: cpp fails sanity checks configure error.
+	# This seems to be an autotools bug, but assuming the bootstrap
+	# environment is the same as the target, this should to be an okay workaround
+	[ ! $CROSS_BUILD ] || export CPP=/usr/bin/cpp
```
pullmoll: Remove this line. Testing for native builds should be `[ -z "$CROSS_BUILD" ]` because $CROSS_BUILD may be undefined, but see below.

srcpkgs/pragha/template
```diff
@@ -0,0 +1,25 @@
...
+	./autogen.sh
```
pullmoll: Try if `CPP=$CPP ./autogen.sh` fixes the problem.

common/shlibs
VargMon: I was unable to cross compile to i686 without adding to `shlibs` I'm supposing that this _could_ be related to the autotools bug. Please, if I may ask, for a possible solution to properly fix this, or some suggestions (should I add a devel subpackage??)

srcpkgs/pragha/template
VargMon: Will do, thanks

srcpkgs/pragha/template
VargMon: I'd worried about that..

common/shlibs
pullmoll: Our builders don't cross compile for i686 but use a native environment, i.e. `./xbps-src -m masterdir-i686 binary-bootstrap i686`.
I think the problem you saw may rather be related to using the wrong (host = x86_64) CPP.

pullmoll: Unless there will be plugins depending on a pragha-devel package you don't need to create a subpackage. If you do, you will need to move the `usr/include/pragha/plugins/pragha-plugin-macros.h` to this subpackage. I don't see a shared library or other files to move.

srcpkgs/pragha/template
VargMon: Still ran into the same error. Looking for a better solution. ;)

VargMon: * Then the one I found... :!

pullmoll:
FWIW here's a template that works for me and creates a pragha-devel subpackage. For this don't forget to symlink from pragha to pragha-devel in srcpkgs.
```
# Template file for 'pragha'
pkgname=pragha
version=1.3.3
revision=1
build_style=gnu-configure
hostmakedepends="automake libtool xfce4-dev-tools glib-devel gettext-devel pkg-config"
makedepends="
 gtk+3-devel glib-devel gstreamer1-devel sqlite-devel taglib-devel
 libpeas-devel libxfce4ui-devel totem-pl-parser-devel gst-plugins-base1-devel
 libnotify-devel libcdio-devel libcdio-paranoia-devel libcddb-devel
 libkeybinder3-devel glyr-devel rygel-devel grilo-devel libgudev-devel libmtp-devel"
short_desc="Pragha is a Lightweight Music Player for GNU/Linux"
maintainer="VargMon <vargmon98@gmail.com>"
license="GPL-3"
homepage="https://pragha-music-player.github.io/"
distfiles="https://github.com/pragha-music-player/pragha/archive/v1.3.3.tar.gz"
checksum=71babb217115f362fc8565072cd70d94f02d58b346f00d2d06e40cecaaf282c8

pre_configure() {
        CPP=$CPP ./autogen.sh
}

pragha-devel_package() {
        pkg_install() {
                vmove usr/include/pragha/plugins
                vmove usr/lib/pragha/plugins/devices/libdeviceclient.so
        }
}
```

VargMon:
Complete weirdness but I was able to compile without adding to `shlibs`... so no longer necessary... hold on a sec... having to run a few for tests...

pullmoll:
Dito. Cross compiling fails due to the dependency on libpeas-devel. I'll whip up a template that disables pragha-devel for cross builds.

VargMon:
I thought about a *-plugins subpackage but eh... I don't want to over complicated it if I can help it... :]

VargMon:
Before I commit the requested changes.... should I also enable `only_for_archs="i686 x86_64"`???

pullmoll:
> Before I commit the requested changes.... should I also enable only_for_archs="i686 x86_64"???

No, because the package would build just fine on a e.g. RasPI.
`only_for_archs` is meant **only** for cases where the source contains e.g. x86 assembler that can't be used on another target.

VargMon:
Right.... thought as much... but thought I should ask anyway. Okay well, it appears to _cross_ compile (I'm on native x86_64) fine to i686... but I'm getting an odd error on `xbps-install` SHA256 hash is not valid Numerical result out of range I figure it's because it's not signed... if not errrrr...

pullmoll:
If you already built locally, you'll have to `-f` force build to register the same version and revision.
Or you could zap your binpkgs directory or at least the x86_64-repodata file.

I already verified it cross compiles, but without any plugins. As I said: I have no clue if pragha makes any sense without the plugins.

VargMon:
I like the subpackage idea (I automatically think *-plugins but if not *-devel  is okay I guess) like you proposed.... personally I think it's a fine media player even without a plugins.

VargMon:
Also I just had to update my local repo data so it's fine :)

srcpkgs/pragha/template
```diff
@@ -0,0 +1,29 @@
...
+makedepends="
+ gtk+3-devel glib-devel gstreamer1-devel sqlite-devel taglib-devel
+ libpeas-devel libxfce4ui-devel totem-pl-parser-devel gst-plugins-base1-devel
+ libnotify-devel libcdio-devel libcdio-paranoia-devel libcddb-devel
+ libkeybinder3-devel glyr-devel libgudev-devel libmtp-devel"
+short_desc="Pragha is a Lightweight Music Player for GNU/Linux"
+maintainer="VargMon <vargmon98@gmail.com>"
+license="GPL-3"
+homepage="https://pragha-music-player.github.io/"
+distfiles="https://github.com/pragha-music-player/pragha/archive/v1.3.3.tar.gz"
```
pullmoll: Replace 1.3.3 with ${version}.

srcpkgs/pragha/template
```diff
@@ -0,0 +1,29 @@
...
+	# Bug: cpp fails sanity checks configure error.
+	# This seems to be an autotools bug, but assuming the bootstrap
+	# environment is the same as the target, this should to be an okay workaround
+	if [ -n $CROSS_BUILD ]; then
+		export  CPP=/usr/bin/cpp
+		./autogen.sh
+	else
+		./autogen.sh
+	fi
+}
```
pullmoll: This adds complexity for no reason. As you can see in my example template always specyfing CPP=$CPP works perfectly fine.

srcpkgs/pragha/template
```diff
@@ -0,0 +1,29 @@
+# Template file for 'pragha'
+pkgname=pragha
+version=1.3.3
+revision=1
+build_style=gnu-configure
+hostmakedepends="autoconf automake libtool xfce4-dev-tools glib-devel gettext-devel pkg-config"
```
pullmoll: autoconf is a dependency of automake. Remove it.
You could have taken a look at my example template.

srcpkgs/pragha/template
VargMon: Thanks forgot about that one

srcpkgs/pragha/template
VargMon: Aye aye. Okay turns out that if `[ -z $CROSS_BUILD ] || export  CPP=/usr/bin/cpp` is set before cross compiling it seems to cross build with plugins.

srcpkgs/pragha/template
VargMon: True, but that didn't work at all for me. However, `[ -z $CROSS_BUILD ] || export  CPP=/usr/bin/cpp` did. Would this be okay?

srcpkgs/pragha/template
pullmoll: I somehow doubt that using the host /usr/bin/cpp for cross builds is a good idea per se.
I'll give it a try locally to see if indeed it makes the plugins being built.

pullmoll: Oh, I see. You were talking cross building for i686. As I said this is not important because you can build natively for i686.
For cross building for e.g. aarch64 you have to disable the dependency on libpeas-devel as I did in the example template and if it's missing, plugins are not built.

VargMon: Alright, I removed the dumb autoconf dep, and it in all my attempts that `./configure` script doesn't see rygel-server (this one I can understand) and oddly doesn't see libglyr... (I think it a versioning thing)

VargMon:
:)

PR #12441, opened 2018-03-12
keepassxc: update to 2.3.1.

Piraty:


Piraty:
[appearently](12716), auto-type doesn't work, but `configure`picks it up correctly.

PR #12495, opened 2018-03-14
New package: bcnc-0.9.11

faulesocke:
A nice tool I use regularly for controlling CNC machines.


srcpkgs/bcnc/template
```diff
@@ -0,0 +1,30 @@
+# Template file for 'bcnc'
+pkgname=bcnc
+version=0.9.11
+revision=1
+wrksrc=bCNC-$version
+makedepends="python"
```
cr6git: `hostmakedepends="python"`

PR #12500, opened 2018-03-14
New package: tekaim-1.5.0

cr6git:
:camera_flash: screenshot tool. updated to the latest release.


cr6git:
@pullmoll oups, just noticed it, I must update to 1.6.0 !

PR #12504, opened 2018-03-14
Remove kfilemetadata and revdeps

ebfe:
See #12152


Johnnynator:
Doesn't tomahawk needs a revbump so the package in the repo also doesn't have any KDE deps any more?

Also I started to create a [branch to remove kde4](https://github.com/Johnnynator/void-packages/commits/remove-kde4) but Callgra is still giving me troubles.

ebfe:
> Doesn't tomahawk needs a revbump so the package in the repo also doesn't have any KDE deps any more?

Correct, missed the default

PR #12514, opened 2018-03-14
firefox-esr: update to 52.7.1.

cr6git:


cr6git:
See: :fox_face: [52.7.1/releasenotes](https://www.mozilla.org/en-US/firefox/52.7.1/releasenotes/).

PR #12546, opened 2018-03-15
distcc: update upstream URIs

mvf:
Fixes source fetch failure due to HTTP 403


Vaelatern:
but in the future random ${pkgname} in urls often just proves annoying.

mvf:
Noted. It's not random though, it's just that both the organization and the repo are called "distcc" and `${pkgname}` only makes sense for the repo part. Lots of templates seem to use it like that. But I have no strong preference.

PR #12547, opened 2018-03-15
New package: qgit-2.7

mvf:


srcpkgs/qgit/template
```diff
@@ -0,0 +1,18 @@
...
+depends="git"
+short_desc="Qt-based Git GUI"
+maintainer="Matthias von Faber <mvf@gmx.eu>"
+license="GPL-2"
+homepage="https://github.com/tibirna/qgit"
+distfiles="${homepage}/archive/${pkgname}-${version}.tar.gz"
+checksum=183d116b2fb38c6a76c99577f6cc86f0241f8b6bdc74042be4cb8631567f5e9e
+
+if [ "$CROSS_BUILD" ]; then
+	hostmakedepends+=" qt5-devel"
```
ebfe: If you really need the full qt5-devel you can just add it without the conditional.

mvf: Thanks, good point. `qt5-host-tools` and `qt5-qmake` should actually be enough.

Issue #12538, opened 2018-03-15
tigervnc and turbovnc need perl dependency


xianwenchen:
tigervnc and turbovnc need perl dependency, which is missing right now.

PR #12549, opened 2018-03-15
turbovnc: add missing perl dependency

cr6git:
resolves [#12538](https://github.com/voidlinux/void-packages/issues/12538).


cr6git:
[504234d](https://github.com/voidlinux/void-packages/commit/504234d2c2ad0d5f4b4e303474fede2691e2dd27)

Issue #12503, opened 2018-03-14
rust: rustc segfaults while building crate encoding_c


Gottox:
### System

* xuname:  
  Void 4.15.6_2 x86_64 GenuineIntel uptodate rrrrmd
* package:  
  rust-1.24.0_1

### Expected behavior
rustc should build the crate encoding_c

### Actual behavior
Segfault of rustc during build.

### Steps to reproduce the behavior
```
xbps-install cargo
git clone https://github.com/hsivonen/encoding_c
cd encoding_c
cargo build --verbose
```

The stacktrace of the segfaulting call: https://gist.github.com/Gottox/9e3d2e8ab32c3dc16036bf4cba0f342b

rustc from rustup works in both versions, 1.24.0 and 1.24.1:
```
rustc --version
rustc 1.24.1 (d3ae9a9e0 2018-02-27)
rustc --version
rustc 1.24.0 (4d90ac38c 2018-02-12)
```

ebfe:
> #0 __strlen_avx2 () at ../sysdeps/x86_64/multiarch/strlen-avx2.S:62

hmm, old cpu?

Gottox:
calling rustc with `-C debuginfo=0` avoids the error.

pullmoll:
FWIW segfault happens on my box as well
```
cat /proc/cpuinfo | grep model | head -n2
model		: 1
model name	: AMD Ryzen Threadripper 1950X 16-Core Processor
```

pullmoll:
Added the following patch as `srcpkgs/rust/patches/fix-nullptr-LinkageName.patch` and the resulting `rustc` does not crash anymore. Of course I don't know what I'm doing there, just avoiding a NULL pointer in favor of a fixed string "(NULL)".
```
--- rustc-1.24.0-src/src/rustllvm/RustWrapper.cpp       2018-02-12 19:51:18.000000000 +0100
+++ rustc-1.24.0-src/src/rustllvm/RustWrapper.cpp       2018-03-14 21:03:42.277533250 +0100
@@ -725,7 +725,7 @@
   return wrap(VarExpr);
 #else
   return wrap(Builder->createGlobalVariable(
-      unwrapDI<DIDescriptor>(Context), Name, LinkageName,
+      unwrapDI<DIDescriptor>(Context), Name, LinkageName ? LinkageName : "(NULL)",
       unwrapDI<DIFile>(File), LineNo, unwrapDI<DIType>(Ty), IsLocalToUnit,
[...]

PR #12561, opened 2018-03-15
rust: Bump revision to rebuild for x86_64-musl

Arzte:
Resolves this: https://github.com/voidlinux/void-packages/issues/12503#issuecomment-373412700


jnbr:
this does not solve the problem. https://repo.voidlinux.eu/distfiles/cargo-0.24.0-x86_64-unknown-linux-musl.tar.gz needs to be rebuilt against the current libssl.

Arzte:
Was completely wrong, sorry.

chneukirchen:
Use cargo-0.25 then.

PR #12576, opened 2018-03-15
[WIP] New package: wiimms-iso-tools-v3.01a

beta-phenylethylamine:


srcpkgs/wiimms-iso-tools/template
```diff
@@ -0,0 +1,24 @@
+# Template file for 'wiimms-iso-tools'
+pkgname=wiimms-iso-tools
+version=v3.01a
+revision=1
+build_style=gnu-makefile
+make_use_env=true
+makedepends="zlib-devel fuse-devel"
+depends="zlib fuse"
```
maxice8: shouldn't be required, void will track library dependencies automatically

srcpkgs/wiimms-iso-tools/template
```diff
@@ -0,0 +1,24 @@
...
+short_desc="Command line tools to manipulate Wii/GameCube ISOs and WBFS containers"
+maintainer="Phenethylamine <beta.phenylethylamine@gmail.com>"
+license="GPL-2"
+wrksrc="${pkgname}"
+homepage="https://wit.wiimm.de/"
+distfiles="https://opensvn.wiimm.de/viewvc/wii/trunk/wiimms-iso-tools/?view=tar>wiimms.tar.gz"
```
maxice8: Isn't there a stable tarball that can be used ?

beta-phenylethylamine: I was having trouble finding one before but I think I've actually figured out a way. One second.

beta-phenylethylamine: This source is actually behaving really oddly even if I do specify the exact revision. Give me a bit to figure this out.

beta-phenylethylamine: Even when retrieving the same file from the same link, I'm getting a slightly different file each which is causing the checksum to fail.

maxice8: That is a know problem with stuff that auto-generate archives, i very strong suggest using a release tarball, if you really can't then you can use prefix the checksum with @ to indicate that it is auto-generated and so it will check the file contents.

beta-phenylethylamine:
Okay, so I'm running into a weird problem with this. Every time I fetch the tarball from source (even once it's been changed to what should be a stable tarball), it's slightly different, despite, upon unarchiving it, running diff -rs between the two directories reports all files the same.

maxice8:
This is a know problem with CVS that auto-generate tarballs ( the timestamps differ, which cause different checksums ), i ***EXTREMELY*** strongly suggest using a proper release tarball.

beta-phenylethylamine:
I'll sniff around for a proper release tarball but I wasn't able to find one before unfortunately :\

beta-phenylethylamine:
Yeah. There's not a proper source release tarball anywhere on the site. There are linux 32 bit and 64 bit tarballs but those are actual binary releases.

maxice8:
On Fri, Mar 16, 2018 at 02:18:00AM +0000, Phenethylamine wrote:
> Yeah. There's not a proper source release tarball anywhere on the site. There
> are linux 32 bit and 64 bit tarballs but those are actual binary releases.
> 
> —
> 

upstream is buggy, like mozilla was when they didn't release a 59.0 tarball, i'd open a bugreport

beta-phenylethylamine:
Done. Should I close this pull request until this is fixed upstream, or just let it stay open for now?

maxice8:
On Fri, Mar 16, 2018 at 02:35:13AM +0000, Phenethylamine wrote:
> Done. Should I close this pull request until this is fixed upstream, or just
> let it stay open for now?
> 
> —
> 

keep it open, put [WIP] in the title.

beta-phenylethylamine:
Upstream has declared it WONTFIX so I'm closing this pull request

PR #12590, opened 2018-03-16
cutemarked: fix cross compilation

newbluemoon:
Just needed some hostmakedepends.


srcpkgs/cutemarked/template
```diff
@@ -15,3 +14,7 @@ license="GPL-2"
 homepage="http://cloose.github.io/CuteMarkEd/"
 distfiles="https://github.com/cloose/${pkgname}/archive/v${version}.tar.gz"
 checksum=78a41808c2f0452375810abdff76eeaaee012f8d1368a2b8772ec6b4d2ceeec8
+
+if [ -n "$CROSS_BUILD" ]; then
+	hostmakedepends+=" qt5-devel qt5-webkit-devel"
```
ebfe: This doesn't look right. qt5-webkit-devel contains no binaries or similar that could be used in the build.

newbluemoon: Without it I get

    Project ERROR: Unknown module(s) in QT: webkitwidgets

newbluemoon:
Hmm, just thought qt packages are a bit special in this respect. When modules aren’t found in cross builds the corresponding -devel package is added to `hostmakedepends`. Like e.g. for ghostrider, lumina, orion, etc.

ebfe:
> Hmm, just thought qt packages are a bit special in this respect. 

cc @pullmoll

pullmoll:
The qmake build style always had its issues with cross building.
I guess there should be some way to generate a local `.qmake.cache` or `.qmake.conf` which tells qmake to use a non-empty `QT_SYSROOT` where to find packages. `qmake -query` prints `QT_SYSROOT:` twice (without value) so there may still something be wrong when qmake is built.

I never managed to get cross qt5 submodule detection right and this is why in many templates we have to install (some) dependencies in the host for qmake to detect their availability.

Issue #12585, opened 2018-03-16
fcitx-4.2.9.6_2 segfault on musl


yshui:
Seems fcitx segfaults twice. Once in the program, then again in the segfault handler when it tries to get a stack trace.

newbluemoon:
Did 4.2.9.6_1 work? Then perhaps the change to use libexecinfo is the culprit.

yshui:
@newbluemoon Yep, _1 works

newbluemoon:
I reverted it to use the execinfo patch again instead of libexecinfo. Perhaps you could already build the referenced PR and test it?

newbluemoon:
@yshui No need to test anymore, it’s already merged. :)

yshui:
Works now!

PR #12594, opened 2018-03-16
fcitx: revert using libexecinfo

newbluemoon:
The change to use libexecinfo in favor of patching it out seems to cause a segmentation fault, see #12585. This reverts it to use the patch again.


pullmoll:
FWIW There are warnings when compiling libexecinfo for e.g. armv7l-musl:
```
stacktraverse.c:11:20: warning: unsupported argument to '__builtin_return_address'
     case 0: return __builtin_return_address(1);
                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~
```
There are no such warnings for x86_64-musl, i686-musl or aarch64-musl, but perhaps these warnings indicate something is wrong with how libexecinfo tries to traverse the stack.

newbluemoon:
Would you suggest to better not use libexecinfo then?

pullmoll:
I'm not sure, but we should keep an eye on packages using libexecinfo and bug reports for *-musl.
I don't feel like trying to debug what actually was causing the segfault in fcitx.
Perhaps one of the other packages using libexecinfo is easier to test?
Currently we use it in `julia`, `telegram-desktop`, `ddcutil`, `remmina` and `varnish`, none of which I know about or use :P

newbluemoon:
Ok, thanks. I’ll keep that in mind. :)
Trying to track this down I don’t think I can be of much help. ;)

PR #12616, opened 2018-03-17
smplayer: update to 18.2.2.

Hoshpak:


cr6git:
@Hoshpak Please check the `homepage` when you update things... The current `homepage` is wrong.

PR #9419, opened 2017-11-21
New package: marble5-17.12.2

Chocimier:
There is a warning `=> WARNING: marble5-devel-17.08.3_1: libmarbledeclarative.so not found in common/shlibs!` although the file is given in shlibs.

PR #12620, opened 2018-03-17
calligra: update to 3.1.0 [ci skip]

Johnnynator:
includes #9419

atm calligra-gemini is segfaulting (tested on x86_64 musl and glibc) and it would be nice if someone could help me fixing it.

```
APP_FLOW:  Flow app (for Desktop)  |  Not yet ported to Qt5/KF5
APP_BRAINDUMP:  Braindump app (for Desktop)  |  Disabled, will (probably) be removed from Calligra / No maintainer currently
```

Also the build takes [too long](https://travis-ci.org/Johnnynator/void-packages/builds/354694350) for Travis


Johnnynator:
Weird it only shows this behavior outside of a kde5 desktop.

Johnnynator:
It works as intended inside of kde5. Outside of that it shows some weird behavior here and there which you can also observe in other distros.

PR #12646, opened 2018-03-18
matio: update to 1.5.12

eichf:


common/shlibs
```diff
@@ -3077,7 +3077,7 @@ libembb_mtapi_c.so embb-devel-1.0.0_1
 libembb_mtapi_network_c.so embb-devel-1.0.0_1
 libembb_base_cpp.so embb-devel-1.0.0_1
 libembb_base_c.so embb-devel-1.0.0_1
-libmatio.so.4 matio-1.5.11_1
+libmatio.so.4 matio-1.5.12_1
```
ebfe: This only needs to change if the soname changes. It is used to track the first version that introduced the specific soname.

PR #12647, opened 2018-03-18
appstream-glib: add missing depends on -devel

maxice8:


srcpkgs/appstream-glib/template
```diff
@@ -27,7 +27,8 @@ else
 fi
 
 appstream-glib-devel_package() {
-	depends="${sourcepkg}>=${version}_${revision} libarchive-devel gcab-devel gdk-pixbuf-devel"
+	depends="${sourcepkg}>=${version}_${revision} gcab-devel gdk-pixbuf-devel
+	libarchive-devel libglib-devel libsoup-devel libuuid-devel libyaml-devel"
```
ebfe: glib-devel is pulled in by gcab-devel already, the rest are from Requires.private are these really required?

maxice8:
On Sun, Mar 18, 2018 at 05:36:39PM +0000, Michael Gehring wrote:
> @ebfe commented on this pull request.
> 
> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
> 
> In srcpkgs/appstream-glib/template:
> 
> > @@ -27,7 +27,8 @@ else
>  fi
> 
>  appstream-glib-devel_package() {
> -       depends="${sourcepkg}>=${version}_${revision} libarchive-devel gcab-devel gdk-pixbuf-devel"
> +       depends="${sourcepkg}>=${version}_${revision} gcab-devel gdk-pixbuf-devel
> +       libarchive-devel libglib-devel libsoup-devel libuuid-devel libyaml-devel"
> 
> glib-devel is pulled in by gcab-devel already, the rest are from
> Requires.private are these really required?
> 
> —
> 

yes, Requires.private are required when using pkg-config --cflags, i prefer being explicit and declaring them all, it also prevents breakage happening in case gcab-devel somehow drops glib-devel dependency.

ebfe:
> yes, Requires.private are required when using pkg-config --cflags, 

Have an example? 

> i prefer being explicit and declaring them all, it also prevents breakage happening in case gcab-devel somehow drops glib-devel dependency.

We do it the exact oppsite way everywhere else. There is even a tool for
it (xmindep).

maxice8:
> Have an example?

```
$ cat /usr/lib/pkgconfig/libgcab-1.0.pc | grep Requires 
Requires: gio-2.0
Requires.private: zlib
$ xbps-remove -Fy zlib-devel
$ pkg-config --cflags libgcab-1.0
Package zlib was not found in the pkg-config search path.
Perhaps you should add the directory containing `zlib.pc'
to the PKG_CONFIG_PATH environment variable
Package 'zlib', required by 'libgcab-1.0', not found
```

> We do it the exact oppsite way everywhere else. There is even a tool for it (xmindep).

oh, i wasn't aware of that, sorry :S , i'll see what i can do with it

maxice8:
can you guide me ? i don't understand how to use xmindep.

```
$ xmindep appstream-glib
appstream-glib
```

ebfe:
> $ xmindep appstream-glib
> appstream-glib

$ xmindep gcab-devel glib-devel
gcab-devel

maxice8:
ok, updated, is there a tool that automatically extracts from a template or should i make a shellscript wrapper ?

ebfe:
> ok, updated, is there a tool that automatically extracts from a template or should i make a shellscript 
> wrapper ?

I don't know a tool, also i'm not entirely convinced these private deps should be added to the packages if the only need is pkg-config --cflags.

maxice8:
>  I don't know a tool, also i'm not entirely convinced these private deps should be added to the packages if the only need is pkg-config --cflags.

the deps are also present on the Debian package

ebfe:
> >  I don't know a tool, also i'm not entirely convinced these private deps should be added to the packages if the only need is pkg-config --cflags.
> the deps are also present on the Debian package

Yes, but at the same time they aren't really hard dependencies. Packages
that ship a zsh completion file don't depend on zsh, packages that ship
cmake configs in /usr/lib/cmake/ don't depend on cmake etc. Not saying
this isn't usefule in some way, it's just not how it's currently handled
in other places.

PR #12662, opened 2018-03-19
noto-fonts-ttf: update to 20180322, split -extra subpkg

cr6git:
Same as Arch [PKGBUILD](https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/noto-fonts).


cr6git:
Updated to the latest version.

PR #12681, opened 2018-03-19
New package: imagescan-plugin-networkscan-1.1.1

gspe:
Epson Imagescan plugin to enable network scan


gspe:
Yeah you are right, but this message is just an extract of `/usr/share/doc/imagescan-plugin-networkscan/README` file, that is installed with the package.
Maybe I can make this message shorter and just link to this file.

pullmoll:
Way too much text for my taste :) I modified it before pushing.

PR #12690, opened 2018-03-19
gparted: update to 0.31.0.

cr6git:


maxice8:
Dependencies can be minimized and sorted with `xmindep` 

```
hostmakedepends="gnome-doc-utils intltool pkg-config"
makedepends="glib-devel gnome-doc-utils gtkmm2-devel libparted-devel"
depends="desktop-file-utils hicolor-icon-theme"
```

PR #12691, opened 2018-03-19
New package: mpv-mpris-0.1

subosito:
MPRIS plugin for mpv


srcpkgs/mpv-mpris/INSTALL.msg
```diff
@@ -0,0 +1,9 @@
+>>> To load this plugin, specify the following option when starting mpv:
+
+    --script /usr/lib/mpv/mpris.so
```
maxice8: I don't think this is necessary.

srcpkgs/mpv-mpris/INSTALL.msg
```diff
@@ -0,0 +1,9 @@
...
+
+>>> To load this plugin automatically whenever mpv runs, do the following:
+
+    mkdir -p $HOME/.config/mpv/scripts
+    ln -s /usr/lib/mpv/mpris.so $HOME/.config/mpv/scripts/mpris.so
```
maxice8: This would fail on my system which uses XDG_CONFIG_HOME to set it to $HOME/etc/mpv , don't be afraid to assume competence of the user.

srcpkgs/mpv-mpris/template
```diff
@@ -0,0 +1,25 @@
+# Template file for 'mpv-mpris'
+pkgname=mpv-mpris
+version=0.1
+revision=1
+only_for_archs="i686 x86_64"
+build_style=gnu-makefile
+hostmakedepends="pkg-config"
+makedepends="libglib-devel mpv-devel"
+depends="mpv glib"
```
maxice8: these shouldn't be necessary, mpv and glib should be automatically added via shlibs

srcpkgs/mpv-mpris/template
```diff
@@ -0,0 +1,25 @@
+# Template file for 'mpv-mpris'
+pkgname=mpv-mpris
+version=0.1
+revision=1
+only_for_archs="i686 x86_64"
```
maxice8: please do not use these, prefer to use 

   arch1) broken="why it is broken goes here"
esac

having it not be supported upstream can be considered a reason to add to broken.

srcpkgs/mpv-mpris/template
```diff
@@ -0,0 +1,25 @@
...
+distfiles="https://github.com/hoyon/mpv-mpris/archive/v${version}.tar.gz"
+checksum=47afa6064fdd7552079d1574644bb1b16b782e96ed6aeebe24269e00b90a381b
+
+do_build() {
+	make mpris.so
+}
+
+do_install() {
+	vlicense LICENSE
+	vmkdir usr/lib/mpv
```
maxice8: prefer using usr/lib/${pkgname} , this prevents polluting the libdir of other packages.

srcpkgs/mpv-mpris/template
```diff
@@ -0,0 +1,25 @@
...
+depends="mpv glib"
+short_desc="MPRIS plugin for mpv"
+maintainer="Alif Rachmawadi <arch@subosito.com>"
+license="MIT"
+homepage="https://github.com/hoyon/mpv-mpris"
+distfiles="https://github.com/hoyon/mpv-mpris/archive/v${version}.tar.gz"
+checksum=47afa6064fdd7552079d1574644bb1b16b782e96ed6aeebe24269e00b90a381b
+
+do_build() {
+	make mpris.so
```
maxice8: you can use make_build_target="mpris.so"

maxice8:
Here is a patch that fixes the template

```
diff --git a/srcpkgs/mpv-mpris/template b/srcpkgs/mpv-mpris/template
index f9f2b11c21..c736d5960b 100644
--- a/srcpkgs/mpv-mpris/template
+++ b/srcpkgs/mpv-mpris/template
@@ -2,24 +2,21 @@
 pkgname=mpv-mpris
 version=0.1
 revision=1
-only_for_archs="i686 x86_64"
 build_style=gnu-makefile
+make_use_env=yes
 hostmakedepends="pkg-config"
 makedepends="libglib-devel mpv-devel"
-depends="mpv glib"
+depends="mpv"
 short_desc="MPRIS plugin for mpv"
 maintainer="Alif Rachmawadi <arch@subosito.com>"
 license="MIT"
 homepage="https://github.com/hoyon/mpv-mpris"
 distfiles="https://github.com/hoyon/mpv-mpris/archive/v${version}.tar.gz"
 checksum=47afa6064fdd7552079d1574644bb1b16b782e96ed6aeebe24269e00b90a381b
-
-do_build() {
-	make mpris.so
-}
+make_build_target="mpris.so"
 
 do_install() {
 	vlicense LICENSE
-	vmkdir usr/lib/mpv
-	vinstall mpris.so 0644 usr/lib/mpv
+	vmkdir usr/lib/${pkgname}
+	vinstall mpris.so 0644 usr/lib/${pkgname}
 }
```

it can be applied with `patch -p1 < patch.p1`

---

Here is a patch that needs to be added to srcpkgs/mpv-mpris/patches and finish with the suffix .patch, this fixes cross-compilation

```
--- Makefile
+++ Makefile
@@ -1,5 +1,5 @@
-CFLAGS=-std=c99 -Wall -Wextra -O2 `pkg-config --cflags gio-2.0 gio-unix-2.0 mpv`
-LDFLAGS=`pkg-config --libs gio-2.0 gio-unix-2.0`
+CFLAGS += -std=c99 -Wall -Wextra -O2 $(shell pkg-config --cflags gio-2.0 gio-unix-2.0 mpv)
+LDFLAGS += $(shell pkg-config --libs gio-2.0 gio-unix-2.0)
 
 mpris.so: mpris.c
 	gcc mpris.c -o mpris.so $(CFLAGS) $(LDFLAGS) -shared -fPIC
```

Vaelatern:
INSTALL.msg files are annoying. Is it absolutely necessary? Every line of it?

subosito:
@maxice8 Thanks for the suggestions and fixes. I've implemented that :+1: 
@Vaelatern Yea, I don't think it's necessary, just removed it.

maxice8:
On Tue, Mar 20, 2018 at 05:23:04AM +0000, Alif Rachmawadi wrote:
> @maxice8 Thanks for the suggestions and fixes. I've implemented that 👍
> @Vaelatern Yea, I don't think it's necessary, just removed it.
> 
> —
> You are receiving this because you were mentioned.
> 

have them as only one commit.

subosito:
@maxice8 Done

srcpkgs/mpv-mpris/patches/fix_build.patch
```diff
@@ -0,0 +1,10 @@
+--- Makefile
++++ Makefile
+@@ -1,5 +1,5 @@
+-CFLAGS=-std=c99 -Wall -Wextra -O2 `pkg-config --cflags gio-2.0 gio-unix-2.0 mpv`
+-LDFLAGS=`pkg-config --libs gio-2.0 gio-unix-2.0`
++CFLAGS += -std=c99 -Wall -Wextra -O2 $(shell pkg-config --cflags gio-2.0 gio-unix-2.0 mpv)
++LDFLAGS += $(shell pkg-config --libs gio-2.0 gio-unix-2.0)
+ 
+ mpris.so: mpris.c
+ 	gcc mpris.c -o mpris.so $(CFLAGS) $(LDFLAGS) -shared -fPIC
```
Vaelatern: Should use `$(CC)` instead of just calling `gcc`

srcpkgs/mpv-mpris/template
```diff
@@ -0,0 +1,22 @@
+# Template file for 'mpv-mpris'
+pkgname=mpv-mpris
+version=0.1
+revision=1
+build_style=gnu-makefile
+make_use_env=yes
+hostmakedepends="pkg-config"
+makedepends="libglib-devel mpv-devel"
+depends="mpv"
```
Vaelatern: I don't think it actually needs `mpv-devel`

maxice8: https://github.com/hoyon/mpv-mpris/blob/master/mpris.c#L3

```
$ xlocate mpv/client.h
mpv-devel-0.27.2_1	/usr/include/mpv/client.h
```

Vaelatern: Fair, it does.

Vaelatern:
With a change to $(CC) it should become able to cross build, and make Travis green.

subosito:
Thanks for the suggestion @Vaelatern, just updated to `$(CC)`. Awesome :+1:

srcpkgs/mpv-mpris/template
```diff
@@ -0,0 +1,22 @@
...
+short_desc="MPRIS plugin for mpv"
+maintainer="Alif Rachmawadi <arch@subosito.com>"
+license="MIT"
+homepage="https://github.com/hoyon/mpv-mpris"
+distfiles="https://github.com/hoyon/mpv-mpris/archive/v${version}.tar.gz"
+checksum=47afa6064fdd7552079d1574644bb1b16b782e96ed6aeebe24269e00b90a381b
+make_build_target="mpris.so"
```
Vaelatern: Upstream does not ask for this line. And it is superfluous.

subosito: Default `install` has additional steps:

```
install: mpris.so
	mkdir -p $(HOME)/.config/mpv/scripts
	cp mpris.so -t $(HOME)/.config/mpv/scripts
```

Which is preferred way of doing this? using default build target or using custom build one? CC @maxice8

Vaelatern: Fair, I'll merge it thus.

subosito: Awesome, thanks @Vaelatern

PR #12694, opened 2018-03-19
xforms: update download url

abenson:


ebfe:
No need to revbump if the tarball didn't change.

abenson:
Fixed.

PR #12712, opened 2018-03-20
openssh: add gssapi vopt

PeterFaiman:
Also remove --without-ssh1, which is not a valid configure arg.


Gottox:
Thanks @PeterFaiman!

PR #12721, opened 2018-03-20
racket: add support for musl

pbui:
To support musl, I ported the libucontext library from Alpine Linux along with a few patches.

I tested cross-compilation with x86_64-musl and armv7l-musl but have not actually executed the resulting binaries.


xxzozaxx:
while True : "Thank you"

pbui:
@cr6git Oops, I've added `vlicense` to libucontext.

Issue #12328, opened 2018-03-09
supercollider 3.9.1 won't launch


jarbowski:
### System
x86_64
* xuname:  
Void 4.14.22_2 x86_64 GenuineIntel uptodate rrrmFF

  *output of ``xuname`` (part of xtools)*
* package:  
supercollider 3.9.1
  *affected package(s) including the version*

### Actual behavior

*** Error in `scide': munmap_chunk(): invalid pointer: 0x000055f40ce560c0 ***

### Steps to reproduce the behavior
launch supercollider (scide)

mvf:
Confirmed. Starting with 3.9.1, SuperCollider's heap is all messed up. `scide` dies trying to parse its configuration file at `$HOME/.config/SuperCollider/sc_ide_conf.yaml`:

```
Starting program: /usr/bin/scide 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
[New Thread 0x7fffe331f700 (LWP 4244)]
[New Thread 0x7fffe24d5700 (LWP 4245)]
*** Error in `/usr/bin/scide': munmap_chunk(): invalid pointer: 0x00005555559310c0 ***

Thread 1 "scide" received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
51	../sysdeps/unix/sysv/linux/raise.c: そのようなファイルやディレクトリはありません.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff279effd in __GI_abort () at abort.c:90
#2  0x00007ffff27df9dd in __libc_message (action=action@entry=do_abort, fmt=fmt@entry=0x7ffff28e5d78 "*** Error in `%s': %s: 0x%s ***\n") at ../sysdeps/posix/libc_fatal.c:181
#3  0x00007ffff27e606c in malloc_printerr (action=<optimized out>, str=0x7ffff28e5dd0 "munmap_chunk(): invalid pointer", ptr=<optimized out>, ar_ptr=<optimized out>) at malloc.c:5425
#4  0x00007ffff317c345 in std::string::reserve(unsigned long) () from /usr/lib/libstdc++.so.6
[...]

PR #12726, opened 2018-03-20
supercollider: fix cross

mvf:
This should unbreak `supercollider` builds for all platforms and links against shared `boost` and `yaml-cpp` instead of the bundled static versions. The different linking avoids the immediate `scide` startup crash on `x86_64` glibc (#12328), but unfortunately not the `sclang` crash. Both glibc binaries probably remain subtly broken. The `x86_64-musl` versions seem to start up fine though.


srcpkgs/supercollider/template
```diff
@@ -1,19 +1,41 @@
...
-	*-musl) broken="Doesn't compile under musl becasue of glibc dependant code in SC_StringBuffer.cpp";;
+	*-musl) makedepends+=" libglob-devel";;
 esac
+
+pre_configure() {
+	sed -i 's/yaml/yaml-cpp/g' CMakeLists.txt editors/sc-ide/CMakeLists.txt
+	case "$XBPS_TARGET_MACHINE" in
+	*-musl)
+		sed -i \
+		 -e '/include/s,glob\.h,libglob/glob\.h,g' \
```
pullmoll: Does this work? AFAIK you have to escape the dot (.) in the replace string.

mvf: It does work, but I'm as surprised as you that it does. Fixed, thanks!

Issue #12716, opened 2018-03-20
keepassxc no auto-type


AngelFreak:
### System

* xuname:  
  Void 4.14.26_1 x86_64 GenuineIntel uptodate rrrmFF
* package:  
  keepassxc 2.3.1

### Expected behavior
Auto-Type menu in application.
Auto-Type working
### Actual behavior
No Auto-Type menu in application.
Auto-Type is not working
### Steps to reproduce the behavior
xbps-install -Su keepassxc

### Notes

Auto-Type was working fine before system update.

Piraty:
true.
I am investigating

PR #12729, opened 2018-03-20
keepassxc: fix auto-type

Piraty:
fixes https://github.com/voidlinux/void-packages/issues/12716


lemmi:
Thanks for taking this on. Just a short question: why do we need so many buildoptions and why are some of them disabled? They don't look like they introduce a lot of dependencies.

PR #12732, opened 2018-03-20
Update gstreamer to 1.14.0

Hoshpak:


srcpkgs/gstreamer-vaapi/template
```diff
@@ -12,4 +12,4 @@ maintainer="Juan RP <xtraeme@voidlinux.eu>"
 license="LGPL-2.1"
 homepage="https://gstreamer.freedesktop.org"
 distfiles="${homepage}/src/gstreamer-vaapi/gstreamer-vaapi-${version}.tar.xz"
-checksum=1c2d77242e1f30c4d1394636cae9f6877228a017960fca96881e0080d8b6e9c9
+checksum=e4e31f085ef289bf1049398f641345979d20a1b11a80285744bba98504991df2
```
ebfe: s/bad1/base1/ ?

Hoshpak: The package requires gstreamer-codecparsers which is part of the bad package.

PR #12735, opened 2018-03-20
New package: mellowplayer-3.3.5

gspe:


srcpkgs/mellowplayer/template
```diff
@@ -0,0 +1,18 @@
...
+version=3.3.5
+revision=1
+wrksrc=MellowPlayer-${version}
+build_style=cmake
+hostmakedepends="pkg-config"
+makedepends="qt5-declarative-devel qt5-quickcontrols2-devel
+			 qt5-plugin-mysql qt5-plugin-pgsql qt5-plugin-sqlite
+			 qt5-plugin-tds qt5-plugin-odbc
+			 qt5-svg-devel qt5-webengine-devel qt5-webchannel-devel
+			 qt5-location-devel qt5-tools-devel libnotify-devel"
```
Johnnynator: intend them with only one space.

cr6git:
@gspe you need to add `qt5-host-tools` in the `hostmakedepends` for cross-compilation.

gspe:
Thanks, I always forgot on first push

srcpkgs/mellowplayer/template
```diff
@@ -0,0 +1,22 @@
+# Template file for 'mellowplayer'
+pkgname=mellowplayer
```
bra1nwave: Shouldn't the name be `MellowPlayer`?

gspe: No it's not necessary

bra1nwave: Well, the [manual](https://github.com/voidlinux/void-packages/blob/master/Manual.md#programs) says otherwise:
> For those packages the upstream packages name should be used. Remember that in contrast to many other distributions, void doesn't lowercase package names.

PR #12737, opened 2018-03-20
tesseract-ocr: use the right tessdata version for 3.05

newbluemoon:
The data files used are for tesseract version 4.00 (still beta), but we are on 3.05. According to their [wiki](https://github.com/tesseract-ocr/tesseract/wiki/Data-Files) this is not supported (they have a legacy mode, that’s why I think it works except for the problems below). The latest commit in the [tessdata repository](https://github.com/tesseract-ocr/tessdata/commits/master) we can use is from 2016-03-08. After that the format changed (we’re currently using the commit from 2016-12-28).

The first problem is that this causes lots of error messages like `ParamsModel::Incomplete line` for the languages ben, chi_tra, deu, fra, ita, jpn, kor, nld, spa, tel, tha, and vie. At least gscan2pdf seems to choke on
that – no problems with the version from 2016-03-08.

Another problem is with ‘ara’ and ‘hin’. Using 3.05 those languages still need the cube data files which have been removed from the newer archive. See the [wiki](https://github.com/tesseract-ocr/tesseract/wiki/Data-Files#cube-data-files-for-version-304305). Without this data tesseract segfaults.

So I propose to use the tessdata from 2016-03-08.
                                                                                                                                                                                                        
cc @pullmoll


pullmoll:
Yes, thank you @newbluemoon ! I should have realized this earlier because we need those files in a project I'm doing paid work for as well and there we had to change the data files some time ago.

PR #12746, opened 2018-03-21
smplayer: update to 18.3.0.

Hoshpak:


cr6git:
What about including this patch from Debian, which disables updates checking mechanism?
````
Description: Disable updates checking mechanism.
Author: Alessio Treglia <alessio@debian.org>
Forwarded: not-needed

--- a/src/smplayer.pro
+++ b/src/smplayer.pro
@@ -29,7 +29,7 @@ DEFINES += MINIGUI
 DEFINES += MPCGUI
 DEFINES += SKINS
 DEFINES += MPRIS2
-DEFINES += UPDATE_CHECKER
+#DEFINES += UPDATE_CHECKER
 #DEFINES += CHECK_UPGRADED
 DEFINES += AUTO_SHUTDOWN_PC
 #DEFINES += CAPTURE_STREAM
````

Hoshpak:
I haven't encountered any update messages yet but if this nags users that our version is outdated, it would certainly be a good idea to disable it, as far as I remember we also did it for mumble and kodi. Just create a PR to build the package with that patch.

PR #12747, opened 2018-03-21
beignet: fix llvm6 build.

maxice8:


chneukirchen:
Seems to fail on x86_64-musl according to travis.

maxice8:
should work now https://travis-ci.org/maxice8/void-packages/jobs/356461141

PR #12748, opened 2018-03-21
lilypond: fix cross compilation

newbluemoon:
The issue in cross building lilypond is its documentation: it uses help2man to generate the man pages and also creates the texinfo files using the lilypond binary.

My error was that I used the wrong option to disable the documentation which only partially disabled it and so man pages and texinfo files were created. But this failed in a cross build, of course. That’s fixed now. Man pages and texinfo doc are now also contained in the separate lilypond-doc (they were excluded to avoid confilict with the lilypond package). And I cleaned up the template a bit.

Tested the cross build on a Raspberry Pi (armv7l) and everything was fine.

cc @lemmi


lemmi:
I'm building this locally right now, since travis doesn't like the huge output.

newbluemoon:
Is it ok to use `[skip ci]` for future updates with this package?

lemmi:
Sure, just ping me, so I don't overlook the PR.

newbluemoon:
Ok. Thanks! :)

PR #12755, opened 2018-03-21
New package: trelby-2.2.0

abenson:


abenson:
I don't like that it puts most of its files in `/opt/trelby`. I can rework it to put them under `/usr/share/trelby`.

Anyone have an opinion?

Issue #9830, opened 2017-12-07
android-studio-3.0.1_1 fails to install with cross-vpkg-dummy conflicts


Ethan-ks:
### System

* xuname:  
Void 4.13.13_1 x86_64 GenuineIntel uptodate rrrm
* package:  
android-studio-3.0.1_1

### Expected behavior
Android studio install normally

### Actual behavior
install fails with 

    CONFLICT: cross-vpkg-dummy-0.25_1 with installed pkg kernel-libc-headers-4.9.8_3 (matched by kernel-libc-headers>=0)
    CONFLICT: cross-vpkg-dummy-0.25_1 with installed pkg binutils-2.29.1_1 (matched by binutils>=0)
    CONFLICT: cross-vpkg-dummy-0.25_1 with installed pkg libgcc-7.2.0_4 (matched by libgcc>=0)
    CONFLICT: cross-vpkg-dummy-0.25_1 with installed pkg libstdc++-7.2.0_4 (matched by libstdc++>=0)
    CONFLICT: cross-vpkg-dummy-0.25_1 with installed pkg libstdc++-devel-7.2.0_4 (matched by libstdc++-devel>=0)
    CONFLICT: cross-vpkg-dummy-0.25_1 with installed pkg libgfortran-7.2.0_4 (matched by libgfortran>=0)
    CONFLICT: cross-vpkg-dummy-0.25_1 with installed pkg libgomp-7.2.0_4 (matched by libgomp>=0) 
    CONFLICT: cross-vpkg-dummy-0.25_1 with installed pkg gcc-7.2.0_4 (matched by gcc>=0)
    CONFLICT: cross-vpkg-dummy-0.25_1 with installed pkg glibc-2.26_3 (matched by glibc>=0)
    CONFLICT: cross-vpkg-dummy-0.25_1 with installed pkg glibc-devel-2.26_3 (matched by glibc-devel>=0)
    Transaction aborted due to conflicting packages.

### Steps to reproduce the behavior
On my system, attempting to install with 
`# xbps-install android-studio` 
and 
`# xbps-install --repository=hostdir/binpkgs android-studio`
after building the package in my local clone of the void packages repo with xbps-src

I do not have another void machine to test this behaviour on.

the output of `xbps-install -d -n android-studio` is as follows:

    [DEBUG] XBPS: 0.51 API: 20151030 GIT: UNSET
    [DEBUG] Processing configuration directory: /etc/xbps.d
[...]

PR #12766, opened 2018-03-21
android-studio: fix dependency conflict

Pretch:
android-studio package fails to install, see
https://github.com/voidlinux/void-packages/issues/9830

### DIFF
--- a/srcpkgs/android-studio/template
+++ b/srcpkgs/android-studio/template
@@ -21,6 +21,8 @@ license="Apache-2.0"
 homepage="http://tools.android.com/"
 nopie=yes
 nostrip=yes
+noverifyrdeps=yes
+depends="openjdk-jre libgcc libXtst libXrender libXi libXext libX11 glibc freetype alsa-lib"
 
 do_install() {
 	unzip android-studio-ide-${_studio_build}-linux.zip


Pretch:
Perhaps `revision` bump is missing.

pullmoll:
If you run `xbps-rindex -Rx android-studio` you will see the dependencies which it currently has:
```
libGL>=0
musl>=0.9.9_1
glibc>=2.25_1
openjdk-jre>=8u20_1
freetype>=2.3.9_1
libgcc>=4.4.0_1
libXext>=1.0.5_1
libX11>=1.2_1
libXrender>=0.9.4_1
libXtst>=1.0.3_1
libXi>=1.2.1_1
alsa-lib>=1.0.20_1
```
So these are the same which you explicitly list, except for `musl>=0.9.9_1`.

Running `sudo xbps-install -d -S android-studio` reveals that the reason for the failing dependency tracking is the dependency on `libc.so` which is provided by the `musl` package, but not by `glibc`.

I don't know if the workaround to set `noverifyrdeps=yes` and explicitly listing the dependencies is a good choice to deal with the underlying problem, @voidlinux/pkg-committers ?

pullmoll:
@voidlinux/pkg-committers what happens, if we add
```
shlib_provides="libc.so"
```
to `srcpkgs/glibc/template` ?

Pretch:
@pullmoll, If I understand correctly linking directly to `libc.so` is not valid in terms of glibc.

I checked my Debian sid chroot with `objdump -p bin/* usr/bin usr/sbin 2>/dev/null | grep libc.so | sort | uniq`:
```
  NEEDED               libc.so.6
  required from libc.so.6:
```

Same on Void Linux with glibc.

I belive `libc.so` entries come from Android shared libraries included in archive. See https://github.com/voidlinux/void-packages/issues/9830#issuecomment-369981172

Those are for Android system. Should we collect deps for them?

Pretch:
@pullmoll 
> Running sudo xbps-install -d -S android-studio reveals that the reason for the failing dependency tracking is the dependency on libc.so which is provided by the musl package, but not by glibc.

Actually glibc doesn't provide `libc.so` as shared library. It is ld script that is used at linking time.

pullmoll:
If there are libs inside the android-studio binary package which are not required for the target system (x86_64 or i686), they should be removed from the `$DESTDIR/usr/lib/...` in `post_install()`.
I'll have a look at it to see if I can get rid of that wrong dependency this way.

Pretch:
@pullmoll My guess: `libperfa.so` are added to Android .apk to provide perfomance metrics while in development. So it is part of Android Studio functionality.

pullmoll:
I'm not really acquainted with Android or smart phones at all. I don't even have one.
The problem here seems to be that libperfa.so is detected as a runtime dependency for the host, while it is meant to be run on a target device or emulation?

IMO we need to find a way to exclude specific libs from the reverse dependency scanning happening in `04-generate-runtime-deps.sh`. Then adding `libperfa.so` to such a list should result in a useable list of runtime dependencies.

Pretch:
> IMO we need to find a way to exclude specific libs from the reverse dependency scanning happening in 04-generate-runtime-deps.sh. Then adding libperfa.so to such a list should result in a useable list of runtime dependencies.

Besides `04-generate-runtime-deps.sh` does not test shared library architecture, should it?

(Just checked: Android Studio 3.1 released already)

ebfe:
Yes, it doesn't check architecture. Yes it should. Current hack to work around bogus dependencies is to add an common/shlibs entry (see 7cece6a8dd3d0b079f1aa4f9e0e8130a8fa0a706 for example)

pullmoll:
Got it:
```
=> /opt/android-studio/plugins/android/resources/perfd/armeabi-v7a/perfd depends on libc.so
=> /opt/android-studio/plugins/android/resources/perfd/x86/perfd depends on libc.so
=> /opt/android-studio/plugins/android/resources/perfd/arm64-v8a/perfd depends on libc.so
```

Pretch:
@pullmoll, is it okay to use filename only? Perhaps it is better to use glob or full path (in case `a/libwhatever.so` needed but not `b/libwhatever.so`)?

pullmoll:
@Pretch  I just wanted to write that down :)
Full path would be the safest, yes.

I think we could add this to the the `common/…` scripts, even if it's just another hack.
IMO better than modifying templates to entirely skip the dependency scanning and fail on the next update due to a `depends=…` entry missing.

Pretch:
@pullmoll, so how exactly we will do that?

pullmoll:
@Pretch: it's there and sitting in the pipe. I could install my locally built android-studio-3.0.1_2 here. The builders are currently stalled, so it will take some time.

Now I need to document `skiprdeps` in the `Manual.md` file...

pullmoll:
You're welcome. I uploaded my locally built xbps to https://repo.voidlinux.de/current/nonfree in case you want to try using `xbps-install -R <thatrepo>`.

PR #12800, opened 2018-03-22
qt5: fix webengine on musl

Johnnynator:
I just copied the musl-sandbox.patch from chromium.

[ci skip]


pullmoll:
I'll also add back a webengine depedency now that libvpx-devel is >=1.7.0.
Built locally for all possible target architectures and now building again with libvpx.

pullmoll:
It's working fine in a VM with qupzilla. Let's hope the builders can take the hit :P

PR #12802, opened 2018-03-22
browserpass: update to 2.0.16.

maxice8:


Johnnynator:
Are you ever reading the changelog of browserpass?
Half of your browserpass version bumps are not really needed.

>  There are no changes to the host app.
>
>     Feel free to keep using the previous version.

PR #12806, opened 2018-03-22
New package: leocad-18.02

teldra:


srcpkgs/leocad/template
```diff
@@ -0,0 +1,23 @@
...
+build_style=qmake
+hostmakedepends="unzip"
+makedepends="qt5-tools-devel"
+depends="hicolor-icon-theme qt5"
+short_desc="A CAD program for creating virtual LEGO models"
+maintainer="teldra <teldra@rotce.de>"
+license="GPL-2"
+homepage="http://leocad.org"
+distfiles="https://github.com/leozide/${pkgname}/archive/v${version}.tar.gz https://github.com/leozide/${pkgname}/releases/download/v${version}/Library-Linux-${_lib_version}.zip"
+checksum="496567a9f05930b3366b38053b6de4f2c9d917280079eeb45d08581784903ca1 55063566f8c887b04f280e32aa138fee2388c267bfc5e7b0f51d5cdcff54c9b4"
```
teldra: @the-maldridge here complaints xlint. maybe because its a zip from github. but thats the source.

Johnnynator: pls put only one distfile per line.

Also that warning is mainly about using the automated .zip instead of the automated .tar.gz archive and not about some arbitrary zip that is needed with some additional files. So that xlint warning can be safely ignored.

srcpkgs/leocad/template
```diff
@@ -0,0 +1,26 @@
+# Template file for 'leocad'
+pkgname=leocad
+version=18.02
+revision=1
+_lib_version=11331
+build_style=qmake
+hostmakedepends="qt5-tools-devel unzip"
+makedepends="qt5-devel libGL"
+depends="qt5-tools hicolor-icon-theme qt5"
```
Johnnynator: qt5 does get automatically detected by xbps-src and I would be surprised if qt5-tools is really a needed runtime dependency, you most likely can remove it.

srcpkgs/leocad/template
```diff
@@ -0,0 +1,26 @@
...
+short_desc="A CAD program for creating virtual LEGO models"
+maintainer="teldra <teldra@rotce.de>"
+license="GPL-2"
+homepage="http://leocad.org"
+distfiles="https://github.com/leozide/${pkgname}/archive/v${version}.tar.gz \
```
Johnnynator: You don't have to escape a new line inside of quotes.

Johnnynator:
I would also recommend adding `configure_args="DISABLE_UPDATE_CHECK=1"`

srcpkgs/leocad/template
```diff
@@ -0,0 +1,27 @@
+# Template file for 'leocad'
+pkgname=leocad
+version=18.02
+revision=1
+_lib_version=11331
+build_style=qmake
+hostmakedepends="qt5-tools-devel unzip"
+makedepends="qt5-devel libGL"
+depends="hicolor-icon-theme"
```
the-maldridge: Does it actually depend on hicolor specifically, or just an icon theme in general?

the-maldridge:
Not many comments, you've already gotten feedback that solved a lot of issues.  Your travis checks are still failing due to whitespace issues, so you'll need to fix those.

srcpkgs/leocad/template
teldra: i found a PKGBUILD from upstream dev in sorces. there are no depends on icons

the-maldridge:
:shipit:

PR #12809, opened 2018-03-22
volctl: change maintainer

teldra:
was a temporary email


the-maldridge:
You may wish to also amend the git mailmap file, so that github and other git tools understand the switch if your email you commit with has changed.

PR #12815, opened 2018-03-23
atom: update to 1.25.0

wpbirney:


wpbirney:
failed... will dig into it

wpbirney:
also removed the check for the package-lock as that is built into atoms build system now

PR #12860, opened 2018-03-23
blender: update to 2.79b.

jnbr:


jnbr:
Travis doesn't like the amount of warnings spilled by the blender build. Local builds for x86_64, x86_64-musl and i686 succeed.

srcpkgs/blender/template
```diff
@@ -15,7 +15,7 @@ maintainer="Enno Boland <gottox@voidlinux.eu>"
 license="GPL-2"
 homepage="http://blender.org"
```
cr6git: `https://www.blender.org`

cr6git:
In [source/](http://download.blender.org/source/) there is a new version dated yesterday:
````
blender-2.79b.tar.gz | 2018-Mar-23
````

jnbr:
I'll update this PR to `2.79b` when there is a official release for it.

jnbr:
> I'll update this PR to 2.79b when there is a official release for it.

Done

PR #12862, opened 2018-03-24
neomutt: update to 20180323.

maxice8:


Vaelatern:
No!
They fixed cross building. You didn't.
I'll fix it. Don't touch it.

PR #12864, opened 2018-03-24
firefox: fix i686 build by reducing debug info. [CI Skip]

jnbr:
Building firefox on i686 currently fails because rustc runs out of memory when building gkrust  [build_log](https://build.voidlinux.eu/builders/i686_builder/builds/8396/steps/shell_3/logs/stdio).
Reducing debug info reduces rustc's memory consumption which makes the build succeed on i686.


jnbr:
problem is solved by updating firefox to 59.0.1.

PR #12873, opened 2018-03-24
firefox-esr: update to 52.7.3.

cr6git:


maxice8:
don't forget about `xtrimdeps` ;)

Gottox:
:+1: Let's see how ff-59 builds and update esr afterwards.

cr6git:
@maxice8 I trimmed even more.

cr6git:
@Gottox Upated to the latest release.

srcpkgs/firefox-esr/template
```diff
@@ -25,19 +24,12 @@ makedepends="
 depends="nss>=3.27 desktop-file-utils hicolor-icon-theme"
 conflicts="firefox>=0"
 
-build_options="alsa dbus gtk3 rust pulseaudio startup_notification xscreensaver"
```
Gottox: Why remove rust?

cr6git: Useless.

Gottox: why useless?

Gottox: I'm not a fan of disabling rust if there's no need to. Rust is basicly a security feature of firefox. See https://hacks.mozilla.org/2016/07/shipping-rust-in-firefox/

cr6git: I currently run this version compiled locally on x86_64-musl, it works flawlessly.
For reference: in [FreeBSD](https://www.freshports.org/www/firefox-esr) `RUST=of`

cr6git: @Gottox for security matters? oh! it'a okay, I will amend my commit. :innocent:

cr6git: @Gottox I asked the same question to **asergi** [here](https://github.com/voidlinux/void-packages/pull/6760#issuecomment-388053697)... Why rust is enabled only for x86_64*? Not available for 32-bit? Thanks.

cr6git: I found the solution.

PR #12879, opened 2018-03-24
an: rebuild missing package in repos, fix template

cr6git:
For whatever reason this package is missing in repos:
````
$ sudo xbps-install -S an
[*] Updating `https://repo.voidlinux.eu/current/musl/x86_64-musl-repodata' ...
Unable to locate 'an' in repository pool.
````

### CODE REVIEW

**Comment by Duncaen** [comment]
Its not necessary to increase the revision if you just want to trigger the build.

File: srcpkgs/an/template [hunk 1]
```diff
@@ -3,7 +3,7 @@ pkgname=an
 version=1.2
 revision=1
 build_style=gnu-makefile
-make_install_args='INSTALLDIR=$(DESTDIR)/usr/bin'
+make_install_args="INSTALLDIR=$(DESTDIR)/usr/bin"
```
ebfe: This doesn't work and messes with the masterdir. Also see package content.

cr6git: @ebfe I'm sure your comment makes sense, but I don't get it. I don't see what is wrong here?

Johnnynator: Hes probably saying that (DESTDIR) should be ${DESTDIR}.

cr6git: @Johnnynator oh!! :hushed: good catch! I just noticed it.

cr6git: @Johnnynator it's even worse: now the binary is missing in the resulting package??

Johnnynator: Actually $() is the way GNU Make handles variables. So it was correct. You have to use `'` instead `"` so the $() doesn't get executed.

Johnnynator: As you can see in the travis log
```
=> an-1.2_1: running post-install hook: 98-lib32 ...
/void-packages/srcpkgs/an/template: line 6: DESTDIR: command not found
```

ebfe: http://wiki.bash-hackers.org/syntax/quoting `'` vs '`"`

cr6git: I've found another solution: `sed` this wrong Makefile.
Also, hostmakedepends="icu-devel" wasn't needed, I will remove it.
I think it's necessary to bumb the revision then...

PR #12880, opened 2018-03-24
urlwatch: update to 2.9.

cr6git:


maxice8:
```
depends="python3-appdirs python3-keyring python3-minidb python3-requests
 python3-yaml"
```

also license is 'BSD-3-Clause'

PR #12881, opened 2018-03-24
virtualbox-ose: updated to 5.2.8

alekseyt:
virtualbox-ose-guest and virtualbox-ose-guest-dkms can't build modules currently (i have kernel 4.15.12_1), DKMS build fails. I've bumped version of virtualbox-ose, seems to be fixed now, i was able to install packages, DKMS build was successful, display and shared folders work.


alekseyt:
```
The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).

The job has been terminated
```

Guess it was a bad idea to submit a pull request.

Hoshpak:
If the build for x86_64 succeeds for you, there's nothing to worry about. Travis always terminates builds when they are running for too long or produce too much output, it doesn't indicate that something is wrong. In fact, I run into this issue for most of the PRs I open. Your build looks fine until Travis forcefully terminates it.

Issue #5930, opened 2017-03-08
Updating rpi-kernel to support aarch64 - guidance needed


E100Beta:
Hello!

### Problem
Some of Raspberry Pi 3 hardware (most notably Wi-Fi) does not work if using `aarch64` version of Void. 

It works on `armv7`, which uses kernel from Raspberry Pi Github `rpi-kernel`, while `aarch64` uses regular kernel.

### Solution
I tried to:
* Update `rpi-kernel` to 4.10.1 (from 4.4.47).
* Add support for `aarch64` to it.
* Make temlate look more like current `linux4.10` template (this is still in progress).

I successfully updated and installed it, wifi works, I did not encounter any problems as of yet.

### Questions
The kernel is an important part of the system, and switching to this new kernel requires modifying `config.txt`, aaaand I am a newbie with experience of two PRs so far, so I wanted to ask some things before I make a PR:
* Are there packages that need to be updated with a kernel?
* Is there a smart way to modify current process of kernel configuration on RPi3?  In particular substitution of `config.txt` file.
* Is there a good reason why `aarch64`'s way of handling kernel and firmware stuff is different from that of `armv7` and `armv6`?

Thanks in advance for any help and information!

pullmoll:
I think that aarch64 kernels may actually be used on servers or desktops with a *normal* (read: BIOS, UEFI or coreboot like) booting scheme, i.e. not the somewhat special u-boot scheme which many tinker boards like the RPIs, Ordroid, BeagleBone, etc. use. These boards don't use /boot partition, but have special devices (NAND flash or similar) where the kernel and sometimes rootfs are loaded from.

The armv6 and armv7 kernels are not too close to mainstream, rather based on kernel forked quite some time in the past, which is now plastered with special drivers and devices for handling NAND flash, WiFfi or bluetooth devices, voltage and clock controllers and, most importantly, the GPIO stuff which these boards usually have and which needs to be initialized very early.

You can have your own configurations for the kernels and use them instead of the default. This is described somewhere in the templates IIRC.

I don't know if kernel 4.10.x has sufficient support to bring up a RPI3 and if so, how this is done - with or without u-boot initializing the hardware first. It looks like I'm not the only non-professional in this area taking the number of replies for a measure ;-P

[...]

PR #12889, opened 2018-03-25
Rpi-kernel: add proper aarch64 support for RPi3 via foundation's kernel (wifi works!)

Piraty:
## What
Provide proper support for RPi3 in 64bit mode (aarch64), by using RaspberryPi foundation's [kernel tree](https://github.com/raspberrypi/linux) instead of the currently used [mainline kernel](https://github.com/voidlinux/void-packages/blob/69beef74c30412c4a742fada10d753d87004b8b4/srcpkgs/rpi3-base/template#L11).
Here i describe what i think is neccesary for the transition from current aarch64 mainline kernel to the raspberrypi foundation's kernel.

## Motivation
The motivation is to get a better working RPi3 board in 64bit mode, the most notably desired feature would probably be wifi (which is not working with the current mainline kernel provided by Void Linux).
Others do it already, for example [Gentoo](https://wiki.gentoo.org/wiki/Raspberry_Pi_3_64_bit_Install).
Since I already [upstreamed the necessary configuration options](https://github.com/raspberrypi/linux/pull/2425) for wifi to work by default, that is now an easy task.

## How it works
- Compile the foundation's kernel for aarch64 by extending current `rpi-kernel` template.
- Use recent firmware for it to work well
- Extend `rpi-base`accordingly
- No need for `arm_control=0x200` or a special `kernel=<file>` line in config.txt, because if `kernel8-img` is found on the vfat partition, the board will go into 64bit mode automatically.
- (extra:) extend void-mklive so that proper rpi3 images are offered by Void Linux (see: ).

## How to test
1. clone this PR / my branch.
1. crossbuild `rpi-base` and `rpi-kernel` (and `rpi-firmware`) for `aarch64` or `aarch64-musl`
1. adapt [void-mklive](https://github.com/voidlinux/void-mklive) so it actually pulls the packages you built in the previous step from your local hostdir/binpkgs repo. Make sure to patch mkplatformfs.sh.in to use `rpi-base`instead of `rpi3-base`.
1. (in void-mklive) run `make void-aarch64(-musl)-ROOTFS-<YYYYMMDD>.tar.xz`
1. (in void-mklive) run `make void-rpi3(-musl)-PLATFORMFS-<YYYYMMDD>.tar.xz`
1. expand the resulting tarball onto you sdcard and boot up the pi3 in 64bit mode with working wifi.

## What has been tested
- [x] Wifi
- [x] `/sys/devices/virtual/thermal/thermal_zone0/temp`
- [x] analogue Sound
- [x] Framebuffer
- [ ] ...?

## Transition
- Users running the previously provided rpi3-images with mainline aarch64 variant on their RPi3 should be ok, because they have rpi3-base. A revbump and a deprecation notice could be added if desired.
- `vkpurge` got aarch64 support to handle the kernel collection of mainline kernel. `rpi-kernel`s don't pile up as they are replaced on every update, so that may be looked into...

## Feedback
I would like feedback from maintainers and users as well. I spent some time in the system already, and besides some software i am used to (currently) not being available on aarch64 repos, i didn't find issues that bother me.
I may post `dmesg`'s output, so some driver / system issues can be sorted out, but i can't do that on my own, so i will be happy about feedback.
This probably solves https://github.com/voidlinux/void-packages/issues/5930 .


chneukirchen:
I don't have rpi3 hardware handy to test this, but I guess it sounds useful.

Not sure why the mklive changes were merged before this, but :+1:

chneukirchen:
Also should rpi-kernel track the current stable kernel version?

Piraty:
Just for reference, we track the latest Raspberry Pi LTS kernel, see f87bd2f094cc9ebcd1418e977c95782ea3f99b71.

Issue #9481, opened 2017-11-23
notepadqq x86_64-musl broken


cr6git:
### System

* xuname:  `Void 4.13.15_1 x86_64-musl GenuineIntel uptodate rrnFFFFFFFFFF`

* package:  `notepadqq-1.2.0_1`

### Expected behavior
works seamlessly on `x86_64` , should work on `x86_64-musl` too. 
### Actual behavior
segfault. 
````
$ notepadqq
Missing icon:  "go-next"
Missing icon:  "go-previous"
Segmentation fault
````
😢

**Comment by cr6git** [comment]
I just tried to rebuild the package on my system, here's the [complete log](https://gist.github.com/cr6git/0bbbc91f8cd7d3bf5b30b7b08f461299) ... 
Well, it doesn't make it any better.

**Comment by newbluemoon** [comment]
Is your musl system up-to-date? Perhaps this is due to the recent Qt5 issue? Qt5 revision 10 was just released.

**Comment by newbluemoon** [comment]
Do you have the icons on your musl system?

    find /usr/share/icons -name "go-next.*"

**Comment by cr6git** [comment]
Yes, icons are there.

**Comment by Johnnynator** [comment]
Do other qt5 applications work for you?

**Comment by ebfe** [comment]
```
#0  0x00007fa54aee499c in WTFCrash () from /lib/libQt5WebKit.so.5
#1  0x00007fa54acf58a5 in JSC::Interpreter::execute(JSC::ProgramExecutable*, JSC::ExecState*, JSC::JSObject*) () from /lib/libQt5WebKit.so.5
#2  0x00007fa54ae030c7 in JSC::evaluate(JSC::ExecState*, JSC::SourceCode const&, JSC::JSValue, JSC::JSValue*) () from /lib/libQt5WebKit.so.5
#3  0x00007fa549a60c5f in WebCore::ScriptController::evaluateInWorld(WebCore::ScriptSourceCode const&, WebCore::DOMWrapperWorld*) ()
   from /lib/libQt5WebKit.so.5
#4  0x00007fa549a60f43 in WebCore::ScriptController::evaluate(WebCore::ScriptSourceCode const&) () from /lib/libQt5WebKit.so.5
#5  0x00007fa54aa4a8c8 in WebCore::ScriptElement::executeScript(WebCore::ScriptSourceCode const&) [clone .part.1192] () from /lib/libQt5WebKit.so.5
[...]

PR #12927, opened 2018-03-26
notepadqq: broken on musl, Orphaned.

cr6git:
[#9481](https://github.com/voidlinux/void-packages/issues/9481)

I will not maintain a package that I can't even use! :frowning_face:


Gottox:
Wise choice. I need to learn that too :smile:

cr6git:
@Gottox Well, I spent the night reading articles about C++, and so Qt5 ... And from what I read, the choice of C++ for developping a Desktop Environment **was not wise**...

cr6git:
However, I learned that the term "Template" comes from C++.

PR #12944, opened 2018-03-27
perl-IPC-Run: update to 0.97.

maxice8:


srcpkgs/perl-IPC-Run/template
```diff
@@ -1,16 +1,16 @@
...
-version=0.96
+version=0.97
 revision=1
 wrksrc="${pkgname/perl-/}-${version}"
 build_style=perl-module
 hostmakedepends="perl"
 makedepends="${hostmakedepends}"
 depends="${makedepends}"
-noarch=yes
 short_desc="system() and background procs w/ piping, redirs, ptys"
```
cr6git: `srcpkgs/perl-IPC-Run/template:10: short_desc should start uppercase`

PR #12945, opened 2018-03-27
python-httplib2: update to 0.11.1.

maxice8:


srcpkgs/python-httplib2/template
```diff
@@ -12,7 +12,7 @@ maintainer="Juan RP <xtraeme@voidlinux.eu>"
 homepage="https://github.com/httplib2/httplib2"
 license="MIT"
```
cr6git: `srcpkgs/python-httplib2/template:13: license 'MIT', but no use of vlicense`

PR #12946, opened 2018-03-27
firefox: update to 59.0.2.

cr6git:


cr6git:
I can build the package locally, but Travis fails for some unknown reason...

PR #12950, opened 2018-03-27
gtkd: refactor into separate libraries and add libgtkdgl.

maxice8:
What this PR does.

- Separates each library into it's own subpkg, with libgtkd-3.so staying on the main package, all subpkgs depend on gtkd but each subpackage has it's own set of dependencies.
- keep only gtkd-devel which requires all libraries.
- add `libgtkdgl`
- use SPDX identifier on license.

thanks to the rpm.spec file which showed me i was missing gtkdgl and dissuading me from creating a -devel package for each library.


maxice8:
oh yeah i also added tests.

PR #12965, opened 2018-03-27
Upgrade to latest release

unixsheikh:
I have upgraded the terminal package and have successfully build and tested it on x64.


maxice8:
Use the proper commit style.

`<pkgname>: update to <version>.`

srcpkgs/terminator/template
```diff
@@ -1,18 +1,18 @@
...
 pkgname=terminator
-version=1.0
-revision=2
+version=1.91
+revision=1
 build_style=python2-module
 hostmakedepends="python intltool"
 makedepends="vte-python"
-depends="desktop-file-utils vte-python python-psutil"
+depends="desktop-file-utils vte-python python-psutil vte3"
```
maxice8: Seems like something to be added to makedepends and be found directly by shlib-depends

unixsheikh: Yes, sorry, new to Void :)

maxice8: no need to actually change, terminator is a python terminal so we can't use shlib depends.

Gottox:
Thanks for your contribution!

Unfortunately terminator 1.91 is the unfinished development version and I can't see a good reason for merging it. I'd like to wait till the actual 2.0 release is out.

unixsheikh:
There is a bug in the current version which messes up neovim and possibly also vim (I haven't tested that). The new version fixes that problem.

void-terminator

PR #12976, opened 2018-03-28
fontconfig: update to 2.13.0.

maxice8:


srcpkgs/fontconfig/template
```diff
@@ -26,13 +27,21 @@ post_install() {
...
 		vmove usr/lib/pkgconfig
-		vmove usr/lib/*.so
 		vmove usr/share/man/man3
+		vmove "usr/lib/*.so"
+		vmove usr/share/gettext
+	}
+}
+
+fontconfig-doc_package() {
+	short_desc="Library for configuring and customizing font access - documentation"
```
pullmoll: Probably noarch=yes

common/shlibs
```diff
@@ -108,7 +108,7 @@ libgdk_pixbuf-2.0.so.0 gdk-pixbuf-2.22.0_1
 libgdk_pixbuf_xlib-2.0.so.0 gdk-pixbuf-xlib-2.30.8_1
 libgailutil.so.18 gtk+-2.16.0_1
 libfreetype.so.6 freetype-2.3.9_1
-libfontconfig.so.1 fontconfig-2.6.0_1
+libfontconfig.so.1 fontconfig-2.13.0_1
```
pullmoll: This isn't required. Is it for cosmetics or because you separted the -doc subpackage?

srcpkgs/fontconfig/template
```diff
@@ -26,13 +27,22 @@ post_install() {
 }
 
 fontconfig-devel_package() {
-	depends="${makedepends} ${sourcepkg}>=${version}_${revision}"
-	short_desc+=" - development files"
+	short_desc="Library for configuring and customizing font access - development files"
```
pullmoll: Why do you repeat the original short_desc instead of adding to it?

srcpkgs/fontconfig/template
```diff
@@ -26,13 +27,22 @@ post_install() {
...
-		vmove usr/lib/*.so
 		vmove usr/share/man/man3
+		vmove "usr/lib/*.so"
+		vmove usr/share/gettext
+	}
+}
+
+fontconfig-doc_package() {
+	noarch=yes
+	short_desc="Library for configuring and customizing font access - documentation"
```
pullmoll: Dito: why not `short_desc+=" - documentation"`?

common/shlibs
maxice8: Problably leftovers, i'll remove.

PR #12977, opened 2018-03-28
meld: update to 3.19.0.

maxice8:


Gottox:
3.19 is the dev version. Are there any reasons why we shouldn't wait for the stable?

maxice8:
On Wed, Mar 28, 2018 at 07:38:19AM +0000, Enno Boland wrote:
> 3.19 is the dev version. Are there any reasons why we shouldn't wait for the
> stable?
> 
> —
> 

my bad, didn't check.

PR #12980, opened 2018-03-28
vscode: update to 1.22.1

julio641742:


srcpkgs/vscode/template
```diff
@@ -5,9 +5,9 @@ revision=1
 only_for_archs="i686 x86_64"
 nocross=yes
 nostrip=yes
-hostmakedepends="pkg-config python nodejs"
-makedepends="libX11-devel libxkbfile-devel libsecret-devel"
-depends="GConf libXtst libxkbfile nss"
+hostmakedepends="nodejs pkg-config python"
+makedepends="libsecret-devel libX11-devel libxkbfile-devel"
+depends="GConf libxkbfile libXScrnSaver libXtst nss"
```
pullmoll: Ì believe (some of) these should be auto detected.
Take a close look at the `04-generate-runtime-deps` stage output and remove depends which are listed there.

PR #12985, opened 2018-03-28
common/xbps-src: add stacktraces on build errors

Gottox:
This commit makes xbps-src print stacktraces when an error occured:
screenshot from 2018-03-28 11-47-03

### DIFF
--- a/common/xbps-src/shutils/common.sh
+++ b/common/xbps-src/shutils/common.sh
@@ -80,9 +80,18 @@ run_step() {
 }
 
 error_func() {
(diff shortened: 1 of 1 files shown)


ebfe:
Maybe only print the first line by default and add an option for a full stacktrace? This is quite noisy for normal usage (where the reason for an error is usually a compiler error and the xbps-src stacktrace  just adds to the noise)

Gottox:
compressed the output. the default error case has now only two lines:

screenshot from 2018-03-28 21-38-11

PR #12989, opened 2018-03-28
New package: neovim-plug-0.9.1

noblehelm:


Duncaen:
Any reason to not use vim-plug from your users config directory? It even has a mechanism to update itself.

noblehelm:
I can't find anything related to vim-plug in my system, even after installing normal vim. Where can I find it?

Duncaen:
I mean like this https://github.com/junegunn/vim-plug#vim, placing it in `~/.vim/autoload/plug.vim` or `~/.local/share/nvim/site/autoload/plug.vim`.

noblehelm:
Oh, I get it. It was because I had this mindset that it would be better for files not owned by the user to be handled by the package manager, to make it easy to remember what/where are they installed and remove them. But I yeah, I can do as you advised, and I can close the PR you if want.

Duncaen:
Its not my decision, maybe someone @voidlinux/pkg-committers wants it merged.

Vaelatern:
I don't want it merged. Maybe someone else does.

noblehelm:
I'll close this PR. I realized, due to the nature of packages in Void, it is best not to litter the repositories with niches packages like this.

PR #13019, opened 2018-03-29
lowdown: fix musl compilation

maxice8:
Tested on a temporary x86_64-musl container, not thoroughly tested, merge with caution.


Duncaen:
Its not a musl problem, its a problem with the configure script while cross compiling.

maxice8:
manually defined HAVE___PROGNAME on configure.local , works ?

PR #13023, opened 2018-03-29
memcached: update to 1.5.7.

maxice8:


maxice8:
thanks to Alpine Linux people for fixing seccomp musl support.

PR #13051, opened 2018-03-30
iptables: check xtables lock

Nicop06:
This will avoid `iptables-restore` issues if another service is using iptables. This is useful when used with `sshguard` for instance. Without this change, the firewall rules are not successfully restored.


Nicop06:
This change will improve the probability of the restore to work, although the main problem with sshguard is that runit does not have package dependency. In fact, if we have a look at `sshguard` runit script [here](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/sshguard/files/sshguard-socklog/run), we can see that we rely on `sv check` which only checks that the service is scheduled to run (is `supervise` subfolder` has been created) but not that it is running. After looking at runit source code, I did not find any obvious way to be startup dependency, and I believe this is one of the main missing feature compared to other service init system such as openrc.

Nicop06:
This change will improve the probability of the restore to work, although the main problem with sshguard is that runit does not have package dependency. In fact, if we have a look at `sshguard` runit script [here](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/sshguard/files/sshguard-socklog/run), we can see that we rely on `sv check` which only checks that the service is scheduled to run (is `supervise` subfolder` has been created) but not that it is running. After looking at runit source code, I did not find any obvious way to be startup dependency, and I believe this is one of the main missing feature compared to other service init system such as openrc.

srcpkgs/iptables/files/ip6tables/run
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 [ ! -e /etc/iptables/ip6tables.rules ] && exit 0
-ip6tables-restore /etc/iptables/ip6tables.rules
+ip6tables-restore /etc/iptables/ip6tables.rules -w
```
Gottox: What do you think about using `|| exit 1` instead?

This way the process cannot silently fail if the lock is acquired indefinitly.

Nicop06: Do you mean together with the `-w` like that:

```
ip6tables-restore /etc/iptables/ip6tables.rules -w $TIMEOUT || exit 1
```

With `TIMEOUT` defined in `conf` or with a reasonable value such as 3. Without any option it will wait indefinitely for the lock.

Or just `|| exit 1` and rely on `runsv` to restart the service?

Piraty:
maybe don't provide the service at all and advise users to place its contents in `/etc/rc.local`? After all, it's not a real service if just pauses...

Nicop06:
@Piraty This could be a solution. I would even suggest to add a `/etc/rc.d/` folder where every executable file could be run added. This would mimic the behavior of `/etc/runit/core-services` for third party startup scripts.

Nicop06:
I just updated the pull request so that it now wait for 3 seconds and then exit if the lock is not release.

Nicop06:
Repo moved

PR #13052, opened 2018-03-30
New package: gnome-tweaks-3.28.0.

maxice8:
@Gottox seems we have missed gnome-tweak-tool changing into gnome-tweaks, i changed gnome-tweak-tool into a transitive package that depends on gnome-tweaks, users that update gnome-tweak-tool to 3.28.0 will install gnome-tweaks, users that install gnome first time will get only gnome-tweaks.

after sometime we can delete gnome-tweak-tool


srcpkgs/gnome-tweak-tool/template
```diff
@@ -1,17 +1,10 @@
 # Template file for 'gnome-tweak-tool'
 pkgname=gnome-tweak-tool
```
Gottox: make this a subpkg of gnome-tweaks.

maxice8: a meta subpkg ?

Gottox: See upower. (Please not that we should not use provides/replaces here, as it had some sideeffects for upower)

Gottox: You can also leave build_style=meta out, it makes no sense for subpackages.

maxice8: ok done.

Issue #11804, opened 2018-02-22
Manual.md: replaces is not documented


yopito:
usage samples: `srcpkgs/xorg-server/template`, `srcpkgs/yq-go/template`, ...

kiarash96:
Also `provides`

PR #13064, opened 2018-03-30
Manual.md: document Transitional packages

maxice8:
I saw 2 implementations of transitional packages.

1. old package stays but is turned into meta-package that points to new package ( lemonbuddy -> polybar )
2. old package is removed, new subpackage is created with name of old package that points to n ew package ( gnome-tweak-tool -> gnome-tweaks, upower0 -> upower )

i documented regarding 2, also having different implementations of such trivialities is annoying.


maxice8:
i am also not sure if to call them `transitive` or `dummy` or `transitive dummy`

Hoshpak:
I would call them transitional packages as I don't recognize any transitivity there.

maxice8:
On Fri, Mar 30, 2018 at 10:06:07AM +0000, Helmut Pozimski wrote:
> I would call them transitional packages as I don't recognize any transitivity
> there.
> 
> —
> 

yeah, i updated the PR

Gottox:
@voidlinux/pkg-committers I like this approach. comments?

Manual.md
```diff
@@ -1227,6 +1228,16 @@ The following variables influence how Haskell packages are built:
 - `make_build_args`: This is passed as-is to `stack build ...`, so
   you can add your `--flag ...` parameters there.
 
+<a id="pkgs_transitional"></a>
+### Transitional packages
+
+Transitional packages are used to move users into a new package. It is commonly
+used when upstream changes it's name and Void Linux changes the package name to
+reflect such change. A transitive package must be an empty subpackage of the
```
ebfe: s/transitive/transitional/

Manual.md
```diff
@@ -1227,6 +1228,16 @@ The following variables influence how Haskell packages are built:
...
+package of the new main package, on the subpackage, it must have `noarch=yes`, 
+`short_desc+"= - transitional package"` and
+`depends="${pkgname}-${version}_${revision}"`
```
ebfe: s/pkgname/sourcepkg/

maxice8:
Thanks for catching, will fix when I get home in about 1 hour

maxice8:
Will fix whenever I get home eta 1h

ebfe:
Maybe it would be good to indentify/explain why these are needed at all. Why doesn't replaces=(/provides=) work? Also see #11804.

These transitional packages have the problem that you never know when they can be safely removed, so i'm not sure if they should be encouraged.

Manual.md
```diff
@@ -1227,6 +1228,16 @@ The following variables influence how Haskell packages are built:
 - `make_build_args`: This is passed as-is to `stack build ...`, so
   you can add your `--flag ...` parameters there.
 
+<a id="pkgs_transitional"></a>
+### Transitional packages
+
+Transitional packages are used to move users into a new package. It is commonly
+used when upstream changes it's name and Void Linux changes the package name to
```
newbluemoon: I think it must be “its name”.

maxice8:
As far as I can tell  replaces provides doesn't update people with the old
package

Manual.md
```diff
@@ -1227,6 +1228,16 @@ The following variables influence how Haskell packages are built:
...
+reflect such change. A transitive package must be an empty subpackage of the
+package of the new main package, on the subpackage, it must have `noarch=yes`,
```
newbluemoon: Is there a “package” too much? And maybe split into two sentences for better readability?

ebfe:
> As far as I can tell replaces provides doesn't update people with the old package

But that's exactly what replaces is for. replaces=oldname>=0 and revbumping all revdeps so they refer to the new name should take care of that. As far as i understand it the only reason to use transitional packages is to avoid the revbump of the revdeps. But all in all replaces should be the "default" rename strategy as it also takes care of removing the old package from the system and doesn't clobber the repo with dummy packages.

maxice8:
oh, i wasn't aware of that (no docs), would be really good if this knowledge was added to the docs.

maxice8:
@ebfe using replaces= also require changes to the package that is going to be removed ?

Duncaen:
Transitional packages are good to replace manually installed packages without rdevs.
If nothing pulls in a package into the transaction it would never be considered for `replaces=`.
The problem with them is that transitional packages would replace the manually installed package and install the package its replaced by as automatically installed.

ebfe:
> If nothing pulls in a package into the transaction it would never be considered for replaces=.

If that's the case then maybe that should be fixed instead of adding transitional packages.

ebfe:
> > If nothing pulls in a package into the transaction it would never be considered for replaces=.
> 
> If that's the case then maybe that should be fixed instead of adding transitional packages.

Which of course isn't trivial to do since it's used in combination with
provides for virtual packages which relies on this
behaviour. And is at the same time broken by it ((releated bug
https://github.com/voidlinux/xbps/issues/257). Which leads us back to #11804.

Manual.md
```diff
@@ -1227,6 +1228,16 @@ The following variables influence how Haskell packages are built:
 - `make_build_args`: This is passed as-is to `stack build ...`, so
   you can add your `--flag ...` parameters there.
 
+<a id="pkgs_transitional"></a>
+### Transitional packages
+
+Transitional packages are used to move users into a new package. It is commonly
+used when upstream changes it's name and Void Linux changes the name to reflect
```
pullmoll: s/it's/its/

maxice8: fixed,, thanks

maxice8:
soooo, what is the current plan ? document replaces= / provides= and make Transitional packages documentation just explain how to use those to make a transitional package ?

ebfe:
I think it became clear that renaming packages isn't a 'triviality' so I propose:

```markdown
# Renaming packages

Don't do it unless you know what you are doing.
```

until replaces/provides semantics are clear and implemented in a sane way in xbps. (only half joking)

maxice8:
> # Renaming packages
>
> Don't do it unless you know what you are doing.

Nobody does.

I'll shelf this until xbps people sort it out.

Issue #13068, opened 2018-03-30
Gajim package requires setuptools


dimaburmistrov:
### System

* xuname:  
  Void 4.15.12_1 x86_64 GenuineIntel uptodate r
* package:  
  ii gajim-1.0.0_1                           Full featured Jabber/XMPP client

### Expected behavior
Gajim is running after launching it. Gajim package fullfils all necessary dependencies in its requirements.

### Actual behavior
Failed with exception:

`Traceback (most recent call last):
  File "/usr/bin/gajim", line 6, in <module>
    from pkg_resources import load_entry_point
ModuleNotFoundError: No module named 'pkg_resources'`

### Steps to reproduce the behavior
Install minimal system without setuptools for Python 3.

After installing `python3-setuptools-39.0.1_1.noarch.xbps` works as expected.

dimaburmistrov:
I think this should fix the issue: https://github.com/voidlinux/void-packages/pull/13071

PR #13071, opened 2018-03-30
Add missing dependency python3-setuptools for Gajim

dimaburmistrov:
Fix for issue #13068


srcpkgs/gajim/template
```diff
@@ -5,7 +5,7 @@ revision=1
 noarch=yes
 build_style=python3-module
 hostmakedepends="python3 python3-setuptools"
```
maxice8: The dependencies can be minimized to 

```
depends="python3-gobject python3-nbxmpp python3-pyasn1 python3-setuptools"
```

dimaburmistrov:
Updated to satisfy comments.

PR #13086, opened 2018-03-30
opera: update to 52.0.2871.64.

cr6git:


srcpkgs/opera/template
```diff
@@ -12,14 +13,12 @@ maintainer="Diogo Leal <diogo@diogoleal.com>"
 license="Proprietary"
 homepage="https://www.opera.com/computer"
 distfiles="https://get.geo.opera.com/pub/opera/desktop/${version}/linux/${pkgname}-stable_${version}_amd64.deb"
-checksum=d97c56b61a27e1cfb1be02cc75c8b8da72f61bfd4be174f91aef2352aa6944df
+checksum=ebbb1520a8951597218db6448cbdbb2beb2b67012c907cbf0f81db54aa2e2e99
 
-do_extract() {
```
ebfe: What's wrong with doing the extraction in do_extract?

cr6git: Nothing wrong. however, taking advantage of the `build_style=fetch` we can reduce this line:
````
ar x ${XBPS_SRCDISTDIR}/${pkgname}-${version}/${pkgname}-stable_${version}_amd64.deb
````

To this:
````
ar x ${pkgname}-stable_${version}_amd64.deb
````

cr6git: @ebfe it's okay, I will change that and update to the latest release! :+1:

PR #13088, opened 2018-03-30
nnn: trim down dependencies, fix short_desc & license

cr6git:


srcpkgs/nnn/template
```diff
@@ -4,8 +4,8 @@ version=1.7
 revision=1
 build_style=gnu-makefile
 hostmakedepends="pkg-config"
-makedepends="ncurses-devel readline-devel"
-short_desc="A lightweight terminal file browser"
+makedepends="readline-devel"
+short_desc="The missing terminal file browser for X"
 maintainer="Dennis Kraus <kraus@posteo.de>"
 license="BSD"
```
maxice8: Whie you're at it also use the SPDX identifier for the license.

srcpkgs/nnn/template
```diff
@@ -4,8 +4,8 @@ version=1.7
 revision=1
```
maxice8: needs revbump

PR #13089, opened 2018-03-30
Further kde4 removal

ebfe:
@Gottox do you still care about autokey? (depends on kde-python)
@Hoshpak do you still care about knemo? (depends on kde-workspace)


Hoshpak:
@ebfe knemo is only relevant in the context of a KDE4 desktop, feel free to remove it.

PR #13093, opened 2018-03-31
nsjail: update to 2.5, don't git clone, fix musl, install manpage

mvf:


srcpkgs/nsjail/template
```diff
@@ -1,28 +1,36 @@
 # Template file for 'nsjail'
 pkgname=nsjail
-version=2.4
-revision=2
+version=2.5
+revision=1
 build_style=gnu-makefile
-hostmakedepends="libnl3-devel git pkg-config protobuf-c bison flex"
-makedepends="libnl3-devel libcap-devel protobuf-devel"
+hostmakedepends="bison flex pkg-config protobuf-devel"
```
cr6git: You can omit `bison` which is a dependency of `flex`.

srcpkgs/nsjail/template
```diff
@@ -1,28 +1,36 @@
...
+makedepends="libnl3-devel protobuf-devel"
 short_desc="A light-weight process isolation tool"
 maintainer="Andrea Brancaleoni <abc@pompel.me>"
 license="Apache-2.0"
 homepage="https://github.com/google/nsjail"
```
cr6git: `homepage="http://nsjail.com"`

mvf: Good point, updated.

srcpkgs/nsjail/template
mvf: IMHO that's an implementation detail. The build process calls both, so it should depend on both. `flex` (the program) doesn't really depend on `bison` either, I don't know why the Void package does. Looks like a bug to me.

cr6git: If details are not important it's okay...

cr6git: I'm probably wrong. Make a PR to change this then! :+1:

maxice8:
2.6 is available.

PR #13102, opened 2018-03-31
Update libsrtp to 2.1.0

Hoshpak:


ebfe:
From the release dates this looks like a downgrade:

https://github.com/cisco/libsrtp/releases

ebfe:
Never mind, just wired wording

PR #13105, opened 2018-03-31
New package: kid3-3.6.0

SolitudeSF:
dont really know how i should go about package splitting here. it also has kde build option.


SolitudeSF:
i have no clue how to fix that error

SolitudeSF:
added a hacky workaround to stop ldd from failing on crosscompiling

PR #13106, opened 2018-03-31
New package: lf-r3

TeddyDD:
lf is terminal file manager similar to ranger but simpler
https://github.com/gokcehan/lf


srcpkgs/lf/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'lf'
+pkgname=lf
+version=1
```
cr6git: `version=r1`

TeddyDD: So version does not have to be a number? Okay :)

srcpkgs/lf/template
```diff
@@ -0,0 +1,19 @@
...
+revision=1
+build_style=go
+go_import_path="github.com/gokcehan/${pkgname}"
+wrksrc="${pkgname}-r${version}"
```
cr6git: then you can omit `wrksrc="...`

srcpkgs/lf/template
```diff
@@ -0,0 +1,19 @@
...
+hostmakedepends="git"
+short_desc="Terminal file manager"
+maintainer="Daniel Lewan <vision360.daniel@gmail.com>"
+license="MIT"
+homepage="https://${go_import_path}"
```
cr6git: `homepage="https://godoc.org/github.com/gokcehan/lf`

srcpkgs/lf/template
```diff
@@ -0,0 +1,19 @@
...
+distfiles="${homepage}/archive/r${version}.tar.gz"
```
cr6git: distfiles=`https://github.com/gokcehan/lf/archive/${version}.tar.gz`

srcpkgs/lf/template
cr6git: version is what upstream provides.

srcpkgs/lf/template
cr6git: change that.

srcpkgs/lf/template
cr6git: change that.

srcpkgs/lf/template
TeddyDD: Not sure about that, readme in repo is more informative, has screenshots, links to community chat and list of features.

srcpkgs/lf/template
TeddyDD: Okay, thanks. I saw some packages omniting "v" from versions number and assumed this should be stripped too..

TeddyDD:
@cr6git I'd prefer not to change homepage. Let me know if I should untalgle distfile and homepage into literal adress, without varialbes.

cr6git:
@TeddyDD don't worry, maybe someday a developer will merge your stuff.

TeddyDD:
Three updates so far :rofl:

pullmoll:
15e9121cc826c34459f34b287323c57f37146a9f

PR #13109, opened 2018-03-31
python-tornado: update to 5.0.1.

pbui:
- Python 2 no longer requires certifi
- Python 2 now requires python-futures


asergi:
Tornado 5 introduces backwards-incompatible changes, so it hasn't been updated on purpose:
- jupyter is already compatible with tornado 5
- pyzmq >= 17 is compatible with tornado 5
- mopidy, salt and wpull (abandoned?) are not compatible with tornado 5 yet

TL;DR Tornado will be updated to 5 in due course.

PR #13122, opened 2018-03-31
firefox: use system default locale

wine:
Firefox 59 uses the system language in UI only if the pref `intl.locale.requested` is set to an empty string.

https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Internationalization#Testing_out_your_extension


asergi:
Merged manually in b431f22f6d7d7071752db3e59947b050b4d3b9e8. Thanks.

PR #13130, opened 2018-04-01
New package: gotop-1.2.2

SolitudeSF:


srcpkgs/gotop/template
```diff
@@ -0,0 +1,16 @@
+# Template file for 'gotop'
+pkgname=gotop
+version=1.2.2
+revision=1
+build_style=go
+go_import_path="github.com/cjbassi/gotop"
+short_desc="Terminal based graphical activity monitor"
+maintainer="SolitudeSF <solitudesf@protonmail.com>"
+license="AGPL-3"
```
maxice8: SPDX identifier is AGPL-3.0

PR #13132, opened 2018-04-01
inxi: update to 3.0.07.1 (3.0.07-1), use tarball instead of gitrev.

cr6git:


cr6git:
Updated to the latest release.

PR #13150, opened 2018-04-02
New package: nerd-fonts-1.2.0

dan-santana:


pullmoll:
Whoa! `v2.0.0.tar.gz` is more than `2.0 GiB (2,125,832,713)`.
I'll rather wait for some @voidlinux/pkg-committers to confirm this PR can be merged.

the-maldridge:
Way too big!

For packages of this size you would ideally have a data package as well since this appears to have executables as well as data.

pullmoll:
The package is `noarch=yes` so it will be in the repo once.

dan-santana:
@the-maldridge the scripts are only as helper scripts to be sourced by the shell in order to render the characters using environment variables (e.g. `echo $i_weather_day_cloudy_gusts` => ).

The package is >2G because all fonts are installed by the package. I could split it (one subpackage per font) but it would be ~45 subpackages and make `nerd-fonts` as only a meta-package. Not sure if this amount of subpackages would be any problem, but I'm sure users with low bandwidth surely will be happy with this. What do you think?

the-maldridge:
I am infinitely more likely to entertain this package as a meta-package with subpkgs than as one giant package.

pullmoll:
FWIW the resulting xbps file is 256,314,456 bytes "only" :)
IMO we can merge this hoping the builders can handle it.

Gottox:
I'm with @the-maldridge here.

Also I'm not sure about any licensing issues.

Gottox:
* [Anonymous Pro](https://www.marksimonson.com/fonts/view/anonymous-pro) is OFL, which is a copyleft license.
* [plex](https://github.com/IBM/plex/blob/master/LICENSE.txt) is OFL
* [Fantasque Sans Mono](https://github.com/belluzj/fantasque-sans#author-and-license) is OFL
* Stopped looking, there are more issues, I guess.

Vaelatern:
https://github.com/ryanoasis/nerd-fonts/issues/7

PR #13170, opened 2018-04-02
New package: postgis-2.4.4

ArniDagur:
I built and installed the package on my own machine and it worked fine. This is my first package though, so take a look first before merging.


srcpkgs/postgis/template
```diff
@@ -0,0 +1,14 @@
+# Template file for 'postgis'
+pkgname=postgis
+version=2.4.3
+revision=1
+build_style=gnu-configure
+hostmakedepends="perl"
+makedepends="geos proj libgdal libxml2 json-c postgresql postgresql-libs-devel libxml2-devel geos-devel proj-devel libgdal-devel libgdal-tools"
+depends=""
```
maxice8: remove empty variable.

srcpkgs/postgis/template
```diff
@@ -0,0 +1,14 @@
+# Template file for 'postgis'
+pkgname=postgis
+version=2.4.3
+revision=1
+build_style=gnu-configure
+hostmakedepends="perl"
+makedepends="geos proj libgdal libxml2 json-c postgresql postgresql-libs-devel libxml2-devel geos-devel proj-devel libgdal-devel libgdal-tools"
```
maxice8: -devel packages automatically include the main package.

srcpkgs/postgis/template
```diff
@@ -0,0 +1,14 @@
...
+depends=""
+short_desc="A spatial database extender for PostgreSQL object-relational database"
```
maxice8: I think only `PostgreSQL` fits instead of the whole name

maxice8:
Everything should be in 1 commit.

srcpkgs/postgis/template
```diff
@@ -0,0 +1,13 @@
+# Template file for 'postgis'
+pkgname=postgis
+version=2.4.3
+revision=1
+build_style=gnu-configure
+hostmakedepends="perl"
+makedepends="geos proj libgdal libxml2 json-c postgresql-libs-devel libxml2-devel geos-devel proj-devel libgdal-devel libgdal-tools"
```
maxice8: don't include the package without -devel if you have it as -devel

ArniDagur:
It only seems to build correctly on x86_64 and i686. https://travis-ci.org/voidlinux/void-packages/builds/361444499

maxice8:
Read the build log, it can't find the binary `pg_config` , which is provided by `postgresql-libs-devel`.

Since it is a binary to be ran on the system that builds it needs to be added to `hostmakedepends`

Read https://github.com/voidlinux/void-packages/blob/master/Manual.md and the build logs.

PR #13183, opened 2018-04-03
monero-core: update to 0.12.0.0, rename to monero-gui

Hoshpak:
following upstream's decision to rename the application, provides transitional
monero-core package. Add INSTALL.msg to inform users about the possibility
to either use a remote node or run a local monero node.


wpbirney:
this is definitly needed, was just about to make a PR myself. The current version in the repo is incompatible with any current wallets

PR #13196, opened 2018-04-03
st: fix PIE build

Nicop06:


maxice8:
it seems fine here, am i missing something ?

Duncaen:
And its enabled in our gcc by default iirc.

maxice8:
On Tue, Apr 03, 2018 at 10:31:41PM +0000, Duncan Overbruck wrote:
> And its enabled in our gcc by default iirc.
> 
> —
> 

Running with checksec -f returns all positive.

Johnnynator:
@Nicop06 be sure that your masterdir is up to date (`./xbps-src bootstrap-update`)

PR #13198, opened 2018-04-03
notify-send: add alternatives between libnotify and notify-send.sh

maxice8:


maxice8:
This most likely won't work  since notify-send is the name of the binary.

biopsin:
hm, bash version of dunstify? have to give it a spin..tnx

PR #13199, opened 2018-04-04
alsa stack: update to 1.1.6.

maxice8:


cr6git:
https://travis-ci.org/voidlinux/void-packages/jobs/362297046#L815

`ftp://ftp.alsa-project.org/pub/lib/alsa-lib-1.1.6.tar.bz2: No route to host`

This is only a temporary glitch. Try again, it should work now.

PR #13204, opened 2018-04-04
python-PyQt5: fix pyrcc5 + pyuic5 + pylupdate5 wrappers

newbluemoon:
Unlike python-PyQt4 the cross triplet isn’t present in the wrappers and therefore doesn’t need to be removed. So I removed the removal because similar to #13203 it causes problems in native builds when `$XBPS_CROSS_TRIPLET` is empty (and `sed` finds a match it shouldn’t).

The pyrcc5, pyuic5, and pylupdate5 wrappers use python’s `-m` option to search for the module in the module path (whereas in PyQt4 an absolute path is used). While cross-builds are ok in a native build e.g. `/usr/bin/python3-pyrcc5` currently looks like this (and is broken):

```
#!bin/sh
exec usr/bin/python3 -m PyQt5.pyrcc_main ${1+"$@"}
```

but should be

```
#!/bin/sh
exec /usr/bin/python3 -m PyQt5.pyrcc_main ${1+"$@"}
```


asergi:
Superseded by 922e8adb3b3db27526453fca8a1b2800b9bfcb1a.

PR #13205, opened 2018-04-04
sway: add elogind to wlc in order to support TTY switching in sway

julio641742:
and add xorg-server-xwayland to sway for default
plus this fix #12817


pullmoll:
Please rebase and squash commits into one.

Issue #6879, opened 2017-06-28
RFC: Apparmor/LSM Implementation


coredumb:
Hello,

I open this tracking issue after having discussed the case with @the-maldridge on IRC.
The PR is: https://github.com/voidlinux/void-packages/pull/6009
This is dependant of https://github.com/voidlinux/void-runit/pull/67 and probably https://github.com/voidlinux/void-runit/pull/55

As discussed providing mandatory access control as an option at distribution level requires to provide sane userlands accordingly.
I've thus started to create void specific AppArmor profiles that cover a basic Void installation - This is a work in progress.

I'm not against enabling other LSM in the kernel config but in my opinion it has to make sense in terms of userlands. This means that enabling SELinux in the kernel config would require a certain amount of work to provide userland packages and a solid set of targeted rules.

Cheers,
-O.

CameronNemo:
@olivier-mauras you should coordinate with upstream AppArmor for the fixes and changes you make to apply their default profiles to Void

olivier-mauras:
Got some free time these days to work on that.
Plan is:

- [x] Update linux4.14 config to support both SELinux and AppArmor
- [ ] Ugrade AppArmor userland to latest version from package from previous PR
- [ ] Add SELinux userland packages - So far audit, libsepol, libselinux packages ready
- [ ] Validate best way to integrate with init - custom mountpoint should be sufficient AFAICT

From there, decide how to integrate AA profiles and SELinux rules

olivier-mauras:
For clarity I'll also split the whole thing into 3 PR
- lsm_implementation_kernel
- lsm_implementation_userland_apparmor
- lsm_implementation_userland_selinux

scadu:
It seems working so far. Didn't test docker yet though. Well done!

CameronNemo:
the dnsmasq apparmor profile isn't being applied on my system because the path is `/usr/sbin/dnsmasq` and I am executing dnsmasq from a script and the path is `/usr/bin/dnsmasq`

PR #13231, opened 2018-04-05
Add SELinux and AppArmor in kernel but keep them disabled by default

olivier-mauras:
See https://github.com/voidlinux/void-packages/issues/6879 for details


chneukirchen:
FWIW, Debian seems to compile in both by default, Arch has AppArmor.

olivier-mauras:
Having them enabled without any userland installed is pointless.
It makes more sense to enable them explicitely from the command line when userland is installed.

M3rsy:
How can enable apparmor and Selinux thanks

olivier-mauras:
Right now only AppArmor has working userlands, SELinux is still WIP.

For AppArmor then, update your kernel latest version, install apparmor package. 
Install message will tell you what to do, this should be straightforward.

PR #13236, opened 2018-04-05
neofetch: update to 3.4.0

konimex:


srcpkgs/neofetch/template
```diff
@@ -10,7 +10,7 @@ maintainer="Muhammad Herdiansyah <herdiansyah@netc.eu>"
 homepage="https://github.com/dylanaraps/neofetch"
 license="MIT"
```
maxice8: please put license= above homepage=

konimex: done

PR #13239, opened 2018-04-06
New Package: passff-1.1

teldra:


srcpkgs/passff/template
```diff
@@ -0,0 +1,39 @@
+# Template file for 'passff'
+pkgname=passff
+version=1.1
+revision=1
+_file_id=904741
+wrksrc="${pkgname}-${version}/src"
```
Johnnynator: should probably be `build_wrksrc=src`

PR #13243, opened 2018-04-06
dhcpcd: update to 7.0.3.

maxice8:


maxice8:
On Thu, May 03, 2018 at 08:19:24AM +0000, cr6git wrote:
> Change the title: dhcpcd: update to 7.0.4. 😉
> 
> I hope this will get merged soon.
> 
> —
> 

i won't waste my time changing github titles, it is not port of my workflow and some void linux devs already deal directly with commits.

PR #13244, opened 2018-04-06
spotify: update x86_64 pkg to 1.0.77.

cr6git:
See [pool/non-free/s/spotify-client/](http://repository.spotify.com/pool/non-free/s/spotify-client/)


teh-jazzman:
:+1: I was hoping to just sit out this inconsistency and wait for the next version, but I guess they removed the previous version already.

acobaugh:
Any chance of pushing this through again? I see it was reverted in 0a8b518. Currently no way to install the spotify package on x86_64.

PR #13257, opened 2018-04-07
tekaim: update to 1.6.0.

cr6git:


srcpkgs/tekaim/template
```diff
@@ -1,18 +1,17 @@
...
 revision=1
 noarch=yes
 build_style=python3-module
 hostmakedepends="python3"
-depends="python3 curl maim xclip"
+depends="curl maim python3 xclip"
 short_desc="Simple tool to take and upload screenshots to teknik.io"
 maintainer="cr6git <quark6@protonmail.com>"
 homepage="https://github.com/IvanFon/tekaim"
 license="GPL-3"
```
maxice8: SPDX identifier, and move above homepage

cr6git: Thanks for reminding me.

PR #13259, opened 2018-04-07
New package: xcb-util-errors-1.0

julio641742:


srcpkgs/xcb-util-errors/template
```diff
@@ -0,0 +1,28 @@
...
+version=1.0
+revision=1
+build_style=gnu-configure
+configure_args="--disable-static"
+hostmakedepends="pkg-config m4"
+makedepends="libxcb-devel"
+short_desc="XCB errors library"
+maintainer="Julio Galvan <juliogalvan@protonmail.com>"
+license="MIT"
+homepage="http://xcb.freedesktop.org"
```
maxice8: xcb.freedesktop.org moved to https

PR #13260, opened 2018-04-07
burp2-server: update to 2.2.0

yopito:


yopito:
"latest" version in fact; I regulary update burp this way for a while almost without any trouble since months: small updates of code each month, not such a big deal.      
Last stable is 2.1.32 and previous stable was 2.0.54, that was about with 15 months ago. 
But I might provide packaging for both "channels" for burp if needed.

maxice8:
Void Linux avoids development versions and prefer to stay on stable
branches, see nginx and bind

PR #13290, opened 2018-04-08
New package: simavr-1.6

pienjo:
simavr - a lean and mean Atmel AVR simulator for linux


srcpkgs/simavr/template
```diff
@@ -0,0 +1,45 @@
...
+}
+
+# simavr-devel is a subpackage
+simavr-devel_package() {
+	short_desc+= " - development files"
+	depends="${sourcepkg}>=${version}_${revision} elfutils-devel"
+	pkg_install() {
+		vmove usr/include
+		vmove usr/lib/*.a
+		vmove usr/lib/*.so
```
maxice8: quote shell wildcard statements

srcpkgs/simavr/template
```diff
@@ -0,0 +1,45 @@
...
+	# good thing as it is required at runtime.
+
+	mkdir -p ${DESTDIR}/usr/avr/include
+	mv ${DESTDIR}/usr/include/simavr/avr ${DESTDIR}/usr/avr/include/simavr
+
+	# Also remove the associated pkgconfig file
+	rm ${DESTDIR}/usr/lib/pkgconfig/simavr-avr.pc
+}
+
+# simavr-devel is a subpackage
```
maxice8: unnecessary comment

srcpkgs/simavr/template
```diff
@@ -0,0 +1,45 @@
...
+revision=1
+build_style=gnu-makefile
+make_build_args+="RELEASE=1"
+make_build_target="build-simavr build-parts"
+make_use_env=1
+hostmakedepends="pkg-config"
+makedepends="libelf elfutils-devel glu-devel libfreeglut-devel"
+short_desc="A lean, mean and hackable AVR simulator"
+maintainer="Martijn van Buul <martijn.van.buul@gmail.com>"
+license="GPL-3"
```
maxice8: use SPDX identifier

srcpkgs/simavr/template
pienjo: Which one - the comment at line 35 (which I'll immediately admit is unnecessary, consider it gone), the comment at line 31 (which is probably unnecessary), or the comment of line 22 onward (which might be a bit too verbose, but I feel it has a purpose, as it explains why the package deviates from upstream)

srcpkgs/simavr/template
pienjo: Done. "GPL-3" is the default from xnew; it might be prudent to update that then.

pienjo:
I'll migrate this to the new organisation

PR #13296, opened 2018-04-08
GPaste: update to 3.28.1.

maxice8:


cr6git:
http://www.imagination-land.org/posts/2018-04-13-gpaste-3.28.2-released.html

PR #13306, opened 2018-04-09
cbatticon-gtk3: update to 1.6.8.

maxice8:


cr6git:
https://github.com/void-linux/void-packages/pull/353

PR #13308, opened 2018-04-09
bird: update to 2.0.2.

jnbr:


jnbr:
Travis fails to fetch sources from `ftp://bird.network.cz` reliable, so most builds are failing.
On local machine, builds for all architectures are passing.

jnbr:
closed in favor of #14020

PR #13309, opened 2018-04-09
Add AppArmor userland

olivier-mauras:
See tracking details here: https://github.com/voidlinux/void-packages/issues/6879
Needs #13231


olivier-mauras:
Continuing discussion from #6009 and IRC here.
sv service is there for user convenience to reload all profiles at will.
I had initially thought about using it upon profiles updates, but I think it makes more sense only reloading the updated profile with apparmor_parser from post install/update of packages that provide said profile.

That's why the service is not actually deployed by the package, let's discuss about the best way before merging :)

PR #13314, opened 2018-04-09
yaml-cpp: update to 0.6.2.

Chocimier:


srcpkgs/yaml-cpp/template
```diff
@@ -1,17 +1,17 @@
 # Template file for 'yaml-cpp'
 pkgname=yaml-cpp
-version=0.5.3
-revision=2
-wrksrc=${pkgname}-release-${version}
+version=0.6.2
+revision=1
+wrksrc=${pkgname}-${pkgname}-${version}
```
maxice8: variable should be quoted.

Chocimier:
Done by https://github.com/void-linux/void-packages/pull/3241

PR #13326, opened 2018-04-09
libtorrent-rasterbar: update to 1.1.7.

maxice8:


srcpkgs/libtorrent-rasterbar/template
```diff
@@ -31,12 +32,12 @@ libtorrent-rasterbar-python_package() {
 	}
 }
 libtorrent-rasterbar-devel_package() {
-	depends="${sourcepkg}>=${version}_${revision} libressl-devel boost-devel geoip-devel"
+	depends="${sourcepkg}-${version}_${revision} libressl-devel boost-devel geoip-devel"
```
asergi: Unnecessary in this case.

PR #13328, opened 2018-04-09
abcde: Fixed musicbrainz backend

faulesocke:
These packages are all required to make abcde work with the default configuration.

Another option is to not provide these at all and change the default config to not use musicbrainz. I asked in the IRC what would be the best thing to do and was told I should package the additional perl packages.


srcpkgs/perl-Class-Accessor/template
```diff
@@ -0,0 +1,15 @@
+# Template file for 'perl-Class-Accessor'
+pkgname=perl-Class-Accessor
+_dist=Class-Accessor
```
maxice8: uncessary, just do ${pkgname//perl/}.

srcpkgs/perl-Class-Accessor/template
```diff
@@ -0,0 +1,15 @@
...
+version=0.51
+revision=1
+wrksrc=$_dist-$version
```
maxice8: double quotes required.

srcpkgs/perl-Class-Accessor/template
```diff
@@ -0,0 +1,15 @@
...
+build_style=perl-module
+hostmakedepends="perl"
+depends="perl"
+short_desc="Automated accessor generation"
+maintainer="Urs Schulz <voidpkgs@ursschulz.de>"
+license="GPL-1 PerlArtistic"
```
maxice8: Artistic-1.0-Perl, GPL-1.0-or-later

srcpkgs/perl-Class-Accessor/template
```diff
@@ -0,0 +1,15 @@
...
+homepage="http://search.cpan.org/dist/$_dist/"
+distfiles="$CPAN_SITE/Class/$_dist-$version.tar.gz"
```
maxice8: ditto

srcpkgs/perl-Mojolicious/template
```diff
@@ -0,0 +1,16 @@
+# Template file for 'perl-Mojolicious'
+pkgname=perl-Mojolicious
+_dist=Mojolicious
```
maxice8: See the comments on perl-Class-Accessor, make all changes that also apply.

srcpkgs/perl-MusicBrainz-DiscID/template
```diff
@@ -0,0 +1,17 @@
+# Template file for 'perl-musicbrainz-discid'
+pkgname=perl-MusicBrainz-DiscID
+_dist="MusicBrainz-DiscID"
```
maxice8: See the comments on perl-Class-Accessor, make all changes that also apply.

srcpkgs/perl-WebService-MusicBrainz/template
```diff
@@ -0,0 +1,15 @@
+# Template file for 'perl-WebService-MusicBrainz'
+pkgname=perl-WebService-MusicBrainz
+_dist=WebService-MusicBrainz
```
maxice8: See the comments on perl-Class-Accessor, make all changes that also apply.

faulesocke:
Now they are nice and shiny.

PR #13333, opened 2018-04-09
sddm: update to 0.17.0.

maxice8:


maxice8:
No clue about that i just tested against a vanilla kde5 setup and it worked fine.

Hoshpak:
I finally managed to pull your branch and build the package. As expected, sddm is still broken when ConsoleKit2 is present. So please don't update this package until either sddm has been made compatible with ConsoleKit2 or ConsoleKit2 has implemented an elogind compatible interface.

PR #13334, opened 2018-04-09
krita: update to 4.0.1.

cr6git:


asergi:
Drop all unnecessary changes, but `version` and `checksum`.

cr6git:
All right.

cr6git:
@asergi the checksum has changed? weird.

PR #13349, opened 2018-04-10
pjproject: update to 2.7.2.

Hoshpak:


srcpkgs/pjproject/template
```diff
@@ -23,7 +23,7 @@ maintainer="Christopher Brannon <chris@the-brannons.com>"
 license="GPL-2"
```
maxice8: use spdx identifier

srcpkgs/pjproject/template
```diff
@@ -2,8 +2,8 @@
 # no-third-party.patch comes from Alpine,
```
maxice8: superfluous comments

PR #13353, opened 2018-04-10
New package: latexmk-4.55

faulesocke:
A really nice little script that automatically runs latex the correct number of times, auto-detects changes in source files, updates the previewer, etc.


cr6git:
`srcpkgs/latexmk/template:16: indent with tabs`

jnbr:
If `texlive-bin` is installed, `latexmk` can be added by running `tlmgr install latexmk`. Does it make sense to package this, given `latex` is only provided by `texlive-bin`?

faulesocke:
Oh, you're right.

PR #13363, opened 2018-04-10
hostfw: update to v0.6.2

abenson:


Duncaen:
You usually need `create_wrksrc=yes` if you download multiple distfiles and don't move them out of the builddir, otherwise `/builddir` has some leftover files after building the package.

abenson:
Okay. Made that change, thanks.

PR #13374, opened 2018-04-11
Update iana-etc build script

### LINKED ISSUE #13242: iana-etc: outdated
Our iana-etc lists are almost 10 years old now.
Instead, we should fetch from the official IANA lists, e.g. what Arch does:
https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/iana-etc

(I'd stick the xml into files/ for reproducability.)

winks:
Fixed #13242 

For now, a direct port of the suggested Arch version

  * not using CSV instead of XML
  * not caching/saving the 2 XML files to files (because I'm not sure what the best flow would be, manual wget und just processing the files from the git checkout?)
  * haven't investigated if some lines in `services` should be purged, it's a lot longer now


winks:
About that build failure, I had this error once, building locally - trying to investigate why it would sometimes create `/builddir/iana-etc-20180405` inside `masterdir` and sometimes not. I thought this was done automatically.

srcpkgs/iana-etc/template
```diff
@@ -12,6 +12,10 @@ checksum="5b0a0d6bd70fb3f8c8dfc09ad712f18ff03bfe54c8f2413569ed3d365f8ede5d 4992f
 distfiles="https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xml https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xml"
 skip_extraction="service-names-port-numbers.xml protocol-numbers.xml"
 
+do_extract() {
```
pullmoll: You should probably use `create_wrksrc=yes` instead.

winks: Thanks, I need to RTFM more :)

winks:
Added script, added files.

Maybe it makes sense to store them compressed with `xz -9` in `/usr/share/`?

```
-rw-r--r-- 1 florian 3.5M Nov 17 11:23 port-numbers.iana
-rw-r--r-- 1 florian  41K Nov 17 11:23 protocol-numbers.iana

-rw-r--r-- 1 florian 315K Nov 17 11:23 port-numbers.iana.xz
-rw-r--r-- 1 florian 6.7K Nov 17 11:23 protocol-numbers.iana.xz
```

chneukirchen:
I don't think we need to keep these .xml files at all? Then we can also just store the results. :)

PR #13380, opened 2018-04-11
edb-debugger: update to qt5

DirectorX:


Johnnynator:
Why not change edb to qt5? It works fine there.

PR #13391, opened 2018-04-11
New package: intellij-idea-ultimate-edition-2018.1.1

rien:
The 'ultimate' version of `intellij-idea-community-edition`.


asiekierka:
Should we be distributing the Ultimate edition, since it's technically paid software with a free trial?

rien:
Good question. This software is free for me because I'm a student, but I would understand it if my PR got rejected because of this.

I would argue that we _do_ want to distribute this. Mainly because Void does not have a 'user repository' by design so there does not exist a proper way to install this piece of software.

Both the [Quality Requirements](https://github.com/voidlinux/void-packages/blob/master/Manual.md#quality_requirements) or the contributing guide do not explicitly allow or disallow paid software. So I suggest to mention Void's policy about such packages there if this PR gets rejected because of this reason

newbluemoon:
@asiekierka If redistribution isn’t allowed I think there could still be template, but it has to use `restricted=yes` (so the builders won’t create a package).

cr6git:
[#14381](https://github.com/voidlinux/void-packages/pull/14381) / [#14527](https://github.com/voidlinux/void-packages/pull/14527)

PR #13402, opened 2018-04-12
riot-desktop: update to 0.14.0

jtl999:
I've also changed the `riot-desktop.sh` wrapper to pass any of it's arguments to the electron process. For example a valid command line argument for the electron process is `--disable-smooth-scrolling` which does as it says. Let me know if I should instead have that as a separate pull request.


jtl999:
Hmm. Builds fine in my system which is also x86_64 arch but fails with an integrity check error for the hash of `fsevents` npm package, which is a dependency. 

Interesting

Vaelatern:
obsolete

Vaelatern:
but I would like your improvements to the .sh file.

Issue #12571, opened 2018-03-15
FreeCAD 0.17 tracking issue


faulesocke:
Hey guys,

I'm not sure whether it's usual here to have a tracking issue for more complex changes but I make one anyway. When reading through their forums, it seems that FreeCAD 0.17 can be expected very soon. Therefore I invested a day in compiling to make sure the current git version compiles fine in void (so that when 0.17 is released it will hopefully compile fine as well).

I had to make two new packages, vtk and med which I will PR in the next hours (currently polishing and recompiling to make sure everything works). I don't expect them to get accepted immediately, especially med is still a little bit ugly.

Here are the Steps that need to be done before we can switch to the new FreeCAD version:
- [x] Package vtk #11993
- [x] Package med #12572
- [x] Check freecad package for unnecessary dependencies

faulesocke:
Piraty already has done a vtk package: #11993

yopito:
hello, nice !  
Disclaimer: I've packaged FreeCAD for VoidLinux, but I don't have any CAD culture ...

They were some discussion on IRC 2 weeks ago about it with @Piraty  and @karl-nilsson .  
I don't know about `med` package, but FreeCAD 0.17 is also supposed needs OpenCascade (OCCT) instead of OCE. What's your opinion about it ? 

I have some work on vtk (also @Piraty) and OCCT packaging but not yet usable by others.
I will ping here as soon I will have something usable.

Piraty:
Yes, freecad update is expected in near future, as my info from their irc channel suggests.
For the sake of completeness: here is my [freecad-git branch](https://github.com/Piraty/void-packages/commits/freecad-git), which makes use of the mentioned vtk PR.
As far as my findings go, no other dependency than `vtk` and `libmed` were neccesary for a succesful build of most recent freecad, and the build output does not suggest anything like `occt` is expected.
I worked on `libmed` already (see my freecad-git branch), no need to package the whole med suite.

faulesocke:
@yopito My build worked just fine without OpenCascade. Does it add any extra features to freecad?

karl-nilsson:
@faulesocke OpenCascade is the CAD kernel FreeCAD uses, so I'd be extremely surprised if your build functioned without it installed.

iirc, the next FreeCAD release will depend on a newer version of OpenCascade than the current oce package provides

oce is working on releasing a new version, but there's no estimated release date
[...]

PR #13404, opened 2018-04-12
freecad: update to 0.17

Piraty:
- vtk is a mandatory dependency (already merged, see #11993)
 - libmed is a mandatory dependency (supersedes #12572)
 - occt is not mandatory, but recommended by freecad devs (see https://github.com/voidlinux/void-packages/issues/12571#issuecomment-377906827)
 - SMDS patch is not needed anymore
 - freecad's transition to python3 is not complete, maybe in 0.18
 - in order to use qt5, Void needs pyside2 etc. first, which is not officially released yet

I tested this a bit for a few days on a x86_64 glibc system


faulesocke:
I've looked at it, builds fine. The templates look OK to me, so IMO this can get merged.

When I have time during the weekend, I will look at med again to see whether we can fix compilation with MPI there.

yopito:
Nice work, so huge stuff ... I will give a try ASAP to help. 
* oce: dedicated commit and revbump on oce package in order to "publish" the conflict (with occt).

karl-nilsson:
I successfully built and installed all 3 packages.

One question: why is coin3-doc a build and runtime dependency?

Piraty:
Thanks both of you, I corrected accordingly.

yopito:
about `conflicts`: Uneeded to add it on both sides, only on `occt` is sufficient.

PR #13435, opened 2018-04-13
ZFS now requires perl to build the DKMS module.

zdykstra:
This should resolve the configure step failing in a minimal chroot / fresh install with out perl. 

Upstream commit that broke the build: 

`https://github.com/zfsonlinux/zfs/blob/e9a77290081b578144e9a911ad734f67274df82f/scripts/enum-extract.pl`


Vaelatern:
Sorry, this PR does not understand hostmakedepends. I've confirmed I do have perl on my testing machine, and will try to verify this PR.

zdykstra:
Did I generate the PR incorrectly? It's my second day using Void, so I'm very new to the whole stack - sorry!

Johnnynator:
If dkms needs perl you need to add it to depends so it gets installed on your machine. Hostmakdepends and makedepends are only used from xbps-src. (Read the manual.md for the desc of these)

zdykstra:
Sorry about that! I've moved perl to depends, with out a specific version requirement.

PR #13436, opened 2018-04-13
kakoune: update to 2018.04.13

SolitudeSF:
~~this is the first release. sadly, versioning scheme differs from merged prerelease so it counts as downgrade.~~
added `reverts`


Johnnynator:
If it counts as downgrade why not add use `reverts`?

scadu:
@Johnnynator why it would cound as downgrade? Release date is actually a snapshot from master branch at the 13rd April.

PR #13438, opened 2018-04-13
kde-cli-tools: add symlink for kdesu / fix cross-build

### LINKED ISSUE #13376: Package Issue: kdesu package does not include binary
### System

* xuname:  
  *Void 4.15.15_1 x86_64 GenuineIntel uptodate rrrmFFF*
* package:  
  *kdesu-5.43.0_1*

### Expected behavior
```kdesu``` should be a binary located somewhere in $PATH (e.g /usr/bin or /usr/local/bin)
### Actual behavior
There is no executable binary in $PATH which prevents applications like octoxbps from using it
example ```bash: kdesu: command not found```
```
$ xbps-install -S kdesu
$ kdesu -c <command>
```

### PR DESCRIPTION
**Pull request by newbluemoon**
closes #13376 

According to #13376 some applications seem to expect `kdesu` in `/usr/bin/` and cannot find it in `/usr/libexec/kf5/`. For KDE4 this symlink was created in kde-runtime and seems to have been overlooked in KDE5.

### CODE REVIEW

File: srcpkgs/kde-cli-tools/template [hunk 1]
```diff
@@ -14,3 +14,11 @@ license="LGPL-2"
 homepage="https://projects.kde.org/projects/plasma/kde-cli-tools"
```
cr6git: the homepage points to https://cgit.kde.org/kde-cli-tools.git

srcpkgs/kde-cli-tools/template
```diff
@@ -1,7 +1,7 @@
 # Template file for 'kde-cli-tools'
 pkgname=kde-cli-tools
 version=5.12.0
-revision=1
+revision=2
 build_style=cmake
 configure_args="-DBUILD_TESTING=OFF"
 hostmakedepends="extra-cmake-modules pkg-config"
```
cr6git: you can remove `makedepends="kdoctools"` it's already pulled by the other `k***-devel` packages.

newbluemoon:
Thanks @cr6git for pointing this out. :)
I just assumed the template was ok and didn’t check any further.

srcpkgs/kde-cli-tools/template
```diff
@@ -1,16 +1,24 @@
...
+homepage="https://cgit.kde.org/kde-cli-tools.git"
 distfiles="http://download.kde.org/stable/plasma/${version}/${pkgname}-${version}.tar.xz"
 checksum=9e27eff6a73935b504d7a8fbf35dd8fba353943da0378df6c2f0422c07bf7288
+
+if [ -n "$CROSS_BUILD" ]; then
+	hostmakedepends+=" kdoctools python qt5-host-tools qt5-qmake"
+fi
+
+post_install() {
+	ln -sf ../lib/libexec/kf5/kdesu ${DESTDIR}/usr/bin
```
Gottox: This introduces a conflict with kde-runtime where the same was done. We can solve this with alternatives though.

newbluemoon: Sorry, somehow I was under the assumption #13089 had already been merged. So if kde-runtime is about to get removed (if that’ll be not too far in the future) would it be better to wait?

Gottox:
ETOOMUCHPRS, let's get kde4 removal merged first then.

PR #13443, opened 2018-04-13
xbps-src: update-check: handle underscores in upstream version string

Chocimier:
Affects also few other packages, e.g. luafilesystem


srcpkgs/lowdown/update
```diff
@@ -1,4 +1 @@
-# This pattern is incomplete, it finds out the versions like '0_3_1' but
-# comparison with 'xbps-uhelper cmpver' results in it being always treated
-# as a lower version
 pattern="VERSION_\K[\d_]+"
```
maxice8: i think it should be `\d\.` instead of `\d_`

maxice8: note: i didn't test.

PR #13449, opened 2018-04-13
Update qalculate to 2.4.0

Hoshpak:


SolitudeSF:
gnuplot is an optional dependency, could it be removed?

SolitudeSF:
also, running `xmindep` on `qalculate` gives this reduced dependecy list
```
cln-devel libcurl-devel libxml2-devel mpfr-devel readline-devel
```
and `gdk-pixbuf-devel` in `qalculate-gtk` is redundant too.

PR #13462, opened 2018-04-14
slock: add config.h support

jmi2k:


srcpkgs/slock/template
```diff
@@ -11,6 +11,10 @@ homepage="http://tools.suckless.org/slock"
 distfiles="http://dl.suckless.org/tools/slock-${version}.tar.gz"
 checksum=b53849dbc60109a987d7a49b8da197305c29307fd74c12dc18af0d3044392e6a
 
+pre_build() {
+	cp ${FILESDIR}/config.h config.h
```
Duncaen: Can we do something like `[ -e ${FILESDIR}/config.h ] && cp ${FILESDIR}/config.h config.h` instead and keep the default `config.h` out of the repository?
The package contains the config file already and adding it to our repository just wastes space and complicated updating.

PR #13473, opened 2018-04-14
freexl: fix name in shlibs

abenson:
SONAME: libfreexl.so.1 <-> UNKNOWN PKG PLEASE FIX!


abenson:
Added another shlibs fix to the same PR since it's a part of the same package effort.

PR #13475, opened 2018-04-14
papirus-icon-theme: update to 20180501.

cr6git:
For a package (typically an icon-theme) containing millions of files, it's faster to extract the tarball directly to `${DESTDIR}`. This is why I modified the `do_install()` stage.

Correct me if I am wrong, of course.
ping @gspe


cr6git:
`The log length has exceeded the limit of 4 MB`

gspe:
Hi,
for sure you are more expert and have more experience  that me,I still have a lot of thing to learn so I agree with you.

cr6git:
I may have experience, but I am certainly not an expert. :laughing:

cr6git:
I just updated to 20180501.

cr6git:
@gspe I have the impression that my pull request got lost in the long list of PRs. :unamused: 

I close.

Issue #13230, opened 2018-04-05
Python 2 EOL


chneukirchen:
Python 2 will be end of lifed in 2020.

I think we should start moving most things to Python 3 when it is easy to do; ideally an ordinary install won't contain Python 2 at all anymore.

This is a tracking issue.

- [ ] dstat (https://github.com/dagwieers/dstat/issues/118)
- [ ] perf/linux-tools (should work with 4.16: https://github.com/torvalds/linux/commit/66dfdff03d196e51322c6a85c0d8db8bb2bdd655#diff-91fa7c5a8167518a0f1ca25cd77b9615)
- [ ] scons (should work with Python 3, but what about the sconscripts?)
- [ ] transmission-remote-cli (upstream dead, tremc uses Python 3)

Hoshpak:
scons works with python 3 but the scripts indeed need to patched/ported. While doing the last scons update, I tried to build it with python 3 and selected 2 random packages using scons as build system. Both failed to execute a build with scons under python 3, at least one of them used the print statement instead of the function. So we need to check all packages in the repository using the scons build style if they still work with python 3.

The question is, if we should aspire to patch all scripts ourselves or ask upstream to port their stuff and only patch if the upstream developers don't respond or refuse to do it.

chneukirchen:
I guess keeping scons as py2 is ok then for now.

maxice8:
i'm transitioning packages to python3 as i update them, they mention this issue.

maxice8:
@cr6git for python-modules (basically anything python-*) no, just stuff that uses either python or python3 should be made to use python3, see my #13275 #13228 #13245

PR #13484, opened 2018-04-14
New package: sshuttle-0.78.4

dan-santana:


jnbr:
Having a package for python2 and python3 makes sense, if it provides a python library. This doesn't seem to be the case. Is there a good reason to have sshuttle for both, python2 and python3? I would consider renaming it to sshuttle and use python3 only.

dan-santana:
But @jnbr you are right, according to [this issue](https://github.com/voidlinux/void-packages/issues/13230) it's better to remove Python 2 builds if that's possible (and it's certainly possible in this case).

PR #13495, opened 2018-04-15
udisks2: update to 2.7.6.

maxice8:
In this PR;

- Fix building `dmraid` and it's -devel dependencies
- Remove series files and tidy up patches for `dmraid`
- Fix building  `volume_key` by adding python to hostmakedepends
- Fix `volume_key-devel` by adding  patch that removes erroneous config.h include
- add new package `libblockdev`
- update `udisks2` to 2.7.6, `gnome-disks-utility` uses them and until now it would silently fail to use `udisks2` during configuration.


maxice8:
IN THIS PR: Travis fails to cross-compile LUL

PR #13511, opened 2018-04-15
New package: pypanel-2.4

cr6git:


asergi:
Do you really use it? Or just for the sake of it? I would like to avoid importing Python2-only dead things.

cr6git:
just for the sake. :stuck_out_tongue: I understand, no problem.

PR #13521, opened 2018-04-16
New package: hunspell-da_DK-2.3_1

unixsheikh:


cr6git:
Your template is not properly formatted.
Read [STYLE.md](https://github.com/maxice8/void-packages/blob/0949fe18d00db245b4f5641afbb8d71a53166cfa/STYLE.md) by @maxice8

maxice8:
note: STYLE isn't official (i'm trying to get it to be) and can change at any moment until it becomes official, at that point all revisions will be marked under `REVISIONS`

cr6git:
@maxice8 I know, and I would be happy to see this become official. Your work a good source of inspiration.

Gottox:
Get it in non the less :)

unixsheikh:
This is a copy paste from a template already in Void. Also if this is the case, then there is a LOT of packages with templates that isn't properly formattet.

Duncaen:
@unixsheikh It was merged, I don't see what the problem is.
Can you please tag a release in your repository, the distfile tracks master instead of a tagged release or a specific revision, which is going to break the template at the point you commit something.

unixsheikh:
@Duncaen, thanks! I have just tagged it at version 2.1. Do I need to update the template and make a new PR?

Duncaen:
Yes the distfiles should have `${version}` in the url, but whats the reason that you host this in your git repository instead of using some upstream link?
`hostmakedepends="unzip"` is not necessary.
`license="..."` should be comma separated.

unixsheikh:
I couldn't locate the upstream link for that specific package in any supported format. I looked at some of the other hunspell packages and noticed different sources too.

I haven't had the time to dive into the more complicated setups with xbps-src yet, but I suspect that it's possible to use the upstream main package with all the languages and then build from that.

I also looked at the Debian package which also seems to be put together differently https://packages.debian.org/da/stretch/hunspell-da. So I just decided to volunteer at keeping that package updated on Void.

Regarding the license - should be comma separated, I already had that in, but was told to remove the commas, I cannot remember who said that.

PR #13540, opened 2018-04-17
Fix adobe-flash-plugin

crocket:
It fixes broken download URLs.


cr6git:
[#13362](https://github.com/voidlinux/void-packages/pull/13362)

PR #13555, opened 2018-04-17
New package: editorconfig-0.12.2

dan-santana:


srcpkgs/editorconfig/template
```diff
@@ -0,0 +1,29 @@
+# Template file for 'editorconfig'
+pkgname=editorconfig
+version=0.12.2
+revision=1
+wrksrc=editorconfig-core-c-${version}
+build_style=cmake
+makedepends="pcre-devel"
+depends="libpcre"
```
Duncaen: `depends` is not necessary, `xbps-src` detects the runtime dependency automatically https://travis-ci.org/voidlinux/void-packages/jobs/367685708#L974.

srcpkgs/editorconfig/template
```diff
@@ -0,0 +1,29 @@
...
+short_desc="EditorConfig core library written in C"
+maintainer="Daniel Santana <daniel@santana.tech>"
+license="Simplified BSD"
```
Duncaen: Please use `BSD-2-Clause` if it is the same license as `BSD 2-Clause "Simplified" License` on https://spdx.org/licenses/.

Issue #12219, opened 2018-03-06
make void-packages license= follow SPDX


maxice8:
#### What

SPDX provides a [list of licenses](https://spdx.org/licenses/) with short identifier and the license text.

#### Why

make easier for maintainers to add licenses and for people to search for the license itself, as an example we currently have in the license= field of package, `BSD` which is ambiguous, `BSD-3-Clause` ( the one provided by SPDX ), `BSD-3-clause`, `2-clause-BSD`, `2-Clause-BSD` ( the biggest one) and even `BSD-style`, we also have `GPL-ish`, `MIT-style` and `Unlicense`, `unlicense` (SPDX), and `The unlicense`.

#### How

a. Enforce license= fields on new packages and on package updates that comply with SPDX short name identifiers.
b. Modify `base-files` to include the licenses with their proper names.
c. Modify current templates to comply with SPDX identifiers ( this is will be costly as it wil require personally checking 7444 packages. )
d. modify `xlint` to deal with the new licenses ( ways on how xlint can help maintain quality on checking license= is to discussed )
e. modify manual.md to tell people that can read manuals of the SPDX and how to use it.

Vaelatern:
Except `BSD` ;)

ebfe:
With a mapping of

```
        "GPL-3": "GPL-3.0",
        "GPL-2": "GPL-2.0",
        "LGPL-2": "LGPL-2.0",
        "LGPL-3": "LGPL-3.0",
        "3-clause-BSD": "BSD-3-Clause",
        "3-Clause-BSD": "BSD-3-Clause",
        "2-clause-BSD": "BSD-2-Clause",
        "2-Clause-BSD": "BSD-2-Clause",
        "zlib": "Zlib",
```

It's already down to 1406 https://gist.github.com/20329ebd7e6161191e447b57dd4335d4

For the templates that actually use vlicense (unfortunately this was a "change
when touch"-thing so not all have it), i think most "BSD" and similar can be
automatically identified by running licensechecker on the license file.

maxice8:
```
 "GPL-3": "GPL-3.0",
        "GPL-2": "GPL-2.0",
        "LGPL-2": "LGPL-2.0",
        "LGPL-3": "LGPL-3.0",
```
those are deprecated GPL-x.x-only GPL-x.x-or-later is the one currently on the list

maxice8:
[...]

PR #13556, opened 2018-04-17
terminology: update to 1.2.0

pulux:


pulux:
what's wrong with: `2-clause-BSD` statement?

this is only a PR I'm not the maintainer.

maxice8:
BSD-2-Clause is the correct short-identifier for SPDX https://spdx.org/licenses/

it is currently optional but i'm pushing for Void Linux to alpine-ish policies (including using SPDX short-identifiers) regarding license= https://github.com/voidlinux/void-packages/issues/12219

PR #13584, opened 2018-04-18
lily: update to 1.4, fix homepage.

cr6git:
`=> WARNING: lily-devel-1.4_1: liblily.so not found in common/shlibs!`

I guess this is not important.


Gottox:
Can be ignored if nothing outside of this package links against the file.

PR #13589, opened 2018-04-18
gcompris-qt: update to 0.90, qml-box2d: update to 0.0.20180406.

cr6git:
Thanks to @newbluemoon for his help.


cr6git:
`Connection reset by peer` :confused: ?

PR #13600, opened 2018-04-18
nomad: update to 0.8.1.

cr6git:


cr6git:
[f591f7e](https://github.com/voidlinux/void-packages/commit/f591f7e41398952f30bbb9e097674493fc39fa83)

PR #13610, opened 2018-04-19
New package: sparse-0.5.2

Chocimier:


chneukirchen:
Move the LD to toplevel. Also, should there be a -devel package?

Chocimier:
superseded by https://github.com/void-linux/void-packages/pull/708

PR #13615, opened 2018-04-19
freecad: fix musl build

Piraty:
sorry, forgot to actually fix musl build issues when i removed the `broken` option in #13404.
It now compiles on x86_64-musl, but i don't run a desktop on that arch, so i didn't test the actual behviour.


srcpkgs/freecad/patches/002-execinfo.patch
```diff
@@ -0,0 +1,29 @@
...
++#ifdef __GLIBC__
+     void *callstack[128];
+     size_t nMaxFrames = sizeof(callstack) / sizeof(callstack[0]);
+     size_t nFrames = backtrace(callstack, nMaxFrames);
+@@ -1105,6 +1108,8 @@
+     }
+ 
+     free(symbols);
++#endif
++    std::cerr << "sorry, no backtrace on musl libc";
```
Gottox: this always returns the error message, even on glibc.

srcpkgs/freecad/patches/010-salomesmesh-execinfo.patch
```diff
@@ -0,0 +1,29 @@
...
++#ifdef __GLIBC__
+     void *array[50];
+     size_t size;
+     char **strings;
+@@ -123,6 +126,8 @@
+       }
+ 
+     free (strings);
++#endif
++   std::cerr << "sorry, no backtrace on musl libc";
```
Gottox: Same here

newbluemoon:
For musl there’s a libexecinfo package. You could try if that worked and then the patches wouldn’t be necessary.

pullmoll:
FYI fixed med by creating a hdf5-18 package.

Piraty:
thanks @pullmoll .
The patches proposed in this PR solve build issue on musl for now. I didn't succeed in making the cmake/linker use libexeinfo correctly, so backtrace won't be available.

PR #13619, opened 2018-04-19
texstudio: rebuild against qt5.10 - fixes glitch

Piraty:
Rebuilding with the new qt fixes the glitch described in https://github.com/texstudio-org/texstudio/issues/82 (which is also the reason I [didn't update to 2.12.8](https://github.com/voidlinux/void-packages/pull/11408#issuecomment-369501500))


Piraty:
hmm it doesn't fix the glitch afterall...

PR #13628, opened 2018-04-19
New package: goxel-0.7.3

Johnnynator:


srcpkgs/goxel/template
```diff
@@ -0,0 +1,25 @@
...
+homepage="https://guillaumechereau.github.io/goxel/"
+distfiles="https://github.com/guillaumechereau/goxel/archive/v${version}.tar.gz"
+checksum=56790d96d9d52d37765ac5c59f924c007885daaa6fb2129baea51dff5380c15c
+
+case "$XBPS_TARGET_MACHINE" in
+	*-musl) makedepends+=" argp-standalone";;
+esac
+
+# For now cross-compilation fails
+nocross=yes
```
maxice8: mind putting a more descriptive value ?

example:  `nocross="tries to run compiled code on host"`

cr6git:
````
nocross="
 cc1plus: error: bad value ('armv8-a') for '-march=' switch
 g++: error: unrecognized command line option '-mfpu=vfpv3'
 g++: error: unrecognized command line option '-mfloat-abi=hard'"
````

File: srcpkgs/goxel/template [hunk 2]
```diff
@@ -0,0 +1,25 @@
...
+
+do_install() {
```
cr6git: You can add the icon and the .desktop file:
````
do_install() {
	sed -i 's|${SNAP}/icon.png|/usr/share/icons/goxel.png|' snap/gui/goxel.desktop
	vbin goxel
	vinstall snap/gui/goxel.desktop 644 usr/share/applications
	vinstall icon.png 644 usr/share/icons goxel.png
}
````

File: srcpkgs/goxel/template (hunk 1 above)
Comment (Johnnynator) [reply srcpkgs/goxel/template]: Actually I just updated cr6git's template we worked like half a year on and didn't update this. Will do that.

**Comment by newbluemoon** [comment]
The problem with cross-compilation is that the `SConstruct` script somehow doesn’t respect `$CC` and `$CXX`. It works when these are set explicitly, e.g. like

```
	pre_build() {
		sed -i -e "/conf = env.Configure()/ a env.Replace(CC = \"$CC\")\nenv.Replace(CXX = \"$CXX\")" SConstruct
	}
fi
```

PR #13671, opened 2018-04-20
texstudio: update to 2.12.8

Piraty:
Void's current version of texstudio has a glitch, that was introduced with qt5.10
see: https://github.com/texstudio-org/texstudio/issues/43
(which is also the reason I [didn't update to 2.12.8](https://github.com/voidlinux/void-packages/pull/11408#issuecomment-369501500))

This downgrade fixes the glitch.

I'm not quite sure if i applied the `reverts` field correctly.


pullmoll:
`reverts=<version>` needs to immediately follow `version=`.

Piraty:
done

Piraty:
upstream fixed the qt issue (https://github.com/texstudio-org/texstudio/commit/4cf5c2eb2a8a0353335c0ed6d6be4b3f19c950c3), so i added the fix +  updated to 2.12.8 finally.

PR #13675, opened 2018-04-20
update_check.sh: add update check for kde

Johnnynator:
Atm most kde applications have their own update file and most kde frameworks templates don't have one, so the update check doesn't work.
This should make the manual update file for kde applications obsolete and add update check to kde frameworks.

~~(tested only whith kcalcore, kate5 and extra-cmake-modules but should work with everything else too)~~ Should work always

~~Edit: It seems to break a few old kde4 packages like akonadi.~~


maxice8:
+1

PR #13693, opened 2018-04-20
filezilla: fix update file

Chocimier:


bra1nwave:
Duplicate: https://github.com/voidlinux/void-packages/pull/13313

Chocimier:
Oh, I didn' t see that.There is not update file there, however.

PR #13701, opened 2018-04-21
pulseaudio: split OSS emulation wrapper into new package

AluisioASG:
`padsp` allows running OSS-only applications directly under PulseAudio. This PR enables building the wrapper script and preload library and moves them into a new subpackage, together with man page and shell completion.

I've verified it compiles on x86_64, i686-musl (with a backported patch) and armv7hf.


TeddyDD:
any chance to get this merged/moved to new repo? I tried to pull this to test if it works but it fails to fetch

PR #13709, opened 2018-04-21
New package: python-miniupnpc-2.0.2

faulesocke:
A Python UPnP library (bindings for miniupnpc).


newbluemoon:
python-setuptools and python3-setuptools have to go to `hostmakedepends`.

PR #13713, opened 2018-04-21
unrar: update to 5.6.3.

cr6git:


cr6git:
musl is ok but glibc does not agree, funny...

PR #13731, opened 2018-04-22
font-symbola: update to 11.00.

cr6git:


Johnnynator:
`Software on this site is free strictly for   personal, non-commercial use`
It seems like our license was already wrong before.

Johnnynator:
Also I can not find any older license, so I would suggest to remove this font completely from the repo.

ChALkeR:
@Johnnynator The old license notice is in web archive, it was «In lieu of a licence; fonts and documents in this site are free for any use; George Douros», and it was there with v.10.23.

v11.0 was released when the new license notice was there. Not sure what license did v.10.24 have.

Refs:
* http://web.archive.org/web/20180129230141/http://users.teilar.gr/~g1951d/
* https://bugs.archlinux.org/task/58886

Johnnynator:
@ChALkeR Ok. Thanks for the clarification.

PR #13732, opened 2018-04-22
file: rebuild due to not detected shlibs

newbluemoon:
Something seems to have gone wrong with the `04-generate-runtime-deps` and `06-shlib-provides` hooks in cross builds when *file* and its subpackages (libmagic) have been created.

Checking `props.plist` in `libmagic-5.33_1.*.xbps` the section shlib-provides is missing for armv6l and armv7l and additionally the section shlib-requires for aarch64, aarch64-musl, armv6l-musl, and armv7l-musl.

In a local rebuild everything got detected correctly.

I just don’t know if it was just a glitch in the package or some other problem. However, checked a few other packages of the same date and couldn’t find a problem.


newbluemoon:
Errors due to Travis not liking ftp.

PR #13733, opened 2018-04-22
linux: update to 4.16.

cr6git:
Why not?


srcpkgs/linux/template
```diff
@@ -1,10 +1,10 @@
 # Template file for 'linux'
 pkgname=linux
-version=4.15
+version=4.16
 revision=1
 build_style=meta
-homepage="http://www.voidlinux.eu/"
-short_desc="The Linux kernel meta package"
+homepage="https://www.voidlinux.eu/"
+short_desc="Linux kernel meta package"
```
asergi: Unnecessary.

Gottox:
4.15 is eol.

PR #13736, opened 2018-04-22
atari800: update to 4.0.0.

cr6git:
:rainbow:


cr6git:
Replaced by https://github.com/void-linux/void-packages/pull/326

PR #13740, opened 2018-04-22
common/environment/fetch/misc.sh: add KDE_SITE.

maxice8:
There are 272 occourences of download.kde.org/stable in distfiles, to avoid repetition add a KDE_SITE variable to be used.

There is also a PR for `xlint` detection on chneukirchen/xtools#87


chneukirchen:
Do they have official mirrors?

Johnnynator:
Yes. https://download.kde.org/extra/download-mirrors.html

PR #13741, opened 2018-04-22
New package: empty-0.6.20b

stb55h:


srcpkgs/empty/template
```diff
@@ -0,0 +1,20 @@
+# Template file for 'empty'
+pkgname=empty
+version=0.6.20b
+revision=1
+build_style=gnu-makefile
+short_desc="Run processes under pseudeo-terminal. TCL/Expect replacement"
+maintainer="stb <stb1010101b@mailbox.org>"
+license="BSD"
```
maxice8: Please use a proper SPDX identifier.

PR #13754, opened 2018-04-23
libwaylandpp: update to 0.2.3.

maxice8:
remove wayland-scanner++ and simply turn it into a subpkg of libwaylandpp


maxice8:
Travis, Episode 2: Travis can't handle the package itself being added to hostmakedepends.

PR #13757, opened 2018-04-23
instead: update to 3.2.1.

maxice8:
fun fact: you don't need to \ a newline inside a dependency group (' ')


SolitudeSF:
`homepage` changed to [`https://instead-hub.github.io`](https://instead-hub.github.io)
you removed sdl1 from `makedeps` from sdl2 option

maxice8:
derp, forgot to push.

PR #13771, opened 2018-04-23
nvidia: update to 390.48

freshprince:
this doesn't fix broken drivers on linux4.16 but it does fix screen flickering when composition pipeline is on on 4.15

### DIFF
--- a/common/shlibs
+++ b/common/shlibs
@@ -143,7 +143,7 @@ libnvidia-gtk2.so.346.47 nvidia-gtklibs-346.47_1 ignore
 libnvidia-gtk3.so.346.47 nvidia-gtklibs-346.47_1 ignore
 libnvidia-glcore.so.346.47 nvidia340-libs-340.46_1 ignore
 libnvidia-glsi.so.346.72 nvidia-libs-346.72_1 ignore
-libnvidia-fatbinaryloader.so.390.25 nvidia-libs-390.25_1 ignore
+libnvidia-fatbinaryloader.so.390.48 nvidia-libs-390.48_1 ignore
 libglapi.so.0 libglapi-7.11_1
 libgbm.so.1 libgbm-9.0_1
 librsvg-2.so.2 librsvg-2.26.0_1
--- a/srcpkgs/nvidia/files/nvidia-swiotlb-linux_4_16.patch
+++ b/srcpkgs/nvidia/files/nvidia-swiotlb-linux_4_16.patch
@@ -1,0 +1,14 @@
+disable check for swiotlb for kernels newer as 4.16 because it uses a symbol
+that isn't exported anymore
+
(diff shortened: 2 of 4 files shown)


ackalker:
@freshprince If by "broken drivers on linux4.16" you mean the missing symbol on `nvidia` module load, then there is a workaround patch by `mlau` on the [NVIDIA Developer Forums](https://devtalk.nvidia.com/default/topic/1030082/linux/kernel-4-16-rc1-breaks-latest-drivers-unknown-symbol-swiotlb_map_sg_attrs-/) which allows the `nvidia` kernel module to load. Perhaps you could test and add this to your PR?
If you want to hold off on fixing that, then there is a prioroty P1 bug report about it on the [Kernel.org bugzilla](https://bugzilla.kernel.org/show_bug.cgi?id=198997) which you might want to watch.

freshprince:
thanks for your feedback @ackalker.

I added a patch that disables the swiotlb check for kernels newer 4.16 which "fixes" the driver for me.  I only added the patch for x86_64 since I can't test this i686.

ackalker:
You're welcome, @freshprince.

AngelFreak:
Any news on this ? 
I'm really missing this path.

abenson:
I'm currently testing out this PR, and so far so good. No problems, and the flicker issue (although minor for me) is gone.

kaervin:
I tried this too, it works on my x64 machine.

AngelFreak:
I now tried this as well.

Works great. No issues so far.

kaervin:
Now using this for some time I think there are some frame-pacing/frame-rate issues.
Although I don't know if it's just my feeling, but I don't think things were this bad before updating to this (or 390.25)
Sometimes things just slow down too much.
NVidia really fucked this one up...

freshprince:
@kaervin what kernel are you running on? could you try a kernel <4.16 to see if that issue persists? maybe it's swiotlb missing.

abenson:
[396.24 is out](http://www.nvidia.com/download/driverResults.aspx/133859/en-us)

freshprince:
> 396.24 is out

Afaik we only use the long lived branch.

kaervin:
Ok, since switching to the newer drivers (390.25) I always had some weird image corruption issues.
Sometimes in chromium some images would get corrupted (previews of visited sites, saved images, and more). This issue even persists in 396.24.
I'll try an older Kernel tomorrow, and see if this persists.

freshprince:
@kaervin that might be this bug: https://bugs.chromium.org/p/chromium/issues/detail?id=817400

I have the same behaviour in chromium.

asergi:
@freshprince please apply the patch unconditionally.

freshprince:
@asergi 
> please apply the patch unconditionally.
done and rebased.

PR #13779, opened 2018-04-24
openbox: fix distfiles, checksum, add openbox-devel short_desc.

cr6git:
... :zipper_mouth_face:


Vaelatern:
I will not accept this with third party untrusted un-upstreamed code, even as an optional build option.

Vaelatern:
Unless another @voidlinux/pkg-committers wants to accept such things, I think we did with freetype (?) and it caused no end of trouble.

srcpkgs/openbox/template
```diff
@@ -16,14 +16,24 @@ conf_files="
 short_desc="Standards compliant, fast, light-weight, extensible window manager"
 maintainer="Juan RP <xtraeme@voidlinux.eu>"
 license="GPL-2"
-homepage="http://www.openbox.org"
-distfiles="http://icculus.org/openbox/releases/$pkgname-$version.tar.gz"
-checksum=8b4ac0760018c77c0044fab06a4f0c510ba87eae934d9983b10878483bde7ef7
+homepage="http://openbox.org/wiki/Main_Page"
```
Vaelatern: Let them redirect it. No reason to do this.

srcpkgs/openbox/template
```diff
@@ -37,6 +47,7 @@ post_install() {
 }
 openbox-devel_package() {
 	depends="${makedepends} libopenbox>=${version}_${revision}"
+	short_desc="Openbox development files"
```
Vaelatern: :+1:

srcpkgs/openbox/template
cr6git: OK.

cr6git: OK!

cr6git:
@Vaelatern I will never go against you. I removed rounded corners stuff.

PR #13803, opened 2018-04-24
Update kiwix-{lib,tools}, libzim and enable libaria2 for aria2.

Johnnynator:


srcpkgs/libzim/template
```diff
@@ -1,17 +1,17 @@
 # Template file for 'libzim'
 pkgname=libzim
-version=3.2.0
+version=3.3.0
```
maxice8: see https://github.com/voidlinux/void-packages/pull/13785

Johnnynator: It is needed for updating kiwix-lib. libzim, kiwix-lib and kiwix-tools do get new releases at the same time always so it is best to update them at once.

maxice8: Understandable, at least import the patch that disables the nightmare that is -Werror , fixes compilation.

maxice8: and while you're also i mport the homepage change.

srcpkgs/aria2/template
```diff
@@ -1,17 +1,36 @@
...
 depends="ca-certificates"
 short_desc="Lightweight multi-protocol/multi-source command-line download utility"
 maintainer="Juan RP <xtraeme@voidlinux.eu>"
 license="GPL-2"
 homepage="https://aria2.github.io/"
-distfiles="https://github.com/tatsuhiro-t/aria2/releases/download/release-${version}/${pkgname}-${version}.tar.xz"
+distfiles="https://github.com/aria2/aria2/releases/download/release-${version}/${pkgname}-${version}.tar.xz"
 checksum=2539e4844f55a1f1f5c46ad42744335266053a69162e964d9a2d80a362c75e1b
+
+libaria2_package() {
```
maxice8: missing symlink according to travis. ditto for libaria2-devel

srcpkgs/libzim/template
Johnnynator: Did compile fine locally without removing Werror for me. WIll see if it will error because of that on any other Arch.

srcpkgs/aria2/template
Johnnynator: Yep, also just notticed. Forget to add them to the commit

srcpkgs/libzim/template
maxice8: if i remember correctly it fails on aarch64-musl

maxice8: as a thumbs up always compile locally and against aarch64-musl because it is the widest net you can cast to get build errors.

Johnnynator: Nah musl is fine. it seems to be that glibc is the cause. Was failing locally on x86_64 glibc. So I will add the patch.

Johnnynator:
Everything should be fine now. Travis will fail for cross because xbps-src -C does not work if you need to compile a package for host and target.

PR #13811, opened 2018-04-24
kde5-baseapps: update to 18.04.0.

Chocimier:


maxice8:
Please use ${KDE_SITE} on distfiles.

maxice8:
and SPDX short-identifiers too

Chocimier:
Licensing here is extra fun. GPL and LGPL is mixed in same tarballs. Few files include licensing header comment that allow exact version, few allow any later, and few has following header, having no spdx identifier:
```   
   This library is free software; you can redistribute it and/or
   modify it under the terms of the GNU Lesser General Public
   License as published by the Free Software Foundation; either
   version 2.1 of the License, or (at your option) version 3, or any
   later version accepted by the membership of KDE e.V. (or its
   successor approved by the membership of KDE e.V.), which shall
   act as a proxy defined in Section 6 of version 3 of the license.
```

maxice8:
LGPL-3.0-or-later-as-approved-by-kde-e-V-or-its-successor :^)

PR #13834, opened 2018-04-25
Fix xtools template

crocket:
vimdiff from vim is required by xdiff


srcpkgs/xtools/template
```diff
@@ -1,10 +1,10 @@
 # Template file for 'xtools'
 pkgname=xtools
 version=0.51
-revision=1
+revision=2
 noarch=yes
 build_style=gnu-makefile
-depends="bash curl findutils git make xbps"
+depends="bash curl findutils git make xbps vim"
```
Vaelatern: And what about other flavors of vim?

crocket: neovim doesn't come with vimdiff, but neovim can do diff with a different command.
Supporting other flavors of vim is out of scope for this pull request.

crocket: I investigate further and discovered that `vim` is actually a symlink to either nvim or vim.
I don't know how to deal with vim flavors in `depends`, yet.

Perhaps, `virtualpkg` in `xbps.d` is the way to handle vim flavors?

Duncaen:
I don't think its really necessary to have the hard dependencies, some user might not use `xdiff` at all and don't want to have vim.
Its easy enough to notice that vim is missing if you use `xdiff`.

crocket:
What about optional dependency?

Duncaen:
There are no optional dependencies.

Duncaen:
None, its really not hard to use `find` or `ls` or the output of `xbps-install` and your preferred editor.
`xdiff` is just a small helper for people who can and want to use `vimdiff`.

crocket:
Ok, I think I can make vimdiff or vim an optional dependency in xdiff.

PR #13836, opened 2018-04-25
vlc: update to 3.0.2.

jnbr:


jnbr:
Tested on x86_64 only.

jnbr:
Some libs which should be in th libvlc package are in vlc. Waiting for qt5 to be build on armv{6,7} / aarch before pushing pushing the fix.

jnbr:
i686 fails because it builds cmake (3.11.1 is missing on repo) which removes pkg-config after it is done, so phonon-qt5-backend-vlc can't use it.

PR #13850, opened 2018-04-25
libmwaw: update to 0.3.14.

bra1nwave:


srcpkgs/libmwaw/template
```diff
@@ -11,7 +11,7 @@ maintainer="bra1nwave <brainwave@openmailbox.org>"
 license="LGPL-2.1, MPL-2"
```
maxice8: please use SPDX short-identifier, LGPL-2.1-or-later, MPL-2.0

bra1nwave: Forgot about it. Thanks.

PR #13853, opened 2018-04-25
virtualbox-ose: update to 5.2.10.

Hoshpak:


hanspolo:
`xmindep` tells me, that many of the `makedepends` aren't necessary.

PR #13854, opened 2018-04-25
Update python-SecretStorage and split python and python3 packages

Hoshpak:


asergi:
Please include/update `python-keyring` as well. Then I'll start the review.

Hoshpak:
alright, added python-keyring

PR #13860, opened 2018-04-25
wpa_gui: rebuild (missing aarch64-musl package)

Piraty:


maxice8:
While you're at it.

homepage should be https://w1.fi/wpa_supplicant/
use SPDX identifier for the license
and update the maintainer to 'Leah Neukirchen <leah@vuxu.org>'

Piraty:
regarding license, the tarball has `README` (which is used by wpa_supplicant) and `COPYING`, both containing BSD-based texts but different ones.

maxice8:
On Wed, Apr 25, 2018 at 09:39:29PM +0000, Piraty wrote:
> regarding license, the tarball has README (which is used by wpa_supplicant) and
> COPYING, both containing BSD-based texts but different ones.
> 
> —
> 

use the one from COPYING

maxice8:
please squash them into one commit.

Piraty:
squashed

PR #13873, opened 2018-04-26
cura: update to 3.3.1.

maxice8:


bra1nwave:
Does the new version of `cura` start up without segfaulting?
I'm asking because there is currently a Bug in `PyQt5-5.10.1`, which crashes `cura`.
See here: https://github.com/Ultimaker/Cura/issues/3438

maxice8:
On Thu, Apr 26, 2018 at 01:58:00PM +0000, bra1nwave wrote:
> Does the new version of cura start up without segfaulting?
> I'm asking because there is currently a Bug in PyQt5-5.10.1, which crashes
> cura.
> See here: Ultimaker/Cura#3438
> 
> —
> 

nope still segfaults

bra1nwave:
I think we can use the patch from [archlinux](https://git.archlinux.org/svntogit/packages.git/plain/trunk/pyqt5-cura-crash.patch?h=packages/pyqt5).

bra1nwave:
Awesome. Thank you!

maxice8:
the libressl update has halted this until the dependencies of python-PyQt5 that also depend on libressl get rebuilt.

maxice8:
done, patch added.

PR #13879, opened 2018-04-26
wayland: avoid conflict with libwayland-egl .so symlink

maxice8:


asergi:
8f823d2c527d93cc83d009537d24ab2b73abf1bc

PR #13882, opened 2018-04-26
guitarix2: update to 0.37.0, change default build_options, fix license.

cr6git:


srcpkgs/guitarix2/template
```diff
@@ -1,35 +1,51 @@
...
 
 case "$XBPS_TARGET_MACHINE" in
 	i686*|x86_64*) ;;
 	*) configure_args+=" --disable-sse" ;;
 esac
 
+if [ -z "$build_option_avahi" ]; then
+	configure_args+=" --no-avahi"
+else
+	makedepends+=" avahi-glib-libs-devel"
```
newbluemoon: I think you can avoid branching when using something like

```
configure_args="... $(vopt_if avahi '' '--no-avahi')"
makedepends="... $(vopt_if avahi avahi-glib-libs-devel)"
```
and similar for option `bluez`.

cr6git: oh yes, I just forgot this possibility! :sweat_smile:

cr6git:
Replaced by https://github.com/void-linux/void-packages/pull/285

PR #13890, opened 2018-04-26
New package: stubby-0.2.2_1, getdns-1.4.1_1

codebam:


maxice8:
Title should be only stubby since it is the main package.

maxice8:
see here for naming commits https://wiki.voidlinux.eu/Xbps-src

srcpkgs/stubby/template
```diff
@@ -0,0 +1,20 @@
+# Template file for 'stubby'
+pkgname=stubby
+version=0.2.2
+revision=1
+build_style=gnu-configure
+makedepends="getdns libyaml-devel automake autoconf"
```
maxice8: autoconf is brought in by automake, you can remove it.

use `xmindep` from xtools to remove implicit package dependencies.

or even `xtrimdeps` from [here](https://github.com/maxice8/xtools-devel)

srcpkgs/getdns/template
```diff
@@ -0,0 +1,16 @@
+# Template file for 'getdns'
+pkgname=getdns
+version=1.4.1
+revision=1
+build_style=gnu-configure
+makedepends="libressl-devel libidn-devel unbound-devel"
```
maxice8: makedepends="libidn-devel unbound-devel"

is enough

srcpkgs/stubby/template
```diff
@@ -0,0 +1,20 @@
...
+short_desc="DNS Privacy stub resolver"
+maintainer="codebam <codebam@riseup.net>"
+license="BSD-3-Clause"
+homepage="https://github.com/getdnsapi/stubby"
+distfiles="$homepage/archive/v$version.tar.gz"
+checksum="d1418d2c4df3b2f32fac668423630f07dbfb38f6f8d402ddfda9304f16263b03"
+
+pre_configure() {
+	autoreconf -vfi
```
maxice8: use autoreconf -fi

maxice8:
stubby provides a configuration file on /etc/stubby/stubby.yml

create a conf_files entry for it.

---

/etc/stubby/stubby.yml.example is an example file so it should be in /usr/share/examples/${pkgname}

add an entry in post_install() that moves the file from ${DESTDIR}/etc/stubby/stubby.yml.example to ${DESTDIR}/usr/share/examples/stubby

codebam:
I did everything you said @maxice8, let me know if there's anything else I need to fix :smiley:

srcpkgs/stubby/template
```diff
@@ -0,0 +1,23 @@
...
+pkgname=stubby
+version=0.2.2
+revision=1
+build_style=gnu-configure
+makedepends="getdns-devel libyaml-devel automake"
+short_desc="DNS Privacy stub resolver"
+maintainer="codebam <codebam@riseup.net>"
+license="BSD-3-Clause"
+homepage="https://github.com/getdnsapi/stubby"
+distfiles="$homepage/archive/v${version}.tar.gz"
```
maxice8: homepage should be in brackets too, all variables inside a bigger variable should be.

srcpkgs/stubby/template
```diff
@@ -0,0 +1,23 @@
...
+checksum="d1418d2c4df3b2f32fac668423630f07dbfb38f6f8d402ddfda9304f16263b03"
+conf_files="/etc/stubby/stubby.yml"
+
+pre_configure() {
+	autoreconf -fi
+}
+
+post_install() {
+	vlicense COPYING
+	mkdir -p ${DESTDIR}/usr/share/examples/stubby/
```
maxice8: use vmkdir, always prefer v* functions.

maxice8:
seems like cross-compilation is borked.

codebam:
yeah I noticed that too, I don't know how to fix it though. I could try cross compiling and see what error it throws

PR #13897, opened 2018-04-27
python: import new patch to support libressl-2.7

pbui:
Patch based on this upstream commit:

https://github.com/python/cpython/commit/edd541897b9c28ee0d0f0131746aa5f19665a104


pbui:
Fixed now.

Issue #13450, opened 2018-04-13
[Proposal] change ext2/ext3 kernels config


cr6git:
Void kernels for example [linux4.15](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/linux4.15/files/x86_64-dotconfig#L7751) or [linux4.9](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/linux4.9/files/x86_64-dotconfig#L7255) have these options enabled:
````
CONFIG_EXT2_FS=m
CONFIG_EXT3_FS=m
````
I think we could just use `CONFIG_EXT4_USE_FOR_EXT2=y` instead.

As is the case with Arch kernels [linux](https://git.archlinux.org/svntogit/packages.git/tree/trunk/config?h=packages/linux#n8615) and [linux-lts](https://git.archlinux.org/svntogit/packages.git/tree/trunk/config?h=packages/linux-lts#n7535) for reference:
````
# CONFIG_EXT2_FS is not set
# CONFIG_EXT3_FS is not set
CONFIG_EXT4_FS=m
CONFIG_EXT4_USE_FOR_EXT2=y
````
The same for Debian kernels 4.15 and 4.9.

chneukirchen:
Wow, I think we had this already. Strong approve.

cr6git:
While we are at it, I have another question for you @chneukirchen 
You know we use this [smsc95xx-mac.patch](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/linux4.15/patches/smsc95xx-mac.patch) for our kernels.
However during the build there's this warning:
````
  CC [M]  drivers/net/usb/smsc95xx.o
drivers/net/usb/smsc95xx.c: In function 'smsc95xx_init_mac_address':
drivers/net/usb/smsc95xx.c:971:2: warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
  const u8 *mac_addr;
  ^~~~~
  CC [M]  drivers/net/usb/gl620a.o
````
What do you think? is that a problem? thanks.

chneukirchen:
The warning is harmless. The patch could be updated I guess to move the declaration before the first line of code.

pbui:
I've updated #13812 to include these changes.

PR #13899, opened 2018-04-27
linux4.4: update to 4.4.130.

cr6git:
[#12377](https://github.com/voidlinux/void-packages/pull/12377), [#13450](https://github.com/voidlinux/void-packages/issues/13450).


cr6git:
It might be best to wait till [4.4.130](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable-rc.git/) gets released.

PR #13903, opened 2018-04-27
php: update to 7.2.5.

maxice8:


DirectorX:
we may want add more php extensions... like memcache{,d} etc

PR #13909, opened 2018-04-27
cantata: update to 2.3.0.

Hoshpak:


srcpkgs/cantata/template
```diff
@@ -16,4 +16,4 @@ maintainer="LockeAnarchist <emanuel@openmailbox.org>"
 license="GPL-3"
```
maxice8: please use SPDX indetifier.

cr6git: `GPL-3.0-or-later`

cr6git:
@Hoshpak why you closed your PR ?

Hoshpak:
@cr6git I'm fed up with that stupid licensing stuff and don't want to invest any more time into amending the PRs.

PR #13919, opened 2018-04-28
sanoid: update to 1.4.18

emacsomancer:


maxice8:
https://github.com/voidlinux/void-packages/pull/13865

PR #13920, opened 2018-04-28
syslinux: Add /etc/kernel.d scripts.

travankor:


srcpkgs/syslinux/INSTALL.msg
```diff
@@ -4,3 +4,7 @@ Some optional packages must be installed for additional functionality:
  - gptfdisk: GPT support
  - efibootmgr: EFI support
  - dosfstools: EFI support
+
+Syslinux kernel.d hooks uses 'LABEL void' and 'LABEL voidfallback' to update kernel version for the latest kernel and last known working kernel, respectively.
+Do not use 'LINUX','KERNEL','INITRD', or 'initrd' for the MENU LABEL entry.
+See https://www.syslinux.org/wiki/index.php?title=SYSLINUX#How_do_I_Configure_SYSLINUX.3F for more information on how to setup syslinux.cfg.
```
newbluemoon: I think a line shouldn’t exceed a width of 80 characters.

travankor:
@newbluemoon Thanks, I fixed the formatting on the latest commit. Please let me know if there are any other issues.

I should mention that these scripts will not interfere with other boot menu options (ie: other Linux distros or Void installs with different boot parameters) on the syslinux.cfg file, so using syslinux should be as flexible as before.

PR #13921, opened 2018-04-28
wine: update to 3.7.

Hoshpak:


srcpkgs/wine/template
```diff
@@ -9,7 +9,7 @@ maintainer="Andrea Brancaleoni <abc@pompel.me>"
 license="LGPL-2.1"
 homepage="http://www.winehq.org/"
```
maxice8: winehq is now https

srcpkgs/wine/template
```diff
@@ -9,7 +9,7 @@ maintainer="Andrea Brancaleoni <abc@pompel.me>"
 license="LGPL-2.1"
```
maxice8: please use SPDX identifier

PR #13924, opened 2018-04-28
hplip: update to 3.18.4.

Hoshpak:


srcpkgs/hplip/template
```diff
@@ -30,7 +30,7 @@ maintainer="Juan RP <xtraeme@voidlinux.eu>"
 homepage="https://developers.hp.com/hp-linux-imaging-and-printing"
 license="GPL-2"
```
maxice8: please use SPDX identifier

PR #13936, opened 2018-04-28
libreoffice: remove old extensions

phili90:
These extensions are from 2012 and have been updated many times since. Including the extensions with the installation of the application conflicts with the internal extension manager. They do not update and if you attempt to install the newer version it fails.

Languagetool current version is 4.1
Mysql connector current version is 4.3

I have never done a pull request on a package before. I hope I did it right.

Thank you.


Johnnynator:
2 Minor things:
Use `packagename: change` as commit name so e.g. `libreoffice: remove old extensions`.
Also try to use only one commit per package.

phili90:
Thank you Johnnynator.

Johnnynator:
Also you can also remove the two corresonding lines in `skip_extraction="`

Johnnynator:
you can `git reset HEAD~3` to remove the last 3 commits, then `git commit srcpkgs/libreoffice/template` to recreate one single commit and the `git push -f` to force override whatever is in the repo on github.

Johnnynator:
Oh also run `xgensum srcpkgs/libreoffice/template` before that because the removal of the 2 distfiles made the checksums order wrong.

pullmoll:
It seems this info was wrong. I removed the two packages from the pre-download list and as you [can see in the build log](https://build.voidlinux.eu/builders/x86_64_builder/builds/9096/steps/shell_3/logs/stdio) libreoffice downloads exactly the same files on its own now: `b63e6340a02ff1cacfeadb2c42286161-JLanguageTool-1.7.0.tar.bz2` and `7239a4430efd4d0189c4f24df67f08e5-mysql-connector-c++-1.1.4.tar.gz`. This means that the revbump will effectively be useless :-P

PR #13947, opened 2018-04-29
Texlive2018

Johnnynator:
Updated with support for aarch64 and x86_64-musl.
The texlive installer doesn't detect musl so I add `-force-platform x86_64-linuxmusl` to the `INSTALL`.
Everything else is just copied form the 2017 template.

@chneukirchen


Piraty:
well done.
looks like texlive finally ships precompiled binaries for musl, [see here](https://www.tug.org/texlive/doc/texlive-en/texlive-en.html#x1-240003.2.1).
Probably worth to mention @svenper

Johnnynator:
> Would it be possible to make the paths relative in the profile or would this fuck up some other things?
I don't that would work properly, but I will take a look.

Johnnynator:
I updated the profile to use relative Path but haven't verified yet if it doesn't break runtime.

svenper:
Nice! `musl` support allows us to stay up do date with binaries more than once a year. Unless we really want dynamically linked binaries...

PR #13964, opened 2018-04-29
nodejs: update to 10.1.0.

jannisch:


jannisch:
Should be fine.

I once experienced an issue with npm downgrading itself but it does not seem to be related to this update after reinstalling and testing everything the last two weeks.

PR #13970, opened 2018-04-30
flare-engine: update to 1.04

diogoleal:


cr6git:
it closes [#13712](https://github.com/voidlinux/void-packages/pull/13712).

PR #13998, opened 2018-05-01
linux3.18: update to 3.18.108.

cr6git:
[#13450](https://github.com/voidlinux/void-packages/issues/13450)


cr6git:
EOL anyway...

PR #14001, opened 2018-05-01
xz: update to 5.2.4, fix license.

cr6git:


srcpkgs/xz/template
```diff
@@ -1,18 +1,15 @@
 # Template file for 'xz'
 pkgname=xz
-version=5.2.3
+version=5.2.4
 revision=1
 bootstrap=yes
 build_style=gnu-configure
 short_desc="The XZ compression utilities"
 maintainer="Juan RP <xtraeme@voidlinux.eu>"
 license="Public domain, GPL-2, GPL-3, LGPL-2.1"
```
maxice8: SPDX identifiers

PR #14002, opened 2018-05-01
[CI skip] Freecad and occt, vtk, med: various fixes

yopito:
Changes:
* `med`: store its .h files in `/usr/include/med` instead of polluting `/usr/include/`
* `occt`: fix musl patch of  `OSD_signal.cxx`: assign values to data as much as possible
*  fix check-update on various packages
* various small gliches


srcpkgs/med/template
```diff
@@ -1,19 +1,23 @@
...
 build_style=gnu-configure
-configure_args="--with-swig=yes --with-hdf5-bin=/usr/lib/hdf5-18/bin"
+configure_args="--with-swig=yes --with-hdf5-bin=/usr/lib/hdf5-18/bin
+ --includedir=/usr/include/med"
 hostmakedepends="gcc-fortran swig"
 makedepends="hdf5-18-devel python-devel tk-devel"
+depends="tk" # xmdump* are wish scripts
 short_desc="Data Modelization and Exchanges"
 maintainer="Piraty <piraty1@inbox.ru>"
 license="GPL-3, LGPL-3"
```
maxice8: SPDX identifiers

srcpkgs/med/template
```diff
@@ -1,19 +1,23 @@
...
 homepage="https://www.salome-platform.org"
 distfiles="http://files.salome-platform.org/Salome/other/med-${version}.tar.gz"
 checksum=dd631ef813838bc7413ff0dd6461d7a0d725bcfababdf772ece67610a8d22588
 
+nocross=yes # hdf5 is nocross
```
maxice8: no need for comment, you can make the value of nocross it.

srcpkgs/occt/template
```diff
@@ -11,13 +11,17 @@ makedepends="freetype-devel glu-devel freeimage-devel gl2ps-devel tbb-devel
  tcl-devel tk-devel"
 short_desc="OpenCASCADE Technology - library for CAD/CAM/CAE applications"
 maintainer="Piraty <piraty1@inbox.ru>"
-license="LGPL-2.1"
+license="LGPL-2.1,OCCT"
```
maxice8: SPDX identifiers

srcpkgs/vtk/template
```diff
@@ -35,7 +38,10 @@ homepage="https://www.vtk.org"
 distfiles="https://www.vtk.org/files/release/${version:0:3}/VTK-${version}.tar.gz"
 checksum=6e269f07b64fb13774f5925161fb4e1f379f4e6a0131c8408c555f6b58ef3cb7
 
+nocross=yes # hdf5 is nocross
```
maxice8: ditto

srcpkgs/vtk/template
```diff
@@ -35,7 +38,10 @@ homepage="https://www.vtk.org"
 distfiles="https://www.vtk.org/files/release/${version:0:3}/VTK-${version}.tar.gz"
```
maxice8: SPDX identifier is BSD-3-Clause

srcpkgs/vtk/template
Piraty: Isn't it enough to only mark hdf5 nocross, and not packages depending on it too? In a crossbuild, xbps-src would then fail on hdf5 ofc.

yopito:
Thanks for the review. Here my updated version.  
@Piraty : I use `nocross` as a shortcut, even if not strictly required.

yopito:
update:
* rebase from current `master` branch
* label with `[ci skip`]: freecad is too big for CI build

PR #14003, opened 2018-05-01
bird: update to 1.6.4

hanspolo:


srcpkgs/bird/template
```diff
@@ -12,7 +12,7 @@ hostmakedepends="flex"
 makedepends="ncurses-devel readline-devel"
 configure_args="--sbindir=/usr/bin"
```
maxice8: configure_args after build_style, hostmakedepends after configure_args, makedepends after hostmakedepends.

license should be SPDX identifier.

PR #14012, opened 2018-05-01
font-iosevka: update to 1.14.2.

diogoleal:


srcpkgs/font-iosevka/template
```diff
@@ -1,6 +1,6 @@
 # Template file for 'font-iosevka'
 pkgname=font-iosevka
-version=1.14.1
+version=1.14.2
 revision=1
 create_wrksrc=yes
 noarch="yes"
```
maxice8: don't need to be quoted

srcpkgs/font-iosevka/template
```diff
@@ -12,7 +12,7 @@ maintainer="Diogo Leal <diogo@diogoleal.com>"
 license="SIL Open Font License"
```
maxice8: please use SPDX identifier

PR #14019, opened 2018-05-01
New package: woff2-1.0.2.

jnbr:


srcpkgs/woff2/template
```diff
@@ -0,0 +1,26 @@
...
+post_install() {
+	vlicense LICENSE
+}
+
+woff2-devel_package() {
+	short-desc+=" - development files"
+	depends="${sourcepkg}>=${version}_${revision}"
+	pkg_install() {
+		vmove usr/include
+		vmove usr/lib/*.so
```
maxice8: must be quoted.

PR #14020, opened 2018-05-01
New package: bird2-2.0.2

hanspolo:


jnbr:
What do you think about running it under a different user than root (e.g. _bird)?

hanspolo:
This is a good idea.
I'm offline for some hours and will fix it afterwards or tomorrow

PR #14023, opened 2018-05-01
common: show value of nocross when it is set.

maxice8:
Manual.md says:

"This should be set to a string describing why it fails, or a link to
 a travis buildlog demonstrating the failure."

it should be ok to tell why it is nocross. at worse we can add a check that doesn't show if nocross is set to 'yes'


maxice8:
before merging there is something that needs to be resolved.

Should we check if nocross is set to 'yes' (the most common value) and don't show the message or should we declare nocross=yes to be broken and unleash someone to go around fixing it ?

PR #14026, opened 2018-05-01
python-Flask: update to 1.0.1

pulux:


srcpkgs/python-Flask/template
```diff
@@ -13,7 +13,7 @@ maintainer="Markus Berger <pulux@pf4sh.de>"
 homepage="http://flask.pocoo.org"
 license="3-clause-BSD"
```
maxice8: BSD-3-Clause, move it above homepage.

srcpkgs/python-Flask/template
```diff
@@ -1,7 +1,7 @@
 # Template file for 'python-Flask'
 pkgname=python-Flask
-version=0.12.2
-revision=2
+version=1.0.1
```
cr6git: Latest version is 1.0.2

pulux:
I think, we should wait for a stable version.

pulux:
for older software with maybe breaks with 1.0.x we should use first #14129 and version 1.0.x has many releases in the last 8 days.

PR #14027, opened 2018-05-01
epson-inkjet-printer-escpr: update to 1.6.20.

cr6git:


srcpkgs/epson-inkjet-printer-escpr/template
```diff
@@ -10,5 +10,5 @@ short_desc="Epson Inkjet Printer Driver (ESC/P-R)"
 maintainer="Juan RP <xtraeme@voidlinux.eu>"
 license="GPL-2"
```
maxice8: SPDX identifier

cr6git: changed to `license="GPL-2.0-or-later"`

Issue #13968, opened 2018-04-30
exfat-dkms needs catching up with upstream in order to work with linux4.16


datenwolf:
### System

* xuname:  Void 4.15.15_1 x86_64 GenuineIntel notuptodate rrrrmd

* package: exfat-dkms-1.2.8_3

### Expected behavior
Build against linux4.16

### Actual behavior
linux4.16 introduced a few minor, but breaking API changed. exfat-nofuse upstream already patched for this in https://github.com/dorimanx/exfat-nofuse/commit/01c30ad52625a7261e1b0d874553b6ca7af25966

### Steps to reproduce the behavior
Install linux4.16 then exfat-dkms-1.2.8_3, module compilation phase will fail.

PR #14037, opened 2018-05-02
exfat-dkms: bump commit for kernel 4.16+

cr6git:
[#13968](https://github.com/voidlinux/void-packages/issues/13968)


srcpkgs/exfat-dkms/template
```diff
@@ -1,15 +1,15 @@
...
 pkgname=exfat-dkms
 version=1.2.8
-revision=3
-_commit=de4c760bc9a05ead83bc3ec6eec6cf1fb106f523
+revision=4
+_commit=01c30ad52625a7261e1b0d874553b6ca7af25966
 wrksrc=exfat-nofuse-${_commit}
 short_desc="Exfat kernel driver (nofuse)"
 maintainer="Andrea Brancaleoni <abc@pompel.me>"
 license="GPL-2"
```
maxice8: SPDX identifier

srcpkgs/exfat-dkms/template
```diff
@@ -1,15 +1,15 @@
...
 homepage="https://github.com/dorimanx/exfat-nofuse"
 distfiles="$homepage/archive/$_commit.tar.gz"
```
maxice8: variables part of a bigger variable should be between brackets.

srcpkgs/exfat-dkms/template
```diff
@@ -1,15 +1,15 @@
 # Template file for 'exfat-dkms'
 pkgname=exfat-dkms
 version=1.2.8
-revision=3
-_commit=de4c760bc9a05ead83bc3ec6eec6cf1fb106f523
+revision=4
+_commit=01c30ad52625a7261e1b0d874553b6ca7af25966
 wrksrc=exfat-nofuse-${_commit}
```
maxice8: must be quoted

cr6git: why? is it new?

cr6git: Sorry, I don't see where it's marked? However you said: "Don't quote unless necessary."

maxice8: > a variable is being expanded

i mean the wrksrc= variable.

cr6git: hmm... not sure the quotes are really needed, but... it's okay, I will add quotes if you want...

maxice8: they aren't really needed in this case since we know that commit won't have any spaces but it is good hygiene to always quote variables being expanded.

srcpkgs/exfat-dkms/template
cr6git: changed to `license="GPL-2.0-or-later"`

cr6git:
Replaced by https://github.com/void-linux/void-packages/pull/276

PR #14039, opened 2018-05-02
docker: update to 18.04.0

hanspolo:


srcpkgs/docker/template
```diff
@@ -13,7 +13,7 @@ maintainer="Andrea Brancaleoni <abc@pompel.me>"
 license="Apache-2.0"
 homepage="http://www.docker.io"
```
maxice8: https://www.docker.com

srcpkgs/docker/template
```diff
@@ -13,7 +13,7 @@ maintainer="Andrea Brancaleoni <abc@pompel.me>"
...
 distfiles="https://github.com/docker/docker-ce/archive/v$_version.tar.gz"
```
maxice8: $_version should be between brackets

PR #14043, opened 2018-05-02
vivaldi: update to 1.15.1147.36.

diogoleal:


cr6git:
https://github.com/void-linux/void-packages/pull/444

PR #14046, opened 2018-05-02
cmus: add sndio output plugin

freshprince:


srcpkgs/cmus/template
```diff
@@ -1,7 +1,7 @@
 # Template file for 'cmus'
 pkgname=cmus
 version=2.7.1
-revision=3
+revision=4
 short_desc="Small, fast and powerful console music player"
 maintainer="Juan RP <xtraeme@voidlinux.eu>"
 license="GPL-2"
```
maxice8: please fix the license to use a SPDX indentifier

srcpkgs/cmus/template
```diff
@@ -12,7 +12,8 @@ checksum=8179a7a843d257ddb585f4c65599844bc0e516fe85e97f6f87a7ceade4eb5165
 hostmakedepends="pkg-config"
 makedepends="ncurses-devel faad2-devel libflac-devel libao-devel libmad-devel
  libmodplug-devel libmp4v2-devel libmpcdec-devel pulseaudio-devel
- opusfile-devel wavpack-devel alsa-lib-devel libvorbis-devel ffmpeg-devel"
```
maxice8: move hostmakedepends and makedepends below revision

maxice8:
Paging the sndio man @Duncaen

srcpkgs/cmus/template
freshprince: Thanks for your feedback but I think I don't understand.  `revision` is in line 4 and `depends` are in line 12 and 13 and clearly below it?

maxice8:
I mean put the directly below

srcpkgs/cmus/template
freshprince: I changed it, but I'm a bit confused now. The [manual](https://github.com/voidlinux/void-packages/blob/master/Manual.md) states that `license` should be

> A string matching any license file available in `/usr/share/licenses`.

And there is no `GPL-2.0` in `/usr/share/licenses` but there is `GPL-2`.

maxice8: yeah sorry we are in transition to adopt SPDX identifiers for licenses, see https://github.com/voidlinux/void-packages/issues/12219

the manual.md is outdated in that regard

srcpkgs/cmus/template
```diff
@@ -1,18 +1,19 @@
 # Template file for 'cmus'
 pkgname=cmus
 version=2.7.1
-revision=3
+revision=4
 short_desc="Small, fast and powerful console music player"
 maintainer="Juan RP <xtraeme@voidlinux.eu>"
-license="GPL-2"
+license="GPL-2.0"
```
maxice8: the SPDX identifier for GPL licenses need -or-later or -only.

srcpkgs/cmus/template
maxice8: i mean directly below revision.

freshprince:
moving to new org

PR #14059, opened 2018-05-02
linux-firmware: update to 20180518.

cr6git:
Using the same commit as [Arch](https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/linux-firmware).


cr6git:
Replaced by https://github.com/void-linux/void-packages/pull/274

PR #14060, opened 2018-05-02
New package: bluez-alsa-1.2.0.

maxice8:


Vaelatern:
Closes #14055 
I'll test locally, see if I can make my audio work.

srcpkgs/ortp/template
```diff
@@ -0,0 +1,24 @@
+# Template file for 'ortp'
+pkgname=ortp
+version=0.27.0
```
Johnnynator: ORTP has a 1.0.2 release https://github.com/BelledonneCommunications/ortp/releases

Johnnynator:
Also both ortp and bctoolbox do generate a .cmake file that are more or less broken. The ones generated while cross compiling can not be used later natively and vice versa.

srcpkgs/bluez-alsa/template
```diff
@@ -0,0 +1,23 @@
...
+distfiles="https://github.com/Arkq/bluez-alsa/archive/v${version}.tar.gz"
+checksum=0a7a27e8cf135ec6bb30ea1a2365736681ced7d8293003c274fa78319f936838
+
+pre_configure() {
+	autoreconf -fi
+}
+
+post_install() {
+	vlicense LICENSE.txt
+}
```
Vaelatern:
```
post_install() {
        vlicense LICENSE.txt
        vsv bluez-alsa
}
```

```
» cat srcpkgs/bluez-alsa/files/bluez-alsa/run
#!/bin/sh

dir=/var/run/bluealsa
user=_bluez_alsa
group=_bluez_alsa

[ ! -d "${dir}" ] && mkdir "${dir}" && chown $user:$group ${dir}
exec chpst -u $user:$group bluealsa
```

```
system_accounts="_bluez_alsa"
```

```
makedepends="libbluetooth-devel alsa-lib-devel glib-devel sbc-devel
 $(vopt_if aac fdk-aac-devel)
 $(vopt_if hcitop "ncurses-devel libbsd-devel")"

build_options="hcitop debug aac"
build_options_default="hcitop debug aac"
```

But these did come from a template where I chose `version=1.2.0.20180402` so up to you

maxice8: hcitop fails to detect ncurses-devel even when it is installed, that is why i disabled it, i also have a dislike for build_options.

maxice8: but yeah i added the service, system_account and enabled --debug by default.

maxice8:
i disabled debug since it requires execinfo.h

maxice8:
oh well, bctoolbox requires CMake to have a upstream-supported workaround for execinfo.h usage, but CMake also produces broken .cmake and .pc files

maxice8:
@Johnnynator can you take a look into fixing *-musl with gnu-configure or making cmake generate correct files ? (the latter is preferred but take what you want)

Johnnynator:
Just use cmake for now, it works as long as you use xbps-src only, and arm stuff always gets cross compiled. I will fix their stuff after my exams. The .pc files can proably be rewritten manually in pre_install or so, fixing cmake isn't easy and will probably end in a few hundreds LOC's. 
The best way for fixing it is probably doing Smth similar as qca-qt5 is doing with their RELATIVE_PATHS cmake option.

maxice8:
@Vaelatern can i put you as maintainer= ?

Vaelatern:
Sure

maxice8:
On Tue, May 29, 2018 at 07:58:39AM -0700, Toyam Cox wrote:
> Sure
> 
> -- 
> https://github.com/voidlinux/void-packages/pull/14060#issuecomment-392808470

ty

PR #14061, opened 2018-05-02
octave: update to 4.4.0

abenson:


srcpkgs/octave/template
```diff
@@ -13,7 +13,7 @@ maintainer="Diogo Leal <diogo@diogoleal.com>"
 license="GPL-3"
```
maxice8: please use spdx identifiers

Piraty:
I run this already, works good

PR #14062, opened 2018-05-02
linux4.16: update to 4.16.7.

cr6git:


cr6git:
@chneukirchen for 4.16 we must **re-enable** these options:
````
# CONFIG_MISC_RTSX is not set
# CONFIG_MISC_RTSX_PCI is not set
# CONFIG_MISC_RTSX_USB is not set
````
We discussed this on the forum.

chneukirchen:
I don't read the forum. Feel free to provide a PR.

PR #14063, opened 2018-05-02
gegl: update to 0.4.0

kartik-ynwa:


srcpkgs/gegl/template
```diff
@@ -1,20 +1,20 @@
...
 configure_args="--disable-docs --without-sdl --without-libopenraw
  --without-libavformat --without-graphviz --without-lua --without-libspiro
  --without-umfpack --with-exiv2 --without-libv4l --disable-introspection"
 hostmakedepends="pkg-config intltool"
 makedepends="libglib-devel libpng-devel babl-devel jasper-devel
- libjpeg-turbo-devel librsvg-devel gtk+-devel exiv2-devel"
+ libjpeg-turbo-devel librsvg-devel gtk+-devel exiv2-devel json-glib-devel"
 short_desc="Graph based image processing framework"
 maintainer="Juan RP <xtraeme@voidlinux.eu>"
 license="GPL-3, LGPL-3"
```
maxice8: use SPDX indentifiers

kartik-ynwa:
Done and done.

kartik-ynwa:
@biopsin The template that I have for GIMP 2.10.0 does not have `gobject-introspection` as a dependency and it builds and works fine. Are you sure you couldn't build GIMP without it?

kartik-ynwa:
@biopsin Here is the link to the template so you can try it out yourself; [GIMP template](https://github.com/kartik-ynwa/void-packages/blob/gimp/srcpkgs/gimp/template)

`gobject-introspection` was used to build `gegl-0.4.0` but not for `gimp-2.10.0`.

kartik-ynwa:
@biopsin Yes it is likely. `libmypaint` needs `gobject-introspection`.

biopsin:
Went down your path and seems --disable-introspection works out, however
one issue in the gegl-devel_package()... vmove usr/lib/*.so ; gimp will not install and complain about unresolvable shlib libgegl-npd-0.4.so and libgegl-sc-0.4.so.
Changing it to ... vmove usr/lib/libgegl-0.4.so solves the issue and gegl-devel will no longer show warning about missing libgegl-npd-0.4.so and libgegl-sc-0.4.so in common/shlibs

kartik-ynwa:
@biopsin bro you are literally a god. I was wondering what was wrong. It made that mistake when I redid the template. Thanks for pointing it out.

newbluemoon:
@kartik-ynwa I think the (real) problem is that you have to use quotes:

    vmove "usr/lib/*.so"

;)

kartik-ynwa:
@biopsin https://github.com/kartik-ynwa/void-packages/tree/gimp_install

This branch builds and installs GIMP successfuly.

kartik-ynwa:
@newbluemoon Oooh that never crossed my mind. Thanks for that. I will keep in mind next time I write bash step.

biopsin:
@newbluemoon : complains again
=> WARNING: gegl-devel-0.4.0_1: libgegl-npd-0.4.so not found in common/shlibs!
=> WARNING: gegl-devel-0.4.0_1: libgegl-sc-0.4.so not found in common/shlibs!

biopsin:
@kartik-ynwa Good job! last thing in you gimp branch bump common/shlibs for all
libgimp-2.0.so.0 libgimp-2.8.10_3 to 2.10.0_1..

newbluemoon:
Ok, sorry, I got it the wrong way and assumed `libgegl-npd-0.4.so` and `libgegl-sc-0.4.so` were symlinks (missing in gegl-devel), but now I see they’re the actual (unversioned) libraries. In this case they of course don’t have to be moved to the -devel subpackage (only `*.so` symlinks have to). :)

PR #14065, opened 2018-05-02
New package: mypaint-brushes-1.3.0

kartik-ynwa:
Dependency for GIMP 2.10.0


srcpkgs/mypaint-brushes/template
```diff
@@ -0,0 +1,21 @@
+# Template file for 'mypaint-brushes'
+pkgname=mypaint-brushes
+version=1.3.0
+revision=1
+build_style=gnu-configure
+hostmakedepends="automake"
+makedepends="libmypaint-devel"
+short_desc="Brushes used by MyPaint and other software using libmypaint"
+maintainer="Kartik Singh <kartik.ynwa@gmail.com>"
+license="CC0-1.0, GPL-2.0, GPL-3.0"
```
maxice8: use SPDX identifiers

kartik-ynwa: Should it be "GPL-2.0-only" and "GPL-3.0-only"? Using this as reference: https://spdx.org/licenses/

maxice8: They should be either -or-later or -only depending on the header files in the source code.

maxice8: on this specific case it is GPL-2.0-or-later and GPL-3.0-only

kartik-ynwa: Aight done

kartik-ynwa:
Also it needs libmypaint to build which I have added in another PR. I didn't have the foresight and requested them separately. Do you think I should move `mypaint-brushes` to the PR for `libmypaint`?

PR #14091, opened 2018-05-03
vopt_onoff: rename to vopt_bool.

maxice8:
true or false is more generic and allows us to use -D\<prop>={true|false} on meson build_style as well.


Gottox:
Naming of the function is wrong then, but I think this could be an okayish tradeoff.

maxice8:
On Thu, May 03, 2018 at 09:13:09AM -0700, Enno Boland wrote:
> Naming of the function is wrong then, but I think this could be an okayish
> tradeoff.
> 
> —
> 

naming is wrong if you only consider cmake -DPROP=ON|OFF you can also think of it as vopt turn on or turn off a feature.

anyways that doesn't matter much and the change is completely invisible for the user unless there is some crazy cmake invocation that doesn't accept standard boolean names.

ebfe:
Having a build_style specific function named vopt_bool is entirely confusing.

Gottox:
As it is currently unused, we can change that.

How about change to usage to `-D$(vopt_bool XXX)`?

Gottox:
or even `-Doption=$(vopt_bool XXX)`

maxice8:
> -D$(vopt_bool XXX)

i'd take this one
 
`-D$(vopt_bool wayland wayland)` | srcpkgs/enlightenment/template

i'd go as far as change it so if <property> is not given it will use <option>

`-D$(vopt_bool wayland)` == `-D$(vopt_bool wayland wayland)`

PR #14092, opened 2018-05-03
flare-engine: update to 1.05

diogoleal:


cr6git:
I took the liberty to update them: https://github.com/void-linux/void-packages/pull/374

PR #14102, opened 2018-05-03
handle changelog.

maxice8:
based on https://github.com/voidlinux/void-packages/pull/4068 , just added an entry to Manual.md


Vaelatern:
Please fix commit message

maxice8:
Will do soon, waiting for my big fat burger in a fast door queue

maxice8:
done

ebfe:
Why does the changelog have to be in plain text? That's impractical and a useless restriction.
This change doesn't make sense without xbps 0.52 you need to bump XBPS_VERSION_REQ.

Vaelatern:
What other than plain text would it be?

Vaelatern:
1441834458

Duncaen:
Well its clear that you don't supply a change log link as binary, the plaintext indicates a url should point to a address that has the content type `text/plain`, which is in many cases not possible.

PR #14103, opened 2018-05-03
libao: update to 1.2.2.

Chocimier:


srcpkgs/libao/template
```diff
@@ -10,9 +10,9 @@ conf_files="/etc/libao.conf"
 homepage="http://xiph.org/ao/"
```
maxice8: homepage should be above distfiles= and the page is now https://

Chocimier:
Is there a tool to check for these?

maxice8:
nope

maxice8:
There isn't at the moment but one can make and contribute to xlint.

PR #14107, opened 2018-05-03
deluge-gtk: patch settings dialog not working for version 1.3.15

jtl999:
https://forum.deluge-torrent.org/viewtopic.php?f=7&t=54517


maxice8:
revbump the package so it is rebuilt.

jtl999:
I Just revbumped and rebased said commit

PR #14116, opened 2018-05-04
New package: tmate-2.2.1

RRvW:


srcpkgs/tmate/template
```diff
@@ -0,0 +1,21 @@
+# Template file for 'tmate'
+pkgname=tmate
+version=2.2.1
+revision=1
+build_style=gnu-configure
+makedepends="libevent-devel ncurses-devel msgpack-devel libssh-devel"
+hostmakedepends="automake pkg-config"
```
maxice8: hostmakedepends should be directly above makedepends.

makedepends can be minimized with `xmindep` to libevent-devel libssh-devel msgpack-devel ncurses-devel

RRvW:
Updated

PR #13787, opened 2018-04-24
frp: update to 0.18.0.

cr6git:

PR #14122, opened 2018-05-04
frp: update to 0.20.0.

cr6git:
closes [#13787](https://github.com/voidlinux/void-packages/pull/13787)


cr6git:
Replaced by https://github.com/void-linux/void-packages/pull/281

Issue #14118, opened 2018-05-04
Spotify 1.0.72_1 install fails


TehDaniel37:
### System

* xuname:  
  `Void 4.16.6_1 x86_64 GenuineIntel notuptodate hold rrDF`
* package:  
  `spotify 1.0.72_1`

### Expected behavior
Install spotify.

### Actual behavior
```bash
$ sudo xbps-install -Sv spotify   
[*] Updating `https://repo.voidlinux.eu/current/x86_64-repodata' ...
x86_64-repodata: 1579KB [avg rate: 33GB/s]
[*] Updating `https://repo.voidlinux.eu/current/nonfree/x86_64-repodata' ...
Found spotify-1.0.72_1 (install) in repository https://repo.voidlinux.eu/current/nonfree

Name    Action    Version           New version            Download size
spotify install   -                 1.0.72_1               - 

Size required on disk:          229B
Space available on disk:        16GB

Do you want to continue? [Y/n] y

[*] Downloading binary packages

[*] Verifying package integrity
spotify-1.0.72_1: verifying RSA signature...

[*] Running transaction tasks
spotify-1.0.72_1: unpacking ...
spotify-1.0.72_1: unpacked file `./usr/share/spotify/pkgdata' (132 bytes)
spotify-1.0.72_1: unpacked file `./usr/bin/spotify' (97 bytes)

[*] Configuring unpacked packages
spotify-1.0.72_1: configuring ...
looking up repository.spotify.com
connecting to repository.spotify.com:80
requesting http://repository.spotify.com/pool/non-free/s/spotify-client/spotify-client_1.0.72.117.g6bd7cc73-35_amd64.deb
http://repository.spotify.com/pool/non-free/s/spotify-client/spotify-client_1.0.72.117.g6bd7cc73-35_amd64.deb: Not Found
Failed downloading spotify client
ERROR: spotify-1.0.72_1: [configure] INSTALL script failed to execute the post ACTION: Operation not permitted
```

### Steps to reproduce the behavior
`# xbps-install -Sv spotify`

Unfortunately I do not have an earlier version of spotify in my cache, so I can't test if this only happens with the latest version.

abenson:
See #13427 ... it hasn't been updated since the versions are different between the two. i686 is still on .72 but x86_64 is up to .79.

cr6git:
In my last pull request I proposed to remove the 32-bit version from the original template, and create a new package `spotify-32bits` for `i686` only.

[...]

Issue #13427, opened 2018-04-13
spotify i686 and x86_64 out of sync


abenson:
Package template for `spotify` is at version 1.0.72.

Versions available as of late March are 1.0.72 for i686 and 1.0.77 for x86_64.

~~~
spotify-client_1.0.72.117.g6bd7cc73-35_i386.deb    27-Feb-2018 13:30            95135550
spotify-client_1.0.77.338.g758ebd78-41_amd64.deb   26-Mar-2018 11:11            96717978
~~~

Can the version different between archs in the template, or do we drop i686, or what?

cr6git:
Note: I proposed this update [#13244](https://github.com/voidlinux/void-packages/pull/13244).

Johnnynator:
The version can be different as long as no  other template depends on that package. E.g. it is fine here for Spotify but e.g. it is error prone in the Spotify SDK library.

PR #14124, opened 2018-05-04
spotify: update to 1.0.79, new package: spotify-32bits-1.0.72

cr6git:
[#13244](https://github.com/voidlinux/void-packages/pull/13244), [#13427](https://github.com/voidlinux/void-packages/issues/13427), [#14118](https://github.com/voidlinux/void-packages/issues/14118).


cr6git:
Problem: https://travis-ci.org/voidlinux/void-packages/jobs/374914994#L672
````
$ common/travis/build.sh $BOOTSTRAP $ARCH
ERROR: requires xbps>=0.52
Bootstrap packages must be updated with 'xbps-src bootstrap-update'
````

File: srcpkgs/spotify/template [hunk 1]
```diff
@@ -1,28 +1,22 @@
 # Template file for 'spotify'
 pkgname=spotify
-version=1.0.72
+version=1.0.79
 revision=1
+_sversion=".223.g92622cc2-21_amd64.deb"
```
Thunderbottom: you need to remove `.deb` since the `INSTALL` file already has it appended, and having `.deb` in here causes it to be 'not found'

srcpkgs/spotify-32bits/INSTALL
```diff
@@ -0,0 +1,41 @@
...
+	ln -sf "/usr/lib/${_LIB}" "/usr/share/spotify/${1}.so.${2}"
+}
+
+if test "$ACTION" = "post"; then
+	. /usr/share/spotify/pkgdata
+	mkdir -p "$_BUILDDIR"
+	(
+		set -e
+		cd "$_BUILDDIR"
+		xbps-uhelper fetch "http://repository.spotify.com/pool/non-free/s/spotify-client/spotify-client_${SVERSION}.deb"
```
Thunderbottom: as you can see, it appends `.deb` in here, so the one in the template isnt needed.

srcpkgs/spotify/template
cr6git: good catch!

Thunderbottom: :)

srcpkgs/spotify-32bits/INSTALL
cr6git: it's done.

PR #14129, opened 2018-05-05
python-Flask: update to 0.12.4

pulux:


cr6git:
your previous PR [#14026](https://github.com/voidlinux/void-packages/pull/14026) was good, you shouldn't have closed it.

PR #14133, opened 2018-05-05
docker-compose: update to 1.21.2

iaroki:


srcpkgs/docker-compose/template
```diff
@@ -15,7 +15,7 @@ maintainer="pancake <pancake@nopcode.org>"
 homepage="https://docs.docker.com/compose/"
 license="Apache-2.0"
```
maxice8: license= directly above homepage.

PR #14141, opened 2018-05-05
nfdump: remove unnecessary shlib

abenson:


maxice8:
change adopted on https://github.com/voidlinux/void-packages/pull/13749

PR #14144, opened 2018-05-05
rpm: update to 4.14.1.

Chocimier:


maxice8:
Please make sure `rinse` is still working as expected as RPM was packaged for it.

PR #14147, opened 2018-05-05
cava: update to 0.6.1.

maxice8:
this adds support for sndio 

@Duncaen mind testing if it works ? sndio doesn't work for me at all.


Vaelatern:
Works with sndio, but I can't get it to input from my output, only from my microphone
Needs various example files installed, I'll take care of that.
sndio seems undocumented from cava's side, but it does link somehow.

maxice8:
On Sun, May 06, 2018 at 09:59:29PM +0000, Toyam Cox wrote:
> Works with sndio, but I can't get it to input from my output, only from my
> microphone
> Needs various example files installed, I'll take care of that.
> sndio seems undocumented from cava's side, but it does link somehow.
> 

thanks.

Vaelatern:
Couldn't get sndio to work quite, but the rest works as before, + example files now installed

PR #14148, opened 2018-05-05
scribus: update to 1.4.7.

Piraty:


Piraty:
moved: https://github.com/void-linux/void-packages/pull/3

PR #14151, opened 2018-05-06
python3-async-timeout: update to 3.0.0.

maxice8:


Gottox:
```
# xbps-query -RX python3-async-timeout
python3-aiohttp-3.1.3_1
```
As this is a major release: does this have implications on the dependant?

PR #14154, opened 2018-05-06
mergerfs: update to 2.24.0.

Hoshpak:


srcpkgs/mergerfs/template
```diff
@@ -1,16 +1,16 @@
 # Template file for 'mergerfs'
 pkgname="mergerfs"
```
maxice8: pkgname= shouldn't be quoted.

Issue #6977, opened 2017-07-07
opensmtpd: conflicts with dma, postfix


ebfe:
All 3 ship:
```
/usr/bin/mailq                                                                                                                                                                                                                
/usr/bin/sendmail  
```

ebfe:
Overlooked `provides=`

Gottox:
But does provides really prevent you from installing both?

ebfe:
Good question, maybe it is the `replaces=` actually. Having opensmtpd installed xbps-installs removes it when trying to install postfix or dma.

opensmtpd:
 ```
provides="smtp-forwarder-0_1 smtp-server-0_1"
replaces="smtp-forwarder>=0 smtp-server>=0"
```

dma:
```
provides="smtp-server-0_1 smtp-forwarder-0_1"
replaces="smtp-server>=0 smtp-forwarder>=0"
```

postfix:
```
provides="smtp-server-0_1"
replaces="smtp-server>=0"
```

chneukirchen:
IMO a perfect candidate for alternatives.

ebfe:
The behaviour of `replaces` `provides` combo is a bit confusing. It will replace an already installed package, but still allows to install several of them when installed explicitely in the same transaction. That is `xbps-install dma postfix` will work as long an none of them is installed yet.

PR #14169, opened 2018-05-06
common/travis: check for file conflicts with other packages

Gottox:
Checks for file conflicts with other packages. Currently untested.


common/travis/show_files.sh
```diff
@@ -8,10 +8,25 @@ if [ "$1" != "$2" ]; then
...
+xlocate -S
+rv=0
 for pkg in $(cat /tmp/templates); do
 	for subpkg in $(xsubpkg $pkg); do
 		/bin/echo -e "\x1b[32mFiles of $subpkg:\x1b[0m"
-		./xbps-src -H $HOME/hostdir $arch show-files "$subpkg"
+		files=$(./xbps-src -H "$HOME/hostdir" "$arch" show-files "$subpkg")
+		echo "$files"
+
+		for file in $files; do
```
Vaelatern: Does this break on files with spaces in the name?

`scratch-1.4.0.7_1       /usr/share/scratch/Projects/Speak Up/1 SmokingKills.sb`

ebfe:
xlocate isn't reliable and only works on x86_64.

Gottox:
@ebfe other suggestions?

ebfe:
Running this on archs != x86_64 will create false positives and only running it
on travis still won't guarantee that no new file conflicts are added. It also
ignores existing (conflicts=/replaces=/reverts=) properties so i don't see
what's the point of runnning it at all. Evaluating these is hard since it isn't always clear what packages can actually be installed at the same time (also see https://github.com/voidlinux/void-packages/issues/11804, https://github.com/voidlinux/xbps/issues/257, https://github.com/voidlinux/void-packages/issues/6977#issuecomment-313810733)

> @ebfe other suggestions?

This has been discussed countless times on IRC, since someone will press the
merge button anyways it's pointless to reiterate here. In short you need 
xbps to handle this.

Vaelatern:
So how would this be done in xbps?

Vaelatern:
How about generating the xlocate db for all our archs?

Gottox:
@ebfe: What do you think about running it on x86 only as long as we don't have support in xbps?

newbluemoon:
> How about generating the xlocate db for all our archs?

But is https://github.com/chneukirchen/xtools/issues/83 solved?

PR #14170, opened 2018-05-06
Plymouth: add udev seat rules

dan-santana:
Currently `plymouth` doesn't work if `elogin` is not installed, as it requires graphical devices tagged as "seat" and dracut will only include the udev rules from `71-seat.rules` (included only in `elogind` package). If it's not installed, Plymouth only runs in text mode.

This patch also includes the ConsoleKit2 udev seat rules to initramfs, as it's named differently (`71-udev-seat.rules`).


dan-santana:
Maybe plymouth install/removal should also trigger `dracut` to recreate the initramfs.

dan-santana:
I tested my setup with and without both of them, and had no issues.

PR #14172, opened 2018-05-06
xapps: update to 1.2.1.

maxice8:


cr6git:
I need this to update [xreader](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/xreader/template).

PR #14173, opened 2018-05-06
New package: PopCorn-0.2

the-maldridge:
The service at popcorn.voidlinux.org isn't live yet, that doesn't even resolve.  This came up out of a discussion on #voidlinux yesterday that it would be cool to have opt-in stats for people that want to contribute.  It might also help us garbage collect old packages over time if we see that they aren't in use.

Quick overview of how this will work:
popcorn.voidlinux.org points to the main build server, which will run the statrepo.  Once per day it will dump the aggregated stats into a directory that is world visible served by nginx.  The logs will get dumped somewhere by rsvlogd and eventually I might turn it into a heat map, but this project is already kind of done as far as I'm concerned.


srcpkgs/PopCorn/template
```diff
@@ -0,0 +1,44 @@
+x# Template file for 'PopCorn'
```
maxice8: extra x ?

the-maldridge: yep, fixed and re-pushed.  Kind of surprised xlint didn't catch that

maxice8: xlint doesn't check for that, maybe it could check for `^# Template file for '$pkgname'`

PR #14179, opened 2018-05-07
grafx2: update to 2.5.

maxice8:


Gottox:
Dup of #14238, @maxice8 agreed to close

PR #14186, opened 2018-05-07
ckb-next: readd sv script

dan-santana:


Gottox:
Hmm... This update will cause ckb to be stopped. Is there a certain reason for renaming?

dan-santana:
The last release renamed all binaries (`ckb` → `ckb-next`) so I renamed the service to keep it consistent. As the `ckb-next-daemon` is *almost* useless without the `ckb-next` client, everyone using this package would have to rename it anyway.

PR #14190, opened 2018-05-07
libfilezilla: update to 0.12.2.

maxice8:


bra1nwave:
I added this `libfilezilla` update here: https://github.com/voidlinux/void-packages/pull/13313

maxice8:
looks good, i'll see if i can make the tests works

PR #14191, opened 2018-05-07
roxterm: update to 3.6.1.

cr6git:


srcpkgs/roxterm/template
```diff
@@ -1,23 +1,14 @@
...
-do_configure() {
-	python2 mscript.py configure --prefix=/usr --cc="$CC" --cpp="$CPP" --cxx="$CXX" --ld="$LD"
-}
-do_build() {
-	python2 mscript.py build --cc="$CC" --cpp="$CPP" --cxx="$CXX" --ld="$LD" --cflags="$CFLAGS" --ldflags="$LDFLAGS"
-}
-do_install() {
-	python2 mscript.py install --destdir=${DESTDIR}
-}
+license="GPL-2.0-or-later, LGPL-3.0"
```
maxice8: LGPL-3.0 is not valid

cr6git: Are you sure? I can see `LGPL-3.0` it in the SPDX License List.

PR #14192, opened 2018-05-07
gitea: update to 1.4.1

iaroki:


srcpkgs/gitea/template
```diff
@@ -28,7 +28,7 @@ maintainer="Michael Aldridge <maldridge@VoidLinux.eu>"
 license="MIT"
 homepage="http://gitea.io"
```
maxice8: gitea.io uses https now

lemmi:
@maxice8 can you add the changelog? https://github.com/go-gitea/gitea/blob/master/CHANGELOG.md

maxice8:
Currently at uni, will be able to when I get home around 23:30 GMT-3

maxice8:
@lemmi derp, this is not my PR, forgot

lemmi:
@iaroki  can you add the changelog to the template? 
```changelog="https://github.com/go-gitea/gitea/blob/master/CHANGELOG.md"```

iaroki:
Added changelog field

PR #14194, opened 2018-05-07
frp: update to 0.18.0

iaroki:


cr6git:
[#14122](https://github.com/voidlinux/void-packages/pull/14122)

PR #14201, opened 2018-05-07
New package: gigolo-0.4.2

MikeWortin:


srcpkgs/gigolo/template
```diff
@@ -0,0 +1,13 @@
...
+pkgname=gigolo
+version=0.4.2
+revision=1
+build_style=gnu-configure
+makedepends="intltool gtk+-devel gvfs-devel pkg-config"
+depends="gtk+ gvfs"
+short_desc="Frontend to manage connections to remote filesystems using GIO/GVFS"
+maintainer="Mike Wortin <mikewortin@gmail.com>"
+license="GPL-2.0-only"
+homepage="http://www.uvena.de/gigolo"
```
Johnnynator: `https`, `http` doesn't work for this site

srcpkgs/gigolo/template
```diff
@@ -0,0 +1,13 @@
+# Template file for 'gigolo'
+pkgname=gigolo
+version=0.4.2
+revision=1
+build_style=gnu-configure
+makedepends="intltool gtk+-devel gvfs-devel pkg-config"
+depends="gtk+ gvfs"
```
Johnnynator: Can probably removed since xbps-src does auto detect lib that are linked against.

srcpkgs/gigolo/template
```diff
@@ -0,0 +1,13 @@
+# Template file for 'gigolo'
+pkgname=gigolo
+version=0.4.2
+revision=1
+build_style=gnu-configure
+makedepends="intltool gtk+-devel gvfs-devel pkg-config"
```
Johnnynator: `intltool` and `pkg-config` should be in `hostmakedepends`

srcpkgs/gigolo/template
Johnnynator:
```
> gigolo-0.4.2_1: running pre-pkg hook: 04-generate-runtime-deps ...
   SONAME: libgtk-x11-2.0.so.0 <-> gtk+>=2.16.0_1
   SONAME: libgdk-x11-2.0.so.0 <-> gtk+>=2.16.0_1
   SONAME: libgio-2.0.so.0 <-> glib>=2.18.0_1
   SONAME: libgobject-2.0.so.0 <-> glib>=2.18.0_1
   SONAME: libglib-2.0.so.0 <-> glib>=2.18.0_1
   SONAME: libX11.so.6 <-> libX11>=1.2_1
   SONAME: libc.so.6 <-> glibc>=2.26_1
```
glib, gtk+, libX11 are auto detected. If gvfs is needed at runtime then let it be the only manual specified dependency.

MikeWortin:
fixed all. now it will probably compile on all arches/libcs

Johnnynator:
Template looks good now.
Also try to use only one commit per package.
```
git reset HEAD^^
xbump gigolo
git push -f
```

PR #14242, opened 2018-05-08
build-style go: always download the package

lemmi:
[MANUAL.md](https://github.com/voidlinux/void-packages/blob/master/Manual.md#pkgs_go) says:

"If set to yes, the package specified via go_import_path will be
 downloaded with go get."

What's a little closer to the truth is that we skip downloading the
distfiles and so when calling `go get` in the build stage the go
toolchain will download the package.

But this way we don't have the source in previous stages to operate on.

From now, when `go_get=yes` is set, the package is downloaded with
`go get -d` and the wrksrc symlink is always set.

Note that this feature is currently unused.


ebfe:
Alternatively adding a do_fetch/do_extract to the build_style that fetches but doesn't build `(go get -n ... might work) if `go_get=yes`. That way the common patch infrastructure would still work. Though overriding these probably has some pitfalls of it's own.

common/build-style/go.sh
```diff
@@ -6,10 +6,11 @@ do_configure() {
         # This isn't really configuration, but its needed by packages
         # that do unusual things with the build where the expect to be
         # able to cd into the $GOSRCPATH
-	if [[ "${go_get}" != "yes" ]]; then
-		mkdir -p "$(dirname ${GOSRCPATH})"
-		ln -fs $PWD "${GOSRCPATH}"
+	if [[ "${go_get}" == "yes" ]]; then
+		go get -d -x tags "${go_build_tags}" ${go_import_path}
```
ebfe: -tags

PR #14250, opened 2018-05-08
goffice: update to 0.10.41

hanspolo:


maxice8:
Just amend the commits your previous Pull request to update goffice to 0.10.41, no need to create a new one.

PR #14254, opened 2018-05-09
imv: update to 3.0.0.

maxice8:


srcpkgs/imv/template
```diff
@@ -1,18 +1,23 @@
...
 revision=1
 build_style=gnu-makefile
-make_build_args="V=1"
-makedepends="SDL2_ttf-devel freeimage-devel fontconfig-devel"
+hostmakedepends="pkg-config asciidoc"
+makedepends="SDL2_ttf-devel fontconfig-devel freeimage-devel"
 short_desc="Image viewer for X11/Wayland"
 maintainer="Arvin Ignaci <arvin.ignaci@gmail.com>"
-license="GPL-2"
+license="MIT"
```
cr6git: https://github.com/eXeC64/imv/blob/master/LICENSE
````
imv is dual-licensed under the MIT license (see LICENSE.MIT), and the
GNU General Public License version 2 or later (see LICENSE.GPL).
````

PR #14255, opened 2018-05-09
Fix dependency on libressl-devel instead of libressl

DBLouis:


Gottox:
Can you state the problem with having libressl without -devel?

DBLouis:
It does not compile and throw an error because of missing headers. Templates for other version of the kernel ([linux4.4](/voidlinux/void-packages/tree/master/srcpkgs/linux4.4/template), [linux4.14](/voidlinux/void-packages/tree/master/srcpkgs/linux4.14/template)) have libressl-devel as dependency, not libressl.

PR #14270, opened 2018-05-09
Fix sfml-devel package

Hoshpak:
CC @thypon


maxice8:
While you're at it please fix the license= it is Zlib

PR #14294, opened 2018-05-10
New package: netbsd-curses-0.2.1

cr6git:
Somebody talked about it on the forum. Not tested.


Johnnynator:
@cr6git you probably have to set `HOSTCC=gcc` for cross compilation (either export it in pre_build or try to set it as `make_build_args=`)

Duncaen:
I don't see the point of packaging this as it is, there is no way to possibly install it as it conflicts with `ncurses`.

cr6git:
@Duncaen you're right, I modified the template.
````
provides="ncurses-${version}_${revision}"
replaces="ncurses>=0"
````

Johnnynator:
@cr6git does this actually work as a drop in replacement without the need of recompilation?

Duncaen:
No that doesn't work.

Johnnynator:
Also the `*.so` should not be in the `-devel` subpackage

cr6git:
@Johnnynator I removed `*.so` from `-devel` subpkg.

PR #14297, opened 2018-05-10
New package: i7z-0.27.2.

maxice8:


cr6git:
[#13461](https://github.com/voidlinux/void-packages/pull/13461)

PR #14299, opened 2018-05-10
xorg-server: update to 1.20.0.

maxice8:


cr6git:
For `xf86-video-fbturbo` and `xf86-video-fbturbo-sunxi`
````
license="MIT, GPL-2.0-or-later"

post_install() {
	vlicense COPYING
}
````

Gottox:
@maxice8 is xorg tested?

PR #14302, opened 2018-05-10
[WIP] common/build-style/go.sh: add do_check(), fix identation of comments on do_configure()

maxice8:


common/build-style/go.sh
```diff
@@ -17,6 +17,10 @@ do_build() {
 	go get -x -tags "${go_build_tags}" -ldflags "${go_ldflags}" ${go_package}
 }
 
+do_check() {
+	go test -v
```
ebfe: That probably won't test anything in most go packages. `go test -v ./...` might work better (but might test more than needed).

maxice8: yeah i will work on this at a future time (anyone feel free to take over), i also need to take care when the go test requires another package.

PR #14309, opened 2018-05-11
New package: xcb-util-errors-1.0

inefarius:
This is an optional dependency for sway, so I thought I'd add it.


maxice8:
duplicate of https://github.com/voidlinux/void-packages/pull/13259

PR #14316, opened 2018-05-11
mate-settings-daemon: add pulseaudio build option, fix license.

cr6git:
PulseAudio not enabled by default. Forum request.


Duncaen:
I don't think this is a good default pulseaudio is what most users want and what works best for most users.
There is exactly one person that asked to remove the dependency and everyone else who doesn't want to use pulse just ignores the dependency, having the library installed does not require the use of pulse.

PR #14317, opened 2018-05-11
 okular: update to 18.04.1.

Chocimier:


srcpkgs/okular/template
```diff
@@ -11,10 +11,10 @@ makedepends="discount-devel djvulibre-devel ebook-tools-devel khtml-devel
 depends="hicolor-icon-theme"
 short_desc="KDE Document Viewer"
 maintainer="Juan RP <xtraeme@voidlinux.eu>"
-license="GPL-2"
+license="GPL-2.0-only"
 homepage="http://kde.org/applications/graphics/okular/"
```
maxice8: homepage is now https://

Chocimier: Is it forbidden to have redirections in templates? If it is, why don't xlint check it yet?

maxice8: no, but it is better to use the actual URL used. xlint can't check for redirections because it would make it incredibly slow and require another dependency (problably curl)

PR #14324, opened 2018-05-11
xbps-src: update-check: update gitlab

Chocimier:


maxice8:
should problably check for packages that have `update` file if it can be removed with these changes

Chocimier:
Others are necessary, mainly for ignoring

PR #14331, opened 2018-05-12
sshguard: use sv start instead of sv check

Nicop06:
`sv start` will actually wait for iptables to start, while `sv check` only that the service is registered (i.e. it is present in `/var/service/`) but not that it is fully started.


Duncaen:
That does not really fix the issue in case of the `iptables` service.
`iptables` should have never been a service as it is not a service.
Without a `check` script `sv check` or `sv start` don't help at all, the service is marked as `up` as soon as `runsv` forks and runs the run script.
You will have the same race condition with `start` and `check`, except that `start` also tries to `up` the service.

Nicop06:
This is true, `iptables` is not a service. Same goes for `alsa` (and certainly many other packages using this trick). But at the same time this is very handy to be able to temporary stop `iptables` with `sv stop iptables` and restart it with `sv start iptables`. Of course, in that case you will lose all the rules created by other services. Most distros I've used solve this new problem by automatically saving the iptables rules on exit, which I usually disable as it gets in the way of sshguard whose purpose it to temporary block some hosts. But it could make sense for some people.

Gottox:
`sv check` does more than just looking for the presence of a service in `/var/service`. See `man sv`:
```
...
       check  Check for the service to be in the state that's been requested.
              Wait up to 7 seconds for the service to reach the requested
              state, then report the status or timeout.  If the requested
              state of the service is up, and the script ./check exists in the
              service directory, sv runs this script to check whether the
              service is up and running; it's considered to be up if ./check
              exits with 0.
...
```

Gottox:
Real fix: add a `check` script to iptables that checks if rules are applied and exit 0 if they are.

cr6git:
@Duncaen another example: `ufw` makes use of `iptables` without having to start the `iptables` service.

Nicop06:
@Gottox 

You are right, using `sv start` does nothing more. I thought it would wait for the program to start but without the `./check` script it does nothing more. Now, the difficulty is to add the `./check` script as this is not trivial.

PR #14332, opened 2018-05-12
Replace sv check by sv start to all runit files

Nicop06:
Here is a rather big pull-request to replace all use of `sv check` by `sv start` to ensure that the depency are started before starting the service [as suggested by aleksey in the void forum](https://forum.voidlinux.eu/t/handling-dependencies-with-runit/5282/8).

I did it for all packages except sshguard as it is already done in #14331. I understand if you prefer a per package aproach. Here is the script I used to generate this pull request, it can totally be adapted to created a new pull-request per package.

```
grep -r -l 'sv check' srcpkgs/ |\
while read file
do
  sed -i 's/^sv check/sv start/' $file
  pkg=$(echo $file|sed 's#srcpkgs/\([^/]*\)/.*#\1#')
  echo $pkg
done |\
sort -u |\
while read pkg
do
  rev=$(sed -n 's/revision=\([0-9]*\)$/\1/p' srcpkgs/$pkg/template)
  let "newrev=$rev+1"
  sed -i "s/revision=$rev/revision=$newrev/" srcpkgs/$pkg/template
done
```


Gottox:
My thoughts:

This won't work as expected as void handles service enabling/disabling a little different: Disabled services aren't known to runitsvdir/runsv at all and therefore `sv start` will fail unless the service is already enabled by the user, in which case it either already runs ( == `sv check` already does the job) or was explicitly stopped ( == possible respawning service depending even if the user explicitly stops it).

chneukirchen:
I don't think we want `sv start` in the scripts, as this will `up` a state with a `./down` file, which are not to be started by the system automatically.

Reading the source, this seems to be the only difference between `start` and `check`, can you elaborate on the issue? Or provide a test case even?

Nicop06:
This is actually useless. The real solution is to write proper `./check` scripts to detect applications state. Because this needs to be done on a per application basis I do not think I will be able to do it for all applications.

PR #14339, opened 2018-05-12
mednafen: update to 1.21.3.

Johnnynator:


srcpkgs/mednafen/template
```diff
@@ -12,7 +12,7 @@ maintainer="Juan RP <xtraeme@voidlinux.eu>"
 license="GPL-2.0-or-later"
 homepage="http://mednafen.sourceforge.net/"
```
cr6git: `homepage="https://mednafen.github.io"`

cr6git:
@Johnnynator there's something I don't understand in the template:
````
configure_args="--with-libcdio --with-libsndfile --enable-alsa"
configure_args+=" ac_cv_sizeof_off_t=8"
````
weird.
I mean, why not this:
````
configure_args="--with-libcdio --with-libsndfile --enable-alsa ac_cv_sizeof_off_t=8"
````
?

Johnnynator:
@cr6git ask whoever did it. I would expect that it was done because `av_size of_off_t` should normally be set correctly by the gnu-configure script but isn't (or smth else), so it looks to me that it was set separate to highlight that it is an attempt to fix a bug. (and just some general configuring)

cr6git:
added by @pullmoll in [bcfc78c](https://github.com/voidlinux/void-packages/commit/bcfc78c804b8effc30874833b8e088b68fcfd508)

Gottox:
style nitpicking.

PR #14340, opened 2018-05-12
lutris: update to 0.4.16

SolitudeSF:
switched `distfiles` to github, since its still not updated on the official site.

also, im not sure, but isn't this thing supposed to be `noarch`?


bra1nwave:
Yeah, it should be `noarch`.

srcpkgs/lutris/template
```diff
@@ -1,15 +1,14 @@
...
+version=0.4.16
 revision=1
 build_style=python3-module
-wrksrc="${pkgname}"
 pycompile_module="${pkgname}"
 hostmakedepends="python3-setuptools python3-gobject"
 depends="python3-dbus python3-gobject python3-yaml xrandr"
 short_desc="Open gaming platform for managing games in a unified way"
 maintainer="Jan Wey. <janwey.git@gmail.com>"
 license="GPL-3"
```
bra1nwave: Please use the spdx identifier `GPL-3.0-or-later`

PR #14342, opened 2018-05-12
[ci skip] krita: update to 4.0.3, fix license.

cr6git:


maxice8:
there is no way this builds in under 50 minutes, add [ci skip] to commit so travis doesn't waste 1 hour hogging.

cr6git:
@maxice8 ok! thank you.

lemmi:
I'll let it run locally on a couple of archs.

PR #14350, opened 2018-05-12
New package: kakasi-2.3.6

asavonic:


srcpkgs/kakasi/template
```diff
@@ -0,0 +1,11 @@
+# Template file for 'kakasi'
+pkgname=kakasi
+version=2.3.6
+revision=1
+build_style=gnu-configure
+short_desc="Tool to convert Kanji characters to Hiragana, Katakana or Romaji"
+maintainer="Andrew Savonichev <andrew.savonichev@gmail.com>"
+homepage="http://kakasi.namazu.org/"
+license="GPL-2"
```
maxice8: use spdx indentifier
license= field directly above homepage= field

asavonic: Thanks! Fixed.

srcpkgs/kakasi/template
```diff
@@ -0,0 +1,11 @@
+# Template file for 'kakasi'
+pkgname=kakasi
+version=2.3.6
+revision=1
+build_style=gnu-configure
+short_desc="Tool to convert Kanji characters to Hiragana, Katakana or Romaji"
+maintainer="Andrew Savonichev <andrew.savonichev@gmail.com>"
+license="GPL-2.0-only"
```
cr6git: `GPL-2.0-or-later`

PR #14351, opened 2018-05-12
hplip: fix license

cr6git:


srcpkgs/hplip/template
```diff
@@ -28,7 +28,7 @@ pycompile_dirs="usr/share/hplip"
 short_desc="HP Linux Imaging and Printing"
 maintainer="Juan RP <xtraeme@voidlinux.eu>"
 homepage="https://developers.hp.com/hp-linux-imaging-and-printing"
-license="GPL-2"
+license="GPL-2.0-only, BSD-3-Clause, MIT"
```
maxice8: license= field right above homepage=

cr6git: oh my god, I totally forgot about...
````
srcpkgs/hplip/template:31: license 'BSD', but no use of vlicense
srcpkgs/hplip/template:31: license 'MIT', but no use of vlicense
````

Comment (maxice8) [reply srcpkgs/hplip/template]: put this on .git/hooks/pre-commit and make it executable

```
#!/bin/sh
#
# this gets run when git commit is called and if it returns anything
# other than 0 it will prevent the commit from being made.
# Send output to stderr
exec 1>&2

git diff --cached --exit-code --quiet && exit 0

porcelain="$(git status --porcelain)"

get() { 
	path="$(git diff --staged --name-only | grep 'template$')"
	sed -n "s/^$1=//p" "$path" 2>/dev/null ||
	# if the sed fails then we get the basename of the dirname
	# we give priority to the upmost result.
	if [ "$1" = "pkgname" ]; then
		basename "$(dirname "$(git diff --staged --name-only | grep srcpkgs/ | head -n 1)")"
	fi
}

out() {
	printf "%s\\n" "$1"
}

if git status --porcelain | grep -q 'template$'; then
	[ "$IGNORE_XLINT" ] || if ! xlint "$(git diff --staged --name-only | grep 'template$')"; then
		exit 1
	fi
fi

# Check if update is done but checksum isn't changed
if git status --porcelain | grep -q 'M.*template$'; then
	git diff --staged | grep -Eq '^+version=|^-version=' &&
	if ! git diff --staged | grep -Eq '^+checksum=|^-checksum='; then
		out "$(get pkgname) being updated but checksum unchanged"
		exit 1
	fi
fi

exit 0
```

it will make you unable to commit unless xlint is satisfied or `XLINT_IGNORE` is set

PR #14363, opened 2018-05-13
rng-tools: update to 6.2

lonwillett:


srcpkgs/rng-tools/template
```diff
@@ -11,7 +11,7 @@ maintainer="Andrea Brancaleoni <abc@pompel.me>"
 license="GPL-3"
```
maxice8: please use spdx identifiers

lonwillett: Fixed. The xbps-src manual should be updated to reflect this policy.

PR #14371, opened 2018-05-14
xurls: add do_check(), fix license and homepage.

maxice8:


srcpkgs/xurls/template
```diff
@@ -7,11 +7,15 @@ go_import_path="github.com/mvdan/${pkgname}"
 go_package="${go_import_path}/cmd/${pkgname}"
 short_desc="Extract urls from text"
 maintainer="dleeram <dleeram@protonmail.com>"
-license="BSD-3"
-homepage="https://${go_import_path}"
+license="BSD-3-Clause"
+homepage="https://www.github.com/vdan/xurls"
```
dleeram: s/vdan/mvdan/, but I don't understand the change

PR #14377, opened 2018-05-14
chromium: update to 66.0.3359.170

freshprince:
see https://chromereleases.googleblog.com/2018/05/stable-channel-update-for-desktop.html
this update fixes a critical sandbox escape


freshprince:
fixed in commit https://github.com/voidlinux/void-packages/commit/3f52c312b2940f4f88f54684a309d0797c198d61

PR #14379, opened 2018-05-14
New package: libinput-gestures-2.36

cr6git:


cr6git:
Replaced by https://github.com/void-linux/void-packages/pull/2300

PR #14384, opened 2018-05-14
postgresql: update to 9.6.9.

arete:


srcpkgs/postgresql/template
```diff
@@ -14,7 +14,7 @@ maintainer="Juan RP <xtraeme@voidlinux.eu>"
 homepage="http://www.postgresql.org"
```
maxice8: postgresql.org is now https

PR #14396, opened 2018-05-14
neomutt: update to 20180512.

freshprince:


maxice8:
cc @Vaelatern 


i'll close my PR since you incorporated the changes

PR #14402, opened 2018-05-14
New package: iwd-0.2.

maxice8:
as requested on the forums, it is quite nifty.

PR #14405, opened 2018-05-14
kernel-libc-headers: fix musl net headers conflicts

Johnnynator:
Needed for #14402


Gottox:
This might be useful for `NetworkManager` too.

PR #14420, opened 2018-05-15
spl: update to 0.7.9.

zdykstra:


Vaelatern:
Have you tested these already?

maxice8:
if you're gonna update spl might as well take zfs with it

srcpkgs/spl/template
```diff
@@ -11,7 +11,7 @@ maintainer="Toyam Cox <Vaelatern@voidlinux.eu>"
 license="GPL-2"
```
maxice8: please use spdx identifier

zdykstra:
[  168.527783] spl: loading out-of-tree module taints kernel.
[  168.537398] SPL: Loaded module v0.7.9-1
[  168.542049] znvpair: module license 'CDDL' taints kernel.
[  168.543575] Disabling lock debugging due to kernel taint
[  170.295554] ZFS: Loaded module v0.7.9-1, ZFS pool version 5000, ZFS filesystem version 5

Seems to work with out issue. Haven't tested booting off of a pool. I can if that's needed.

maxice8:
> No rush, just trying to pick up some of the work on a few packages I use. I opened up a PR for each package, as that's what https://github.com/voidlinux/void-packages/blob/master/CONTRIBUTING.md#starting-a-pull-request seemed to suggest.
--

yeah but spl is a special case of `this package was made only for this other specific package and they have releases that walk in lockstep`, spl was made for zfs and is the only reason it is in the repo.

PR #14249, opened 2018-05-08
linux4.16: Re-enable kernel modules on Linux 4.16

faulesocke:
This re-enables the rtsx kernel modules that were enabled on previous versions. This is required to use the Realtek SD card readers.

This is a squashed and corrected version of #14217. I've correct commit message wording and revbumped the package.

PR #14426, opened 2018-05-15
linux4.16: update to 4.16.9.

freshprince:


cr6git:
You didn't take into account [#14249](https://github.com/voidlinux/void-packages/pull/14249) ?

freshprince:
I can add that patch to this PR if desired.

cr6git:
And we are at 4.16.9 now...

PR #14429, opened 2018-05-15
ca-certificates: update to 20180409.

pbui:


Gottox:
This update broke certain https connections and was reverted.
See https://build.voidlinux.eu/builders/x86_64_builder/builds/9173/steps/shell_3/logs/stdio for details.

srcpkgs/ca-certificates/template
```diff
@@ -13,7 +13,7 @@ maintainer="Juan RP <xtraeme@voidlinux.eu>"
 homepage="https://tracker.debian.org/pkg/ca-certificates"
 license="GPL-2, MPL-2.0"
```
cr6git: `GPL-2.0-or-later`

PR #14435, opened 2018-05-15
dash: update to 0.5.10.1.

maxice8:


Duncaen:
Still broken, this release only contains the patch that was removed from here.

chneukirchen:
We'd need to add 469c5fd4f576 as a patch.

maxice8:
On Wed, May 16, 2018 at 02:17:08AM -0700, Leah Neukirchen wrote:
> We'd need to add 469c5fd4f576 as a patch.
> 
> —
> 

added

PR #14447, opened 2018-05-15
libkcddb: update to 18.04.2.

Chocimier:


Chocimier:
`srcpkgs/libkcddb/template:22: use vlicense`
Vlicense can't be used here, as that is not licence for main package.

PR #14455, opened 2018-05-16
ddrescue: fix license.

cr6git:


cr6git:
````
ERROR: [trans] failed to download `ca-certificates-20170717_2' package from `http://198.204.250.219/current': Not Found
````

Gottox:
It's a fast build. Increase the revision so the package is updated too :)

PR #14462, opened 2018-05-16
New package: you-get-0.4.1077

cr6git:


Piraty:
It can be installed via pip, also upstream recommends ffmpeg as dependency

PR #14464, opened 2018-05-16
telegram-desktop: only show the latet release

Johnnynator:
That fixes the listening of Alpha Versions and also that version of sub packages do also get listed.


maxice8:
don't think the quotes around pkgname= are necessary, but just a nitpick

Johnnynator:
yep. You are right they aren't necessary but also they don't hurt so I will not change them.

PR #14472, opened 2018-05-16
xfce4-terminal: update to 0.8.7.4.

maxice8:


cr6git:
you didn't see [#14453](https://github.com/voidlinux/void-packages/pull/14453) ? it's okay, I let you take care of it.

PR #14473, opened 2018-05-16
python-Flask-OAuthlib: update to 0.9.5.

maxice8:


srcpkgs/python-Flask-OAuthlib/template
```diff
@@ -1,19 +1,20 @@
...
-version=0.9.4
-revision=2
+version=0.9.5
+revision=1
 noarch=yes
 wrksrc="${pkgname#*-}-${version}"
 build_style=python-module
 pycompile_module="flask_oauthlib"
 hostmakedepends="python-setuptools python3-setuptools"
+makedepends="python-devel python3-devel"
```
asergi: Unnecessary.

PR #14497, opened 2018-05-17
New package: font-libertine-5.3.0

pvanier:
Adding the libertine fonts both in OTF and TTF formats.

### DIFF
--- a/srcpkgs/font-libertine-otf/template
+++ b/srcpkgs/font-libertine-otf/template
@@ -1,0 +1,21 @@
+# Template file for 'font-libertine'
+pkgname=font-libertine-otf
+version="5.3.0"
+revision="2012_07_02"
+
+noarch="yes"
+depends="font-util xbps-triggers"
+font_dirs="/usr/share/fonts/OTF"
+create_wrksrc=yes
+
+short_desc="Libertine Fonts - Libre multilingual font family"
+maintainer="Pascal Vanier <pascal.vanier@gmail.com>"
+license="OFL-1.1"
+homepage="http://libertine-fonts.org/"
+distfiles="${SOURCEFORGE_SITE}/project/linuxlibertine/linuxlibertine/${version}/LinLibertineOTF_${version}_${revision}.tgz"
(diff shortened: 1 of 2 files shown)


srcpkgs/font-libertine-otf/template
```diff
@@ -0,0 +1,21 @@
+# Template file for 'font-libertine'
+pkgname=font-libertine-otf
+version="5.3.0"
```
maxice8: should not be quoted

srcpkgs/font-libertine-otf/template
```diff
@@ -0,0 +1,21 @@
+# Template file for 'font-libertine'
```
maxice8: needs to match pkgname=

srcpkgs/font-libertine-otf/template
```diff
@@ -0,0 +1,21 @@
...
+pkgname=font-libertine-otf
+version="5.3.0"
+revision="2012_07_02"
```
maxice8: revision= is used internally by us, it should be 1 by default, and increased when changes are made that need recompilation.

srcpkgs/font-libertine-otf/template
```diff
@@ -0,0 +1,21 @@
...
+
+noarch="yes"
```
maxice8: unquote=, no need for blank space between it and revision

srcpkgs/font-libertine-otf/template
```diff
@@ -0,0 +1,21 @@
...
+depends="font-util xbps-triggers"
+font_dirs="/usr/share/fonts/OTF"
+create_wrksrc=yes
+
```
maxice8: unnecessary blank line

pvanier:
I implemented the requested changes, but I don't understand why the checks have failed.

Johnnynator:
Looks to me that Travis is linking to the wrong build. Also you should have one commit per package. One for `fobt-libertine-otf` and one for `font-libertine-ttf`

pvanier:
I'll make sure to separate packages  for my next package.

PR #14498, opened 2018-05-17
upower: rev bump to fix upower0 replacement

abenson:
Also updated license to SPDX compliance.


Gottox:
A little context before merging:

We removed upower0 and made it pull upower. Unfortunately the update didn't replace all files correctly and left some files from upower0. This bump should force overwriting those files.

PR #14503, opened 2018-05-17
New package: miniserve-0.1.5

cr6git:


cr6git:
Replaced by https://github.com/void-linux/void-packages/pull/949

PR #14513, opened 2018-05-17
New package: lnav-0.8.3

renatoaguiar:


srcpkgs/lnav/template
```diff
@@ -0,0 +1,18 @@
+# Template file for 'lnav'
+pkgname=lnav
+version=0.8.3
+revision=1
+build_style=gnu-configure
+configure_args="--disable-static"
+hostmakedepends="re2c"
+makedepends="pcre-devel sqlite-devel ncurses-devel readline-devel zlib-devel bzip2-devel libcurl-devel"
```
cr6git: `makedepends="libcurl-devel ncurses-devel pcre-devel readline-devel sqlite-devel"`

renatoaguiar: I added dependencies in the order they appear in the README file, but forgot to sort them.

Thanks for noticing that. I'll update the patch soon :)

cr6git: However I have no idea why the Travis build failed?

renatoaguiar: Why did you remove bzip2-devel and zlib-devel?

maxice8: they are implicitly brought by other dependencies, see `xmindep` from xtools.

renatoaguiar: I didn't know xmindep. Thanks :)

Johnnynator:
you should add broken statement for musl since it doesn't compile there. Smth like:
```
 *-musl) broken="https://travis-ci.org/voidlinux/void-packages/jobs/380512280";;
esac
```

PR #14519, opened 2018-05-18
cmake: update to 3.11.2.

maxice8:


srcpkgs/cmake/template
```diff
@@ -1,6 +1,6 @@
 # Template file for 'cmake'
 pkgname=cmake
-version=3.11.1
+version=3.11.2
 revision=1
 build_style=configure
```
cr6git: As `build_style=configure` is defined globally, what's the point of redefining it in:
````
if [ "$CROSS_BUILD" ]; then
	# XXX ugly :-)
	build_style=cmake
	[...]
else
	build_style=configure
	[...]
````
I don't understand.

newbluemoon: @cr6git This is some sort of bootstrap. Cmake builds itself with itself. In a cross build it can use the same cmake version already built for the host.

PR #14528, opened 2018-05-18
mono: update to 5.12.0.226. [NOMERGE]

Hoshpak:


Hoshpak:
The build on x86-64 musl is much slower than on other systems, CSC takes ages to compile the files shipped with mono. I think this requires further investigation and this release shouldn't be merged just yet.

Hoshpak:
The issue hit alpine first and is documented in upstream bug report https://github.com/mono/mono/issues/7167 . I will close this PR for now, subscribed to the issue and will open a new PR once a mono release is that does work again with musl.

PR #9144, opened 2017-11-13
[WIP] android-studio: Update to 3.0.0-8 171.4408382, and fix maintainer

jcjordyn130:

PR #14532, opened 2018-05-18
android-studio: update to 3.1.2.

Chocimier:


Chocimier:
It looks like xlint should accept skiprdeps.

#9144 is obsolete.

PR #14504, opened 2018-05-17
gimp: update to 2.10.0

kartik-ynwa:

PR #14533, opened 2018-05-18
Gimp 2.10

Gottox:
Successor of #14504. cc @Duncaen @kartik-ynwa


DirectorX:
#14504

Piraty:
gimp 2.10.2 relased ([here](https://www.gimp.org/news/2018/05/20/gimp-2-10-2-released/)), at the notes state sth about a bug-tracker-url (first time for `github.com/void-linux/void-packages/issues`?):

> Packagers: set your bug tracker address
>
> As you know, we now have a debug dialog which may pop-up when crashes occur with debug information. This dialog opens our bug tracker in a browser.
> <...>
> This is why our configure script now has the `--with-bug-report-url` option, allowing you to set your own bug tracker web URL. This way, when people click the “Open Bug Tracker” button it will open the package bug tracker instead.

Piraty:
I just built your branch @Gottox, but installation fails.
```
xi gimp
[*] Updating `https://repo.voidlinux.eu/current/x86_64-repodata' ...
gimp-2.10.0_1: broken, unresolvable shlib `libgegl-npd-0.4.so'
libgimp-2.10.0_1: broken, unresolvable shlib `libgegl-npd-0.4.so'
Transaction aborted due to unresolved shlibs.
```
just moving `"usr/lib/libgegl-${version%.*}.so"` might be enough

Gottox:
was gegl installed and updated too?

Gottox:
@Piraty was gegl installed and updated too?

Gottox:
@biopsin it shouldn't

Piraty:
yes, gegl was built before gimp.
As stated, the problem is that the mentioned `libgegl-npd-0.4.so` is moved to `gegl-devel`, so the entry in `common/shlibs is wrong`. that's why it canot resolve.

Here's my simple fix. `libgegl-0.4.so` is the only symlink to a numbered .so, all others don't have versions.
Dunno what's preferred in such case, but all other .so files are kept in the `gegl`package then.
```
diff --git a/srcpkgs/gegl/template b/srcpkgs/gegl/template
index ac9001d290..7b6d43965f 100644
--- a/srcpkgs/gegl/template
+++ b/srcpkgs/gegl/template
@@ -22,6 +22,6 @@ gegl-devel_package() {
        pkg_install() {
                vmove usr/include
                vmove usr/lib/pkgconfig
-           vmove "usr/lib/*.so"
+         vmove "usr/lib/libgegl-${version%.*}.so"
        }
 }
```

Gottox:
Already fixed. :)

kartik-ynwa:
For some reason I get a warning message saying "some fonts failed to load" on startup. Doing `mkdir /usr/share/gimp/2.0/fonts` fixes it.

Piraty:
same here

PR #14552, opened 2018-05-19
i2pd: muted stdout in service file.

5nefarious:
I'm not running a login manager and my standard output gets absolutely spammed without this line. If there is a better way to fix this, please let me know.


cr6git:
I never used it but, did you try to adust these options:
````
--log= Logs destination: stdout, file, syslog (stdout if not set, file
       - otherwise, for compatibility)

--logfile
       Path to logfile (default - autodetect)

--loglevel=
       Log messages above this level (debug, info, warn, error)
````

cr6git:
@5nefarious check [releases](https://github.com/PurpleI2P/i2pd/releases), it's time to update to 2.18.0 anyway.

PR #14553, opened 2018-05-19
caffeine-ng: update to 3.4.2.

cr6git:


srcpkgs/caffeine-ng/template
```diff
@@ -14,7 +14,7 @@ maintainer="cr6git <quark6@protonmail.com>"
 license="GPL-3.0-or-later"
 homepage="https://gitlab.com/DJCrashdummy/caffeine-ng"
```
asergi: https://github.com/caffeine-ng/caffeine-ng/commit/3a86cfb6256a3efedede5c21aaeb824272522924

PR #14554, opened 2018-05-19
lutris: update to 0.4.16_2.

nonchip:
this patch fixes an issue with people using `LD_LIBRARY_PATH` in their lutris system config.

also, after talking to @chneukirchen we realized the original maintainer seems very absent and decided for me to take care of this package since I use it on an almost daily basis and am at least a bit able to figure out my way in the codebase and should be able to test and/or debug PRs.


srcpkgs/lutris/template
```diff
@@ -1,14 +1,14 @@
 # Template file for 'lutris'
 pkgname=lutris
 version=0.4.16
-revision=1
+revision=2
 noarch=yes
 build_style=python3-module
 pycompile_module="${pkgname}"
```
cr6git: Use `lutris` instead of `${pkgname}` you will save four characters ;)

nonchip: i didn't touch that :D
and actually it's nicely modular

asergi:
`0.4.17` is out.

asergi:
Superseded by #14608.

PR #14556, opened 2018-05-19
poco: update to 1.9.0.

Chocimier:


srcpkgs/poco/template
```diff
@@ -1,14 +1,14 @@
...
-version=1.8.0.1
+version=1.9.0
 revision=1
 build_style=cmake
 short_desc="C++ class libraries for building network-based applications"
 maintainer="Julien Dehos <dehos@univ-littoral.fr>"
-license="Boost Software License 1.0"
+license="BSL-1.0"
 homepage="https://pocoproject.org"
 distfiles="https://pocoproject.org/releases/${pkgname}-${version}/${pkgname}-${version}-all.tar.gz"
```
cr6git: Use `poco` instead of `${pkgname}` it's shorter.

srcpkgs/poco/template
```diff
@@ -1,14 +1,14 @@
...
-checksum=cb891540b44b74e844fd0787c796f2699a894dba690329660e29b57efb8f80dd
+checksum=0387bf0f9d313e2311742e1ad0b64e07f2f3e76039eed20e3b9aa9951b88e187
 wrksrc="${pkgname}-${version}-all"
```
cr6git: Move `wrksrc=` under `revision=` and replace `${pkgname}` with `poco` it's shorter.

srcpkgs/poco/template
```diff
@@ -1,14 +1,14 @@
 # Template build file for 'poco'.
```
cr6git: `sed -i "s|Template build file for 'poco'.|Template file for 'poco'|" srcpkgs/poco/template`

srcpkgs/poco/template
Chocimier: > Move wrksrc= under revision=

I won't until there is tool to assert correct order.

srcpkgs/poco/template
Chocimier: Same here, tool should warn on that, especially as there are ~700 templates with wrong header.

PR #14557, opened 2018-05-19
inxi: update to 3.0.12r1.

Chocimier:


srcpkgs/inxi/template
```diff
@@ -1,16 +1,16 @@
 # Template file for 'inxi'
 pkgname=inxi
-version=3.0.07.1
+version=3.0.09.1
 revision=1
 wrksrc=inxi-${version/.1/-1}
```
cr6git: I think it's better if you quote `wrksrc="inxi-${version/.1/-1}"`

cr6git:
3.0.10 is out

cr6git:
Replaced by https://github.com/void-linux/void-packages/pull/441

PR #14558, opened 2018-05-19
armadillo: update to 8.500.1.

jnbr:


srcpkgs/armadillo/template
```diff
@@ -10,7 +10,7 @@ maintainer="Julien Dehos <dehos@lisic.univ-littoral.fr>"
 license="BSD"
```
cr6git: Not valid, see [SPDX](https://spdx.org/licenses/).

jnbr: thanks, was wrong anyway

PR #14560, opened 2018-05-19
mlt: update to 6.8.0.

jnbr:
the musl patch is not needed anymore, instead HAVE_STRTOD_L and HAVE_LOCALE_H have to be set to 1


srcpkgs/mlt/template
```diff
@@ -13,11 +13,14 @@ makedepends="alsa-lib-devel ffmpeg-devel gtk+-devel jack-devel ladspa-sdk
  libvidstab-devel"
 depends="mlt-data>=${version}_${revision} ladspa-sdk-example-plugins"
 short_desc="Multimedia framework, designed for television broadcasting"
+license=" LGPL-2.1-only"
```
cr6git: trailing space.
You removed `GPL-3` ? So what's the point of `configure_args="--enable-gpl3"` ?

srcpkgs/mlt/template
```diff
@@ -13,11 +13,14 @@ makedepends="alsa-lib-devel ffmpeg-devel gtk+-devel jack-devel ladspa-sdk
...
 maintainer="Diogo Leal <diogo@diogoleal.com>"
-license="GPL-3"
 homepage="http://mltframework.org/"
```
cr6git: Use https

srcpkgs/mlt/template
jnbr: I guess it is GPL-3.0 than? Not sure if -only or -or-later

PR #14564, opened 2018-05-20
procps-ng: update to 3.3.15.

maxice8:
Fixes security issues found in http://www.openwall.com/lists/oss-security/2018/05/17/1

[changelog](https://gitlab.com/procps-ng/procps/blob/master/NEWS)


maxice8:
added missing checkdepends= on dejagnu

PR #14570, opened 2018-05-20
New package: custard-1.2

uMeYou:


newbluemoon:
Well, first, when you make changes you should always amend your commit, not add new ones. So that there’s one commit per affected package. If there are multiple commits already you can squash them into one.

Next thing you should use `build_style="gnu-makefile"` then you won’t need a custom `do_install()`. Also, the `make` command to build the package isn’t supposed to be run in the `do_install()` stage.
However, the `Makefile` needs to be adjusted because it doesn’t respect Void’s (cross-)compiler and CFLAGS:

```
pre_build() {
        sed -i -e "s/CC=g++/CC=$CXX/" \
         -e 's/CFLAGS =/CFLAGS +=/' \
         Makefile
}
```

With these changes we also have to set `make_use_env=yes` (see `common/build-style/gnu-makefile.sh`) to make sure the CFLAGS from the `Makefile` are used (else linking fails).

For `short_desc` "Stacking/floating window manager for X.Org" should be sufficient.

And finally add the license in `post_install()`.

To make it clear here’s the template with the changes incorporated:

```
# Template file for 'custard'
pkgname=custard
version=1.2
revision=1
build_style="gnu-makefile"
make_use_env=yes
makedepends="libxcb-devel xcb-util-devel xcb-util-wm-devel xcb-util-xrm-devel"
short_desc="Stacking/floating window manager for X.Org"
maintainer="MeYou <meeyou@tuta.io>"
license="MIT"
homepage="https://github.com/Sweets/custard"
distfiles="https://github.com/Sweets/${pkgname}/archive/v${version}.tar.gz"
checksum=90579bc10eac9fd0330b76a6cce29ddf5ce3075ee7156d56e71f14d5ad538d57

pre_build() {
	sed -i -e "s/CC=g++/CC=$CXX/" \
	 -e 's/CFLAGS =/CFLAGS +=/' \
	 Makefile
}

post_install() {
	vlicense LICENSE
}
```

srcpkgs/custard/template
```diff
@@ -0,0 +1,19 @@
+# Template file for 'custard'
+pkgname=custard
+version=1.2
+revision=1
+makedepends="libxcb-devel xcb-util-devel xcb-util-wm-devel xcb-util-xrm-devel"
```
cr6git: You can reduce to: `makedepends="xcb-util-wm-devel xcb-util-xrm-devel"`

uMeYou:
Thank you for the pointers

PR #14577, opened 2018-05-20
xorg-server: enable glx on aarch64

q66:
This is necessary so that Tegra-based systems can make use of the
accelerated open source graphics stack in Mesa 18.1.

I'm currently in process of porting Void to the Google Pixel C tablet and without this it is not possible to use X11 with OpenGL on the Tegra X1 SoC; with it (and with Mesa updated to 18.1 with the appropriate features enabled) we can use desktop OpenGL 4.3 as well as GLES 3.1.

### DIFF
--- a/srcpkgs/xorg-server/template
+++ b/srcpkgs/xorg-server/template
@@ -1,7 +1,7 @@
 # Template file for 'xorg-server'
 pkgname=xorg-server
 version=1.19.6
-revision=4
+revision=5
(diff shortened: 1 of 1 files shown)


maxice8:
added to https://github.com/voidlinux/void-packages/pull/14299

q66:
Alright, thanks @maxice8. I will abandon this pull request then.

PR #14581, opened 2018-05-20
New package: doomretro-2.6.9

cr6git:
:hurtrealbad:


cr6git:
Replaced by https://github.com/void-linux/void-packages/pull/898

Issue #14592, opened 2018-05-20
zathura-ps has a runtime dependency on ghostscript


copyleftie:
### System

* xuname:  
  Void 4.16.9_blackbox x86_64 GenuineIntel uptodate rFFFFFF
* package:  
 `zathura-ps-0.2.6_1`

### Expected behavior

Installing `zathura-ps` enables PostScript support in `zathura`.

### Actual behavior

`zathura` fails at runtime unless `ghostscript` is manually installed.

### Steps to reproduce the behavior

0. If necessary, uninstall `ghostscript` and install `zathura`.
1. Install `zathura-ps`.
2. Try to open a PostScript file.

```
% zathura testfile.ps
GPL Ghostscript 9.23: Can't find initialization file gs_init.ps.
fatal internal error -100error: Rendering failed (page 1)
# the file will fail to display
```

PR #14594, opened 2018-05-20
zathura-ps: add ghostscript runtime dependency

copyleftie:
#14592


maxice8:
add closes #14592 to commit msg.

copyleftie:
added

PR #14602, opened 2018-05-21
firefox-esr: update to 60.0.1

kimat:


cr6git:
````
=> firefox-esr-60.0.1_1: running post-extract hook: 00-patches ...
/void-packages/common/xbps-src/shutils/common.sh: line 31: cd: /builddir/firefox-60.0.1esr: No such file or directory
=> ERROR: firefox-esr-60.0.1_1: cannot access wrksrc directory [/builddir/firefox-60.0.1esr]
````
Of course you didn't run any tests before opening your PR... :unamused:

Gottox:
I'd like to wait with merging this for two reasons:

* firefox is at 60 as well, so updating esr now will result in a duplicate
* firefox-esr currently the only browser that works half way reliably across different architectures. We need to fix rust cross build first.

chneukirchen:
Please keep ESR at 52 as long as supported, the newer versions break many things.

Gottox:
Such as alsa support?

PR #14604, opened 2018-05-21
runc: update to 1.0.0 rc5

iaroki:
It is 1.0.0 rc5 upstream. Should I just revbump revision or change the version string?


Johnnynator:
You have to bump revidion.

srcpkgs/runc/template
```diff
@@ -1,8 +1,8 @@
 # Template file for 'runc'
 pkgname=runc
 version=1.0.0
-revision=4
-_subver="rc4"
+revision=5
+_subver="rc5"
 _ver="$version-$_subver"
 build_style=go
```
cr6git: with `build_style=go` you don't have to put `go` in the `hostmakedepends`

PR #14607, opened 2018-05-21
common/travis: remove need for -C flag for xbps-src

Johnnynator:
should fix travis linting when a package does either depends on itself in `hostmakedepnds` when cross building or when it pulls in a package as `hostmakedepends` and `makedepends` when cross compiling.

e.g. #13803 or #14590 

Test build log: https://travis-ci.org/Johnnynator/void-packages/builds/381697961


.travis.yml
```diff
@@ -15,7 +15,7 @@ env:
         - ARCH=x86_64 BOOTSTRAP=x86_64
         - ARCH=i686 BOOTSTRAP=i686
         - ARCH=aarch64 BOOTSTRAP=x86_64
-        - ARCH=armv7hf BOOTSTRAP=x86_64
+        - ARCH=armv7l BOOTSTRAP=x86_64
         - ARCH=x86_64-musl BOOTSTRAP=x86_64-musl
         - ARCH=armv6hf-musl BOOTSTRAP=x86_64-musl
```
Gottox: this line should change too, shouldn't it?

Johnnynator: Yep. I updated it

Vaelatern:
:shipit:

PR #14644, opened 2018-05-22
acr: update to 1.6.2

nocriw:


srcpkgs/acr/template
```diff
@@ -1,12 +1,12 @@
...
 noarch=yes
 build_style=gnu-configure
 short_desc="AutoConf Replacement"
-maintainer="Juan RP <xtraeme@voidlinux.eu>"
-license="GPL-3"
-homepage="http://radare.org"
-distfiles="${homepage}/get/acr-${version}.tar.gz"
-checksum=b44c083ce988c62f9ac929784e1de69db01ec56b9e71ea9fcc5bed93f88052c6
+maintainer="criw <contact@thecriw.com>"
+license="GPL-2"
```
Johnnynator: pls use the [SDPX identifier](https://spdx.org/licenses/). In this case it would be`GPL-2.0-or-later`

nocriw: Sorry about that, done :)

Johnnynator: No problem. AFAIK the manual isn't yet updated  and therefore it isn't mentioned anywhere (except in some PR's and somewhere is a issue about it. Nothing obvious)

PR #14647, opened 2018-05-23
gcc: update to 8.1.0

pullmoll:
I think this can be merged, no further fallout so far.
[ci skip]


pullmoll:
Oops! `i686{,-musl}` fail to build qt5 with a new error:
```
MOJO_STATIC_ASSERT(MOJO_ALIGNOF(int64_t) == 8, "int64_t has weird alignment");
```
Unfortunately MOJO doesn't tell me what it thinks that *weird* means :P

pullmoll:
This PR is actually not ready for prime time. Fallouts for i686* and cross builds due to the alignment issues. I'd suggest to wait until at least patches are available for our *biggies*.

PR #14671, opened 2018-05-23
common/update_check: add gnome.gitlab.org

Johnnynator:


Vaelatern:
Why is this not folded into *gitlab.com* ?

Johnnynator:
Because the URL (and url var therefore ) is not the same. But I just noticed that the new pattern doesn't include the URL anymore. If #14324 gets merged I can update this.

Johnnynator:
Done. I also included the remaining commits of #14324

PR #14672, opened 2018-05-23
kdeconnect: update to 1.3.

Johnnynator:


maxice8:
Seems like travis crapped out

Johnnynator:
Looks more like it was some problems with the mirror.

maxice8:
On Thu, May 24, 2018 at 03:08:50AM -0700, John Zimmermann wrote:
> Looks more like it was some problems with the mirror.
> 
> -- 
> https://github.com/voidlinux/void-packages/pull/14672#issuecomment-391661427

either way let's see if it works out now.

Johnnynator:
Nope still fails to fetch the deps.

PR #14683, opened 2018-05-24
New package: buildah-1.0

iaroki:


srcpkgs/buildah/template
```diff
@@ -0,0 +1,44 @@
...
+	vman docs/buildah-images.1
+	vman docs/buildah-inspect.1
+	vman docs/buildah-mount.1
+	vman docs/buildah-push.1
+	vman docs/buildah-rm.1
+	vman docs/buildah-rmi.1
+	vman docs/buildah-run.1
+	vman docs/buildah-tag.1
+	vman docs/buildah-umount.1
+	vman docs/buildah-version.1
```
Johnnynator: wouldn't smth like
```
for _i in docs/*.1; do
   vman $_i
done
```
much more cleaner instead of listing every single man page. And it wouldn't require updating if they add a new man page.

iaroki: yeah, just saw similiar list in other templates

iaroki:
aarch64 and armv7l builds fail with
```
/bin/sh: go-md2man: command not found
```
But it installed in the deps
```
=> skopeo-0.1.30_1: installing target dependency 'go-md2man-1.0.8_3' ...
```

Johnnynator:
`go-md2man` needs to be in hostmakedepends.

Vaelatern:
The provided patches are substantial.

Why do you provide them?

iaroki:
@Vaelatern This patches remove applying SELinux context to files.
Without them build will fail with error about missing `libselinux.so`

Vaelatern:
Is there a feature flag we can trip instead?

iaroki:
As far as I know this changes were added upstream without any option to disable.
It is how this package build in Arch: https://git.archlinux.org/svntogit/community.git/tree/trunk?h=packages/buildah

PR #14686, opened 2018-05-24
R and CRAN packages: updates and questions

wagnerflo:
I've just updated R from 3.4.x to 3.5.0. This requires a bunch of CRAN packages to be rebuilt against the new R version. For these I've bumped `revision` or updated to a new version where available.

All commits are in dependency order but when Travis builds my branches commits in bulk it'll ignore the order and that make the build fail.

How do manage to get the packages to build in the required order? Should I add explicit version numbers to `depends`? How do I even make sure that the new R is built before any CRAN packages?


ebfe:
Commit order is neither respected by travis nor by the actual builder. If all the CRAN packages _directly_depend on R at least the builder should dtrt.

wagnerflo:
All of the CRAN packages depend on R of course, but not on any explicit version. Should I add that?

ebfe:
The issue for this failure is the dependency chain: R-cran-scales -> R-cran-munsell -> R-cran-colorspaces.
xbps-src only tracks dependencies only one level deep. So R-cran-colorspaces isn't necessarily built before R-cran-scales. The usual "fix" is to push the updates in the correct order, or as a workaround make R-cran-scales directly depend on R-cran-colorspaces.

ebfe:
```
https://cran.r-project.org/src/contrib/utf8_1.1.3.tar.gz: Not Found
=> ERROR: R-cran-utf8-1.1.3_2: failed to fetch utf8_1.1.3.tar.gz.
```

wagnerflo:
They were updating packages faster than I could follow. Finally caught up!

PR #14725, opened 2018-05-26
libfcitx-qt5: update to 1.2.2.

maxice8:


Johnnynator:
You probably wanted to update it to 1.2.3 and not just refractor the template?

maxice8:
On Sat, May 26, 2018 at 02:04:06PM -0700, John Zimmermann wrote:
> You probably wanted to update it to 1.2.3 and not just refractor the template?
> 
> -- 
> https://github.com/voidlinux/void-packages/pull/14725#issuecomment-392288013

yes, i'm trying out git hooks to automatically refactor the template in some places, seems like it is meesing things up.

PR #14730, opened 2018-05-27
cava-gui: missing portaudio dependency

nikp123:
Portaudio input support has been added to the project since 0.6.1 and it hasn't been included.


maxice8:
increase revision

nikp123:
Done

PR #14732, opened 2018-05-27
php: update to 7.2.6.

maxice8:


Gottox:
> We can’t connect to the server at www.secure.php.net.

Gottox:
And why change it anyway?

Johnnynator:
@Gottox you need to use https to connect to www.secure.php.net. maxice probably changed the URL so the homepage is using https and not http.

Gottox:
There is no `www.secure.php.net`.

There's `secure.php.net`. But I would prefer `https://php.net` as it works too and is more future proof in case they fix their TLS Setup.

Johnnynator:
Oh yeah my fault. I agree, just `https://php.net` is probably better.

maxice8:
On Mon, May 28, 2018 at 01:31:28AM -0700, John Zimmermann wrote:
> Oh yeah my fault. I agree, just `https://php.net` is probably better.
> 
> -- 
> https://github.com/voidlinux/void-packages/pull/14732#issuecomment-392458803

fixed

PR #14733, opened 2018-05-27
khal: update to 0.9.9.

maxice8:


asergi:
Broken: https://github.com/pimutils/khal/commit/0fea680afa1cb0e13220f98948dbfe95ad22876c

Gottox:
Thanks, feel free to mention me in such cases.

Gottox:
I'm going to be more careful with python merges now :)

asergi:
Fixed in https://github.com/void-linux/void-packages/commit/b20b5f8f586196587e5f52ceb167cc48f8f0e5aa.

PR #14737, opened 2018-05-27
gns3: update to 2.1.6 and add aiohttp2-2.3.10 package

Johnnynator:


Gottox:
i686 needs to be fixed.

Johnnynator:
@Gottox I just set it to broken. Probably i686 is still affected by https://github.com/GNS3/gns3-server/issues/970 . It worked fine on x86_64-musl so I removed only_for_arches without testing every Architecture.

Gottox:
okay. merging for now as I guess it's an easy fix. Thanks for linking the issue.

PR #14742, opened 2018-05-27
chroma: update to 0.4.0

TeddyDD:
Also I changed homepage to literal address so it will be displayed correctly
on https://repo.voidlinux.eu/void-updates/void-updates.txt


TeddyDD:
@Gottox what's about second commit with version update?

Gottox:
applied, thanks for mentioning.

Issue #14765, opened 2018-05-29
Package Request - i8kutils


anarchomoh:
Hi 
Could I please request that i8kutils be packaged for Void?  I have an old Dell laptop and the fan's driving me insane.  This will help in controlling the fan.  Thanks.

https://github.com/vitorafsr/i8kutils

The Debian packages are on: https://launchpad.net/i8kutils

maxice8:
i'll take a look.

abenson:
done did

abenson:
Well, it's not merged yet ;]

PR #14766, opened 2018-05-29
New package: i8kutils-1.43

abenson:
Closes https://github.com/voidlinux/void-packages/issues/14765


srcpkgs/i8kutils/template
```diff
@@ -0,0 +1,32 @@
...
+maintainer="Andrew Benson <abenson+void@gmail.com>"
+license="GPL-2.0-or-later"
+homepage="https://launchpad.net/i8kutils"
+distfiles="https://launchpad.net/i8kutils/trunk/${version}/+download/i8kutils_${version}.tar.xz"
+checksum=881a211c64f79be6a08c88864deda45280f0ce925ed9e2af180db2d93d364dcb
+conf_files="/etc/i8kmon.conf"
+
+case "${XBPS_TARGET_MACHINE}" in
+	i686*) ;;
+	x86_64*) ;;
```
maxice8: i believe this can be replaced with i686*|x86_64*

abenson: That is a cleaner (and works), but I wonder if there is a really clean option

abenson: Other than only_for_archs="i686 i686-musl x86_64 x86_64-musl"

Issue #14750, opened 2018-05-28
Building gnutls fails with p11-kit


pullmoll:
It seems as if the recent update of p11-kit in #14702 broke gnutls.
gnutls has a local header `lib/pkcs11_int.h` which `#define CRYPTOKI_GNU` but that macro isn't checked for in `<p11-kit/pkcs11.h>`. Instead there is a check for `CRYPTOKI_COMPAT` which, when I define it, doesn't solve the problem.

Here's the error output:
```
  CC       pkcs7-output.lo
In file included from ./../pkcs11_int.h:34:0,
                 from x509.c:37:
/usr/include/p11-kit-1/p11-kit/uri.h:99:68: error: unknown type name 'CK_INFO'
                                                              const CK_INFO *info);
                                                                    ^~~~~~~
/usr/include/p11-kit-1/p11-kit/uri.h:104:68: error: unknown type name 'CK_SLOT_INFO'
                                                              const CK_SLOT_INFO *slot_info);
                                                                    ^~~~~~~~~~~~
/usr/include/p11-kit-1/p11-kit/uri.h:113:68: error: unknown type name 'CK_TOKEN_INFO'
                                                              const CK_TOKEN_INFO *token_info);
                                                                    ^~~~~~~~~~~~~
/usr/include/p11-kit-1/p11-kit/uri.h:134:68: error: unknown type name 'CK_ATTRIBUTE'
                                                              const CK_ATTRIBUTE *attrs,
                                                                    ^~~~~~~~~~~~
make[5]: *** [Makefile:1512: x509.lo] Error 1
make[5]: *** Waiting for unfinished jobs....
In file included from ./../pkcs11_int.h:34:0,
                 from verify-high2.c:36:
/usr/include/p11-kit-1/p11-kit/uri.h:99:68: error: unknown type name 'CK_INFO'
                                                              const CK_INFO *info);
                                                                    ^~~~~~~
/usr/include/p11-kit-1/p11-kit/uri.h:104:68: error: unknown type name 'CK_SLOT_INFO'
                                                              const CK_SLOT_INFO *slot_info);
                                                                    ^~~~~~~~~~~~
/usr/include/p11-kit-1/p11-kit/uri.h:113:68: error: unknown type name 'CK_TOKEN_INFO'
                                                              const CK_TOKEN_INFO *token_info);
                                                                    ^~~~~~~~~~~~~
[...]

PR #14798, opened 2018-05-30
p11-kit: update to 0.23.12.

maxice8:


pullmoll:
Did you try if that update works with `gnutls` again, i.e. compatibility to previous versions?
It seems the previous update broke gnults (see #14750).

maxice8:
On Wed, May 30, 2018 at 12:28:32PM -0700, Jürgen Buchmüller wrote:
> Did you try if that update works with `gnutls` again, i.e. compatibility to previous versions?
> It seems the previous update broke gnults (see #14750).
> 
> -- 
> https://github.com/voidlinux/void-packages/pull/14798#issuecomment-393287763

yes i built it successfully in x86_64, didn't test on *-musl or aarch64*|armv* .

the change in specific that fixes it comes from https://github.com/p11-glue/p11-kit/pull/160

maxice8:
On Wed, May 30, 2018 at 12:28:32PM -0700, Jürgen Buchmüller wrote:
> Did you try if that update works with `gnutls` again, i.e. compatibility to previous versions?
> It seems the previous update broke gnults (see #14750).
> 
> -- 
> https://github.com/voidlinux/void-packages/pull/14798#issuecomment-393287763

regardless i would try build again just to be sure.

pullmoll:
I'm trying with my branch gcc81 and see if that solves the problems.

pullmoll:
Yep, it really solves the issue. Thanks!

maxice8:
On Wed, May 30, 2018 at 07:36:28PM +0000, Jürgen Buchmüller wrote:
> Yep, it really solves the issue. Thanks!
> 
> -- 
> https://github.com/voidlinux/void-packages/pull/14798#issuecomment-393290033

<3

pullmoll:
But p11-kit doesn't cross build for `*-musl` .. strange. It cross builds for `*-glibc`, so this has to be some specific check in configure.

maxice8:
On Wed, May 30, 2018 at 07:50:53PM +0000, Jürgen Buchmüller wrote:
> But p11-kit doesn't cross build for `*-musl` .. strange. It cross builds for `*-glibc`, so this has to be some specific check in configure.
> 
> -- 
> https://github.com/voidlinux/void-packages/pull/14798#issuecomment-393295016

I can take a look later (adding tests to neovim).

pullmoll:
FWIW: [this is the change](https://github.com/p11-glue/p11-kit/commit/71b62aa1cdbdec3724c8e451f621309994dc59a0) which makes cross building `p11-kit` fail for `*-musl`.
In case of glibc the code relies on the `extern int __libc_enable_secure;` and configure does not (need to) run the `issetugid()` brokenness test.

That change should probably check for a configure option to override the test like `--with-issetugid-openbsd={no,yes}`. I dunno how to achieve that by patching `configure.ac`, though.

PR #14801, opened 2018-05-30
gede: add `qt5` build option

SolitudeSF:


Johnnynator:
Any reason why you don't set qt5 ad default?

Issue #14804, opened 2018-05-30
package request: sblg


copyleftie:
"The simple offline blog generator", tarball available at https://kristaps.bsd.lv/sblg/index.html

Should be a simple template.

PR #14805, opened 2018-05-30
New package: sblg-0.4.13

Johnnynator:


copyleftie:
is there a reason for `0.4.9`? The current version is `0.4.12`...

Johnnynator:
Actually now. I probably just have miss clicked and copied the wrong url

maxice8:
please add resolves #14804 to commit.

copyleftie:
I've been playing around with this template and found a subtle runtime error on musl. The program uses `strptime()` to parse for timestamps, and uses the flag `%F`, which musl doesn't support. Fortunately, this doesn't cause a crash, just a very confusing refusal to parse timestamps correctly.

This is easy to patch, basically just `sed -i s/%F/%Y-%m-%d/ grok.c`. Or you can use this patch file here: ([strptime-musl.patch.txt](https://github.com/voidlinux/void-packages/files/2086426/strptime-musl.patch.txt)).

I have a template/commit on my own branch if you just want to use that I guess.

Johnnynator:
@copyleftie Ok. I added your patch.

PR #14810, opened 2018-05-31
New package: kcharselect-18.04.1

travankor:


Johnnynator:
you are missing at least `kdoctools` in hostmakedepends for Cross Building.

PR #14811, opened 2018-05-31
p11-kit: disable issetugid(2) brokeness test.

pullmoll:
Can we safely assume that issetugid(2) is _not_ broken with musl libc?
For this upstream patch I set `ac_cv_issetugid_openbsd=no`.


ueno:
I don't understand your problem.  First, `issetugid()` is only used for emulating `getauxval()`, while musl-libc actually provides a working `getauxval()`.  Second, if the configure check detects `issetugid()` is broken, that means it cannot be used for emulating `getauxval()` (see the [issetugid sadness article](https://gist.github.com/nicowilliams/4daf74a3a0c86848d3cbd9d0cdb5e26e) linked from the commit).

Could you please elaborate the problem you are trying to solve (e.g. providing the error logs)? I can successfully run `./configure` and `make check` on Alpine Linux 3.5.2 from http://www.nongnu.org/pretest/.

pullmoll:
Since I have not much experience with the issue behind it, I'd rather leave the decision what to do with this problem to more knowledgeable @voidlinux/pkg-committers. See https://github.com/p11-glue/p11-kit/pull/170.

PR #14812, opened 2018-05-31
spotify: update to 1.0.79

teh-jazzman:
Since the 32bit version apparently is not beeing updated anymore I've made the package 64bit only for the time beeing. Maybe a separate solution for the outdated 32bit version can be found later.


cr6git:
Fine.
It closes [#14124](https://github.com/voidlinux/void-packages/pull/14124)

teh-jazzman:
Oh, sorry, I didn't see that you already had a pull request pending. I did not want to step on your toes there.

cr6git:
No problem.
It is me who stepped on your toe, you are the maintainer of this package after all.

jcjordyn130:
To whoever is doing the PR for this: You need to update it again, and while I'm at it why don't we keep a local copy of the .deb in FILES as to not break the package on upstream updates?

teh-jazzman:
We can't save the .deb file because it's proprietary software and void does not have the right to distribute it. I will create another pull request for the new version later today.

jcjordyn130:
Ah okay, that makes sense.

PR #14816, opened 2018-05-31
julia: update to 0.6.3

non-Jedi:


Johnnynator:
You probably have to set `wrksrc="julia"` since the 0.6.3 tar ball does not contain a versioned julia dir but a unversioned one.

non-Jedi:
@Johnnynator My bad. Was testing simultaneous with creating the PR, and I didn't check that everything completed correctly before opening it. `wrksrc` issue should be fixed now although the rest of the build is still in progress on my machine.

non-Jedi:
Finished local build and did a bit of testing. Everything seems to be in order now.

PR #14830, opened 2018-06-01
backupninja: update to 1.0.2.

Chocimier:
Old homepage links to new.


cr6git:
https://github.com/void-linux/void-packages/pull/1050

PR #14835, opened 2018-06-01
pjproject: update to 2.7.2.

Chocimier:


srcpkgs/pjproject/template
```diff
@@ -2,8 +2,8 @@
 # no-third-party.patch comes from Alpine,
 # and this template is heavily based on Alpine's APKBUILD.
 pkgname=pjproject
-version=2.7.1
-revision=4
+version=2.7.2
+revision=1
 disable_parallel_build=yes
 build_style=gnu-configure
 configure_args="--enable-shared \
```
maxice8: i don't think \ is necessary, but didn't test.

PR #14840, opened 2018-06-01
Added GEOS development dependency

camconn:
Added GEOS headers as a dependency. GEOS is essential to some of GDAL's core functionality, such as using the `OGRGeometry::Centroid` method. Without GEOS, these methods will fail (and Python's SWIG bindings will throw a RuntimeError).

Paging libgdal maintainer @pullmoll


maxice8:
bump revision= so that it rebuilds the package.

pullmoll:
Hmm.. GEOS support is not there when cross building libgdal, probably because geos-config is not found for the host.

PR #14842, opened 2018-06-01
New package: rwc

Anachron:
report when files change 

https://github.com/chneukirchen/rwc


srcpkgs/rwc/template
```diff
@@ -0,0 +1,11 @@
+# Template file for 'rwc'
+pkgname=rwc
+version=0.1
+revision=1
+build_style=gnu-makefile
+short_desc="report when files change"
+maintainer="Anachron <Anachron14@gmx.de>"
+license="GPL-3"
```
Johnnynator: SPDX Identifier.

Anachron: Oh crap, yeah the license is wrong. Thanks, will update shortly (just checking if all builds pass)

Anachron:
`srcpkgs/rwc/template:6: short_desc should start uppercase` <-- I think I hit the smallest possible error I could have done. :dagger:

srcpkgs/rwc/template
```diff
@@ -0,0 +1,11 @@
...
+homepage="https://github.com/chneukirchen/${pkgname}"
+distfiles="https://github.com/chneukirchen/${pkgname}/archive/v${version}.tar.gz"
+checksum="6b1136aba4c857beccca5811615d20184c9aded39150b77c5da0f6e176f53bf0"
```
Chocimier: No need to quote. BTW, there is `xgensum` that can set checksum for you. And `xbump`, that makes a commit for updated or new package.

maxice8:
problably good to change the commit to include the version of the package

New package: rwc-0.1

Anachron:
I've fixed all remaining issues. Thanks!

PR #14743, opened 2018-05-27
uzbl: change maintainer

dan-santana:

PR #14844, opened 2018-06-01
gst-plugins-good1: remove unused gtk+-devel from makedepends; add gtk3 build option

dan-santana:


Duncaen:
Hm this shouldn't change the resulting package, is it really necessary to bump the revision?

dan-santana:
I'm not sure, I just did it as a thumb of rule: if the output of `xbps-query -R <package>` is different, I bump the revision. Actually @Vaelatern asked me to bump the revision on [another PR](https://github.com/voidlinux/void-packages/pull/14743) when changing only the maintainer field too, so I don't know how to proceed in this case.

Vaelatern:
The maintianer field ends up in the final package.
`xbps-query -RS gst-plugins-good1`

dan-santana:
Yes, and it seems that the build options do too (as `build-options`), so should the revision be bumped in this case too?

Duncaen:
Thats the question, the `maintainer` field and other meta data fields have a value, the new `build-options` field for this package has no value at all because the current package has it enabled and the new package would have it enabled anyways.

But I guess we can just bump/rebuild it, its not that the would take ages to build or is large to re-download for a change that has no real effect.

dan-santana:
Got it! In this case, I reverted the revision number.

Vaelatern:
:shipit:

dan-santana:
Moved to https://github.com/void-linux/void-packages/pull/2952 (already merged).

PR #14857, opened 2018-06-02
fix dublicated soname for: procps-ng

pulux:


cr6git:
https://github.com/void-linux/void-packages/pull/146

PR #14862, opened 2018-06-02
rofi: split devel subpackage

SolitudeSF:


srcpkgs/rofi/template
```diff
@@ -17,3 +17,11 @@ checksum=f1bc0911201e528d443141d53de43be97538ee7484f8c278c38e2fab160aa173
 post_install() {
 	vlicense COPYING
 }
+
+rofi-devel_package() {
```
maxice8: it's pkg-config files declares cairo-devel and libglib-devel as Requires.private

SolitudeSF: i suspected that i missed something. thank you.

SolitudeSF: hmm, whats the difference between `libglib-devel` and `glib-devel`, since i compiled a package that depends on rofi-devel using either?

maxice8:
On Sat, Jun 02, 2018 at 02:22:35PM -0700, Solitude wrote:
> SolitudeSF commented on this pull request.
> 
> 
> 
> > @@ -17,3 +17,11 @@ checksum=f1bc0911201e528d443141d53de43be97538ee7484f8c278c38e2fab160aa173
>  post_install() {
>  	vlicense COPYING
>  }
> +
> +rofi-devel_package() {
> 
> hmm, whats the difference between `libglib-devel` and `glib-devel`, since i compiled a package that depends on rofi-devel using either?
> 
> -- 
> https://github.com/voidlinux/void-packages/pull/14862#discussion_r192570794

libglib-devel includes libraries and headers, glib-devel includes binaries used during packaging (like glib-mkenums), documentation and python stuff.

i always use libglib-devel, glib-devel generally appears on hostmakedepends when cross-compiling.

PR #14888, opened 2018-06-04
New package: libksane-18.04.1

travankor:


travankor:
The build is hanging on removing autodeps. Not sure if this is related to the template or not.

Johnnynator:
Nope is not related to your template. It is know that `xbps-remove -o` can take ages under certain circumstances

PR #14909, opened 2018-06-05
New package: dolphin-plugins-18.04.1

travankor:


travankor:
closing since this causes dolphin to crash frequently

PR #14910, opened 2018-06-05
New package: modsecurity-2.9.2.

maxice8:


srcpkgs/modsecurity/template
```diff
@@ -0,0 +1,23 @@
+# Template file for 'ModSecurity'
```
newbluemoon: I think this should match with `pkgname`.

srcpkgs/modsecurity/template
```diff
@@ -0,0 +1,23 @@
...
+homepage="https://www.modsecurity.org"
+distfiles="https://github.com/SpiderLabs/ModSecurity/releases/download/v${version}/modsecurity-${version}.tar.gz"
+checksum=41a8f73476ec891f3a9e8736b98b64ea5c2105f1ce15ea57a1f05b4bf2ffaeb5
+
+pre_configure() {
+	./autogen.sh
+}
+
+pre_build() {
+	sed -ri 's|(hardcode_into_libs)=.*|\1=no|' libtool
```
newbluemoon: Brackets/parentheses need to be escaped here: ...`\(hardcode_into_libs\)`...

PR #14916, opened 2018-06-05
cackey: rebuild against pcsclite

abenson:


Gottox:
Can you be more specific? - Which crash exaclty?

abenson:
It would segfault on load, taking Firefox with it. It seems to have been
something with the pcsclite update a while back.

abenson:
Fixed the commit message.

PR #14923, opened 2018-06-05
New package: perl-Email-MIME-1.946.

maxice8:
cc @chneukirchen


chneukirchen:
perl-Email-Address-XS-1.03_1 should not be noarch.

PR #14946, opened 2018-06-06
python-click-log: update to 0.3.0.

maxice8:


asergi:
This update breaks `vdirsyncer`.

PR #14965, opened 2018-06-06
font-symbola: set nonfree

Johnnynator:
the new License does not allow redistribution.

If anyone finds the old version with also a License then we could downgrade but the official site only hosts the newest version.


Johnnynator:
#13731 see comments there

Chocimier:
What about turning to nonfree instead?

Johnnynator:
@Chocimier you mean `restricted=yes` ? Because `nonfree` is for packages that are nonfree but still allow redistribution. That should also work, but would probably require that someone removes the package manually from the binary repo (?). (idk how the cleaning script works).

Gottox:
moving packages between repositories is basicly the same problem as deleting it from the repository completely. So restricted is fine with me too.

Johnnynator:
The PR sets font-symbola to restricted now.
If for any reason the nonfree method with INSTALL and REMOVE is preferred, it still is in my repo https://github.com/Johnnynator/void-packages/tree/font-symbola-nonfree

PR #14967, opened 2018-06-07
nodm: remove systemd service

cr6git:


maxice8:
Is it really necessary to remove it ?

cr6git:
Probably not.

PR #15011, opened 2018-06-08
Seafile client stack: switch to python3

yopito:
switch all underlying dependencies of seafile-client-qt to python3, that leads to update `libsearpc` package to last version 3.0.8. `seafile-client-qt` package does not require an update. Personal cross-build is fine.


yopito:
update: use variables instead of hardcoded path for python stuff.

PR #15068, opened 2018-06-11
xbps-src: include new chroot implementation (bubblewrap)

thypon:


common/chroot-style/bwrap.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
+#
+# This chroot script uses bubblewrap (see https://github.com/projectatomic/bubblewrap)
```
Gottox: comment needs update ;)

common/chroot-style/bwrap.sh
```diff
@@ -0,0 +1,23 @@
...
+#
+set -e
+readonly MASTERDIR="$1"
+readonly DISTDIR="$2"
+readonly HOSTDIR="$3"
+readonly EXTRA_ARGS="$4"
+shift 4
+
+if ! command -v bwrap >/dev/null 2>&1; then
+	exit 1
```
Gottox: It might be useful to get an error message here. (same goes for `proot`)

common/chroot-style/bwrap.sh
thypon: I was updating while you were commenting :P

common/chroot-style/bwrap.sh
thypon: :+1:

the-maldridge:
Wow, thanks for the quick turnaround on this!  I'll take a look into using it in my container CI commit, hopefully this will clear up the problems I was having there.

PR #15069, opened 2018-06-11
New package: wordwarvi-1.0.2

laserbat:


srcpkgs/wordwarvi/template
```diff
@@ -0,0 +1,21 @@
+# Template file for 'wordwarvi'
+pkgname=wordwarvi
+version=1.0.2
+revision=1
+build_style=gnu-makefile
+hostmakedepends="pkg-config"
+makedepends="portaudio-devel gtk+-devel libvorbis-devel alsa-lib-devel"
+short_desc="Word War vi is a side-scrolling shoot 'em up '80s style arcade game"
+maintainer="Olga Ustuzhanina <me@laserbat.pw>"
+license="GPL-2.0"
```
Johnnynator: Pls use [SPDX Identifier](https://spdx.org/licenses/). `GPL-2.0-or-later`

laserbat: Fixed!

Johnnynator: I meant the newer once `GPL-2.0` is depracted. `GPL-2.0-or-later` would be the correct License.

PR #15138, opened 2018-06-14
New package: olm-2.2.2

non-Jedi:
No packages currently depend on this, but I call it with some local scripts I have, and it'd be nice to see it packaged. In next release, nheko will also depend on this library.


srcpkgs/olm/template
```diff
@@ -0,0 +1,54 @@
+# Template file for 'olm'
+pkgname=olm
+version=2.2.2
+revision=1
+build_style=gnu-makefile
+make_check_target="test"
+hostmakedepends="gcc python-devel python3-devel"
```
maxice8: gcc is in base-devel.

the -devel packages should be in makedepends=

srcpkgs/olm/template
```diff
@@ -0,0 +1,54 @@
+# Template file for 'olm'
+pkgname=olm
+version=2.2.2
+revision=1
+build_style=gnu-makefile
+make_check_target="test"
```
maxice8: don't need to be quoted, it is a fixed string

srcpkgs/olm/template
```diff
@@ -0,0 +1,54 @@
...
+		vmove usr/include
+	}
+}
+
+python-olm_package() {
+  short_desc+=" - python bindings"
+	depends="${sourcepkg}>=${version}_${revision} python"
+	pycompile_module="olm"
+	pkg_install() {
+      vmove ${py2_sitelib}
```
maxice8: poor indentation

srcpkgs/olm/template
```diff
@@ -0,0 +1,54 @@
...
+	}
+}
+
+python3-olm_package() {
+  short_desc+=" - python3 bindings"
```
maxice8: poor identation

srcpkgs/olm/template
```diff
@@ -0,0 +1,54 @@
...
+
+olm-devel_package() {
+	depends="${sourcepkg}>=${version}_${revision}"
+	short_desc+=" - development files"
+	pkg_install() {
+		vmove usr/include
+	}
+}
+
+python-olm_package() {
```
maxice8: bindings follow <pkgname>-<language> rules.

see irssi-perl and weechat-ruby as examples.

non-Jedi:
Thanks for review @maxice8. Comments have been resolved.

srcpkgs/olm/template
```diff
@@ -0,0 +1,54 @@
...
+		mkdir -p ${DESTDIR}${pysl}
+		cp -r ${wrksrc}/python/olm ${DESTDIR}${pysl}
+	done
+}
+
+olm-devel_package() {
+	depends="${sourcepkg}>=${version}_${revision}"
+	short_desc+=" - development files"
+	pkg_install() {
+		vmove usr/include
```
maxice8: add 

vmove "usr/lib/*.so"

srcpkgs/olm/template
```diff
@@ -0,0 +1,54 @@
...
+	}
+}
+
+olm-python_package() {
+	short_desc+=" - python bindings"
```
maxice8: maybe necessary to add pycompile_modules="olm" here, not sure

is it @asergi ?

srcpkgs/olm/template
non-Jedi: done

srcpkgs/olm/template
non-Jedi: I have `pycompile_module=olm` a few lines down, or were you saying it might *not* be necessary?

maxice8: oh, yeah keep it, my bad.

non-Jedi:
FYI, I've added a necessary patch I missed on my first run through to make the python bindings able to find the library.

Travis build seems to be failing for several architectures, and I can't tell why. Any tips on diagnosing?

maxice8:
Hey, we moved to a new github organisation at https://github.com/void-linux/void-packages , can you re-open there ?

PR #15143, opened 2018-06-14
quodlibet: update to 4.1.0.

cr6git:


cr6git:
Replaced by https://github.com/void-linux/void-packages/pull/284

PR #15145, opened 2018-06-14
hunspell-pl_PL: update to 20180612.

cr6git:


cr6git:
Replaced by https://github.com/void-linux/void-packages/pull/277

PR #15149, opened 2018-06-14
fmit: add missing dependency / unbreak

cr6git:
https://github.com/voidlinux/void-packages/commit/be5b4e60209cc70d2dfb748b1fadd1bf46b2a696

`/bin/sh: itstool: command not found`

Good catch @chneukirchen :+1:


maxice8:
mind moving to new repo  at void-linux/void-packages ?

cr6git:
Replaced by https://github.com/void-linux/void-packages/pull/195

PR #15152, opened 2018-06-14
kea: add pgrep to checkdepends

Piraty:
The tests need pgrep

btw, it then fails like this:
```
<...>
Running test: pgsql_tests.sh

START TEST pgsql.lease-init
Wiping whole database keatest
Assertion failure: 0 != 127, for val1=0, val2=127
pgsql_wipe drop failed, expected exit code: 0, actual: 127
FAILED pgsql.lease-init
```


Piraty:
moved: https://github.com/void-linux/void-packages/pull/1

PR #15169, opened 2018-06-15
dnscrypt-proxy: ensure log dir is present

Piraty:
since it wasn't


Piraty:
moved: https://github.com/void-linux/void-packages/pull/5

PR #15170, opened 2018-06-15
lmms: update to 1.2.0rc6.

cr6git:
Tarball available for [rpmalloc](https://github.com/rampantpixels/rpmalloc/releases).


cr6git:
Replaced by https://github.com/void-linux/void-packages/pull/196

